PR #15, opened 2015-09-11
Support for tools:plugins

mikhailmikhaylov:
Please see the resulting changes. All the changes for plugins module are already in tools project


mikhailmikhaylov:
We should not apply this yet.

mikhailmikhaylov:
Now as we have plugins we can merge this to support them

PR #22, opened 2015-09-16
MessageJournal (not storage) and related changes

alexander-yevsyukov:


alexander-yevsyukov:
@AlexanderLitus, have a look at this pull request too. It already has compatibility fix for EngineShould test.

PR #27, opened 2015-09-23
Make AggregateRepository implement MultiHandler.

alexander-yevsyukov:
Added factory methods to MethodMap.

Added test for InMemorySample, which runs under IDEA, but fails to run under Gradle build. @mikhailmikhailov, please investigate.


alexander-yevsyukov:
@mikhailmikhaylov, please review and merge.

PR #35, opened 2015-10-19
Refactoring after the code review

AlexanderLitus:


AlexanderLitus:
Also, java-gae-sample module is empty now. Should I remove it?

alexander-yevsyukov:
@AlexanderLitus, yes, please remove the empty java-gae-sample module.

AlexanderLitus:
There are 2 TODOs related to Datastore which we can fix now (others are related to issue with Windows). 
They are located in DatastoreManager and both are related to connecting to cloud Datastore: one is related to Datastore project ID, second - to credentials usage.

PR #37, opened 2015-11-02
Remove properties file path prefix.

mikhailmikhaylov:


mikhailmikhaylov:
I've already changed the plugin.

PR #42, opened 2015-11-06
Migrate to "io.grpc:grpc-all" and "io.grpc:protoc-gen-grpc-java" 0.9.0

AlexanderLitus:
Migrate to "io.grpc:grpc-all" and "io.grpc:protoc-gen-grpc-java" 0.9.0;
replace internal API usages, use preferred API.

### DIFF
--- a/grpc-sample/src/main/java/org/spine3/sample/server/Server.java
+++ b/grpc-sample/src/main/java/org/spine3/sample/server/Server.java
@@ -19,9 +19,8 @@
  */
 package org.spine3.sample.server;
 
-import io.grpc.ServerImpl;
+import io.grpc.ServerBuilder;
 import io.grpc.ServerServiceDefinition;
-import io.grpc.netty.NettyServerBuilder;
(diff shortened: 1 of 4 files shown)


grpc-sample/src/main/java/org/spine3/sample/server/Server.java
```diff
@@ -114,12 +123,13 @@ public void run() {
...
         public void handle(CommandRequest req, StreamObserver<CommandResult> responseObserver) {
-            final CommandResult reply = Engine.getInstance().process(req);
 
-            responseObserver.onValue(reply);
+            final CommandResult reply = Engine.getInstance().process(req);
+            responseObserver.onNext(reply);
             responseObserver.onCompleted();
         }
 
+        @SuppressWarnings("ReturnOfNull")
```
alexander-yevsyukov: Why do we suppress this warning?

alexander-yevsyukov:
LGTM apart from the question on suppressed warning.

Just remove that suppression and we're good to move on.

PR #44, opened 2015-11-06
Projection documentation

alexander-yevsyukov:


AlexanderLitus:
LGTM, but what about paragraphs in javadocs?
I've found:
http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format
When use </p> tag it seems that indent is too big. Let's make a convention.

PR #49, opened 2015-11-18
Fix "java_outer_classname" option in person.proto (avoid conflict wit…

AlexanderLitus:
…h PersonName message).


mikhailmikhaylov:
LGTM

AlexanderLitus:
PTAL

AlexanderLitus:
There are only double and float in proto3.

AlexanderLitus:
PTAL.

alexander-yevsyukov:
Amount as Any is not a good idea. We obviously need to be definite here.

alexander-yevsyukov:
Please apply the fix to the outer class name and have money matter as a separate branch.

alexander-yevsyukov:
That is, remove adding money definition from this pull request, and then merge it. Money is a separate matter. Please don't mix things. It delays progressing the code.

PR #50, opened 2015-11-18
Money proto

AlexanderLitus:


AlexanderLitus:
PTAL (after this [pull request](https://github.com/SpineEventEngine/core-java/pull/51)).

core-java-sample/build.gradle
```diff
@@ -9,9 +9,7 @@ dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile project(path: ":core-java", configuration: 'testArtifacts')
 
-    compile project(':core-java')
-    compile project(':java-gae')
-
+    compile project(path: ':core-java');
     compile 'commons-io:commons-io:2.4'
```
alexander-yevsyukov: What do we need commons-io for?
Can we avoid this dependency in favor of IO-related classes from Guava?

core-java/src/main/proto/spine/money/money.proto
```diff
@@ -0,0 +1,408 @@
...
+package spine.money;
+
+option java_generate_equals_and_hash = true;
+option java_multiple_files = true;
+option java_outer_classname = "MoneyProto";
+option java_package = "org.spine3.money";
+
+import "google/protobuf/descriptor.proto";
+
+// Represents the amount of money.
```
alexander-yevsyukov: The word “Represents” may be excessive here and below. See Google's JavaDocs and Protobuf docs for examples of describing classes.

core-java/src/main/proto/spine/money/money.proto
```diff
@@ -0,0 +1,408 @@
...
+message Money {
+    // Represents the amount of money (e.g. cents).
```
alexander-yevsyukov: We need to be precise in what we have here. See https://www.wikiwand.com/en/ISO_4217, which defines the term “exponent”. We can refer to this ISO standard and the definition of exponent in explaining what the amount we have here.

core-java/src/main/proto/spine/money/money.proto
```diff
@@ -0,0 +1,408 @@
...
+    // Represents the currency.
+    Currency currency = 2;
+}
+
+message CurrencyOptions {
+    int32 numeric_code = 1;
+}
+
+extend google.protobuf.EnumValueOptions {
+    // TODO:2015-11-20:alexander.litus: change to globally unique field number
```
alexander-yevsyukov: Getting a unique field number require writing to Google. Why do we need a global number?
Can we use project-specific (specific to Spine) code?

core-java/src/main/proto/spine/money/money.proto
```diff
@@ -0,0 +1,408 @@
...
+    CurrencyOptions currency_options = 54321;
```
alexander-yevsyukov: How about naming the extension item just “currency”?

core-java/src/main/proto/spine/money/money.proto
```diff
@@ -0,0 +1,408 @@
...
+message Money {
+    // Represents the amount of money (e.g. cents).
+    sint64 amount = 1;
+
+    // Represents the currency.
+    Currency currency = 2;
+}
+
+message CurrencyOptions {
+    int32 numeric_code = 1;
```
alexander-yevsyukov: Please extend options with all the fields from the table at https://www.wikiwand.com/en/ISO_4217.
We already have the name as the item of the enum. We've got `numeric_code`.
We definitely need `exponent_digits`, as it would serve as the basis for interpreting the `amount` field.

We need to store the default name of the currency somewhere. Why not have it as the `name` or `title` field? Localization to other languages would use this value as a key for language-specific versions.

Information in the `country` field (assuming it can be repeated), would help in listing default currency basing on a geo location of the user.

alexander-yevsyukov: And we definitely need `symbol` field, so that we can display shortened formattings for money, like `$100`.

core-java-sample/build.gradle
AlexanderLitus: There are no such methods in Guava 19. Used java.nio.

core-java/src/main/proto/spine/money/money.proto
AlexanderLitus: OK. I took this from MS sample project.

core-java/src/main/proto/spine/money/money.proto
AlexanderLitus: From [language guide](https://developers.google.com/protocol-buffers/docs/proto#customoptions):
`Since custom options are extensions, they must be assigned field numbers like any other field or extension. In the examples above, we have used field numbers in the range 50000-99999. This range is reserved for internal use within individual organizations, so you can use numbers in this range freely for in-house applications. If you intend to use custom options in public applications, however, then it is important that you make sure that your field numbers are globally unique.`
I've obtained the field number from Google via email.

core-java/src/main/proto/spine/money/money.proto
AlexanderLitus: OK.

core-java/src/main/proto/spine/money/money.proto
AlexanderLitus: OK.

AlexanderLitus:
PTAL.

PR #51, opened 2015-11-20
Move java gae project

AlexanderLitus:


AlexanderLitus:
Should I publish core-java and gae-java projects to Maven central repo?

core-java-sample/build.gradle
```diff
@@ -9,9 +9,7 @@ dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile project(path: ":core-java", configuration: 'testArtifacts')
 
-    compile project(':core-java')
-    compile project(':java-gae')
-
+    compile project(path: ':core-java');
     compile 'commons-io:commons-io:2.4'
```
alexander-yevsyukov: Why do we need this dependency?

alexander-yevsyukov:
> Should I publish core-java and gae-java projects to Maven central repo? 

I don't think we're ready for pushing the code to the Central repo. We'll do it when we have a framework that can be used for evaluation for real projects.

core-java-sample/build.gradle
AlexanderLitus: org.apache.commons.io.FileUtils.deleteDirectory method is used in FileSystemStorageFactory. Unfortunately, there is no such method in java.nio (its delete method can delete only empty dirs).

PR #52, opened 2015-11-24
Improve exception messages, add tests

AlexanderLitus:
Improve exception messages for cases when Entity does not have a public constructor with an ID parameter.

### DIFF
--- a/core-java/src/main/java/org/spine3/server/RepositoryBase.java
+++ b/core-java/src/main/java/org/spine3/server/RepositoryBase.java
@@ -26,6 +26,8 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 
 import static com.google.common.base.Throwables.propagate;
+import static java.lang.reflect.Modifier.isPrivate;
+import static java.lang.reflect.Modifier.isPublic;
 
 /**
  * Abstract implementation of basic features of repositories.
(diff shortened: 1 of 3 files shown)


core-java/src/main/java/org/spine3/server/RepositoryBase.java
```diff
@@ -47,16 +49,29 @@ protected RepositoryBase() {
...
+        }
+    }
 
-            result = entityClass.getConstructor(idClass);
-        } catch (NoSuchMethodException e) {
-            throw propagate(e);
+    private static void checkConstructorAccessModifier(int modifiers, String entityClassName) {
+        if (!isPublic(modifiers)) {
+            final String constructorModifier = isPrivate(modifiers) ? "private." : "protected.";
+            final String message = "Constructor must be public in the " + entityClassName + " class, not " + constructorModifier;
```
alexander-yevsyukov: Please change: " not " with " found: ".

core-java/src/main/java/org/spine3/server/RepositoryBase.java
```diff
@@ -47,16 +49,29 @@ protected RepositoryBase() {
...
+        final Class<E> entityClass = getEntityClass();
+        final Class<I> idClass = getIdClass();
         try {
-            final Class<E> entityClass = getEntityClass();
-            final Class<I> idClass = getIdClass();
+            final Constructor<E> result = entityClass.getDeclaredConstructor(idClass);
+            checkConstructorAccessModifier(result.getModifiers(), entityClass.getSimpleName());
+            return result;
+        } catch (NoSuchMethodException ignored) {
+            throw noSuchConstructorException(entityClass.getSimpleName());
```
alexander-yevsyukov: Having short class name may be confusing. There can be cases of large applications that have the same name of things, but in different bounded contexts. E.g. consider "Order" for in two parts of a larger app that deal with different kinds of orders.

Please pass FQN.

core-java/src/main/java/org/spine3/server/RepositoryBase.java
```diff
@@ -47,16 +49,29 @@ protected RepositoryBase() {
...
+            final String constructorModifier = isPrivate(modifiers) ? "private." : "protected.";
+            final String message = "Constructor must be public in the " + entityClassName + " class, not " + constructorModifier;
+            throw propagate(new IllegalAccessException(message));
         }
-        return result;
+    }
+
+    private static RuntimeException noSuchConstructorException(String entityClassName) {
+        final String message = entityClassName +
+                " class must declare a constructor with a single ID parameter (like in Aggregate class).";
```
alexander-yevsyukov: The part "like in Aggregate class" doesn't sound clear. What did you mean to say?

core-java/src/main/java/org/spine3/server/RepositoryBase.java
```diff
@@ -47,16 +49,29 @@ protected RepositoryBase() {
...
 
     private Constructor<E> getEntityConstructor() {
-        final Constructor<E> result;
+        final Class<E> entityClass = getEntityClass();
+        final Class<I> idClass = getIdClass();
         try {
-            final Class<E> entityClass = getEntityClass();
-            final Class<I> idClass = getIdClass();
+            final Constructor<E> result = entityClass.getDeclaredConstructor(idClass);
+            checkConstructorAccessModifier(result.getModifiers(), entityClass.getSimpleName());
```
alexander-yevsyukov: We need the name of the class only when modifiers are not right. But we get it from the class all the time.
Why cannot we get the class name right before we generate the exception message?
Please also note my comment on using FQN for the entity class name below.

PR #54, opened 2015-12-01
Abstract base classes for process managers and process manager repositories, sample, tests

AlexanderLitus:
I decided to cover sagas with tests after the code review (tested via sample).

PR #55, opened 2015-12-01
Make command handler methods invoking more convenient

AlexanderLitus:
Please review this request after [this one](https://github.com/SpineEventEngine/core-java/pull/54).
The last commit only.


core-java-sample/src/main/java/org/spine3/sample/Application.java
```diff
@@ -29,6 +29,7 @@
 import org.spine3.base.UserId;
 import org.spine3.sample.order.OrderId;
 import org.spine3.sample.order.OrderRepository;
+import org.spine3.sample.saga.SagaRepositorySample;
```
alexander-yevsyukov: Please have package `procman` for everything that was `saga` or `process`.
Please rename Saga to ProcessManager everywhere too.

core-java-sample/src/main/java/org/spine3/sample/Requests.java
```diff
@@ -65,44 +63,37 @@ public static CommandRequest addOrderLine(UserId userId, OrderId orderId) {
...
+    }
 
-        final CommandRequest result = CommandRequest.newBuilder()
-                .setCommand(Messages.toAny(cmd))
-                .setContext(getCommandContext(userId))
-                .build();
-        return result;
+    public static CommandRequest newCommandRequest(UserId userId, Message command) {
+        final CommandContext context = getCommandContext(userId);
+        final CommandRequest.Builder request = CommandRequest.newBuilder()
```
alexander-yevsyukov: We need to have a utility method that takes command and its context and builds CommandRequest from them. See if we have such a method, if not, please create it and use everywhere.

core-java/src/main/java/org/spine3/server/aggregate/AggregateRepositoryBase.java
```diff
@@ -131,7 +137,7 @@ private Method dispatchAsMethod() {
      */
     @Nonnull
     @Override
-    public A load(I id) throws IllegalStateException {
+    public A load(@Nonnull I id) throws IllegalStateException {
```
alexander-yevsyukov: Why do we need this annotation here. The ID cannot be `null` ever, and it's covered by default annotation at the package level.

core-java/src/main/java/org/spine3/server/saga/SagaRepository.java
```diff
@@ -0,0 +1,183 @@
...
+    private static final String EVENT_DISPATCHER_METHOD_NAME = "dispatchEvent";
+
+    /**
+     * Returns a saga ID based on the information from the command and command context.
+     *
+     * @param command command which the saga handles
+     * @param context context of the command
+     * @return a saga ID
+     */
+    protected abstract I getSagaIdOnCommand(Message command, CommandContext context);
```
alexander-yevsyukov: How about `getId(Message, CommandContext)`?
Reasons:
1. We know that this is a command because we have CommandContext as the second parameter.
2. We have this method as `protected` and asking for an ID assumes we get the ID of things managed by the repository (which is ProcessManager) in this case.

core-java/src/main/java/org/spine3/server/saga/SagaRepository.java
```diff
@@ -0,0 +1,183 @@
...
+     * The name of the method used for dispatching events to sagas.
+     *
+     * <p>This constant is used for obtaining {@code Method} instance via reflection.
+     *
+     * @see #dispatchEvent(Message, EventContext)
+     */
+    private static final String EVENT_DISPATCHER_METHOD_NAME = "dispatchEvent";
+
+    /**
+     * Returns a saga ID based on the information from the command and command context.
```
alexander-yevsyukov: The word `information` is redundant.

core-java/src/main/java/org/spine3/server/saga/SagaRepository.java
```diff
@@ -0,0 +1,183 @@
...
+     *
+     * @param command command which the saga handles
```
alexander-yevsyukov: From what I remember from the code, the ID comes in context instances. Why do we need a command instance?

core-java/src/test/java/org/spine3/server/RepositoryBaseShould.java
```diff
@@ -73,7 +74,7 @@ public void store(EntityWithProtectedConstructor obj) {
         }
         @Nullable
         @Override
-        public EntityWithProtectedConstructor load(ProjectId id) {
+        public EntityWithProtectedConstructor load(@Nonnull ProjectId id) {
```
alexander-yevsyukov: Can we have non-nullity defined at the level of package?

core-java/src/main/java/org/spine3/server/aggregate/AggregateRepositoryBase.java
AlexanderLitus: OK.

core-java-sample/src/main/java/org/spine3/sample/Requests.java
AlexanderLitus: OK.

core-java-sample/src/main/java/org/spine3/sample/Application.java
AlexanderLitus: Yes, I've renamed sagas in the other branch.

AlexanderLitus:
Deleted this branch because it is tightly related to [this](https://github.com/SpineEventEngine/core-java/pull/54) pull request and branch. Processed the changes there.

PR #57, opened 2015-12-03
Engine configuration using Builder pattern.

alexander-yevsyukov:


core-java/src/test/java/org/spine3/server/EngineShould.java
```diff
@@ -233,4 +228,77 @@ public void on(TaskAdded event, EventContext context) {
         public void on(ProjectStarted event, EventContext context) {
         }
     }
```
AlexanderLitus: Why not move these inner static classes to the bottom? They are not very interesting, I think.
Other than that, LGTM.

PR #58, opened 2015-12-04
Introduce @Internal annotation and hide command handler methods.

alexander-yevsyukov:
Introduced internal interface for CommandHandlingObject.
MultiHandler is CommandHandlingObject now.
Also avoided instanceof's in scanning for methods in CommandHandlerMethod.
Hide AggregateCommandHandler and ProcessManagerCommandHandler from public API.
Added special case of method object for AggregateRepositoryDispatchMethod. It may help in debug in the future.
Made EntityId a part of public API.

### DIFF
--- a/core-java/src/main/java/org/spine3/server/internal/CommandHandlerMethod.java
+++ b/core-java/src/main/java/org/spine3/server/internal/CommandHandlerMethod.java
@@ -29,12 +29,11 @@ import com.google.protobuf.Message;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.spine3.CommandClass;
+import org.spine3.Internal;
 import org.spine3.base.CommandContext;
 import org.spine3.internal.MessageHandlerMethod;
 import org.spine3.server.Assign;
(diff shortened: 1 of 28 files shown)


core-java/src/main/java/org/spine3/server/internal/CommandHandlerMethod.java
```diff
@@ -138,58 +135,47 @@ protected static boolean acceptsCorrectParams(Method method) {
      */
     @CheckReturnValue
     public static Map<CommandClass, CommandHandlerMethod> scan(Object object) {
```
AlexanderLitus: I'd add to docs something like this:
''
@param object the object implementing {@link CommandHandlingObject} that keeps command handler methods
@see CommandHandlingObject
''
So that client doesn't have to look into the method implementation.
Other than that, LGTM.

alexander-yevsyukov:
@AlexanderLitus, PTAL

alexander-yevsyukov:
@AlexanderLitus, PTAL.

PR #60, opened 2015-12-08
Check entity ID type in Entity constructor.

AlexanderLitus:


core-java/src/main/java/org/spine3/server/internal/EntityId.java
```diff
@@ -60,8 +61,33 @@
...
+     * @param entityId the ID of the entity to check
+     * @throws IllegalArgumentException if the ID is not of one of the supported types
+     */
+    public static <I> void checkType(I entityId) {
+        final Class<?> idClass = entityId.getClass();
+        if (SUPPORTED_TYPES.contains(idClass)) {
+            return;
+        }
+        if (!Message.class.isAssignableFrom(idClass)){
+            throw new IllegalArgumentException("Unsupported entity ID type: " + idClass.getName());
```
alexander-yevsyukov: Please list types that are needed, and then one found.

PR #65, opened 2015-12-10
Code layout improvements

alexander-yevsyukov:


core-java/src/main/proto/spine/base/command_service_new.proto
```diff
@@ -0,0 +1,56 @@
...
+    oneof value {
+        // The command was accepted to execution.
+        google.protobuf.Empty ok = 1;
+
+        // The message of the error occurred upon receiving the request.
+        string error = 2;
+    }
+}
+
+//TODO:2015-12-10:alexander.yevsyukov: Uncomment this block when grpc/src/main/proto/command_service.proto is addressed.
```
AlexanderLitus: What does it mean? Other than that, LGTM.

alexander-yevsyukov: This thing is going to replace currently used CommandResult. By default, the server is going to return just nothing, if the command is accepted to the processing.

PR #66, opened 2015-12-10
Introduce 'client' package

alexander-yevsyukov:
This PR is based on #65, and in addition introduces package for client-side API.


AlexanderLitus:
LGTM.

PR #69, opened 2015-12-18
Aggregate state neutral events

AlexanderLitus:
@alexander-yevsyukov, PTAL.

### DIFF
--- a/core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
+++ b/core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
@@ -59,7 +59,6 @@ import static org.spine3.util.Identifiers.idToAny;
  * @author Mikhail Melnik
  * @author Alexander Yevsyukov
  */
-@SuppressWarnings("ClassWithTooManyMethods")
 public abstract class Aggregate<I, M extends Message> extends Entity<I, M> implements CommandHandlingObject {
 
     /**
@@ -132,22 +131,22 @@ public abstract class Aggregate<I, M extends Message> extends Entity<I, M> imple
      * in the {@link Registry} if it is not registered yet.
      */
     private void init() {
-
-        if (!this.initialized) {
-
-            final Registry registry = Registry.getInstance();
-            final Class<? extends Aggregate> thisClass = getClass();
-
-            // Register this aggregate root class if it wasn't.
-            if (!registry.contains(thisClass)) {
-                registry.register(thisClass);
-            }
-
-            commandHandlers = registry.getCommandHandlers(thisClass);
-            eventAppliers = registry.getEventAppliers(thisClass);
-
-            this.initialized = true;
+        if (this.initialized) {
+            return;
         }
+
+        final Registry registry = Registry.getInstance();
+        final Class<? extends Aggregate> thisClass = getClass();
+
+        // Register this aggregate root class if it wasn't.
+        if (!registry.contains(thisClass)) {
+            registry.register(thisClass);
+        }
+
+        commandHandlers = registry.getCommandHandlers(thisClass);
+        eventAppliers = registry.getEventAppliers(thisClass);
+
+        this.initialized = true;
     }
 
     private void invokeApplier(Message event) throws InvocationTargetException {
@@ -309,22 +308,18 @@ public abstract class Aggregate<I, M extends Message> extends Entity<I, M> imple
      */
     @CheckReturnValue
     protected EventContext createEventContext(CommandId commandId, Message event, M currentState, Timestamp whenModified, int currentVersion) {
-
         final EventId eventId = Events.createId(commandId, whenModified);
-
-        final EventContext.Builder builder = EventContext.newBuilder()
+        final EventContext.Builder result = EventContext.newBuilder()
                 .setEventId(eventId)
                 .setVersion(currentVersion)
                 .setAggregateId(getIdAsAny());
-
-        addEventContextAttributes(builder, commandId, event, currentState, currentVersion);
-
-        return builder.build();
+        addEventContextAttributes(result, commandId, event, currentState, currentVersion);
+        return result.build();
     }
 
     /**
      * Adds custom attributes to an event context builder during the creation of the event context.
-     * <p/>
+     *
      * <p>Does nothing by default. Override this method if you want to add custom attributes to the created context.
      *
      * @param builder        a builder for the event context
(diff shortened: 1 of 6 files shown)


core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -220,22 +225,31 @@ public void play(Iterable<EventRecord> records) throws InvocationTargetException
         }
     }
 
+    /**
+     * Applies events to an aggregate unless they are state-neutral.
+     *
+     * <p>Event applier should call {@link #incrementState(Message)},
```
alexander-yevsyukov: We should not rely on this. Please arrange 'automatic' version advancement. State neutral events should not advance the version. It is assumed that a state-neutral event comes in a sequence of events where first one modifies the state.

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -255,12 +269,21 @@ private void apply(Message event) throws InvocationTargetException {
...
-
         invokeApplier(event);
     }
 
-    private void putUncommitted(EventRecord record) {
-        uncommittedEvents.add(record);
+    /**
+     * Returns a set of classes of state-neutral events.
+     *
+     * <p>An event is state-neutral if there is no need to modify a state of an aggregate on this event.
```
alexander-yevsyukov: The aggregate does not belong to the event. I would use definite articles in this sentence. Please consult with Mary on this.

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -255,12 +269,21 @@ private void apply(Message event) throws InvocationTargetException {
...
+     * It is not required to create applier methods for these events.
```
alexander-yevsyukov: I would say: "Instead of creating empty applier methods for such events, override this method returning immutable set of event classes. `<Code example>`"

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -274,18 +297,56 @@ public void restore(SnapshotOrBuilder snapshot) {
...
     /**
-     * @return immutable view of records for uncommitted events
+     * Returns all uncommitted events (including state-neutral).
+     *
+     * @return immutable view of records for all uncommitted events
+     * @see #getStateNeutralEventClasses()
      */
     @CheckReturnValue
-    public List<EventRecord> getUncommittedEvents() {
+    public List<EventRecord> getAllUncommittedEvents() {
```
alexander-yevsyukov: Why 'all'? Are there methods that return 'some'?

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -274,18 +297,56 @@ public void restore(SnapshotOrBuilder snapshot) {
...
     }
 
     /**
-     * Returns and clears the events that were uncommitted before the call of this method.
+     * Returns uncommitted events (excluding state-neutral).
+     *
+     * @return immutable view of records for applicable uncommitted events
+     * @see #getStateNeutralEventClasses()
+     */
+    public Collection<EventRecord> getApplicableUncommittedEvents() {
```
alexander-yevsyukov: How about naming this method `getStateChangingUncomittedEvents()`?

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -274,18 +297,56 @@ public void restore(SnapshotOrBuilder snapshot) {
...
+        //noinspection LocalVariableNamingConvention
+        final Set<Class<? extends Message>> stateNeutralEventClasses = getStateNeutralEventClasses();
+        final Predicate<EventRecord> isNotStateNeutral = isNotStateNeutralPredicate(stateNeutralEventClasses);
+        final Collection<EventRecord> result = filter(uncommittedEvents, isNotStateNeutral);
+        return result;
+    }
+
+    @SuppressWarnings("MethodParameterNamingConvention") // to be precise
+    private static Predicate<EventRecord> isNotStateNeutralPredicate(
```
alexander-yevsyukov: How about `isStateChanging`?

Have a constant to avoid creation of predicate instances all the time. And do have the method that returns the constant.

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -274,18 +297,56 @@ public void restore(SnapshotOrBuilder snapshot) {
...
+                return !isStateNeutral;
+            }
+        };
+    }
+
+    /**
+     * Returns and clears all the events that were uncommitted before the call of this method.
      *
-     * @return the list of event records
+     * @return the list of event records (including state-neutral)
```
alexander-yevsyukov: There's no need to stress on inclusion of state-neutral events. It's assumed.

core-java/src/main/java/org/spine3/server/aggregate/AggregateRepositoryBase.java
```diff
@@ -182,7 +183,7 @@ public A load(I id) throws IllegalStateException {
      */
     @Override
     public void store(A aggregateRoot) {
-        final List<EventRecord> uncommittedEvents = aggregateRoot.getUncommittedEvents();
+        final Collection<EventRecord> uncommittedEvents = aggregateRoot.getApplicableUncommittedEvents();
```
alexander-yevsyukov: Events are applied in the sequence. Return `Iterable`, if you'd like.

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
alexander-yevsyukov: Why do we want this method being `public`? What's the interest of having only state-changing events _outside_ the aggregate?

core-java/src/main/java/org/spine3/server/aggregate/Aggregate.java
alexander-yevsyukov: We need to have description of an event applier somewhere else. Since this method is private it's not really likely that this documentation is going to be read soon enough. But the user would need to understand how to write an applier from the very beginning.

PR #70, opened 2015-12-21
ClientService defined

alexander-yevsyukov:
This PR defines general API for calling ClientService and drafts how BoundedContext work with it.

Further implementation may result in more classes linked with BoundedContext, but ClientService API is likely to stay as it's now.


mikhailmikhaylov:
LGTM

PR #71, opened 2015-12-28
Add MoneyUtil class.

AlexanderLitus:


alexander-yevsyukov:
It should be in `money` package.

PR #74, opened 2016-01-13
Event store service

alexander-yevsyukov:


alexander-yevsyukov:
@AlexanderLitus 
I commented out `&& beforeSpecified` because IDEA complains about always `true` here. We either need to suppress the warning or comment it out. The latter is easier and better from the code point of view.

Comments are stale if developers don't pay attention to them. If we change this part of the code, we'll change the comment too.

PR #79, opened 2016-01-14
Implement storing entity version and timestamp in storage records

AlexanderLitus:
@alexander-yevsyukov, PTAL.


core-java/src/main/proto/spine/base/entity.proto
```diff
@@ -0,0 +1,47 @@
...
+    // The timestamp of the last entity modification.
+    google.protobuf.Timestamp when_modified = 2;
+
+    // The version of the entity.
+    int32 version = 3;
+
+    reserved 4 to 19;
+
+    // The entity state.
+    google.protobuf.Any entity_state = 20;
```
alexander-yevsyukov: Why not just `state`?

core-java/src/main/proto/spine/server/storage/entity_storage.proto
```diff
@@ -0,0 +1,47 @@
...
+option java_generate_equals_and_hash = true;
+option java_multiple_files = true;
+option java_outer_classname = "EntityStorageProto";
+option java_package = "org.spine3.server.storage";
+
+import public "google/protobuf/timestamp.proto";
+import public "google/protobuf/any.proto";
+
+// Used to store entities in the entity storage.
+message EntityStorageRecord {
```
alexander-yevsyukov: Why do we need to have this type AND `EntityRecord` that have the same set of fields?

If you plan to extend EntityStorageRecord in the future, please specify how.

storage-filesystem/src/main/java/org/spine3/server/storage/filesystem/FsEntityStorage.java
```diff
@@ -88,6 +86,18 @@ public void write(I id, M message) {
 
         final Any any = Messages.toAny(message);
         FsUtil.writeMessage(file, any);
+    }*/
+
+    @Nullable
+    @Override
+    protected EntityStorageRecord read(I id) {
+        // TODO:2016-01-14:alexander.litus: impl
```
alexander-yevsyukov: Please do.

storage-filesystem/src/main/java/org/spine3/server/storage/filesystem/FsEntityStorage.java
```diff
@@ -88,6 +86,18 @@ public void write(I id, M message) {
...
+        return null;
+    }
+
+    @Override
+    protected void write(EntityStorageRecord record) {
+        // TODO:2016-01-14:alexander.litus: impl
```
alexander-yevsyukov: Please do.

core-java/src/main/java/org/spine3/server/BoundedContext.java
```diff
@@ -165,7 +166,7 @@ private void shutDownRepositories() {
      * <p>For instances of {@link AggregateRepository} an instance of {@link AggregateStorage} is created
      * and assigned.
      *
-     * <p>For other types of repositories an instance of {@link org.spine3.server.storage.EntityStorage} is
+     * <p>For other types of repositories an instance of {@link EntityStorage} is
```
alexander-yevsyukov: Let's discuss the way we work with storages. I suppose, we can simplify this and make it more transparent for the developer. Now assigning a storage is a hidden thing, which requires having a storage factory.
Instead, we can accept fully built instance of a repository. This way we would be able adding new kinds of repositories _without_ the need to modify StorageFactory API.

core-java/src/main/java/org/spine3/server/EntityRepository.java
```diff
@@ -41,41 +45,53 @@
...
+        return entity;
+    }
 
-            result.setState(entityState, 0, TimeUtil.getCurrentTime());
-            return result;
-        }
-        // No state stored -> no entity.
-        return null;
+    private EntityRecord toEntityRecord(E entity) {
+        final String idString = idToString(entity.getId());
```
alexander-yevsyukov: Why do we have it as `string`? It can be one of the integer types too. We have a set of supported ID types, and we have the type we deal with defined as generic type parameter.

core-java/src/test/java/org/spine3/server/storage/EntityStorageShould.java
```diff
@@ -23,82 +23,108 @@
...
+import static org.spine3.protobuf.Messages.toAny;
+import static org.spine3.util.Identifiers.newUuid;
 
 @SuppressWarnings({"InstanceMethodNamingConvention", "AbstractClassWithoutAbstractMethods",
-        "ConstructorNotProtectedInAbstractClass", "DuplicateStringLiteralInspection", "ConstantConditions"})
+        "ConstructorNotProtectedInAbstractClass"})
 public abstract class EntityStorageShould {
 
-    private final EntityStorage<String, StringValue> storage;
+    private static final int DEFAULT_ENTITY_VERSION = 5;
```
alexander-yevsyukov: Why 5?

core-java/src/main/java/org/spine3/server/EntityRepository.java
```diff
@@ -41,41 +45,53 @@
 
     @Nullable
     @Override
-    protected EntityStorage<I, M> getStorage() {
```
alexander-yevsyukov: What happens if we pass an EntityRecord with the state, which is not of the same type with those already stored in the repository?

alexander-yevsyukov: As discussed vocally, it's not going to cause problems because we control the passed type from an EntityRepository.

core-java/src/main/proto/spine/server/storage/entity_storage.proto
```diff
@@ -0,0 +1,61 @@
...
+    message Id {
+        oneof type {
+            // The string representation of the ID.
+            string string_value = 1;
+
+            // The integer representation of the ID.
+            int32 int_value = 2;
+
+            // The long representation of the ID.
+            int64 long_value = 3;
```
alexander-yevsyukov: Have `long_value` before `int_value`. Long IDs are going to be used more often.

core-java/src/test/java/org/spine3/server/storage/EntityStorageShould.java
```diff
@@ -23,18 +23,21 @@
...
+import static com.google.protobuf.util.TimeUtil.getCurrentTime;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.spine3.protobuf.Messages.toAny;
+import static org.spine3.server.storage.EntityStorage.*;
+import static org.spine3.util.Identifiers.newUuid;
 
 @SuppressWarnings({"InstanceMethodNamingConvention", "AbstractClassWithoutAbstractMethods",
-        "ConstructorNotProtectedInAbstractClass", "DuplicateStringLiteralInspection", "ConstantConditions"})
+        "ConstructorNotProtectedInAbstractClass"})
```
alexander-yevsyukov: Please have protected constructor.

core-java/src/test/java/org/spine3/server/storage/memory/InMemoryEntityStorageShould.java
```diff
@@ -31,7 +30,8 @@
  */
 public class InMemoryEntityStorageShould extends EntityStorageShould {
 
-    private static final InMemoryEntityStorage<String, StringValue> STORAGE = InMemoryEntityStorage.newInstance();
+    private static final InMemoryEntityStorage<String> STORAGE =
+            InMemoryEntityStorage.newInstance();
```
alexander-yevsyukov: Why on new line?

PR #84, opened 2016-01-24
Have CommandBus object and CommandDispatcher interface

alexander-yevsyukov:
We need to have a concept for classes that do not process commands, but just dispatch them to objects responsible for handling.

This PR introduces `CommandDispatcher` as such thing. It required to restore Command Bus its canonical name, which is a good thing, as it ease the learning curve.


armiol:
The renaming itself LGMT.

However, I'd like to discuss the naming of the BoundedContext class. Right now it means "Restricted Context", but there is also a common class naming a-la "BoundContext", meaning "The context wired to something". Let's discuss what we wanted to imply here.

armiol:
OK, the context is bounded. Resolved.

PR #87, opened 2016-01-27
Distinguish dispatchers and handlers in CommandBus and EventBus

alexander-yevsyukov:


alexander-yevsyukov:
@armiol, FYI.

AlexanderLitus:
Other than my comment, LGTM.

alexander-yevsyukov:
As we discussed, there has to be result of executing a command. In case of routing in a Process Manager, there should be `CommandRouted` event that would contain command requests that were generated for routing.

This is going to be implemented as a separate task.

PR #97, opened 2016-01-29
Rename command request to command

alexander-yevsyukov:


server/src/main/java/org/spine3/server/CommandBus.java
```diff
@@ -121,11 +117,21 @@ protected CommandBus(CommandStore commandStore) {
...
+     * <p>If {@code Command} instance is passed, its message is extracted and validated.
+     *
+     * @param command the command message to check
      * @return the result of {@link Responses#ok()} if the command is supported,
      *         {@link CommandValidation#unsupportedCommand(Message)} otherwise
      */
     public Response validate(Message command) {
+        if (command instanceof Command) {
+            final Command outerCommand = (Command) command;
+            //noinspection AssignmentToMethodParameter
```
mikhailmikhaylov: Didn't we want to move from comment suppression to annotated?

server/src/test/java/org/spine3/server/BoundedContextShould.java
```diff
@@ -152,7 +151,7 @@ public void return_CommandDispatcher() {
     }
 
     @Test(expected = NullPointerException.class)
-    public void throw_NPE_on_null_CommandRequest() {
+    public void throw_NPE_on_null_Command() {
         // noinspection ConstantConditions
```
mikhailmikhaylov: Same here, maybe we should use annotation?

mikhailmikhaylov:
LGTM, but I see comment suppressions and probably changes made by IDEA (inspection profiles).

server/src/test/java/org/spine3/server/BoundedContextShould.java
alexander-yevsyukov: The problem is that there are no annotations for some of the warnings, and if we stick with annotation-only approach suppressions will cover the whole method. I'll fix it here, though.

server/src/main/java/org/spine3/server/CommandBus.java
alexander-yevsyukov: See my comment on this matter below. But I guess, it's better to avoid the suppression and make the intent clearer.

alexander-yevsyukov:
Inspection profile changed to make writing tests easier.

PR #98, opened 2016-01-29
Timestamps refactoring, docs and tests.

AlexanderLitus:


client/src/main/java/org/spine3/protobuf/Timestamps.java
```diff
@@ -38,53 +41,108 @@
...
-    public static boolean isAfter(Timestamp timestamp, Timestamp from) {
-        return (compare(from, timestamp) < 0);
+    /**
+     * Calculates if the {@code isAfter} timestamp is after the {@code point} timestamp.
+     *
+     * @param point the first point in time which is supposed to be before the {@code isAfter} timestamp
+     * @param isAfter the timestamp to check if it is after the {@code point} timestamp
+     * @return true if the {@code isAfter} timestamp is after the {@code point} timestamp, false otherwise
+     */
+    public static boolean isAfter(Timestamp point, Timestamp isAfter) {
```
alexander-yevsyukov: Did you check the usages of this method? From what I see the order of parameters changed.

alexander-yevsyukov: Anyway, the previous naming and the order of parameters was consistent with the rest of methods in this class: the argument in question comes first. Please return the order back and change the naming accordingly.

AlexanderLitus:
@alexander-yevsyukov, PTAL.

AlexanderLitus:
OK, I've removed this method with default context. Anything else?

alexander-yevsyukov:
Please change the min method name length to 2, and it'd be LGTM.

PR #100, opened 2016-01-29
Remove string aggregate ID from aggregate storage record

AlexanderLitus:
Remove string aggregate ID from aggregate storage record (let the client to choose how to store IDs).

Fix and add tests for aggregate storage, fix test factory method.

### DIFF
--- a/server/src/main/java/org/spine3/server/storage/AbstractStorage.java
+++ b/server/src/main/java/org/spine3/server/storage/AbstractStorage.java
@@ -58,7 +58,6 @@ public abstract class AbstractStorage<I, R extends Message> implements AutoClose
      * @param id the ID for the record
      * @param record a record to store
      * @throws IllegalStateException if the storage was closed before
-     * @throws NullPointerException if the ID or record is {@code null}
      */
     public abstract void write(I id, R record);
 
--- a/server/src/main/java/org/spine3/server/storage/AbstractStorage.java
+++ b/server/src/main/java/org/spine3/server/storage/AbstractStorage.java
@@ -58,6 +58,7 @@ public abstract class AbstractStorage<I, R extends Message> implements AutoClose
      * @param id the ID for the record
      * @param record a record to store
      * @throws IllegalStateException if the storage was closed before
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/spine3/server/storage/AbstractStorage.java
```diff
@@ -58,6 +58,7 @@
      * @param id the ID for the record
      * @param record a record to store
      * @throws IllegalStateException if the storage was closed before
+     * @throws NullPointerException if the ID or record is {@code null}
```
alexander-yevsyukov: We use `@ParametersAreNonnullByDefault` on the level of the package, which assumes that everything, which is not `@Nullable` is going to cause NPE when passed.

server/src/main/java/org/spine3/server/storage/AggregateStorage.java
```diff
@@ -83,52 +86,87 @@ public AggregateEvents read(I aggregateId) {
...
 
+    /**
+     * Writes an event record to the storage by an aggregate ID.
+     *
+     * @param id the aggregate ID
+     * @param eventRecord the record to write
+     * @throws NullPointerException if the ID or record is {@code null}
+     * @throws IllegalStateException if the storage is closed
+     */
+    public void writeEvent(I id, EventRecord eventRecord) {
```
alexander-yevsyukov: Please rename the `eventRecord` parameter to `event`. New PR, which contains related renaming is coming.

server/src/main/java/org/spine3/server/storage/AggregateStorage.java
```diff
@@ -83,52 +86,87 @@ public AggregateEvents read(I aggregateId) {
...
+        checkNotClosed();
+        checkNotNull(id, "aggregate id");
+        checkNotNull(eventRecord, "event record");
```
alexander-yevsyukov: Have just "event" here.

alexander-yevsyukov:
Other than that, LGTM.

PR #102, opened 2016-01-29
Rename stream projection to projection

alexander-yevsyukov:


AlexanderLitus:
Other than that, LGTM.

PR #103, opened 2016-02-01
Implement `getDefaultState` in Entity

AlexanderLitus:


server/src/main/java/org/spine3/server/Entity.java
```diff
@@ -50,6 +59,8 @@
 
     private int version;
 
+    private final M defaultState;
```
alexander-yevsyukov: Isn't this state a part of the class, rather than each instance? 

It's immutable and the same for all the instances of the class. The method `getDefaultState()` can and probably should read this state once it's available, but why have it in every instance?

AlexanderLitus: Yes, I've tried to declare it static, but there is a [restriction...](https://docs.oracle.com/javase/tutorial/java/generics/restrictions.html#createStatic)

alexander-yevsyukov: Have map per class. We cannot have default state in every instance. It is memory-expensive.

AlexanderLitus:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov:
LGTM

PR #104, opened 2016-02-01
Make `CommandHandlerMethod` not abstract

alexander-yevsyukov:


alexander-yevsyukov:
@AlexanderLitus 
I've merged your recent changes from `entity-default-state` into this branch as it already added improvement in terms of API clean-up. But please address the comments on keeping default state at the class level.

PR #108, opened 2016-02-08
Add "required" option definition.

mikhailmikhaylov:


client/src/main/proto/spine/base/options.proto
```diff
@@ -0,0 +1,35 @@
...
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+syntax = "proto3";
+
+// We do not define package here to allow users use these options without package, e.g.
```
alexander-yevsyukov: How about wording like this:

// We do not define package for options for brevity of their usage.
// Without the package, it's possible to have `(required = true)` vs `(spine.base.required = true)`.
// For more information on using options, please see documentation of the corresponding types.

client/src/main/proto/spine/base/options.proto
```diff
@@ -0,0 +1,35 @@
...
+
+option java_generate_equals_and_hash = false;
+option java_multiple_files = true;
+option java_outer_classname = "OptionsProto";
+option java_package = "org.spine3.base";
+
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.FieldOptions {
+    bool required = 51001;
```
alexander-yevsyukov: Please add documentation here.

alexander-yevsyukov:
Please add test for failing validation of a command without `required` field specified.

client/src/main/proto/spine/base/options.proto
alexander-yevsyukov: Please also get unique non-project-dependent ID from Protobuf developers. Ask @AlexanderLitus for details.

PR #116, opened 2016-02-19
Add command storage tests.

AlexanderLitus:


server/src/main/java/org/spine3/server/storage/memory/InMemoryCommandStorage.java
```diff
@@ -104,4 +102,10 @@ private void put(CommandId id, CommandStorageRecord record) {
     private CommandStorageRecord get(CommandId id) {
         return storage.get(id);
     }
+
+    private void checkNotContains(CommandId id, String commandId) {
+        if (storage.containsKey(id)) {
```
alexander-yevsyukov: This hints on the fact that we need a separate method `update()`, which would take the same arguments, but assume that there's a record with such ID.

AlexanderLitus: We can add an `update` method, but it is needed only in EntityStorage now. Also, we don't know what method to use: `update` or `insert`, so the `contains` method is also needed (only for entity storage). And implementing these methods in all storages violates the YAGNI principle.
Or we can document `write` method in entity storage that it can also update the record, and that it only inserts one record in other storages.

AlexanderLitus:
@alexander-yevsyukov, PTAL.

server/src/main/java/org/spine3/server/EntityRepository.java
```diff
@@ -81,7 +82,7 @@ public void store(E entity) {
     public E load(I id) {
```
alexander-yevsyukov: JavaDoc? Even if it's simply `{@inheritDoc}` it would be nice.

alexander-yevsyukov: We mainly need to document when we return `null`.

server/src/main/java/org/spine3/server/storage/memory/InMemoryCommandStorage.java
```diff
@@ -41,24 +41,22 @@
     @Override
     public void write(CommandId id, CommandStorageRecord record) {
         checkNotNull(id);
```
alexander-yevsyukov: How about having `checkNotDefault(M extends Message)`, which does non-null and non-default checks? We are going to have it over and over again.

alexander-yevsyukov: We also need `Identifiers.checkValid()` for checking `EventId`, `CommandId`, `AggregateId`, `ProcessManagerId`.

server/src/main/java/org/spine3/server/EntityRepository.java
AlexanderLitus: This is documented in Repository class.

server/src/main/java/org/spine3/server/storage/memory/InMemoryCommandStorage.java
AlexanderLitus: Added for command and event IDs (Aggregate and PM IDs are gone).

AlexanderLitus: Added for event and command IDs (others are gone).

AlexanderLitus:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov:
Please adjust `CommandStorageRecord` so that it works with `target` rather than `aggregate`. So, we’d have `target_id` and `target_id_type` fields.

Also, as we agreed, assume that a command to a `ProcessManager` would be the first field of a command message similarly to aggregate commands.

alexander-yevsyukov:
@AlexanderLitus, any update on this?

AlexanderLitus:
@alexander-yevsyukov, PTAL. I've also made some additional changes.

alexander-yevsyukov:
LGTM

PR #117, opened 2016-02-19
Introduce Values sub-project.

alexander-yevsyukov:
This PR creates a new module called `values` and separates types that used to be a part of `client` module but were not really used in there.

The purpose of the Values sub-project is to provide protobuf types and utilities for business value objects like email addresses, phone numbers, person names, bar-codes etc.

The reason for separating those types from the `client` module is that not all application will use some or even most of types provided by this sub-project.

For some near future, this code is going to be part of `core-java`, but later it would migrate into a separate sub-project with its own deployment artifacts and versioning.


values/src/main/proto/spine/base/email.proto
```diff
@@ -0,0 +1,39 @@
...
+// No validation is performed for the passed value.
+message EmailAddress {
+
+    //TODO:2016-02-20:alexander.yevsyukov: Make it structured, to reflect http://www.ietf.org/rfc/rfc822.txt
+    // A minimum it should have fields like `user`, `domain`, `personal`, `suffix` - the part coming after `+`.
+    // See also InternetAddress from JavaMail: http://docs.oracle.com/javaee/6/api/javax/mail/internet/InternetAddress.html
+
+    // String value of the address.
+    string value = 1;
+}
```
AlexanderLitus: An empty line is needed.

values/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+        resolutionStrategy.cacheChangingModulesFor 0, 'seconds'
+    }
+    dependencies {
+        classpath group: 'org.spine3.tools', name: 'proto-lookup-plugin', version: '1.1', changing: true
+    }
+}
+
+dependencies {
+    compile group: 'io.grpc', name: 'grpc-all', version: '0.13.0'
+    testCompile project(path: ":client", configuration: 'testArtifacts')
```
AlexanderLitus: Does this module depend on these modules?

values/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+                grpc {
+//                    option 'nano=true'
+                }
+            }
+        }
+    }
+}
+
+// required for usage of test classes from other modules
+configurations {
```
AlexanderLitus: It is not needed in this module (all the code after this comment).

AlexanderLitus:
Other than I've mentioned, LGTM.

values/build.gradle
alexander-yevsyukov: We depend on test utilities from `client`.

values/build.gradle
alexander-yevsyukov: We will have another round of dependency analysis after this PR is closed. We may want to have a mini-core, which would form the basis for compiler and values sub-project.

PR #121, opened 2016-03-25
Use state builder in event appliers

alexander-yevsyukov:
... instead of `getState()`, then `toBuilder(), and then `incrementState()`.


server/src/main/java/org/spine3/server/aggregate/EventApplier.java
```diff
@@ -96,7 +98,7 @@ public static boolean isEventApplier(Method method) {
             }
 
             final Class<?>[] parameterTypes = method.getParameterTypes();
-            final boolean hasOneParam = parameterTypes.length == 1;
+            final boolean hasOneParam = parameterTypes.length == NUMBER_OF_PARAMS;
```
AlexanderLitus: It would be great to rename the variable to `isParamCountValid`.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -86,23 +87,36 @@
...
  * @param <I> the type for IDs of this class of aggregates
- * @param <M> the type of the state held by the aggregate
+ * @param <S> the type of the state held by the aggregate
+ * @param <B> the type of the aggregate state builder
  * @author Mikhail Melnik
  * @author Alexander Yevsyukov
  */
-public abstract class Aggregate<I, M extends Message> extends Entity<I, M> implements CommandHandler {
+@SuppressWarnings("ClassWithTooManyMethods")
+public abstract class Aggregate<I, S extends Message, B extends Message.Builder>
```
AlexanderLitus: Can we get the builder type using reflection?

AlexanderLitus:
LGTM.

PR #122, opened 2016-03-25
Proto command messages validation

AlexanderLitus:


client/src/main/proto/spine/validation.proto
```diff
@@ -57,43 +56,173 @@ extend google.protobuf.FieldOptions {
...
+    // The default value is false.
+    bool is = 1;
 
-    // User-defined validation error message format.
+    // A user-defined validation error format message.
+    // The format parameters are the same as for the default message.
     string msg = 2;
 }
 
+// TODO:2016-03-14:alexander.litus: is this option really needed? Why not use min/max instead?
```
alexander-yevsyukov: Let's address this and other TODOs.

client/src/main/proto/spine/validation.proto
```diff
@@ -57,43 +56,173 @@ extend google.protobuf.FieldOptions {
...
+// Applicable only to numbers.
 message DecimalOption {
-    option (default_message) = "Field value must fit into range";
+    option (default_message) = "field value must fit into range";
```
alexander-yevsyukov: Have messages as sentences as we discussed. We probably need to have the range mentioned in the default message format.

client/src/main/proto/spine/validation.proto
```diff
@@ -57,43 +56,173 @@ extend google.protobuf.FieldOptions {
...
+    // The format parameters are:
+    // 1) the simple name of the field;
+    // 2) "or equal to " string (if the `inclusive` parameter is true);
+    // 3) the minimum number;
+    // 4) an actual value of the field.
+    option (default_message) = "'%s' must be greater than %s%s, actual: %s";
+
+    // The minimum value of the field.
+    // The default value is zero.
+    double is = 1;
```
alexander-yevsyukov: `value` name for the field would read better.

client/src/main/proto/spine/validation.proto
```diff
@@ -57,43 +56,173 @@ extend google.protobuf.FieldOptions {
...
+
+    // Specifies if the field can be equal to the minimum value.
+    // The default value is false.
+    bool inclusive = 2;
```
alexander-yevsyukov: Let's assume that it is always `inclusive`.

client/src/main/proto/spine/validation.proto
```diff
@@ -57,43 +56,173 @@ extend google.protobuf.FieldOptions {
...
+
+    // Set this property to true if it is needed to use only the `max` option.
+    // The default value is false.
+    bool ignore = 3;
```
alexander-yevsyukov: Assume if `max` is not specified, then there's no maximum, if the minimum value < 0.  If minimum is less than zero, than zero is the maximum, even if it's not specified.

client/src/main/proto/spine/validation.proto
```diff
@@ -103,49 +232,52 @@ enum Time {
...
-}
-
-
+// A string field value must match the given regular expression.
+// Is applicable only to strings.
+// Repeated fields are supported.
+//
+// Example:
+//
+// string field = 1 [(pattern).regex = FUTURE];
```
alexander-yevsyukov: The string `FUTURE` doesn't look like a natural regexp pattern.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -124,27 +127,37 @@ public void unregister(CommandHandler handler) {
...
      *
-     * @param command the command message to check
+     * @param commandOrMessage the command message to check
      * @return the result of {@link Responses#ok()} if the command is supported,
      *         {@link CommandValidation#unsupportedCommand(Message)} otherwise
      */
-    public Response validate(Message command) {
-        checkNotNull(command);
-        final CommandClass commandClass = CommandClass.of(command);
+    public Response validate(Message commandOrMessage) {
```
alexander-yevsyukov: Let's assume the parameter is only a `commandMessage`.

See the code of the `BoundedContext` that checks validity of the `CommandContext`. It is also a part of command validation. This code may belong to `CommandBus`, if it should be entitled to check for namespaces in a multi-tenant application.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -124,27 +127,37 @@ public void unregister(CommandHandler handler) {
...
-            return Responses.ok();
+        final Message message = toMessage(commandOrMessage);
+        final CommandClass commandClass = CommandClass.of(message);
+        if (isUnsupportedCommand(commandClass)) {
+            return CommandValidation.unsupportedCommand(message);
         }
-
-        if (handlerRegistered(commandClass)) {
-            return Responses.ok();
+        final MessageValidator validator = new MessageValidator(CommandFieldValidatorFactory.newInstance());
```
alexander-yevsyukov: This looks strange. Why do we need to pass the instance of some factory to `MessageValidator`?
Why cannot it be `MessageValidator.newInstance()`?

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -124,27 +127,37 @@ public void unregister(CommandHandler handler) {
...
+        validator.validate(message);
+        if (validator.isMessageInvalid()) {
```
alexander-yevsyukov: How about `!validator.isMessageValid()`?

server/src/main/java/org/spine3/server/entity/EntityPackagesMap.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+
+import static com.google.common.collect.Maps.newHashMap;
+import static org.spine3.io.IoUtil.loadAllProperties;
+
+/**
+ * A map from Protobuf package to {@link EntityType}.
+ * The package contains files for this type of entity (aggregate, etc).
```
alexander-yevsyukov: Please document why we need this data structure.

server/src/main/java/org/spine3/server/validate/CommandFieldValidatorFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Creates {@link FieldValidator}s for commands.
+ *
+ * @author Alexander Litus
+ */
+@Internal
+@SuppressWarnings("RefusedBequest")
+public class CommandFieldValidatorFactory extends FieldValidatorFactory {
+
```
alexander-yevsyukov: Please remove excessive empty line.

server/src/main/java/org/spine3/server/validate/CommandNumberFieldValidator.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param fieldValues field values to validate
+     */
+    /* package */ CommandNumberFieldValidator(FieldDescriptor descriptor, ImmutableList<Number> fieldValues) {
+        super(descriptor, fieldValues);
+    }
+
+    @Override
+    protected void validate() {
+        super.validate();
+        // TODO:2016-03-18:alexander.litus: check type (long or int)
```
alexander-yevsyukov: Resolved?

client/src/main/proto/spine/validation.proto
```diff
@@ -22,11 +22,11 @@ syntax = "proto3";
 // We do not define the package for this file to allow shorter options for user-defined types.
 // This would allow to write...
 //
-//     [(required).value = true];
+//     [(required).is = true];
```
alexander-yevsyukov: `required value` reads better than `required is`.

server/src/main/java/org/spine3/server/validate/FieldValidator.java
```diff
@@ -0,0 +1,199 @@
...
+     * @return {@code true} if the field is not set, {@code false} otherwise
+     */
+    protected boolean isValueNotSet(V value) {
+        throw new UnsupportedOperationException("The method is not implemented.");
+    }
+
+    /**
+     * Returns {@code true} if the validated field is invalid, {@code false} otherwise.
+     */
+    protected boolean isFieldInvalid() {
```
alexander-yevsyukov: Why don't have it `isFieldValid` or just `isValid`? It's the field validator after all.

server/src/main/java/org/spine3/server/command/CommandBus.java
AlexanderLitus: But we have an inspection `Boolean method is always inverted`. And client code needs to know if the message is invalid. So I added `isValid` method and inverted the variable.

AlexanderLitus:
@alexander-yevsyukov, PTAL. Remained: validator API changes (ConstraintViolations) and gradle plugin which creates `entities.properties`.

alexander-yevsyukov:
Please make build check pass. Please also address TODOs in introduced in this PR.

client/src/main/proto/spine/base/failure.proto
```diff
@@ -49,4 +50,11 @@ message Failure {
 
     // Additional information on the failure.
     map<string, google.protobuf.Value> attributes = 4;
-}
\ No newline at end of file
+}
+
+// A failure indicating that a message is invalid according to Spine custom Protobuf options.
```
alexander-yevsyukov: We should not mention the implementation details in this part of the documentation. I would also say that “... a message did not pass validation” or something like that.

server/src/main/java/org/spine3/server/command/CommandValidation.java
```diff
@@ -79,10 +79,10 @@ public static Response invalidCommand(Message command, String errorMessage) {
...
         final Response response = Response.newBuilder()
                 .setError(Error.newBuilder()
-                .setType(CommandValidationError.getDescriptor().getFullName())
-                .setCode(CommandValidationError.INVALID_COMMAND.getNumber())
-                .putAllAttributes(commandTypeAttribute(commandType))
-                .setMessage(errorMessage)).build();
+                    .setType(CommandValidationError.getDescriptor().getFullName())
+                    .setCode(CommandValidationError.INVALID_COMMAND.getNumber())
+                    .putAllAttributes(commandTypeAttribute(commandType))
+                    .setMessage(errorMessage)).build();
```
alexander-yevsyukov: Please have `.build()` on new line.

server/src/test/java/org/spine3/server/validate/MessageValidatorShould.java
```diff
@@ -565,21 +582,9 @@ private void digitsCountTest(double value, boolean isValid) {
...
-    private static void assertMessageIsValid(boolean isValid, MessageValidator validator) {
-        if (isValid) {
-            assertFalse(validator.isMessageInvalid());
-            assertTrue(validator.getErrorMessage().isEmpty());
-        } else {
-            assertTrue(validator.isMessageInvalid());
-            assertFalse(validator.getErrorMessage().isEmpty());
-        }
+        assertEquals(isValid, validator.isMessageValid());
+        assertEquals(!isValid, validator.isMessageInvalid());
```
alexander-yevsyukov: I don't think we need two methods `isMessaseValid()` should be enough.

alexander-yevsyukov: As long the class is named `MessageValidator`, we can have just `isValid()` to make it shorter.

client/src/main/proto/spine/base/field_path.proto
```diff
@@ -0,0 +1,35 @@
...
+
+package spine.base;
+
+option java_package = "org.spine3.base";
+option java_outer_classname = "FieldPathProto";
+option java_multiple_files = true;
+option java_generate_equals_and_hash = true;
+
+
+// A field path, which consists of simple message field names, e.g. ["field1", "field2", ...].
```
alexander-yevsyukov: Have a look at description of `FieldMask` for ideas on describing the field patch concept. Give an example of nested data structure, which may contain a fictitious data structures `CustomerAccount`, which contains `user` record of type `User`, which in turn has `first_name` field of type `string`. In this case, the path would be `user.first_name`. Being structurally represented in `FieldPath` it would have two entries.

Just having `field1` and `field2` in the example makes it hard to understand.

client/src/main/proto/spine/validation.proto
```diff
@@ -47,15 +49,33 @@ extend google.protobuf.MessageOptions {
...
     MinOption min = 51007;
+    DigitsOption digits = 51004;
+    TimeOption when = 51005;
     PatternOption pattern = 51008;
     ValidOption valid = 51009;
 }
 
+// A validation constraint violation for a message field.
+message ConstraintViolation {
+    // An interpolated error message for this constraint violation.
```
alexander-yevsyukov: What does `interpolation` mean?

client/src/main/proto/spine/validation.proto
```diff
@@ -47,15 +49,33 @@ extend google.protobuf.MessageOptions {
...
+    string message = 1;
+
+    // A non-interpolated error message for this constraint violation.
+    string message_template = 2;
```
alexander-yevsyukov: How do we use it? We don't seen to have all the required fields to reconstruct the interpolated message ourself (at the client code).

alexander-yevsyukov:
What about localizing validation error messages? How do we do it?

client/src/main/proto/spine/validation.proto
AlexanderLitus: Just copy-pasted this from Java Bean Validation docs.
In this context:
Interpolation - (in philology) modification of the original text; an insertion of words or phrases that did not appear in the original text by an interpreter.
I'll use a simpler term, though.

server/src/main/java/org/spine3/server/validate/FloatFieldValidator.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    protected Float getAbs(Float value) {
+        final Float abs = abs(value);
+        return abs;
+    }
+
+    @Override
+    protected Any wrapToMsgAndAny(Float value) {
```
alexander-yevsyukov: A better name would be simply `wrap()` with proper documentation. 

It is assumed that plain Java type cannot be put into `Any` without being converted into proper Protobuf instance first. In that sense, we don't wrap the passed value into the message, but convert.

server/src/main/java/org/spine3/server/validate/FieldValidator.java
```diff
@@ -66,14 +65,13 @@
     protected FieldValidator(FieldDescriptor descriptor, ImmutableList<V> values) {
         this.fieldDescriptor = descriptor;
         this.values = values;
-        this.fieldName = fieldDescriptor.getName();
         final FileDescriptor file = fieldDescriptor.getFile();
         this.isFileBelongsToEntity = belongsToEntity(file);
         this.isCommandsFile = isCommandsFile(file);
         this.isFirstField = fieldDescriptor.getIndex() == 0;
         // TODO:2016-03-30:alexander.litus: add full field path (if this field is nested)
```
alexander-yevsyukov: Resolved?

client/src/main/proto/spine/base/command.proto
```diff
@@ -87,11 +87,8 @@ enum CommandValidationError {
     // The command type is not supported by the server.
     UNSUPPORTED_COMMAND = 1;
 
-    // The command is supported but has invalid fields (e.g., marked as "required" but not set).
-    INVALID_COMMAND = 2;
-
```
alexander-yevsyukov: Why do we remove this case?

client/src/main/java/org/spine3/protobuf/Messages.java
```diff
@@ -209,4 +213,55 @@ public static String toJson(Message message) {
         }
     }
 
+    /**
+     * Creates a new StringValue wrapping the passed string.
+     *
+     * @param value the value to wrap
+     * @return a new StringValue instance
+     */
+    public static StringValue newStringValue(String value) {
```
alexander-yevsyukov: Please put these factory methods into a utility class called `Values`.

client/src/main/proto/spine/validation.proto
```diff
@@ -61,19 +61,23 @@ extend google.protobuf.FieldOptions {
 
 // A validation constraint violation for a message field.
 message ConstraintViolation {
-    // An interpolated error message for this constraint violation.
+    // An error message for this constraint violation
+    // (or a template which contains `%s` format specifiers).
```
alexander-yevsyukov: If it contains format placeholders, then it's not a `message`, it's `message_format`, or, better, `format`.

client/src/main/proto/spine/validation.proto
```diff
@@ -61,19 +61,23 @@ extend google.protobuf.FieldOptions {
...
     string message = 1;
 
-    // A non-interpolated error message for this constraint violation.
-    string message_template = 2;
+    // Parameters (as strings) referenced by the format specifiers in the error message format string.
+    repeated string format_param = 2;
```
alexander-yevsyukov: Name it just `param`. Having `format` for the string part of the story would be enough.

client/src/main/proto/spine/validation.proto
```diff
@@ -61,19 +61,23 @@ extend google.protobuf.FieldOptions {
...
 // A validation constraint violation for a message field.
 message ConstraintViolation {
-    // An interpolated error message for this constraint violation.
+    // An error message for this constraint violation
+    // (or a template which contains `%s` format specifiers).
     string message = 1;
 
-    // A non-interpolated error message for this constraint violation.
-    string message_template = 2;
+    // Parameters (as strings) referenced by the format specifiers in the error message format string.
```
alexander-yevsyukov: Why do we have `string` rather than `Value`?

client/src/main/proto/spine/validation.proto
```diff
@@ -112,11 +114,9 @@ message RequiredOption {
 message DecimalMinOption {
     // The default error message format string.
     // The format parameters are:
-    // 1) the simple name of the field;
-    // 2) "or equal to " string (if the `inclusive` parameter is true);
-    // 3) the minimum number;
-    // 4) the actual value of the field.
-    option (default_message) = "'%s' must be greater than %s%s, actual: %s";
+    // 1) "or equal to " string (if the `inclusive` parameter is true);
```
alexander-yevsyukov: What happens if it's false? Is it an empty string then?

client/src/main/proto/spine/validation.proto
AlexanderLitus: 1. `Value` doesn't have `long` kind field, only `double`. And the `long` value can be rounded (if it is greater than max double value). 
2. It is more convenient to use `string` params for message formatting. I think they won`t be used in other ways. Also, we pass the min/max value to the options as strings, too (to support both `long` and `double` types).

client/src/main/proto/spine/base/command.proto
AlexanderLitus: OK, I'll return `INCOMPLETE_COMMAND` status (as it is a technical error).

server/src/main/java/org/spine3/server/validate/FieldValidator.java
AlexanderLitus: Yep.

AlexanderLitus:
@alexander-yevsyukov, PTAL. It is remained to implement the Gradle plugin for generating entities.props. And the build will pass then.
Also, should I write to Google support to obtain globally unique field number(s) for our options now? We need several numbers.

AlexanderLitus:
@alexander-yevsyukov, PTAL. I've requested globally unique field numbers for our custom options, it is remained only to add them.

alexander-yevsyukov:
LGTM

PR #123, opened 2016-04-07
Schedule command delivering

AlexanderLitus:


AlexanderLitus:
@alexander-yevsyukov, PTAL.

client/src/main/proto/spine/base/command.proto
```diff
@@ -77,6 +78,14 @@ message CommandContext {
 
     // The `namespace` attribute must be defined for commands in multitenant applications.
     Namespace namespace = 5;
+
+    oneof time_opts {
+        // The delay to wait before sending the command.
```
alexander-yevsyukov: The delay after which moment?

client/src/main/proto/spine/base/command.proto
```diff
@@ -77,6 +78,14 @@ message CommandContext {
...
+        google.protobuf.Duration delay = 6;
+
+        // The time when the command should be sent.
```
alexander-yevsyukov: Sent, or executed?

client/src/main/proto/spine/base/command.proto
```diff
@@ -77,6 +78,14 @@ message CommandContext {
 
     // The `namespace` attribute must be defined for commands in multitenant applications.
     Namespace namespace = 5;
+
+    oneof time_opts {
```
alexander-yevsyukov: Isn't it `scheduling_options`? Or, `schedule_by`, or just `schedule`?

How do they name such things on existing messaging systems?

server/src/main/java/org/spine3/server/command/DefaultCommandScheduler.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public DefaultCommandScheduler(CommandBus commandBus) {
+        this.commandBus = commandBus;
+    }
+
+    @Override
+    public void schedule(Command command) {
+        checkState(!isShutdown, "Scheduler is shut down.");
+        final long delaySec = getDelaySeconds(command);
+        final Command commandUpdated = removeSchedulingOptions(command);
```
alexander-yevsyukov: Why do we modify the command?

AlexanderLitus: We remove scheduling options so that when we post it again, it is not considered as scheduled.

client/src/main/proto/spine/base/command.proto
AlexanderLitus: After the current time.

AlexanderLitus:
@alexander-yevsyukov, PTAL.

client/src/main/proto/spine/base/command.proto
```diff
@@ -79,13 +79,23 @@ message CommandContext {
...
+}
+
+message Schedule {
+    oneof options {
+        // The amount of time between the current moment and a command delivering to the target.
+        google.protobuf.Duration delay = 1;
 
-        // The time when the command should be sent.
-        google.protobuf.Timestamp sending_time = 7;
+        // The time when the command should be delivered to the target.
```
alexander-yevsyukov: The delivery time depends on the path a command goes after it is posted to `CommandBus`. There could be one or more dispatchers in between, for example. The load time for an aggregate may be unpredictable, etc.

So, in general, we won't be able to guarantee the delivery by the time specified. It would be too complex. What we can promise is that we'd post the command to the `CommandBus` shortly after the time specified.

client/src/main/proto/spine/base/command.proto
```diff
@@ -79,13 +79,23 @@ message CommandContext {
...
+        google.protobuf.Timestamp delivery_time = 2;
     }
+
+    // `true` if the command should be delivered to the target right now, despite the scheduling options set;
+    // `false` otherwise.
+    // This flag is needed to specify that a scheduled command arrived in time, and to retain the scheduling options.
+    bool in_time = 3;
```
alexander-yevsyukov: A command may be several nanoseconds late or even earlier. We need to name the flag differently.

client/src/main/proto/spine/base/command.proto
```diff
@@ -79,13 +79,23 @@ message CommandContext {
...
-    oneof time_opts {
-        // The delay to wait before sending the command.
-        google.protobuf.Duration delay = 6;
+    // The command scheduling options.
+    Schedule schedule = 6;
+}
+
+message Schedule {
+    oneof options {
+        // The amount of time between the current moment and a command delivering to the target.
```
alexander-yevsyukov: `the current moment` isn't clear. Is this a moment at the client, or the moment of receiving the command at the server?

client/src/main/proto/spine/base/command.proto
```diff
@@ -79,13 +79,23 @@ message CommandContext {
...
 
-    oneof time_opts {
-        // The delay to wait before sending the command.
-        google.protobuf.Duration delay = 6;
+    // The command scheduling options.
+    Schedule schedule = 6;
+}
+
+message Schedule {
+    oneof options {
```
alexander-yevsyukov: Since there can be only one option, I'd use it as singular: `option`.

client/src/main/proto/spine/base/command.proto
alexander-yevsyukov: How about just `after`?

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -224,6 +227,15 @@ public Response validate(Message command) {
         return result;
     }
 
+    private void schedule(Command command) {
+        if (scheduler != null) {
+            scheduler.schedule(command);
+        } else {
+            throw new IllegalStateException(
+                    "Scheduled commands are not supported by this command bus (scheduler is not set).");
```
alexander-yevsyukov: Consider having two sentences in the message, instead of having the nested one in braces.

client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -192,28 +193,59 @@ public static String formatMessageTypeAndId(String format, Message commandMessag
...
+     * Checks if the file is for commands.
+     *
+     * @param file a descriptor of a {@code .proto} file to check
+     * @return {@code true} if the file name contains {@link #COMMANDS_FILE_SUBSTRING} substring, {@code false} otherwise
+     */
+    public static boolean isCommandsFile(FileDescriptor file) {
+        final String fqn = file.getName();
+        final int startIndexOfFileName = fqn.lastIndexOf(PROTO_FILE_SEPARATOR) + 1;
+        final String fileName = fqn.substring(startIndexOfFileName);
+        final boolean isCommandsFile = fileName.contains(COMMANDS_FILE_SUBSTRING);
```
alexander-yevsyukov: Can we name the constant `COMMANDS_FILE_SUFFIX` instead of `...SUBSTRING`?

client/src/main/proto/spine/base/command.proto
AlexanderLitus: OK, so I'm removing `delivery_time` for now.

client/src/main/java/org/spine3/base/Commands.java
AlexanderLitus: But it can be a prefix also. I`ll rename it to just `COMMANDS`.

AlexanderLitus:
@alexander-yevsyukov, PTAL.

client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -53,7 +52,7 @@
     /**
      * A substring which the {@code .proto} file containing commands must have in its name.
      */
-    public static final String COMMANDS_FILE_SUBSTRING = "commands";
+    public static final String COMMANDS = "commands";
```
alexander-yevsyukov: “commands” what? From the API point of view it looks this `Commands.COMMANDS`?

Why don't we have it as `Commands.FILE_SUFFIX`, for example?

client/src/main/proto/spine/base/command.proto
```diff
@@ -84,18 +84,13 @@ message CommandContext {
...
-        google.protobuf.Timestamp delivery_time = 2;
-    }
-
-    // `true` if the command should be delivered to the target right now, despite the scheduling options set;
+    // `true` if the command should be delivered to the target right now, despite the delay set;
     // `false` otherwise.
-    // This flag is needed to specify that a scheduled command arrived in time, and to retain the scheduling options.
-    bool in_time = 3;
+    // This flag is needed to specify that the delay is passed, and to retain the info about the delay.
+    bool ignore_delay = 2;
```
alexander-yevsyukov: This name is still confusing. Why cannot we have a set of already scheduled command IDs? If a command posted is already scheduled, it would mean that we need to execute it.

client/src/main/java/org/spine3/base/Commands.java
AlexanderLitus: If we name it `suffix` we have to check the file name with `endsWith()` method. But now it is checked via `contains()` method, for more flexibility (maybe, a client wants to name the file `commands_order` instead of `order_commands` or just `commands`).

alexander-yevsyukov: Suffix is flexible enough.

AlexanderLitus:
PTAL.

server/src/test/java/org/spine3/server/command/ExecutorCommandSchedulerShould.java
```diff
@@ -69,36 +72,29 @@ public void tearDownTest() {
...
-    public void schedule_command_if_delay_is_set() throws InterruptedException {
+    public void schedule_command_if_delay_is_set() {
         final CommandContext context = createCommandContext(DELAY);
         final Command cmd = Commands.create(createProject(newUuid()), context);
 
         scheduler.schedule(cmd);
 
-        assertCommandSent(cmd, DELAY);
+        assertNull(postedCommand);
+        sleep(DELAY_MS + CHECK_OFFSET_MS);
```
alexander-yevsyukov: Please test it using `spy()` or `mock()`. Sleeping threads does not scale well for testing.

AlexanderLitus:
PTAL.

client/src/main/java/org/spine3/protobuf/Durations.java
```diff
@@ -19,7 +19,7 @@
 
 /**
  * Utility class for working with durations in addition to those available from {@link TimeUtil}.
- * 
+ * <p/>
```
alexander-yevsyukov: Please remove the excessive paragraph tag.

alexander-yevsyukov: Can we make IDEA stop adding these nasty tags?

client/src/main/java/org/spine3/protobuf/Durations.java
```diff
@@ -42,56 +42,83 @@ private Durations() {}
...
      * @return a non-null {@code Duration}
      */
     public static Duration ofHours(long hours) {
-        return ofMinutes(safeMultiply(hours, MINUTES_PER_HOUR));
+        final Duration duration = ofMinutes(safeMultiply(hours, MINUTES_PER_HOUR));
+        return duration;
     }
 
+    // Methods for brief computations with Durations like
+    //       add(hours(2), minutes(30));
```
alexander-yevsyukov: Please add some kind of section separator so that this comment is more visible.

server/build.gradle
```diff
@@ -36,9 +36,10 @@ protobuf {
     generateProtoTasks {
         all().each { final task ->
             task.plugins {
-                grpc {
-//                    option 'nano=true'
-                }
+                grpc {}
```
alexander-yevsyukov: Do we need the empty `grpc` block?

mikhailmikhaylov: We do. Adding even empty closure in this section adds protoc option --%CLOSURE_NAME%_out. In our case, this empty block is a signal to generate grpc services

server/src/main/java/org/spine3/server/reflect/MethodMap.java
```diff
@@ -37,20 +38,64 @@
...
+                    throw new DuplicateHandlerMethodException(
+                            declaringClass,
+                            messageClass,
+                            alreadyPresent.getName(),
+                            method.getName());
+                }
+                tempMap.put(messageClass, method);
+            }
+        }
+        final ImmutableMap.Builder<Class<? extends Message>, Method> builder = ImmutableMap.builder();
```
alexander-yevsyukov: Could you refactor so that we don't use the temp map and add to the builder from the very beginning?

values/build.gradle
```diff
@@ -36,9 +36,10 @@ protobuf {
     generateProtoTasks {
         all().each { final task ->
             task.plugins {
-                grpc {
-//                    option 'nano=true'
-                }
+                grpc {}
```
alexander-yevsyukov: Do we need the empty `grpc` section here?

mikhailmikhaylov: No, we don't

server/build.gradle
AlexanderLitus: Yep.

server/src/main/java/org/spine3/server/reflect/MethodMap.java
AlexanderLitus: No, builder does't have `containsKey` method.

On Tue, Apr 19, 2016 at 6:27 PM, Alexander Yevsyukov <
notifications@github.com> wrote:

> In server/src/main/java/org/spine3/server/reflect/MethodMap.java
> <https://github.com/SpineEventEngine/core-java/pull/123#discussion_r60252891>
> :
>
> > +        final Map<Class<? extends Message>, Method> tempMap = Maps.newHashMap();
> > +        for (Method method : declaringClass.getDeclaredMethods()) {
> > +            if (filter.apply(method)) {
> > +                final Class<? extends Message> messageClass = HandlerMethod.getFirstParamType(method);
> > +                if (tempMap.containsKey(messageClass)) {
> > +                    final Method alreadyPresent = tempMap.get(messageClass);
> > +                    throw new DuplicateHandlerMethodException(
> > +                            declaringClass,
> > +                            messageClass,
> > +                            alreadyPresent.getName(),
> > +                            method.getName());
> > +                }
> > +                tempMap.put(messageClass, method);
> > +            }
> > +        }
> > +        final ImmutableMap.Builder<Class<? extends Message>, Method> builder = ImmutableMap.builder();
>
> Could you refactor so that we don't use the temp map and add to the
> builder from the very beginning?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/SpineEventEngine/core-java/pull/123/files/754a16114faf922e17be25e9599d05a3ecb18cab..5be63438f17d662556292d161ffb9deac30697ee#r60252891>
>



-- 
Best,
Alexander Litus

Software Engineer | TeamDev, Ltd.
alexander.litus@teamdev.com  |  www.teamdev.com

AlexanderLitus:
PTAL.

values/build.gradle
AlexanderLitus: Killed.

alexander-yevsyukov:
Please use `times(1)` instead of `atMost(1)` in tests. The change seems to be lost in recent merges.

alexander-yevsyukov:
Other than that, LGTM

PR #130, opened 2016-04-22
Make context param optional in event and command handlers

AlexanderLitus:


alexander-yevsyukov:
@AlexanderLitus, please resolve conflicts.

alexander-yevsyukov:
@eugeniakotlyar, we need a documentation update that reflect the optional nature of context parameters.

server/src/main/java/org/spine3/server/aggregate/EventApplier.java
```diff
@@ -101,8 +102,8 @@ public EventApplier create(Method method) {
 
         private enum Singleton {
             INSTANCE;
-            @SuppressWarnings("NonSerializableFieldInSerializableClass")
-            private final Factory value = new Factory();
+            @SuppressWarnings({"NonSerializableFieldInSerializableClass", "UnnecessarilyQualifiedInnerClassAccess"})
```
alexander-yevsyukov: Please demote this warning to Info level.

server/src/main/java/org/spine3/server/event/Subscribe.java
```diff
@@ -28,12 +31,17 @@
 /**
  * Marks a method as an event subscriber.
  *
- * <p>An event subscriber method must have two parameters. The class of the event will be
- * indicated by the first parameter.
+ * <p>An event subscriber/handler method:
```
alexander-yevsyukov: We need to standardize on terms. It's either `subscriber` or `handler`. If we don't do this, there would be a question: "What is the difference between subscriber and handler?"

alexander-yevsyukov: Let's standardize on `subscriber`. We ‘handle’ commands, but ‘subscribe’ to events.

alexander-yevsyukov: Please rename `EventHandler` to `EventSubscriber`. And `EventHandlerMethod` to `EventSubscriberMethod`.

server/src/main/java/org/spine3/server/reflect/EventHandlerMethod.java
```diff
@@ -117,25 +111,24 @@ public EventHandlerMethod create(Method method) {
 
         private enum Singleton {
             INSTANCE;
-            @SuppressWarnings("NonSerializableFieldInSerializableClass")
-            private final Factory value = new Factory();
+            @SuppressWarnings({"NonSerializableFieldInSerializableClass", "UnnecessarilyQualifiedInnerClassAccess"})
```
alexander-yevsyukov: Please demote this warning to Info level and remove suppression here and in other places.

alexander-yevsyukov: Why do you add the qualifier here, BTW?

server/src/main/java/org/spine3/server/reflect/EventHandlerMethod.java
```diff
@@ -152,15 +145,19 @@ private static boolean isEventHandler(Method method) {
         }
 
         private static boolean acceptsCorrectParams(Method method) {
-            final Class<?>[] parameterTypes = method.getParameterTypes();
-
-            if (parameterTypes.length != EVENT_HANDLER_PARAM_COUNT) {
+            final Class<?>[] paramTypes = method.getParameterTypes();
+            final int paramCount = paramTypes.length;
+            if (paramCount == 1) {
+                final boolean isCorrect = Message.class.isAssignableFrom(paramTypes[MESSAGE_PARAM_INDEX]);
```
alexander-yevsyukov: Extract first param type verification into a variable called `firstParamIsMessage`. You have this check in two branches of the `if`.

server/src/main/java/org/spine3/server/reflect/HandlerMethod.java
```diff
@@ -41,9 +45,14 @@
  *
  * @author Mikhail Melnik
  * @author Alexander Yevsyukov
- * @param <C> the type of the message context or {@code Void} if context is not used
+ * @param <C> the type of the message context or {@link Empty} if context is not used
```
alexander-yevsyukov: Please change the wording `if context` into `if a context parameter is never used`. It would stress on the fact that the parameter may or may not be used, but there are cases where there is only one parameter only.

server/src/main/java/org/spine3/server/reflect/HandlerMethod.java
```diff
@@ -121,13 +129,20 @@ protected int getParamCount() {
...
+    public <R> R invoke(Object target, Message message, C context) throws InvocationTargetException {
         checkNotNull(message);
         checkNotNull(context);
         try {
-            @SuppressWarnings("unchecked")
-            final R result = (R) method.invoke(target, message, context);
-            return result;
+            final int paramCount = getParamCount();
+            if (paramCount == 1) {
+                @SuppressWarnings("unchecked")
```
alexander-yevsyukov: Please document why we can suppress the warning.

server/src/main/java/org/spine3/server/reflect/HandlerMethod.java
```diff
@@ -121,13 +129,20 @@ protected int getParamCount() {
...
+                final R result = (R) method.invoke(target, message);
+                return result;
+            } else {
+                @SuppressWarnings("unchecked")
```
alexander-yevsyukov: Please document why we can suppress the warning.

server/src/main/java/org/spine3/server/reflect/HandlerMethod.java
```diff
@@ -150,19 +165,18 @@ protected int getParamCount() {
      *                                   an {@link Error}. {@code Error} instances are propagated as-is.
      */
     protected <R> R invoke(Object target, Message message) throws InvocationTargetException {
-        checkNotNull(message);
+        return invoke(target, message, getDefaultContextInstance());
```
alexander-yevsyukov: Can we avoid passing the default instance?

server/src/main/java/org/spine3/server/reflect/HandlerMethod.java
```diff
@@ -150,19 +165,18 @@ protected int getParamCount() {
...
+    }
+
+    private C getDefaultContextInstance() {
```
alexander-yevsyukov: If we cannot avoid having this method, we need to cache its result per context class. We need to calculate it only once per class.

testutil/src/main/java/org/spine3/test/Verify.java
```diff
@@ -36,7 +36,7 @@
  *     Verify.assertContains("fred", nameList);  // from new extensions
  *     Verify.assertBefore("fred", "jim", orderedNamesList);  // from new extensions
  * </pre>
- * <p/>
+ *
```
alexander-yevsyukov: Don't we need `<p>` prefix before the following paragraph?

testutil/src/main/java/org/spine3/test/Verify.java
```diff
@@ -61,7 +61,7 @@ private Verify() {}
     /**
      * Mangles the stack trace of {@link AssertionError} so that it looks like its been thrown from the line that
      * called to a custom assertion.
-     * <p/>
+     *
```
alexander-yevsyukov: Don't we need `<p>` prefix before the following paragraph?

server/src/main/java/org/spine3/server/reflect/EventHandlerMethod.java
AlexanderLitus: IDEA complains about that (shows compilation errors though Gradle build passes). I think this is because the base factory class name is the same. 
Also, should I demote `UnnecessarilyQualifiedInnerClassAccess` warning only, or `NonSerializableFieldInSerializableClass` also?

server/src/main/java/org/spine3/server/event/Subscribe.java
AlexanderLitus: But `EventHandlerMethod` extends `HandlerMethod`. Would it be OK if `EventSubscriberMethod` extends `HandlerMethod`?

server/src/main/java/org/spine3/server/reflect/EventHandlerMethod.java
AlexanderLitus: Yes, this way we can avoid one more check. After extracting, I have to check param count one more time before.

server/src/main/java/org/spine3/server/event/Subscribe.java
alexander-yevsyukov: Yes, it would be OK because Event Subscriber is a special kind of Handler Method.

AlexanderLitus:
PTAL.

alexander-yevsyukov:
LGTM

AlexanderLitus:
PTAL: added more tests and some minor changes.

AlexanderLitus:
@mikhailmikhaylov, please review 5 last commits.

mikhailmikhaylov:
@AlexanderLitus LGTM

PR #133, opened 2016-04-27
Support integration events

AlexanderLitus:


server/src/main/proto/spine/server/integration/integration_event.proto
```diff
@@ -0,0 +1,48 @@
...
+    // The context of the event.
+    spine.base.EventContext context = 2;
+
+    // The name of the bounded context, which generated the event.
+    string source = 3;
+}
+
+service IntegrationEventSubscriber {
+
+    rpc Notify(IntegrationEvent) returns (spine.base.Response);
```
AlexanderLitus: As I understand, BoundedContext should implement this service interface. Why `notify` and not `post`?

alexander-yevsyukov: As discussed, the nature of the call to Subscriber is different.

AlexanderLitus:
PTAL.

client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -70,17 +71,21 @@ public static CommandId generateId() {
...
-     * Creates new command context with the current time
+     * Creates a new command context with the current time.
+     *
      * @param userId the actor id
      * @param zoneOffset the offset of the timezone in which the user works
+     * @param boundedContextName the name of the current context
+     * @see CommandFactory
      */
-    public static CommandContext createContext(UserId userId, ZoneOffset zoneOffset) {
+    public static CommandContext createContext(UserId userId, ZoneOffset zoneOffset, String boundedContextName) {
```
alexander-yevsyukov: The last parameter will be used _only_ for integration events. We should not modify the core API for such a minority of cases. 

It would be better to modify the set of fields of `IntegrationEvent` eliminating the `context` field, which has type `EventContext`. It would solve the problem of necessity of having a Bounded Context name.

client/src/main/java/org/spine3/client/CommandFactory.java
```diff
@@ -37,12 +38,13 @@
 public class CommandFactory {
 
     private final UserId actor;
-
     private final ZoneOffset zoneOffset;
+    private final String boundedContextName;
 
-    protected CommandFactory(UserId actor, ZoneOffset zoneOffset) {
+    protected CommandFactory(UserId actor, ZoneOffset zoneOffset, String boundedContextName) {
```
alexander-yevsyukov: Again, we need to avoid this parameter. 

Even if we need info about a  `BoundedContext`, we should pass its instance and have it as a first parameter because it has more “weight” than a user — it doesn't change that often, while users and zones do.

client/src/main/proto/spine/base/event.proto
```diff
@@ -76,8 +76,13 @@ message EventContext {
     // The version of the entity after the event was applied.
     int32 version = 5;
 
+    // An identifier of a software component, which generated the event.
+    // If the source is a Bounded Context, this field is its name.
+    // It can be a name of another BC for integration event.
+    string source = 6;
```
alexander-yevsyukov: We have this information already because we aggregate `CommandContext`.

client/src/main/proto/spine/base/command.proto
```diff
@@ -79,8 +79,12 @@ message CommandContext {
     // The `namespace` attribute must be defined for commands in multitenant applications.
     Namespace namespace = 5;
 
+    // An identifier of a software component, which generated the command.
+    // If the source is a Bounded Context, this field is its name.
+    string source = 6;
```
alexander-yevsyukov: This information may be useful, but it would be better to have it as a client environment definition, which may contain: client type, OS, code version etc.

alexander-yevsyukov: We have this info in `client.proto`, but I'm not sure we need to add it here.

server/src/main/java/org/spine3/server/integration/IntegrationEventBus.java
```diff
@@ -20,39 +20,136 @@
...
+    public static IntegrationEventBus getInstance() {
+        return instance();
     }
 
-    public void post(IntegrationEvent event) {
-        //TODO:2016-04-27:alexander.yevsyukov: Implement
+    /**
+     * Sets a response observer used in {@link IntegrationEventSubscriber#notify(Event, StreamObserver)}.
+     */
+    public void setResponseObserver(StreamObserver<Response> responseObserver) {
```
alexander-yevsyukov: The response observer is passed with the call. We don't set it. See gRPC examples.

alexander-yevsyukov: Why do we need to set the observer from outside?

AlexanderLitus: A client may want to pass his observer to do smth on event delivery or error.

AlexanderLitus:
We discussed that we convert IntegrationEvents to Events. If we are doing so, why not just post them to local event bus? I think it would be easier for a client to define regular event handler methods for integration events.

alexander-yevsyukov:
> just post them to local event bus?

We may post event messages as they are, but we don't have information in `EventContext`.

server/src/main/java/org/spine3/server/BoundedContext.java
```diff
@@ -227,25 +231,56 @@ private Response validateCommand(Command command) {
...
+        return response;
+    }
+
+    private static Event toEvent(IntegrationEvent integrationEvent) {
+        final IntegrationEventContext sourceContext = integrationEvent.getContext();
+        final EventContext context = EventContext.newBuilder()
+                .setEventId(sourceContext.getEventId())
+                .setTimestamp(sourceContext.getTimestamp())
+                .setProducerId(sourceContext.getProducerId())
+                .setBoundedContextName(sourceContext.getBoundedContextName())
```
alexander-yevsyukov: I think the `producerId` is the name of the bounded context, which sent us the event. And we should not have the `boundedContextName` attribute for the regular events at all. *All of them* (apart from integration events) are going to be within this bounded context.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -130,19 +130,15 @@ public void unregister(CommandHandler handler) {
     /**
      * Verifies if the command can be posted to this {@code CommandBus}.
```
alexander-yevsyukov: I think we need to say that the method verifies a command message, not a command.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -130,19 +130,15 @@ public void unregister(CommandHandler handler) {
...
      *
-     * <p>The command can be posted if it has either dispatcher or handler registered with
+     * <p>A command can be posted if it has either dispatcher or handler registered with
```
alexander-yevsyukov: We also need to mention validation prior to describing returned results. Something like:

----
The command message must also satisfy validation constraints defined in its protobuf type.
----

server/src/main/java/org/spine3/server/event/DispatcherRegistry.java
```diff
@@ -66,6 +66,17 @@
         dispatchers.clear();
     }
 
+    /* package */ boolean hasDispatchersFor(EventClass eventClass) {
+        final Set<EventDispatcher> dispatchers = getDispatchers(eventClass);
+        final boolean result = !dispatchers.isEmpty();
+        return result;
+    }
+
+    /* package */ boolean noDispatchersFor(EventClass eventClass) {
```
alexander-yevsyukov: Do we really have many occasions for using this method?

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -130,19 +130,15 @@ public void unregister(CommandHandler handler) {
...
      * the command bus.
```
alexander-yevsyukov: Please change `the command bus` to `this {@code CommandBus}. It would match the syntax of a similar method in `EventBus`.

AlexanderLitus:
> We may post event messages as they are, but we don't have information in EventContext because there is no command.
Yes, when converting `IntegrationEvent
              

.setEventId(sourceContext.getEventId())
                .setTimestamp(sourceContext.getTimestamp())
                .setProducerId(sourceContext.getProducerId())
                .setBoundedContextName(sourceContext.getBoundedContextName())

server/src/main/java/org/spine3/server/event/DispatcherRegistry.java
AlexanderLitus: Only one for now. I'll remove it and change its usage so that it reads better.

AlexanderLitus:
PTAL.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -128,10 +128,10 @@ public void unregister(CommandHandler handler) {
     }
 
     /**
-     * Verifies if the command can be posted to this {@code CommandBus}.
+     * Verifies that the command message can be posted to this {@code CommandBus}.
      *
-     * <p>A command can be posted if it has either dispatcher or handler registered with
-     * the command bus.
+     * <p>A command message can be posted if it has either dispatcher or handler registered with
```
alexander-yevsyukov: We do not post message, we post a whole command. We associate dispatchers or messages with types of command messages.

client/src/test/java/org/spine3/base/EventsShould.java
```diff
@@ -63,4 +83,74 @@ public void return_actor_from_EventContext() {
         // the main reason to have this test is to mark the Events.getActor() as `used`.
```
alexander-yevsyukov: Let's have a good test with some test data and check.

alexander-yevsyukov:
Other than latest minor comments, LGTM.

PR #136, opened 2016-05-03
Projection Repository catch-up  support

alexander-yevsyukov:


server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
```diff
@@ -44,14 +53,67 @@
...
+
+    /**
+     * The current status of the repository.
+     */
+    private Status status = Status.CREATED;
+
+    /**
+     * Underlying storage for projections.
+     */
+    private EntityStorage<I> entityStorage;  //TODO:2016-05-03:alexander.yevsyukov: Why isn't it named projectionStorage?
```
alexander-yevsyukov: @AlexanderLitus, please pay attention to this TODO.
It should probably be `projectionStorage` of `ProjectionStorage` type.

server/src/test/java/org/spine3/server/BoundedContextShould.java
```diff
@@ -261,8 +258,8 @@ public TaskAdded handle(AddTask cmd, CommandContext ctx) {
         private void event(ProjectCreated event) {
 
             final Project newState = Project.newBuilder(getState())
-                    .setProjectId(event.getProjectId())
-                    .setStatus(STATUS_NEW)
```
alexander-yevsyukov: @AlexanderLitus, it's what I told you about yesterday. We wouldn't have status as a string in the production code. So shouldn't we do such a thing in tests.

client/src/main/java/org/spine3/protobuf/TypeToClassMap.java
```diff
@@ -106,13 +106,29 @@ public static ClassName get(TypeName protoType) {
...
+    public static TypeName get(ClassName className) {
+        final TypeName result = NAMES_MAP.inverse()
+                                           .get(className);
+        if (result == null) {
+            throw new IllegalStateException("No Protobuf type found for the Java class " + className);
+        }
+        return result;
+    }
+
+    /**
```
alexander-yevsyukov: @AlexanderLitus, please add test for this.

server/src/main/java/org/spine3/server/type/MessageClass.java
```diff
@@ -51,6 +52,13 @@ public String toString() {
     }
 
     /**
+     * Obtains the {@code ClassName} for this message class.
+     */
+    public ClassName getClassName() {
+        return ClassName.of(value());
+    }
```
alexander-yevsyukov: @AlexanderLitus, please add test for this.

server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
AlexanderLitus: Projection storage is not an Entity storage but uses it. So, this field is not a projection storage, but  an aggregated entity storage used to store projections.

client/src/main/java/org/spine3/protobuf/TypeToClassMap.java
```diff
@@ -74,7 +74,7 @@
      *
      * <p>For example, for a key {@code spine.base.EventId}, there will be the value {@code org.spine3.base.EventId}.
      */
-    private static final Map<TypeName, ClassName> NAMES_MAP = Builder.build();
+    private static final BiMap<TypeName, ClassName> NAMES_MAP = Builder.build();
```
AlexanderLitus: Why do we need to use this Map implementation?

client/src/main/java/org/spine3/protobuf/TypeToClassMap.java
```diff
@@ -106,13 +106,29 @@ public static ClassName get(TypeName protoType) {
...
 
     /**
+     * Returns the protobuf name for the class with the given name.
+     *
+     * @param className the name of the Java class for which to get protobuf type
+     * @return protobuf type name
+     * @throws IllegalStateException if there is no protobuf type for the specified class
+     */
+    public static TypeName get(ClassName className) {
+        final TypeName result = NAMES_MAP.inverse()
```
AlexanderLitus: Why do we need to inverse the map?

AlexanderLitus:
PTAL.

client/src/test/java/org/spine3/protobuf/TypeToClassMapShould.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.spine3.test.Tests.hasPrivateUtilityConstructor;
+
+/**
+ * @author Alexander Litus
+ */
+@SuppressWarnings("InstanceMethodNamingConvention")
+public class TypeToClassMapShould {
+
+    @Test
+    public void test() {
```
alexander-yevsyukov: Why not name it `have_private_utility_ctor()` as we do in other tests?

client/src/main/java/org/spine3/protobuf/TypeToClassMap.java
```diff
@@ -74,7 +74,12 @@
      *
      * <p>For example, for a key {@code spine.base.EventId}, there will be the value {@code org.spine3.base.EventId}.
      */
-    private static final BiMap<TypeName, ClassName> NAMES_MAP = Builder.build();
+    private static final BiMap<TypeName, ClassName> TYPE_TO_CLASS_MAP = Builder.build();
```
alexander-yevsyukov: I think we should use `lowerCamelCase` for this static field. Have a look at [Google Style Guide on constants](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) and [non-constants](https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names). This member doesn't _feel_ like a constant event that it's immutable. Why?

Suppose we want to have a map of types that is updated as new types are dynamically added to the application. We don't do it now, but it might be an option in the future. Or, we'd want to use this class on design time, and add types dynamically as the user works in IDE.  Would the map still be in upper case then?

client/src/main/java/org/spine3/protobuf/TypeToClassMap.java
```diff
@@ -74,7 +74,12 @@
...
+
+    /**
+     * The inverse view of the primary type to class map.
+     */
+    private static final BiMap<ClassName, TypeName> CLASS_TO_TYPE_MAP = TYPE_TO_CLASS_MAP.inverse();
```
alexander-yevsyukov: We don't need to keep the reference of the inverse map. The `BiMap` assumes it. Have a look at `AbstractBiMap` implementation:

```java
   public BiMap<V, K> inverse() {
        return this.inverse;
    }
```

alexander-yevsyukov: Assuming the above, I'd say that `underlyingBiMap` or just `biMap` would be a good candidate for what is now called `TYPE_TO_CLASS_MAP`.

client/src/test/java/org/spine3/protobuf/TypeToClassMapShould.java
AlexanderLitus: Done already.

AlexanderLitus:
PTAL.

alexander-yevsyukov:
LGTM

PR #138, opened 2016-05-06
Store unsupported commands

AlexanderLitus:


server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -280,8 +280,13 @@ private void schedule(Command command) {
...
     }
 
-    private void store(Command request) {
-        commandStore.store(request);
+    /**
+     * Stores a command to the storage.
+     *
+     * @param command a command to store
+     */
+    public void store(Command command) {
```
alexander-yevsyukov: We must not expose this method. If we do, anybody who can post to the Command Bus would be able to store the command.

server/src/main/java/org/spine3/server/BoundedContext.java
```diff
@@ -215,9 +217,17 @@ public void post(Command command, StreamObserver<Response> responseObserver) {
...
             commandBus.post(command);
+        } else if (isUnsupportedCommand(response)) {
+            storeUnsupportedCommand(command, response);
         }
     }
 
+    private void storeUnsupportedCommand(Command command, Response response) {
+        commandBus.store(command);
+        final CommandId id = command.getContext().getCommandId();
+        commandBus.getCommandStatusService().setToError(id, response.getError());
```
alexander-yevsyukov: Can we make storing with one call? We know the status already.

server/src/test/java/org/spine3/server/BoundedContextShould.java
```diff
@@ -163,14 +151,29 @@ public void register_ProjectionRepository() {
...
+        final TestResponseObserver responseObserver = new TestResponseObserver();
+        final CommandBus commandBus = boundedContext.getCommandBus();
+
+        final Command cmd = createProjectCmd();
+        boundedContext.post(cmd, responseObserver);
+
+        final Response response = responseObserver.getResponseHandled();
+        assertTrue(isUnsupportedCommand(response));
+        verify(commandBus, times(1)).store(cmd);
+        verify(commandBus, times(1)).getCommandStatusService();
```
alexander-yevsyukov: We need to test that real methods. “Real” are those that do what we want to test in the test. The `store()` is that kind of methods. The `getCommandStatusService()` is not. We have to call this method in the current implementation, but it is not the “meat” of the functionality under the test.

alexander-yevsyukov:
Since we're extending this functionality in #139, I'm closing this PR.

PR #139, opened 2016-05-08
Store invalid and unsupported commands with error status 

alexander-yevsyukov:


alexander-yevsyukov:
Due to some reason `ProcessManagerShould.dispatch_several_commands()` executes forever. Looking into it now.

client/src/test/java/org/spine3/client/ClientUtilShould.java
```diff
@@ -41,7 +42,7 @@ public void have_private_constructor() {
     public void generate_command_Ids() {
         final CommandId result = generateId();
 
-        assertFalse(result.getUuid().isEmpty());
+        assertFalse(idToString(result).isEmpty());
```
AlexanderLitus: We have to check the `UUID` field explicitly because if `generateId()` returns `null` this test will pass (the string will be `NULL`).

client/src/test/java/org/spine3/client/ClientUtilShould.java
```diff
@@ -41,7 +42,7 @@ public void have_private_constructor() {
     public void generate_command_Ids() {
```
AlexanderLitus: Moved this test to `CommandsShould`.

server/src/main/java/org/spine3/server/command/CommandValidator.java
```diff
@@ -77,7 +77,7 @@ private CommandValidator() {
         result.addAll(messageViolations);
 
         final CommandContext context = command.getContext();
-        final String commandId = idToString(context.getCommandId()).trim();
```
alexander-yevsyukov: What if the command ID was set directly to have some spaces (by mistake)?

If we remove the call to `trim()` — and we are very confident that this is a good idea — then the text of the message and the corresponding constant name should be changed. Now it mentions `_OR_BLANK`, and with the change we do not check for that.

server/src/main/java/org/spine3/server/command/error/CommandException.java
```diff
@@ -37,30 +38,52 @@
...
 
-    public static Map<String, Value> commandTypeAttribute(String commandType) {
-        return ImmutableMap.of(Attribute.COMMAND_TYPE_NAME, Value.newBuilder().setStringValue(commandType).build());
+    /**
+     * Returns a map with a command type attribute.
+     *
+     * @param commandMessage a command message to get the type from
+     */
+    public static Map<String, Value> commandTypeAttribute(Message commandMessage) {
+        final String commandType = commandMessage.getDescriptorForType().getFullName();
```
alexander-yevsyukov: How about using `TypeName` for getting the type name string? It would be a bit more compact.

server/src/main/java/org/spine3/server/command/CommandValidator.java
AlexanderLitus: We have `Command/EventIdToStringConverter`s registered in the converter registry, so I think we can remove explicit command/event ID conversion from `idToString()`.

client/src/main/java/org/spine3/base/Identifiers.java
```diff
@@ -59,10 +59,10 @@
 
     private Identifiers() {}
 
-    /*
-     * Null message or field string representation
+    /**
+     * Null or empty ID string representation.
      */
-    public static final String NULL_ID_OR_FIELD = "NULL";
+    public static final String NULL_OR_EMPTY_ID = "NULL_OR_EMPTY";
```
alexander-yevsyukov: Why do we need this? We cannot have duality here. It's the value of the ID.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -246,32 +270,31 @@ private void dispatch(Command command) {
...
-                commandStatusService.setToError(commandId, Errors.fromThrowable(cause));
-            }
+            onHandlerException(msg, commandId, e);
         }
     }
 
-    private CommandHandler getHandler(CommandClass commandClass) {
-        final CommandHandler handler = handlerRegistry.getHandler(commandClass);
-        return handler;
+    private void onHandlerException(Message msg, CommandId commandId, InvocationTargetException e) {
```
alexander-yevsyukov: It would be better to pass `Throwable` to this method. It would make the purpose of the method more clear.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -97,108 +98,68 @@ public void do_not_accept_null_CommandStore_on_construction() {
...
+        commandBus.register(new AllCommandDispatcher());
+        commandBus.register(new AllCommandDispatcher());
     }
 
-    //
-    // Tests for not overriding handlers by dispatchers and vice versa
-    //-------------------------------------------------------------------
+    /*
+     * Tests for not overriding handlers by dispatchers and vice versa.
+     ******************************/
```
alexander-yevsyukov: Have more stars to align with the text. They are cheap.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -214,69 +175,19 @@ public void do_not_allow_to_register_handler_for_the_command_with_registered_dis
...
-        public void dispatch(Command request) throws Exception {
-        }
-    }
-
     @Test
     public void unregister_handler() {
         commandBus.register(createProjectHandler);
         commandBus.unregister(createProjectHandler);
-        assertFalse(commandBus.isSupportedCommand(CommandClass.of(CreateProject.class)));
+        assertIsSupportedCmd(CreateProject.class, false);
```
alexander-yevsyukov: Why not `assertSupported`?

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -214,69 +175,19 @@ public void do_not_allow_to_register_handler_for_the_command_with_registered_dis
...
-
-        @Override
-        public void dispatch(Command request) throws Exception {
-            dispatcherInvoked = true;
-        }
-
-        public boolean wasDispatcherInvoked() {
-            return dispatcherInvoked;
-        }
+        assertAreSupportedCommands(true, CreateProject.class, AddTask.class);
```
alexander-yevsyukov: Why not have `assertSupported()` which accepts `Iterable` through ellipses?

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -408,9 +319,9 @@ public void onCompleted() {
         }
     }
 
-    //
-    // Tests of command processing
-    //---------------------------------------------
+    /*
+     * Command processing tests.
+     *********************/
```
alexander-yevsyukov: More starts, please.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -610,23 +498,152 @@ public void throw_exception_if_post_scheduled_cmd_and_no_scheduler_is_set() {
...
+        if (isSupported) {
+            assertTrue(commandBus.isSupportedCommand(clazz));
+        } else {
+            assertFalse(commandBus.isSupportedCommand(clazz));
+        }
+    }
+
+    /*
+     * Test command dispatchers.
+     ********************/
```
alexander-yevsyukov: More stars, please.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -610,23 +498,152 @@ public void throw_exception_if_post_scheduled_cmd_and_no_scheduler_is_set() {
...
+
+        @Override
+        public void dispatch(Command request) throws Exception {
+            throw exception;
+        }
+    }
+
+    /*
+     * Test command handlers.
+     ********************/
```
alexander-yevsyukov: More stars, please.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -610,23 +498,152 @@ public void throw_exception_if_post_scheduled_cmd_and_no_scheduler_is_set() {
...
+    private class CreateProjectHandler extends CommandHandler {
+
+        private boolean handlerInvoked = false;
+
+        protected CreateProjectHandler(String id) {
+            super(id, eventBus);
+        }
+
+        @Assign
+        @SuppressWarnings("unused")
```
alexander-yevsyukov: Can we suppress the warning on all methods annotated with `@Assign`?

server/src/main/java/org/spine3/server/command/CommandBus.java
AlexanderLitus: Passed a cause exception.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
AlexanderLitus: Yes, we can have one varargs method. What is `Iterable through ellipses`?..

server/src/test/java/org/spine3/server/command/CommandBusShould.java
AlexanderLitus: I've done it in my IDE, but this had no effect on our inspections profile.

client/src/main/java/org/spine3/validate/Validate.java
```diff
@@ -103,7 +104,7 @@ public static boolean isNotDefault(Message object) {
      */
     public static <M extends Message> M checkNotDefault(M object) {
         checkNotNull(object);
-        checkNotDefault(object, "The message is in the default state: %s", object);
+        checkNotDefault(object, "The message is in the default state: %s", TypeName.of(object).value());
```
alexander-yevsyukov: Having `toString()` in `TypeName` makes the `value()` call redundant. I think we can make the code more terse here.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -185,49 +188,38 @@ public void post(Command command, StreamObserver<Response> responseObserver) {
...
-     *
-     * @param command a command to store
-     * @param exception an exception occurred during command processing
-     */
-    @VisibleForTesting
-    /* package */ void storeWithErrorStatus(Command command, Exception exception) {
-        store(command);
-        //TODO:2016-05-08:alexander.yevsyukov: Support storage method that can make storing command with error in one call.
-        commandStatusService.setToError(command.getContext().getCommandId(), exception);
+    private static StatusRuntimeException newInvalidArgumentException(Exception exception) {
```
alexander-yevsyukov: The method call sounds a bit misleading. We create new `StatusRuntimeException`, not `IAE`.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -179,15 +186,15 @@ public void do_not_allow_to_register_handler_for_the_command_with_registered_dis
...
+        assertSupported(false, CreateProject.class);
     }
 
     @Test
     public void validate_commands_both_dispatched_and_handled() {
         commandBus.register(createProjectHandler);
         commandBus.register(new AddTaskDispatcher());
 
-        assertAreSupportedCommands(true, CreateProject.class, AddTask.class);
+        assertSupported(true, CreateProject.class, AddTask.class);
```
alexander-yevsyukov: What if we have `assertSupported()` and `assertNotSupported()`?

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -212,10 +212,10 @@ private boolean handleValidation(Command command, StreamObserver<Response> respo
     private void handleUnsupported(Command command, StreamObserver<Response> responseObserver) {
         final UnsupportedCommandException exception = new UnsupportedCommandException(command);
         commandStore.store(command, exception);
-        responseObserver.onError(newInvalidArgumentException(exception));
+        responseObserver.onError(newStatusRuntimeException(exception));
     }
 
-    private static StatusRuntimeException newInvalidArgumentException(Exception exception) {
+    private static StatusRuntimeException newStatusRuntimeException(Exception exception) {
```
alexander-yevsyukov: Please rename to `invalidArgumentWithCause()`.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -212,10 +212,10 @@ private boolean handleValidation(Command command, StreamObserver<Response> respo
     private void handleUnsupported(Command command, StreamObserver<Response> responseObserver) {
         final UnsupportedCommandException exception = new UnsupportedCommandException(command);
```
alexander-yevsyukov: Please rename the variable to `unsupportedException`.

alexander-yevsyukov:
LGTM now. I'll merge.

PR #141, opened 2016-05-11
gRPC 0.14.0 migration

mikhailmikhaylov:
I've moved gRPC dependency to root `build.gradle` file. However, I did not apply gRPC plugin to all subprojects to leave us an ability to enable\disable gRPC code generation. It looks like `grpc {}` block in protobuf task plugins sections in `server` and `client` modules.

Please tell me if we want to move these blocks to root `build.gradle` too.


mikhailmikhaylov:
Is there a way to re-run checks? I can't understand why one of them failed

AlexanderLitus:
Yeah, the fourth check always fail. But now there are only three of them and it is OK.

alexander-yevsyukov:
LGTM

PR #142, opened 2016-05-12
Store scheduled commands and re-schedule them

AlexanderLitus:
Store scheduled commands and re-schedule them on CommandBus creation to make scheduling more reliable. Also, refactor scheduler, add/move tests, etc.


client/src/main/proto/spine/base/command.proto
```diff
@@ -95,7 +95,7 @@ message CommandContext {
 // Enumeration of possible failures when validating a command.
 enum CommandValidationError {
     // Reserved value.
-    UNKNOWN = 0;
+    UNDEFINED_COMMAND_VALIDATION_ERROR = 0;
```
alexander-yevsyukov: Why do we need this long constant name assuming the enum is already named `CommandValidationError`?

Why do we want the user to read `CommandValidationError.UNDEFINED_COMMAND_VALIDATION_ERROR`, instead of  `CommandValidationError.UNDEFINED`?

alexander-yevsyukov: Please also note that zero is reserved by Protobuf for internal purposes. Presumably, it would show us that there was an error in enum transfer or the value wasn't set. So, I don't think we should use zero for the framework needs.

client/src/main/proto/spine/base/command.proto
```diff
@@ -110,18 +110,21 @@ enum CommandValidationError {
 // This enumeration defines possible statuses of command processing.
 enum CommandStatus {
     // Reserved value.
-    UNDEFINED = 0;
+    UNDEFINED_COMMAND_STATUS = 0;
```
alexander-yevsyukov: Again, why?

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -80,6 +82,16 @@ private CommandBus(Builder builder) {
         commandStore = builder.getCommandStore();
         scheduler = builder.getScheduler();
         commandStatusService = new CommandStatusService(commandStore);
+        rescheduleCommands();
+    }
+
+    private void rescheduleCommands() {
+        final Iterator<Command> commands = commandStore.load(SCHEDULED);
+        // TODO:2016-05-12:alexander.litus: re-calculate delays and schedule
```
alexander-yevsyukov: When do you plan to finish with this?

client/src/main/proto/spine/base/command.proto
AlexanderLitus: The enum constant with zero value is the default value in Protobuf messages, so it is rather  convenient to define it this way (I've faced this in Lobby). I didn't know that zero is reserved value, I'll read docs.

AlexanderLitus: The problem is that we have to name enum constants differently if they are located in one package, so I throught we can differ them this way.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -215,13 +222,34 @@ private void handleUnsupported(Command command, StreamObserver<Response> respons
...
+    }
+
     private static StatusRuntimeException invalidArgumentWithCause(Exception exception) {
         final StatusRuntimeException result = Status.INVALID_ARGUMENT
                 .withCause(exception)
                 .asRuntimeException();
         return result;
     }
 
+    private static StatusRuntimeException failedPreconditionNoScheduler() {
```
alexander-yevsyukov: I would keep the term `precondition` to parameter checking. Here we talk about illegal state.

client/src/main/proto/spine/base/command.proto
AlexanderLitus: Protoc states that `The first enum value must be zero in proto3.`.

server/src/main/java/org/spine3/server/command/CommandScheduler.java
```diff
@@ -31,32 +36,36 @@
  */
 public abstract class CommandScheduler {
 
+    private static final Set<CommandId> SCHEDULED_COMMAND_IDS = newHashSet();
```
alexander-yevsyukov: Again, this IS NOT A CONSTANT.

server/src/main/java/org/spine3/server/entity/GetTargetIdFromCommand.java
```diff
@@ -63,10 +61,11 @@ private GetTargetIdFromCommand() {
      * throws an exception (in the case if the command is not for an entity)
      */
     @Nullable
-    public static Object asNullableObject(Message commandMessage) {
+    public static <I> I asNullableObject(Message commandMessage) {
         //TODO:2016-05-09:alexander.yevsyukov: return Optional
```
alexander-yevsyukov: Please do return `Optional` here.

server/src/main/java/org/spine3/server/storage/CommandStorage.java
```diff
@@ -72,31 +92,76 @@ public void store(Command command) {
...
+    /**
+     * Loads all commands with the given status.
+     *
+     * @param status a command status to search by
+     * @return commands with the given status
+     * @throws IllegalStateException if the storage is closed
+     */
+    public Iterator<Command> load(CommandStatus status) {
+        checkNotClosed();
+        final Iterator<CommandStorageRecord> recordIterator = read(status);
```
alexander-yevsyukov: We don't really `read` here. We get an iterator over the status. When we read, we load data from a media.

AlexanderLitus:
@alexander-yevsyukov, PTAL. I've created a task to implement a CommandScheduler for AppEngine.

client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -130,7 +129,7 @@ public static CommandId getId(Command command) {
             public boolean apply(@Nullable Command request) {
                 checkNotNull(request);
                 final Timestamp timestamp = getTimestamp(request);
-                return Timestamps.isAfter(timestamp, from);
+                return Timestamps.isLater(timestamp, from);
```
alexander-yevsyukov: Why the  name change?

Anyway, the new name should have been `isLaterThan`.

client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -251,4 +250,44 @@ public static boolean isScheduled(Command command) {
         }
         return false;
     }
+
+    /**
+     * Sets a new scheduling time to {@link CommandContext.Schedule}.
```
alexander-yevsyukov: What is ‘scheduling time’? Is it the time after which the command should be executed? Or, is it the time when the command was put into the scheduled command list?

client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -251,4 +250,44 @@ public static boolean isScheduled(Command command) {
...
+
+    /**
+     * Updates {@link CommandContext.Schedule}.
+     *
+     * @param command a command to update
+     * @param delay a delay to set
+     * @param schedulingTime a scheduling time to set
+     * @return an updated command
+     */
+    public static Command setSchedule(Command command, Duration delay, Timestamp schedulingTime) {
```
alexander-yevsyukov: Having to parameters `delay` and `schedulingTime` makes this method really puzzling. The documentation on the parameters doesn't add any value. It simply repeats the name of the parameters.

client/src/main/java/org/spine3/protobuf/Durations.java
```diff
@@ -304,4 +305,17 @@ public static boolean isNegative(Duration value) {
         final boolean isNegative = nanos < 0;
         return isNegative;
     }
+
+    /**
+     * Returns a duration between two timestamps.
```
alexander-yevsyukov: Have a look at [definition of Interval](http://www.threeten.org/threeten-extra/apidocs/org/threeten/extra/Interval.html) in Joda Time and [the question on a similar concept in Java 8](http://stackoverflow.com/questions/22886030/equivalent-to-jodatime-interval-in-java-8-date-and-time-api).

As long, as we talk about the distance between two timestamps, we need to add a new data type called `Interval` in our Values sub-project. Duration and Interval are quite different things. A Duration is, so to say, is ‘filled in’ with something, while an Interval, is ‘empty’.

client/src/main/java/org/spine3/base/Commands.java
alexander-yevsyukov: The current name is in-line with [naming in new date/time API in Java](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html#isAfter-java.time.chrono.ChronoLocalDate-).

server/src/main/java/org/spine3/server/BoundedContext.java
```diff
@@ -361,7 +361,7 @@ public BoundedContext build() {
             checkNotNull(commandBus, "commandDispatcher must be set");
             checkNotNull(eventBus, "eventBus must be set");
 
-            commandBus.setMultitenant(this.multitenant);
+            commandBus.setIsMultitenant(this.multitenant);
```
alexander-yevsyukov: `setMultitenant` sounds as imperative and is OK for this case. The new name is also harder to read and pronounce. Please revert.

client/src/main/java/org/spine3/base/Commands.java
AlexanderLitus: Yes, but they have a method of a class, so they write:
'time.isAfter(time2)'
And we have a separate function, so to be not confusing, I wanted to change it to:
'isLater(time1, thanTime2)`

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -67,38 +75,43 @@
...
 
     /**
-     * Creates a new builder for command bus.
+     * Creates a new instance.
+     *
+     * @param commandStore a store to save commands
+     * @param scheduler a command scheduler
+     * @param problemLog a problem logger
      */
-    public static Builder newBuilder() {
```
alexander-yevsyukov: Please revert to having builder. Have also setting to multitenancy attribute in the builder and remove the setter. There is no need to change the attribute after the instance is constructed.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -374,58 +404,26 @@ private CommandDispatcher getDispatcher(CommandClass commandClass) {
         return dispatcherRegistry.getDispatcher(commandClass);
     }
 
+    private static CommandScheduler createCommandScheduler() {
+        @SuppressWarnings("AccessOfSystemProperties")
+        final String appEngineVersion = System.getProperty("com.google.appengine.runtime.version");
```
alexander-yevsyukov: This needs to be a feature of a utility class, which is visible to the whole framework. It can be called `Environment` or something like that.

server/src/main/java/org/spine3/server/command/error/CommandExpiredException.java
```diff
@@ -0,0 +1,65 @@
...
+import org.spine3.base.CommandValidationError;
+import org.spine3.base.Commands;
+import org.spine3.base.Error;
+import org.spine3.server.type.CommandClass;
+import org.spine3.type.TypeName;
+
+import static java.lang.String.*;
+
+/**
+ * Exception that is thrown when a scheduled command was not delivered to the target in time.
```
alexander-yevsyukov: Do we have expiration period for a command? If yes, where do we set it?

server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -83,26 +89,31 @@
     private TestResponseObserver responseObserver;
 
     @Before
-    public void setUp() {
+    public void setUpTest() {
```
alexander-yevsyukov: Why have the `Test` suffix? It's common to name these methods as `setUp` and `tearDown`. It is assumed that methods annotated `@Before` and `@After` are configuration and clean-up for each test method.

alexander-yevsyukov: And if you suggest the renaming here, are we going to rename all other similar methods in other framework tests?

client/src/main/java/org/spine3/base/Commands.java
alexander-yevsyukov: OK, let's have `isLaterThan` and `isBeforeThan`.

server/src/main/java/org/spine3/server/aggregate/Apply.java
```diff
@@ -20,22 +20,27 @@
...
-
- * <p>Event appliers are not supposed to be called from the outside of the declaring aggregate class.
- * As such they should be declared {@code private}.
  *
- * <p>The type of the event will be indicated by the first (and only) parameter.
+ * <p>An event applier method:
+ * <ul>
+ *     <li>is annotated with {@link Apply};
+ *     <li>is {@code private};
+ *     <li>is void;
```
alexander-yevsyukov: `{@code void}`?

server/src/main/java/org/spine3/server/aggregate/Apply.java
```diff
@@ -20,22 +20,27 @@
...
+ *     <li>accepts an event {@link Message}.
```
alexander-yevsyukov: It would be better not no link to the `Message` interface. Because it may be understood as a method must accept `Message`, which raises a question, “How do I have multiple appliers?”

AlexanderLitus:
PTAL.

alexander-yevsyukov:
Please also check my earlier comments. Not all of them seem to be addressed.

server/src/main/java/org/spine3/server/aggregate/Apply.java
AlexanderLitus: Changed.

server/src/test/java/org/spine3/server/command/CommandBusShould.java
AlexanderLitus: Removed the suffix.

server/src/main/java/org/spine3/server/command/error/CommandExpiredException.java
AlexanderLitus: Improved docs, please check it out.

server/src/main/java/org/spine3/server/command/CommandBus.java
AlexanderLitus: Done.

server/src/main/java/org/spine3/server/command/CommandBus.java
AlexanderLitus: Discussed that we need the setter in the BoundedContext.

client/src/main/java/org/spine3/base/Commands.java
AlexanderLitus: Improved docs.

client/src/main/java/org/spine3/protobuf/Durations.java
AlexanderLitus: Done.

server/src/main/java/org/spine3/server/BoundedContext.java
AlexanderLitus: OK.

client/src/main/java/org/spine3/base/Commands.java
AlexanderLitus: Improved docs.

server/src/main/java/org/spine3/server/storage/CommandStorage.java
AlexanderLitus: Renamed.

AlexanderLitus:
Processed all comments, check it out.

values/src/main/proto/spine/time/time.proto
```diff
@@ -82,4 +84,16 @@ message OffsetDateTime {
...
+
+// An interval between two points in time.
+message Interval {
+    // A start of the interval.
+    google.protobuf.Timestamp start = 1;
+
+    // An end of the interval.
+    google.protobuf.Timestamp end = 2;
+
+    // A duration between `start` and `end`.
```
alexander-yevsyukov: The duration can be calculated. We have all the needed information for this. There's no reason to store the duration. `Intervals` should have a conversion method, `toDuration(Interval)`.

values/src/main/java/org/spine3/time/Intervals.java
```diff
@@ -0,0 +1,57 @@
...
+    private Intervals() {}
+
+    /**
+     * Returns an interval between two timestamps.
+     *
+     * @param start the first point in time
+     * @param end the second point in time
+     * @return an interval between {@code start} and {@code end}
+     */
+    public static Interval between(Timestamp start, Timestamp end) {
```
alexander-yevsyukov: There should be argument check that verifies that `start` is before of equals to `end`. If it equals, then it returns the `ZERO` constant.

server/src/main/java/org/spine3/util/EnvironmentUtil.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Provides information about the environment (current platform used, etc).
+ *
+ * @author Alexander Litus
+ */
+public class EnvironmentUtil {
+
+    private static final String APP_ENGINE_RUNTIME_VERSION = "com.google.appengine.runtime.version";
+
+    private EnvironmentUtil() {}
```
alexander-yevsyukov: Please have singleton instance of the `Environment` class, which should have the following methods:

`isAppEngine()`
`getAppEngineVersion()`

That will do for now. Have the constructor with the `protected` access. We may want to have derived classes. For example, for specifying minimal compatible configuration.

server/src/main/java/org/spine3/server/command/error/CommandExpiredException.java
```diff
@@ -32,6 +32,8 @@
 /**
  * Exception that is thrown when a scheduled command was not delivered to the target in time.
  *
+ * <p>For example, in the case when a server restarted when it was the time to deliver a scheduled command.
```
alexander-yevsyukov: Why don't we deliver it after when the server is started back again? What is the standard practice for such cases in messaging systems?

alexander-yevsyukov: Create a separate task for addressing this issue. We need to close this PR soon.

server/src/main/java/org/spine3/server/validate/MessageFieldValidator.java
```diff
@@ -123,8 +123,8 @@ private void validateTimestamps() {
 
     private static boolean isTimeInvalid(Timestamp time, Time when, Timestamp now) {
```
alexander-yevsyukov: Please add Javadoc for the method.

AlexanderLitus:
PTAL.

server/src/main/java/org/spine3/server/command/error/CommandExpiredException.java
AlexanderLitus: Created.

server/src/main/java/org/spine3/util/Environment.java
```diff
@@ -25,20 +25,46 @@
...
+public class Environment {
 
-    private static final String APP_ENGINE_RUNTIME_VERSION = "com.google.appengine.runtime.version";
+    /**
+     * The key of the Google AppEngine runtime version system property.
+     */
+    public static final String APP_ENGINE_RUNTIME_VERSION_KEY = "com.google.appengine.runtime.version";
+
+    @SuppressWarnings("AccessOfSystemProperties")
+    private static final String APP_ENGINE_RUNTIME_VERSION = System.getProperty(APP_ENGINE_RUNTIME_VERSION_KEY);
```
alexander-yevsyukov: Again, this isn't a constant. It's final, but not a constant.

server/src/main/java/org/spine3/util/Environment.java
```diff
@@ -25,20 +25,46 @@
...
+                !APP_ENGINE_RUNTIME_VERSION.isEmpty();
         return isVersionPresent;
     }
+
+    /**
+     * Returns the current Google AppEngine version
+     * retrieved by the key {@link Environment#APP_ENGINE_RUNTIME_VERSION_KEY}.
+     */
+    public String getAppEngineVersion() {
+        return APP_ENGINE_RUNTIME_VERSION;
```
alexander-yevsyukov: What do we return, if we're not under AppEngine?

AlexanderLitus: `null` is returned, added docs.

AlexanderLitus:
PTAL.

server/src/main/java/org/spine3/util/Environment.java
```diff
@@ -49,17 +49,18 @@ public static Environment getInstance() {
...
+        final boolean isVersionPresent = (appEngineRuntimeVersion != null) &&
+                !appEngineRuntimeVersion.isEmpty();
         return isVersionPresent;
     }
 
     /**
      * Returns the current Google AppEngine version
-     * retrieved by the key {@link Environment#APP_ENGINE_RUNTIME_VERSION_KEY}.
+     * retrieved by the key {@link Environment#APP_ENGINE_RUNTIME_VERSION_KEY}
+     * or {@code null} if the program is running not on the AppEngine.
```
alexander-yevsyukov: Annotate the method as `@Nullable` then.

alexander-yevsyukov: `@return` Javadoc tag is applicable in this case.

server/src/main/java/org/spine3/util/Environment.java
```diff
@@ -49,17 +49,18 @@ public static Environment getInstance() {
...
-                !APP_ENGINE_RUNTIME_VERSION.isEmpty();
+        final boolean isVersionPresent = (appEngineRuntimeVersion != null) &&
+                !appEngineRuntimeVersion.isEmpty();
         return isVersionPresent;
     }
 
     /**
      * Returns the current Google AppEngine version
-     * retrieved by the key {@link Environment#APP_ENGINE_RUNTIME_VERSION_KEY}.
+     * retrieved by the key {@link Environment#APP_ENGINE_RUNTIME_VERSION_KEY}
```
alexander-yevsyukov: We do not need to expose this `private static final` constant here. It is not supposed to be visible to public API user.

alexander-yevsyukov:
Please address last comments, then it would LGTM.

PR #145, opened 2016-05-26
Client service — facade for client apps

alexander-yevsyukov:


server/src/main/java/org/spine3/server/BoundedContext.java
```diff
@@ -64,7 +65,9 @@
  * @author Alexander Yevsyukov
  * @author Mikhail Melnik
  */
-public class BoundedContext implements ClientService, IntegrationEventSubscriber, AutoCloseable {
+public class BoundedContext implements org.spine3.client.grpc.ClientServiceGrpc.ClientService,
```
mikhailmikhaylov: Do we use not imported class here for some purpose?

mikhailmikhaylov:
LGTM

server/src/main/java/org/spine3/server/BoundedContext.java
alexander-yevsyukov: This interface is no longer implemented — no need.

PR #147, opened 2016-05-30
Make multitenancy clear. Packages for generic subdomains.

alexander-yevsyukov:


client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -80,15 +81,19 @@ public static CommandId generateId() {
      *
      * @param userId the actor id
      * @param zoneOffset the offset of the timezone in which the user works
+     * @param tenantId the ID of the tenant or null for single-tenant applications
      * @see CommandFactory
      */
-    public static CommandContext createContext(UserId userId, ZoneOffset zoneOffset) {
+    public static CommandContext createContext(UserId userId, ZoneOffset zoneOffset, @Nullable TenantId tenantId) {
```
AlexanderLitus: Why not just overload this method instead of adding a @Nullable parameter?

examples/src/main/java/org/spine3/examples/aggregate/ClientApp.java
```diff
@@ -82,16 +87,56 @@ public void onCompleted() {
...
+                .setAuthor("Eric Evans")
+                .setTitle("Domain Driven Design.")
+                .setPrice(price)
+                .build();
+        final int quantity = 1;
+        final double totalPrice = book.getPrice() * quantity;
+        final OrderLine orderLine = OrderLine.newBuilder()
+                .setProductId(Messages.toAny(book.getBookId()))
+                .setQuantity(quantity)
+                .setTotal(totalPrice)
```
AlexanderLitus: I'd rather use Money here.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -192,6 +193,11 @@ public void post(Command command, StreamObserver<Response> responseObserver) {
             scheduleAndStore(command, responseObserver);
             return;
         }
+
+        if (isMultitenant) {
+            CurrentTenant.set(command.getContext().getTenantId());
```
AlexanderLitus: Please add a test for this case.

server/src/main/java/org/spine3/server/users/CurrentTenant.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+@SPI
+public class CurrentTenant {
+
+    private static final ThreadLocal<TenantId> threadLocal = new ThreadLocal<>();
+
+    /**
+     * Have only static methods for this class.
+     */
+    private CurrentTenant() {}
```
AlexanderLitus: Please add a test for this case.

server/src/main/java/org/spine3/server/users/CurrentTenant.java
```diff
@@ -22,11 +22,12 @@
...
-import org.spine3.validate.Validate;
 
 import javax.annotation.Nullable;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.validate.Validate.isNotDefault;
 
 /**
  * This class allows to set a tenant when handling a command or a query
```
AlexanderLitus: Maybe we should provide some links for those who are not familiar with multitenant architecture?

server/src/test/java/org/spine3/server/users/CurrentTenantShould.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class CurrentTenantShould {
+
+    @SuppressWarnings("ConstantConditions")
+    @Test(expected = NullPointerException.class)
+    public void reject_null_value() {
+        CurrentTenant.set(null);
```
AlexanderLitus: We can create a method `null()` which simply returns `null`, and we can use it in such cases without warnings suppression.

server/src/main/java/org/spine3/server/users/CurrentTenant.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * This class allows to set a tenant when handling a command or a query
+ * in a multitenant application.
+ *
+ * @author Alexander Yevsyukov
+ */
+@SPI
+public class CurrentTenant {
+
+    private static final ThreadLocal<TenantId> threadLocal = new ThreadLocal<>();
```
AlexanderLitus: Please add docs about why it is ThreadLocal.

AlexanderLitus:
PTAL.

alexander-yevsyukov:
LGTM

PR #150, opened 2016-06-09
Injecting events

AlexanderLitus:


AlexanderLitus:
I removed TODOs from `Aggregate`.

`applyForTest()` is used in Lobby for testing event appliers. I spent some time trying to use `TestOrderAggregate` and define methods which call real appliers via reflection; this approach worked before, but when we added a restriction that `getBuilder()` method can be called only from applier, this approach stopped working. So I started using real aggregate in tests.

`incrementStateForTest()` is also used in Lobby for constructing aggregates for tests. When I used `TestOrderAggregate`, `incrementState` was overridden to make it accessible from tests. Then I started using real `OrderAggregate` in tests, so I defined this method.

alexander-yevsyukov:
Then we need tests in the core-java do that the methods are used.

AlexanderLitus:
LGTM. 
PTAL at my changes.

PR #151, opened 2016-06-10
Use `StreamObserver.onError()` while posting integration events

AlexanderLitus:
Use `StreamObserver.onError()` instead of returning responses while posting integration events.
Add  `internal`, `SPI`, `experimental` and `beta` custom opts.
Also refactor CommandBus to reduce the number of methods.


AlexanderLitus:
PTAL.

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -268,18 +270,28 @@ public void run() {
...
+     *          {@link StreamObserver#onError(Throwable)} is called if an event is unsupported or invalid
+     * @return {@code true} if event is supported and valid and can be posted, {@code false} otherwise
      */
-    public Response validate(Message event) {
+    public boolean validate(Message event, StreamObserver<Response> responseObserver) {
         final EventClass eventClass = EventClass.of(event);
         if (isUnsupportedEvent(eventClass)) {
-            return unsupportedEvent(event);
+            final UnsupportedEventException unsupportedEvent = new UnsupportedEventException(event);
+            // TODO:2016-06-09:alexander.litus: consider storing an event with `Error` status
```
alexander-yevsyukov: No need to store events with errors, at least now. We store commands with error status because they come from outside, and we need to track this history. For EventBus we control all the code and can detect the error in the posting size after we call `onError()`.

Please remove TODOs related to your suggestion.

server/src/main/java/org/spine3/server/event/error/EventException.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Map;
+
+/**
+ * A base for exceptions related to events.
+ *
+ * @author Alexander Litus
+ */
+public abstract class EventException extends RuntimeException {
+
+    @SuppressWarnings("NonSerializableFieldInSerializableClass")
```
alexander-yevsyukov: If you have `GeneratedMessage` it will be serializable.

server/src/main/java/org/spine3/server/event/error/EventException.java
```diff
@@ -0,0 +1,90 @@
...
+                                 .setStringValue(type)
+                                 .build();
+        return ImmutableMap.of(Attribute.EVENT_TYPE_NAME, value);
+    }
+
+    /**
+     * Returns a related event.
+     */
+    public Message getEventMessage() {
+        return eventMessage;
```
alexander-yevsyukov: Please add tests that include getters into coverage.

alexander-yevsyukov:
Please add field options and message options called `internal` and `experimental`, which would allow to annotate fields and types in our API.

Annotate the internal fields you're introducing in this PR with `internal` annotation.

alexander-yevsyukov:
LGTM

PR #155, opened 2016-06-13
Exclude slf4j-jdk14 from the published dependencies. Use it only for tests.

armiol:
@alexander-yevsyukov PTAL.


alexander-yevsyukov:
LGTM

PR #157, opened 2016-06-17
Event enrichments

alexander-yevsyukov:


alexander-yevsyukov:
Please improve the code coverage.

client/src/main/java/org/spine3/protobuf/Messages.java
```diff
@@ -212,35 +211,47 @@ public static GenericDescriptor getClassDescriptor(Class<? extends Message> claz
...
-                defaultObject = fieldDescriptor.getDefaultValue();
-                return defaultObject.getClass();
+                return ByteString.class;
             case ENUM:
-                //TODO:2016-06-17:alexander.yevsyukov: Figure out how to treat enums.
-                throw Exceptions.unsupported(
-                        "Enums in mapping are not yet supported. Discovered: " + fieldDescriptor.getFullName());
+                final String enumTypeNameStr = field.getEnumType().getFullName();
+                final TypeName enumTypeName = TypeName.of(enumTypeNameStr);
+                return tryConvertToClass(enumTypeName);
```
alexander-yevsyukov: Why `try`? Event if it would throw an exception it may be better expressed in Javadoc for the method.

client/src/main/java/org/spine3/protobuf/Messages.java
```diff
@@ -212,35 +211,47 @@ public static GenericDescriptor getClassDescriptor(Class<? extends Message> claz
...
-                try {
-                    final Class<? extends Message> result = toMessageClass(typeName);
-                    return result;
-                } catch (ClassNotFoundException e) {
-                    propagate(e);
-                }
-                break;
+                return tryConvertToClass(typeName);
+        }
+        throw new IllegalStateException("Unknown field type discovered: " + field.getFullName());
```
alexander-yevsyukov: Please describe this outcome in Javadoc too.

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
```diff
@@ -243,10 +244,9 @@ private Builder() {}
          * @throws IllegalArgumentException if the builder already has a function, which has the same couple of
          * source and target classes
          */
-        private <M extends Message, E extends Message> void checkDuplicate(EnrichmentFunction<M, E> function) {
-            final Optional<EnrichmentFunction<? extends Message, ? extends Message>> duplicate =
-                    FluentIterable.from(functions)
-                                  .firstMatch(SameTransition.asFor(function));
+        private <M, E> void checkDuplicate(EnrichmentFunction<M, E> function) {
```
alexander-yevsyukov: Why don't we use `S` and `T` here for consistency of renaming you suggested?

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
```diff
@@ -227,8 +227,9 @@ private Builder() {}
             return this;
         }
 
-        public <M extends Message, E extends Message> Builder addFieldEnrichment(Class<M> eventMessageClass,
-        Class<E> enrichmentClass, Function<M, E> function) {
+        public <M, E> Builder addFieldEnrichment(Class<M> eventMessageClass,
```
alexander-yevsyukov: If it comes to field enrichment, the parameter should be probably called `fieldClass` instead of `eventMessageClass`.

Consider using `S` and `T` generic type names as you suggested.

server/src/test/proto/spine/test/event/events.proto
```diff
@@ -23,13 +23,18 @@ package spine.test.event;
...
 
 option java_package="org.spine3.test.event";
 option java_outer_classname = "EventEventsProto";
 option java_multiple_files = true;
 
 message ProjectCreated {
     ProjectId project_id = 1;
+
+    message Enrichment {
+        string project_name = 2 [(by) = "project_id"];
```
alexander-yevsyukov: Cannot a nested type have field indexes independent from the outer type?

client/src/main/java/org/spine3/protobuf/Messages.java
AlexanderLitus: `try` prefix is used to distinguish this method from `public toMessageClass()`. I find it better than `innerToMessageClass()` or something. I use `try` prefix for such methods as Bob Martin does in his great book, `Clean Code`.

AlexanderLitus:
I think that CodeCov changed the way it calculated the coverage. Because we had 90% or so, and now we have 75.06%... Though I covered Enricher and other new methods good enough. And some results are stale. I'll add some more tests.

client/src/main/java/org/spine3/base/Events.java
```diff
@@ -90,11 +91,20 @@ public static Timestamp getTimestamp(Event event) {
     }
 
     /**
-     * Creates {@code Event} instance with the passed event and context.
+     * Creates a new {@code Event} instance with the passed event message and context.
      */
+    @SuppressWarnings("OverloadedMethodsWithSameNumberOfParameters")
     public static Event createEvent(Message event, EventContext context) {
```
alexander-yevsyukov: `Any` is also a `Message`. How do we ensure that a correct method is called?

client/src/main/proto/spine/event_annotations.proto
```diff
@@ -50,15 +50,7 @@ extend google.protobuf.FieldOptions {
...
-    // <pre>
-    //     spine.people.PersonName name = 1 [(by) = "MyEvent.user_id"];
-    // </pre>
-    //
-    // If the source message is not in the same package, the type name prefixing the field reference
-    // must be fully-qualified.
+    // The enrichment message can be defined anywhere and with any name.
     //
     string by = 59100;
 }
```
alexander-yevsyukov: Please add CR.

client/src/main/proto/spine/event_annotations.proto
```diff
@@ -50,15 +50,7 @@ extend google.protobuf.FieldOptions {
...
-    // If the enrichment message is not enclosed into the event message, the source field name must be prefixed
-    // with the source event message type name:
-    //
-    // <pre>
-    //     spine.people.PersonName name = 1 [(by) = "MyEvent.user_id"];
-    // </pre>
-    //
-    // If the source message is not in the same package, the type name prefixing the field reference
-    // must be fully-qualified.
+    // The enrichment message can be defined anywhere and with any name.
```
alexander-yevsyukov: But how do we define the referenced field name in those cases?

client/src/main/java/org/spine3/base/Events.java
AlexanderLitus: The correct implementation is selected automatically, if we pass Any, then one with any is called; if we pass a message, which is not Any, then the method with msg param is called.

client/src/main/proto/spine/event_annotations.proto
AlexanderLitus: Only the field name should be specified as we get it from the descriptor and the FQN is not needed.

server/src/main/java/org/spine3/server/event/enrich/EventMessageEnricher.java
```diff
@@ -21,103 +21,108 @@
...
-            final Descriptors.FieldDescriptor targetField = fieldMap.get(srcField);
-            if (targetValue != null) {
-                builder.setField(targetField, targetValue);
+        setFields(builder, message);
+        @SuppressWarnings("unchecked") // types are checked during the initialization and validation
+        final T result = (T) builder.build();
+        return result;
+    }
+
+    @SuppressWarnings({"ConstantConditions", "unchecked", "MethodWithMultipleLoops"})
```
alexander-yevsyukov: It would be better to have `unchecked` suppressed in the smallest scope.

alexander-yevsyukov:
Please see that examples are excluded from coverage.

AlexanderLitus:
PTAL. Covered enricher, event bus and new util methods.

client/src/main/proto/spine/event_annotations.proto
```diff
@@ -51,6 +51,7 @@ extend google.protobuf.FieldOptions {
     // </pre>
     //
     // The enrichment message can be defined anywhere and with any name.
+    // The value of the option is the name of the source event field, exactly the same as in the definition.
```
alexander-yevsyukov: As we discussed over Skype, please implement the definition of source field references using type name and FQN, if an enrichment is in another package, and revert the documentation of such abilities here.

AlexanderLitus:
PTAL.

client/src/main/java/org/spine3/protobuf/Messages.java
```diff
@@ -121,14 +121,19 @@ public static Any toAny(TypeName type, ByteString value) {
...
      * @see #fromAny(Any) that uses the same convention
      */
-    public static <T extends Message> Class<T> toMessageClass(TypeName messageType) throws ClassNotFoundException {
+    public static <T extends Message> Class<T> toMessageClass(TypeName messageType) {
         final ClassName className = TypeToClassMap.get(messageType);
-        @SuppressWarnings("unchecked")
-        final Class<T> result = (Class<T>) Class.forName(className.value());
-        return result;
+        try {
+            @SuppressWarnings("unchecked")
```
alexander-yevsyukov: Please document why is it safe to suppress?

client/src/main/java/org/spine3/protobuf/EntityPackagesMap.java
```diff
@@ -62,7 +62,7 @@
      * <p>Example:
      * <p>{@code com.example.order} - {@code AGGREGATE}
      */
-    private static final Map<String, EntityType> PACKAGES_MAP = buildPackagesMap();
+    private static final Map<String, EntityType> packagesMap = buildPackagesMap();
```
alexander-yevsyukov: Do we update the map after we build it? If not, let's have `ImmutableMap`.

client/src/main/java/org/spine3/protobuf/EntityPackagesMap.java
```diff
@@ -62,7 +62,7 @@
      * <p>Example:
```
alexander-yevsyukov: Please have a paragraph after abstract line of the property. Then you'll be able to have `Example:` in the same paragraph.

client/src/main/java/org/spine3/protobuf/EntityPackagesMap.java
```diff
@@ -62,7 +62,7 @@
...
      * <p>{@code com.example.order} - {@code AGGREGATE}
```
alexander-yevsyukov: People may want to have multiple aggregates per package. It would be safer to map from message type to an entity type.

client/src/main/java/org/spine3/protobuf/Messages.java
```diff
@@ -95,23 +95,23 @@ public static Any toAny(TypeName type, ByteString value) {
...
-            result = any.unpack(messageClass);
-
-        } catch (ClassNotFoundException ignored) {
-            throw new UnknownTypeException(typeStr);
+            final T result = any.unpack(messageClass);
+            return result;
+        } catch (RuntimeException e) {
+            if (e.getCause() instanceof ClassNotFoundException) {
+                // noinspection ThrowInsideCatchBlockWhichIgnoresCaughtException
+                throw new UnknownTypeException(typeStr);
```
alexander-yevsyukov: Why don't we set the cause?

client/src/main/java/org/spine3/protobuf/Messages.java
```diff
@@ -121,14 +121,19 @@ public static Any toAny(TypeName type, ByteString value) {
...
         final ClassName className = TypeToClassMap.get(messageType);
-        @SuppressWarnings("unchecked")
-        final Class<T> result = (Class<T>) Class.forName(className.value());
-        return result;
+        try {
+            @SuppressWarnings("unchecked")
+            final Class<T> result = (Class<T>) Class.forName(className.value());
+            return result;
+        } catch (ClassNotFoundException e) {
+            throw propagate(e);
```
alexander-yevsyukov: In some places we simply propagate, but in other we throw `UnknownTypeException`. We need to be consistent in handling unknown types.

client/src/main/proto/spine/event_annotations.proto
```diff
@@ -36,8 +36,36 @@ option java_package = "org.spine3.annotations";
...
+    // If the enrichment message is enclosed into the target event message, this option can be omitted.
+    //
+    // If the event message to enrich is in another package, the type name prefixing the field reference
+    // must be fully-qualified:
+    // <pre>
+    // message MyEventEnrichment {
+    //     option (enrichment_for) = "com.company.MyEvent";
+    //     ...
+    // }
+    // </pre>
```
alexander-yevsyukov: Please also document the case when a reference in `by` can be qualified with a type name.

server/src/main/java/org/spine3/server/ClientService.java
```diff
@@ -112,7 +112,7 @@ public void run() {
     }
 
     @VisibleForTesting
-    /* package */ boolean isShutdown() {
+    public boolean isShutdown() {
```
alexander-yevsyukov: Who is going to call it from outside?

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
```diff
@@ -106,8 +108,8 @@ private EventEnricher(Builder builder) {
      * <p>An event can be enriched if the following conditions are met:
      *
      * <ol>
-     *     <li>There is one or more functions registered for an {@link EnrichmentFunction} where
-     *     the passed class is the {@code source}.
+     *     <li>There is one or more enrichments defined in Protobuf using {@code enrichment_for} and/or {@code by} options.
+     *     <li>There is one or more field enrichment functions registered for the class of the passed event.
```
alexander-yevsyukov: What happens if there is at least one `enrichment_for` option and no `by` references in a type?

alexander-yevsyukov: What happens if the user specifies two or more types in the `enrichment_for` option? How do we resolve `by` references then?

server/src/main/java/org/spine3/server/event/enrich/EventEnrichmentsMap.java
```diff
@@ -0,0 +1,82 @@
...
+/* package */ class EventEnrichmentsMap {
+
+    /**
+     * A path to the file which contains enrichment and event Protobuf type names.
+     * Is generated by Gradle during the build process.
+     */
+    private static final String PROPS_FILE_PATH = "enrichments.properties";
+
+    /**
+     * A map from an event enrichment Protobuf type name to the corresponding event to enrich type name.
```
alexander-yevsyukov: This Javadoc duplicates one in the class. Why do we need it?

server/src/main/java/org/spine3/server/event/enrich/EventEnrichmentsMap.java
```diff
@@ -0,0 +1,82 @@
...
+import org.spine3.type.TypeName;
+
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+
+import static org.spine3.io.IoUtil.loadAllProperties;
+
+/**
+ * A map from an event enrichment Protobuf type name to the corresponding event to enrich type name.
```
alexander-yevsyukov: The description isn't quite clear because of repeated `to`. Please also have paragraph mark after the class abstract sentence.

server/src/main/java/org/spine3/server/event/enrich/EventEnrichmentsMap.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * <p>Example:
+     * <p>{@code proto.type.MyEventEnrichment} - {@code proto.type.MyEvent}
+     */
+    private static final ImmutableMap<TypeName, TypeName> enrichmentsMap = buildEnrichmentsMap();
+
+    private EventEnrichmentsMap() {}
+
+    /**
+     * Returns the map instance.
```
alexander-yevsyukov: Please say that the returned instance is immutable.

server/src/main/java/org/spine3/server/event/enrich/EventEnrichmentsMap.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    @SuppressWarnings("ReturnOfCollectionOrArrayField") // is immutable
+    public static Map<TypeName, TypeName> getInstance() {
```
alexander-yevsyukov: Do we really need it `public`, assuming the class has package access?

server/src/test/proto/spine/test/event/enrichment/enrichment.proto
```diff
@@ -0,0 +1,33 @@
...
+
+option java_package="org.spine3.test.event.enrichment";
+option java_multiple_files = true;
+
+import "spine/event_annotations.proto";
+
+message ProjectCreatedEnrichmentAnotherPackage {
+    option (enrichment_for) = "spine.test.event.ProjectCreated";
+
+    string project_name = 1 [(by) = "project_id"];
```
alexander-yevsyukov: Please have a test data where we have FQN name right in the `by` specification.

server/src/test/proto/spine/test/event/events.proto
```diff
@@ -41,6 +41,12 @@ message ProjectCreated {
     }
 }
 
+message ProjectCreatedSeparateEnrichment {
+    option (enrichment_for) = "ProjectCreated";
+
+    string project_name = 1 [(by) = "project_id"];
```
alexander-yevsyukov: Please have a test data where we have same package type name right in the `by` specification.

client/src/main/java/org/spine3/protobuf/EntityPackagesMap.java
AlexanderLitus: But we have a convention that there is a separate package for aggregate's files. Each package contains files with aggregate state, `commands.proto` (which must have exactly this name) and `events.proto`.

server/src/main/java/org/spine3/server/ClientService.java
AlexanderLitus: Is needed in tests, which I moved to the separate package from the root one (because we have separate messages and factories for tests).

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
AlexanderLitus: > What happens if there is at least one enrichment_for option and no by references in a type?

An exception is thrown.


>  What happens if the user specifies two or more types in the enrichment_for option? How do we resolve by references then?

Currently, there is no support for this. I can create a task for this and we'll do it later. Or, if this feature is needed in Projects, I'll do it in the next PR.
 We can support such option:

`option (enrichment_for) = "ProjectCreated, ProjectStarted";`

and say that `by` options should be non-qualified.

AlexanderLitus:
PTAL.

client/src/main/proto/spine/event_annotations.proto
```diff
@@ -38,25 +38,19 @@ import "google/protobuf/descriptor.proto";
 
 extend google.protobuf.MessageOptions {
     // An instruction to associate an enrichment message with the event message to enrich.
+    // If the event message to enrich is in another package, the event type name must be fully-qualified.
```
alexander-yevsyukov: It would be better to say that the type name should be qualified according to Protobuf conventions. It may not be a full package name, just the path that it needed to resolve the type name.

server/src/test/java/org/spine3/server/event/enrich/EventEnrichmentsMapShould.java
```diff
@@ -48,12 +54,19 @@ public void return_map_instance() {
...
+    public void contain_event_types_by_enrichment_type_names() {
+        getEventTypeByEnrichmentTypeTest(ProjectCreated.class, ProjectCreated.Enrichment.class);
+        getEventTypeByEnrichmentTypeTest(ProjectCreated.class, ProjectCreatedSeparateEnrichment.class);
+        getEventTypeByEnrichmentTypeTest(ProjectCreated.class, ProjectCreatedEnrichmentAnotherPackage.class);
+        getEventTypeByEnrichmentTypeTest(ProjectCreated.class, ProjectCreatedEnrichmentAnotherPackageFqn.class);
+        getEventTypeByEnrichmentTypeTest(ProjectCreated.class, ProjectCreatedEnrichmentAnotherPackageFqnAndMsgOpt.class);
+        getEventTypeByEnrichmentTypeTest(ProjectStarted.class, ProjectStarted.Enrichment.class);
+    }
 
+    private static void getEventTypeByEnrichmentTypeTest(Class<? extends Message> expectedEventClass,
```
alexander-yevsyukov: Please rename the method to `assertEventTypeByEnrichmentType`.

server/src/test/java/org/spine3/server/event/enrich/EventEnrichmentsMapShould.java
```diff
@@ -48,12 +54,19 @@ public void return_map_instance() {
     }
 
     @Test
-    public void return_event_type_by_enrichment_type_name() {
-        final TypeName enrichmentType = TypeName.of(ProjectCreated.Enrichment.getDescriptor());
+    public void contain_event_types_by_enrichment_type_names() {
```
alexander-yevsyukov: Please have separate methods for each case.

AlexanderLitus:
Done, PTAL.

alexander-yevsyukov:
LGTM

PR #159, opened 2016-07-06
Use EventContext for enrichment

AlexanderLitus:
Allow `context` reference in `by` field option.

### DIFF
--- a/client/src/main/proto/spine/base/event.proto
+++ b/client/src/main/proto/spine/base/event.proto
@@ -88,7 +88,11 @@ message EventContext {
 
     // Service attributes that the system wants to store in addition to the domain attributes
     // defined in the event message.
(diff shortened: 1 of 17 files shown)


client/src/main/proto/spine/base/event.proto
```diff
@@ -88,7 +88,11 @@ message EventContext {
 
     // Service attributes that the system wants to store in addition to the domain attributes
     // defined in the event message.
-    map<string, google.protobuf.Any> attributes = 8;
+    Attributes attributes = 8;
+
+    message Attributes {
+        map<string, google.protobuf.Any> attributes = 1;
```
alexander-yevsyukov: Please rename the field to `entries`.

alexander-yevsyukov:
Other than that, LGTM.

PR #161, opened 2016-07-07
Custom type url

alexander-yevsyukov:


alexander-yevsyukov:
This set of changes is preparation for using `type_url` file option when packing into `Any`.

alexander-yevsyukov:
@AlexanderLitus, please address TODOs added in the changes of this PR. 
Please also notice TODOs in the `EntityPackages` class and let's discuss.

client/src/main/java/org/spine3/protobuf/KnownTypes.java
```diff
@@ -204,9 +204,6 @@ private Builder putProperties(Properties properties) {
             return this;
         }
 
-        //TODO:2016-07-07:alexander.yevsyukov: Can we do what the below method does automatically?
-        // We can compose the same map on compile time by analyzing proto files packed with Protobuf's JAR.
-
         /**
```
alexander-yevsyukov: Please remind me, what was the reason for not doing it the way suggested in the TODO?

Maintaining Protobuf changes on our side can be done, but it's manual work, and we have to remember about it.

client/src/main/proto/spine/base/command.proto
```diff
@@ -26,6 +26,8 @@ option java_outer_classname = "CommandProto";
 option java_multiple_files = true;
 option java_generate_equals_and_hash = true;
 
+option (type_url_prefix) = "type.spine3.org";
```
alexander-yevsyukov: Please put this option immediately before the `java_package` option. No blank line is needed in between.

client/src/main/proto/spine/base/error.proto
```diff
@@ -26,8 +26,12 @@ option java_multiple_files = true;
 option java_outer_classname = "ErrorProto";
 option java_package = "org.spine3.base";
 
+option (type_url_prefix) = "type.spine3.org";
+
 import "google/protobuf/struct.proto";
+
 import "spine/validate/constraint_violation.proto";
+import "spine/annotations.proto";
```
alexander-yevsyukov: Import of annotations should be the first line in Spine-related block everywhere.

client/src/main/java/org/spine3/type/TypeName.java
```diff
@@ -144,6 +144,7 @@ private static String getTypeName(String typeUrl)
         return parts[1];
     }
 
+    //TODO:2016-07-07:alexander.yevsyukov: Support custom type URLs in this method.
```
alexander-yevsyukov: @AlexanderLitus, please address this TODO.

client/src/main/java/org/spine3/protobuf/AnyPacker.java
```diff
@@ -0,0 +1,85 @@
...
+    private AnyPacker() {}
+
+    /**
+     * Wraps {@link Message} object inside of {@link Any} instance.
+     *
+     * @param message message that should be put inside the {@link Any} instance.
+     * @return the instance of {@link Any} object that wraps given message.
+     */
+    public static Any pack(Message message) {
+        final Any result = Any.pack(message);
```
alexander-yevsyukov: @AlexanderLitus, here we need to see if the message has `type_url_prefix` option in its file and if yes call `Any.pack(Message, String)`.

client/src/main/java/org/spine3/protobuf/KnownTypes.java
AlexanderLitus: I remember that we found a way to get all Java classes from `protobuf` package using open-source Reflections lib. But then we removed this lib because we didn't want to depend on it because of one needed method. And we decided to just put needed proto classes (such as wrappers) to the map manually. 
Primarily, this was needed to support wrappers usage in tests. And that's why we put only wrappers to the map. Anyway, we should declare which classes to put (because we don't need all of them from that package), and the easiest way to do it is in `KnownTypes.java`.

AlexanderLitus: Also, using Reflections lib is much slower, and we need to have a fast startup.

client/src/main/proto/spine/base/command.proto
AlexanderLitus: Oh. I'd better delegate this stuff)

client/src/main/java/org/spine3/protobuf/TypeUrl.java
```diff
@@ -0,0 +1,207 @@
...
+
+    public String getPrefix() {
+        return prefix;
+    }
+
+    public String getTypeName() {
+        return typeName;
+    }
+
+    /** Returns the simple name of the Protobuf type (not fully qualified name). */
```
alexander-yevsyukov: It is better to give an example here. Saying “not fully qualified name” doesn't make it clearer.

client/src/main/java/org/spine3/protobuf/TypeUrl.java
```diff
@@ -0,0 +1,207 @@
...
+    private static final Pattern TYPE_URL_SEPARATOR_PATTERN = Pattern.compile(SEPARATOR);
+
+    private static final String PROTOBUF_PACKAGE_SEPARATOR = ".";
+    private static final Pattern PROTOBUF_PACKAGE_SEPARATOR_PATTERN = Pattern.compile('\\' + PROTOBUF_PACKAGE_SEPARATOR);
+
+    public static final String GOOGLE_TYPE_URL_PREFIX = "type.googleapis.com";
+
+    private static final String GOOGLE_PROTOBUF_PACKAGE = "google.protobuf";
+
+    private final String prefix;
```
alexander-yevsyukov: Is it type URL prefix?

client/src/main/java/org/spine3/protobuf/TypeUrl.java
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * @see Any#getTypeUrl()
+ * @see Descriptors.FileDescriptor#getFullName()
+ * @author Alexander Yevsyukov
+ */
+public final class TypeUrl extends StringTypeValue {
+
+    // TODO:2016-07-08:alexander.litus: implement toString() and equals()
+
+    public static final String SEPARATOR = "/";
```
alexander-yevsyukov: Why do we need it public?

client/src/main/java/org/spine3/protobuf/TypeUrl.java
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * Creates a new instance from the passed type URL or type name.
+     *
+     * @param typeUrlOrName the type URL of the Protobuf message type or its fully-qualified name
+     */
+    @Internal
+    public static TypeUrl of(String typeUrlOrName) { // TODO:2016-07-08:alexander.litus:  tests
+        checkNotEmptyOrBlank(typeUrlOrName, "type URL or name");
+        return isTypeUrl(typeUrlOrName) ?
```
alexander-yevsyukov: Please extract `result` variable.

client/src/main/java/org/spine3/protobuf/TypeUrl.java
AlexanderLitus: Made package-local.

client/src/main/java/org/spine3/protobuf/TypeUrl.java
AlexanderLitus: Added docs.

AlexanderLitus:
PTAL.

alexander-yevsyukov:
Renaming getter methods in `KnownTypes` was a right thing! Now the code reads better.
LGTM.

PR #164, opened 2016-07-19
Add and refactor tests, fix bugs

AlexanderLitus:


AlexanderLitus:
@armiol, PTAL. I'll add ClientService tests soon.

client/src/main/java/org/spine3/util/Exceptions.java
```diff
@@ -74,4 +76,30 @@ public static UnsupportedOperationException unsupported(String message) throws U
...
+    /**
+     * Sets a throwable as the cause of a {@link RuntimeException} and throws it.
+     *
+     * <p>Can be used instead of {@link Throwables#propagate(Throwable)} if it is needed to wrap the param anyway.
+     * Google's implementation throws {@link RuntimeException} and {@link Error} as-is.
+     *
+     * @param throwable to wrap
+     * @return always throws an exception, the return type is for convenience
+     */
+    public static RuntimeException wrapped(Throwable throwable) {
```
armiol: Let's discuss why we need that anyway.

client/src/main/java/org/spine3/validate/Validate.java
```diff
@@ -164,16 +164,22 @@ public static String checkNotEmptyOrBlank(String stringToCheck, String fieldName
...
-     * @throws IllegalArgumentException if the timestamp {@code seconds} value is less or equal to zero
+     * @throws IllegalArgumentException if any of the requirements are not met
      */
-    public static Timestamp checkTimestamp(Timestamp timestamp, String fieldName) {
-        checkNotNull(timestamp);
-        final long seconds = timestamp.getSeconds();
-        checkArgument(seconds > 0, fieldName + " is invalid.");
+    public static Timestamp checkIsPositive(Timestamp timestamp, String nameToLog) {
+        checkNotNull(timestamp, nameToLog + " is null.");
+        checkArgument(timestamp.getSeconds() > 0, nameToLog + " must have positive seconds.");
```
armiol: `"... must have positive number of seconds..."`

client/src/main/java/org/spine3/validate/Validate.java
```diff
@@ -164,16 +164,22 @@ public static String checkNotEmptyOrBlank(String stringToCheck, String fieldName
...
+        checkArgument(timestamp.getNanos() >= 0, nameToLog + " must have positive nanoseconds.");
```
armiol: `"... must not have negative number of seconds"`.

Please note the condition is NOT strict, i.e. `>=`, so you cannot call it `positive`.

client/src/test/java/org/spine3/protobuf/MessageFieldShould.java
```diff
@@ -20,18 +20,106 @@
...
+    private static final int STR_VALUE_FIELD_INDEX = 0;
+
+    @SuppressWarnings("DuplicateStringLiteralInspection")
+    public static final String STR_VALUE_FIELD_NAME = "value";
+
+    private final StringValue stringValue = newStringValue(newUuid());
+
+    @Test
+    public void have_index() {
+        final int index = 5;
```
armiol: Do you check for a numeric value to pass? Why do you call it `index` then? It looks like a proto field index, judging by its name.

client/src/test/java/org/spine3/protobuf/MessageFieldShould.java
```diff
@@ -20,18 +20,106 @@
...
+        final TestMessageField field = new TestMessageField(STR_VALUE_FIELD_INDEX);
+        field.setIsFieldAvailable(false);
+
+        field.getValue(stringValue);
+    }
+
+    @Test(expected = ArrayIndexOutOfBoundsException.class)
+    public void throw_exception_if_no_field_by_given_index() {
+        @SuppressWarnings("MagicNumber")
+        final TestMessageField field = new TestMessageField(256);
```
armiol: How do you know `256` is big enough?

client/src/test/java/org/spine3/protobuf/TimestampsShould.java
```diff
@@ -48,14 +49,24 @@ public void have_private_constructor() {
         assertTrue(hasPrivateUtilityConstructor(Timestamps.class));
     }
 
+    @Test
+    public void do_not_throw_exception_if_timestamp_is_valid() {
```
armiol: Following the Should paradigm, I would name this method `not_throw_exception...`.

In your case the test reads `Timestamps should do not throw ...`, which is does not make much sense in English.

server/src/main/java/org/spine3/server/storage/memory/InMemoryCommandStorage.java
```diff
@@ -137,6 +129,15 @@ private void put(CommandId id, CommandStorageRecord record) {
...
     private CommandStorageRecord get(CommandId id) {
-        return storage.get(id);
+        final CommandStorageRecord record = storage.get(id);
+        if (record == null) {
+            return CommandStorageRecord.getDefaultInstance();
+        }
+        return record;
+    }
+
+    private static CommandStorageRecord check(CommandStorageRecord record, CommandId id) {
```
armiol: Let bring some specifics into this `check` method name.

server/src/test/java/org/spine3/server/aggregate/AggregateRepositoryShould.java
```diff
@@ -72,17 +107,148 @@ public void return_default_aggregate_instance_if_no_aggregate_found() {
...
+                          .store(any(ProjectAggregate.class));
+
+        repositorySpy.dispatch(cmd);
+
+        verify(commandStore).updateStatus(getId(cmd), exception);
+    }
+
+    @Test
+    public void set_cmd_status_to_error_if_got_exception_on_dispatching() {
+        final Command cmd = Given.Command.createProject(); // TODO:2016-07-22:alexander.litus: DRY
```
armiol: Please process this TODO item before PR is closed.

server/src/test/java/org/spine3/server/bc/BoundedContextShould.java
```diff
@@ -149,11 +146,25 @@ public void notify_integration_event_subscribers() {
     }
 
     @Test
+    @SuppressWarnings("unchecked") // have to cast to use Mockito
```
armiol: Let's suppress this warning for a statement only, not for a whole method.

server/src/test/java/org/spine3/server/command/CommandHandlerShould.java
```diff
@@ -0,0 +1,217 @@
...
+        assertHandles(Given.Command.startProject());
+    }
+
+    @Test
+    public void post_generated_events_to_event_bus() {
+        final Command cmd = Given.Command.startProject();
+
+        handler.handle(cmd);
+
+        final int postedEventCount = 2;
```
armiol: How do you know there is going to be `2`?

server/src/test/java/org/spine3/server/entity/EntityShould.java
```diff
@@ -185,11 +185,81 @@ public void return_id_protobuf_type_name() {
...
     }
 
-    private static Project newProject() {
-        final Project.Builder project = Project.newBuilder()
-                                               .setId(Given.AggregateId.newProjectId())
-                                               .setStatus(Project.Status.CREATED);
-        return project.build();
+    @Test
+    public void generate_non_zero_hash_code() {
+        final int hashCode = entityWithState.hashCode();
```
armiol: What is wrong with the `hashCode()` returning zero? 

Besides, how do you check all the possible source object states to ensure none of them gives zero as the `hashCode()` invocation result?

server/src/test/java/org/spine3/server/event/EventBusShould.java
```diff
@@ -302,10 +291,31 @@ public void do_not_enrich_event_if_it_cannot_be_enriched() {
...
+        final Throwable cause = responseObserver.getThrowable().getCause();
+        assertEquals(exceptionClass, cause.getClass());
+        assertNull(responseObserver.getResponse());
+    }
+
+    private static class ProjectCreatedSubscriber extends EventSubscriber {
+
+        private Event eventHandled;
 
+        // OK for the test class.
```
armiol: What is OK?

server/src/test/java/org/spine3/server/projection/ProjectionRepositoryShould.java
```diff
@@ -188,6 +142,41 @@ public void return_entity_storage() {
...
+    @Test
+    public void updates_status_to_CLOSED_on_close() throws Exception {
+        repository.close();
+
+        assertEquals(CLOSED, repository.getStatus());
+    }
+
+    @Test
+    public void return_true_if_status_is_ONLINE() {
+        assertTrue(repository.isOnline());
```
armiol: Consider adding a comment, that you've set it `online` in `@Before`.

server/src/test/proto/spine/test/aggregate/events.proto
```diff
@@ -28,10 +28,10 @@ option java_multiple_files = true;
 
 import "spine/annotations.proto";
 import "spine/test/aggregate/project.proto";
-import "spine/test/aggregate/task.proto";
 
 message ProjectCreated {
     ProjectId project_id = 1;
+    string name = 2;
```
armiol: Looks strange. I would understand the fact that just the `CreateProject` command has a name _and_ ID (which is also suspicious). But why the event has both?

armiol:
@AlexanderLitus Let's discuss some of my comments in person.

client/src/test/java/org/spine3/protobuf/MessageFieldShould.java
AlexanderLitus: It is really big, but I can use a bigger number to be 300% sure...

server/src/test/java/org/spine3/server/bc/BoundedContextShould.java
AlexanderLitus: Extracted a method with suppression.

server/src/test/proto/spine/test/aggregate/events.proto
AlexanderLitus: Yeah, but imagine we have the case when the client sets IDs (actually, it would be rather painful to get rid of the ID in this command, as it is the most popular one in tests).

AlexanderLitus: And the client can also set the project's name.

server/src/test/java/org/spine3/server/event/EventBusShould.java
AlexanderLitus: I guess this is an old comment about some suppression.

server/src/test/java/org/spine3/server/entity/EntityShould.java
AlexanderLitus: In this case, a hash code is generated using an ID, state, etc. An ID is a non-empty String, therefore, the hash code of the entity cannot be zero. 
I guess it is better to rename a method to `generate_non_zero_hash_code_if_entity_has_non_empty_id_and_state`.

client/src/test/java/org/spine3/protobuf/MessageFieldShould.java
AlexanderLitus: I think `Integer.MAX_VALUE` would be enough =)

client/src/test/java/org/spine3/protobuf/MessageFieldShould.java
AlexanderLitus: Yes, it is the index of the Protobuf message.

AlexanderLitus:
PTAL.

server/src/test/java/org/spine3/server/bc/BoundedContextShould.java
```diff
@@ -170,6 +171,29 @@ public void tell_if_set_multitenant() {
...
 
+    @Test
+    public void assign_storage_during_registration_if_repository_does_not_have_storage() {
+        final ProjectAggregateRepository repository = new ProjectAggregateRepository(boundedContext);
+        boundedContext.register(repository);
+        assertTrue(repository.storageAssigned());
+    }
+
+    @Test
+    public void do_not_change_storage_during_registration_if_a_repository_has_one() {
```
armiol: Please remove `do` in the method name.

armiol:
@AlexanderLitus LGTM except for one comment.

PR #171, opened 2016-08-12
Move examples in a separate repo

alexander-yevsyukov:
Examples have been moved to https://github.com/SpineEventEngine/examples-java.


alexander-yevsyukov:
@armiol, it should be noted that sub-projects with the `sample` suffix in the `tools` repository are not examples. They are test data for the plugins. I will apply proper naming in that repository. A separate problem is making those tests work without deploying to a Maven repository.

armiol:
@alexander-yevsyukov LGTM.

PR #172, opened 2016-08-22
UserAccount and related types

alexander-yevsyukov:
This PR does not implement services defined in Protobuf. It is expected that after the types are finalized these services will be added in a separate PR.


alexander-yevsyukov:
These `UserAccount` and other message types define most of the attributes that we needed in Projects v1. In Projects these attributes were gathered from GApps, GAE, and other Google APIs (e.g. `com.google.gdata.data.appsforyourdomain.Login`, `com.google.appengine.api.users.UserService`, etc.) 

The layout is different, though. The idea of adding linked account identities is borrowed from `FirebaseUser` and `UserInfo`.

client/src/test/java/org/spine3/users/UserAccountsShould.java
```diff
@@ -18,38 +18,16 @@
...
-
-        assertEquals(expected, userId);
-    }
-
-    @Test(expected = NullPointerException.class)
-    public void do_not_accept_null_UseId_value() {
-        newUserId(Tests.<String>nullRef());
-    }
-}
+}
```
armiol: Newline is missing.

armiol:
@alexander-yevsyukov as discussed, some `User`-related code should probably be extracted as a separate module. Also, the services should probably be built on top of Spine read-side `QueryService` and `SubscriptionService`.

Other than that, LGTM.

PR #182, opened 2016-09-22
Require alternative fields

alexander-yevsyukov:


client/src/main/proto/spine/validate.proto
```diff
@@ -41,57 +41,112 @@ import "google/protobuf/descriptor.proto";
...
+    // Unlike the `required` field option which always require corresponding field,
+    // this message option allows to require alternative fields.
+    //
+    // Field names are separated using the pipe (`|`) symbol. The combination of fields is defined
+    // using the ampersand (`&`) symbol.
+    //
+    // For example:
+    //
+    // message PersonName {
+    //    option (required_field) = "given_name|honorific_prefix & family_name";
```
armiol: @alexander-yevsyukov Let's discuss this.  I would suggest adding parenthesises. I.e.

`given_name|honorific_prefix & family_name` is different from `given_name| (honorific_prefix & family_name)`.

server/src/main/java/org/spine3/server/validate/AlternativeFieldValidator.java
```diff
@@ -0,0 +1,241 @@
...
+            }
+            return fieldName;
+        }
+
+        /* package */ ImmutableList<String> getFieldNames() {
+            if (fieldNames == null) {
+                throw new IllegalStateException("The option is not a combination, but a single field.");
+            }
+            // It is OK to suppress as we're using ImmutableList.
+            //noinspection ReturnOfCollectionOrArrayField
```
armiol: I would suppress this warning with the annotation to the method name.

armiol:
@alexander-yevsyukov please refer to a single comment I left.

server/src/main/java/org/spine3/server/validate/AlternativeFieldValidator.java
alexander-yevsyukov: Done

armiol:
@alexander-yevsyukov LGTM.

PR #191, opened 2016-10-07
Add convenience wrapper for getting strings from `ValueMismatch`.

alexander-yevsyukov:


client/src/test/java/org/spine3/base/MismatchShould.java
```diff
@@ -83,7 +85,6 @@ public void return_mismatch_object_with_string_values() {
     public void return_mismatch_object_with_int32_values() {
         final int expected = 0;
         final int actual = 1;
-        final int requested = 2;
```
dmytro-kashcheiev: If we remove unnecessary 'requsted' code from class. 
Maybe remove `private static final String REQUESTED` as well.

alexander-yevsyukov: Well spotted. Fixed.

armiol:
@alexander-yevsyukov LGTM.

PR #198, opened 2016-10-12
Field masks cleanup 

dmdashenkov:
No coverage improvement can be provided for `FieldMasks` at the moment since `catch` blocks are not to be executed within the standard flow. The only possibility I see for those pieces of code to be run is if the protobuf definitions are compiled improperly.


dmdashenkov:
@armiol PTAL

PR #200, opened 2016-10-14
Update version to 0.6.0.

armiol:


armiol:
@alexander-yevsyukov PTAL.

PR #204, opened 2016-10-19
Sharing extensions

alexander-yevsyukov:
This PR defines only Protobuf types. Construction and validation of corresponding Java objects would be done in a separate PR.


users/src/main/proto/spine/sharing/sharing.proto
```diff
@@ -31,45 +31,103 @@ import "spine/annotations.proto";
...
+message IndirectPermission {
+
+    // A level of visibility.
+    Visibility visibility = 1;
+
+    // A level of access.
+    AccessLevel level = 2;
+}
+
+// Access level for one subject.
```
armiol: `...for a single subject` would read better.

users/src/main/proto/spine/sharing/sharing.proto
```diff
@@ -31,45 +31,103 @@ import "spine/annotations.proto";
...
+//
+// An object can be shared to a wide audience (such as organization or all users of the system),
+// or only to specific people.
+//
+// If a subject is not added to a sharing schema, the level is `AccessLevel.NONE`.
 //
-// If a user or a group is not added to the sharing schema, the level is [SharingLevel.NONE].
-message Sharing {
-    repeated Permission permissions = 2;
+message SharingSchema {
```
armiol: To my mind, it looks more like a `SharingRule` or `SharingPolicy`.

As I feel, `Schema` term defines a wider aspect. Such as "what rules can be applied to a particular organization".

users/src/main/proto/spine/sharing/sharing.proto
alexander-yevsyukov: Changed.

PR #205, opened 2016-10-19
Minor fix of test data

dmdashenkov:
Set event ID to `AggregateStorageRecord` when generating test data.


dmdashenkov:
@armiol PTAL.
After that I'll reassign this to @alexander-yevsyukov if it's needed.

PR #206, opened 2016-10-19
Allow taking sharing settings from parent object

alexander-yevsyukov:


users/src/main/proto/spine/sharing/sharing.proto
```diff
@@ -117,17 +117,30 @@ message Permissions {
...
 // If a subject is not added to sharing settings of an object, the level is `AccessLevel.NONE`.
 //
 message SharingSettings {
-    oneof audience {
-        // All users of the system can access.
-        IndirectPermission public = 1;
 
-        // All organization members can access.
-        IndirectPermission organization = 2;
+    oneof mode {
```
armiol: As discussed, we need to include both `take_from_parent` and `settings` simultaneously.

This will allow us to simulate Drive functionality by overriding the parent `SharingSettings` on a child object.

PR #209, opened 2016-11-14
Simplified Field Filter

dmdashenkov:
Now we use `string field_path` instead of `FildMask field` hence the field must be only one.


dmdashenkov:
@armiol PTAL

dmdashenkov:
@armiol As I said, we don't even use this filters yet. There are no actual usages of class `FieldFilter` besides the `EventFilter` internal generated by protobuf plugin usages.

dmdashenkov:
@armiol, please, see my response and proceed your review.

armiol:
@dmdashenkov LGTM.

PR #210, opened 2016-11-16
Make `CommandBus` configurable

armiol:
Summary of changes:

* Migrate to `Builder`-based approach from `CommandBus#newInstance()`.
* Remove hard-coded thread spawning and replace it with a configuration option.

### DIFF
--- a/server/src/main/java/org/spine3/server/command/CommandBus.java
+++ b/server/src/main/java/org/spine3/server/command/CommandBus.java
@@ -502,8 +502,10 @@ public class CommandBus implements AutoCloseable {
 
         private Builder() {}
 
+        /**
+         * Builds an instance of {@link CommandBus}.
+         */
         public CommandBus build() {
-
             checkNotNull(commandStore, "CommandStore must be set");
 
             if(commandScheduler == null) {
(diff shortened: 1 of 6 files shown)


armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -103,35 +103,50 @@
...
     /**
-     * Creates a new instance.
+     * Determines whether the manual thread spawning is allowed within current runtime environment.
      *
-     * @param commandStore a store to save commands
-     * @return a new instance
+     * <p>If set to {@code true}, {@code CommandBus} will be running some of internal processing in parallel
+     * to improve performance.
+     */
+    private final boolean isThreadSpawnAllowed;
```
alexander-yevsyukov: Why do we need to name the field in the same way as its getter?

alexander-yevsyukov: As discussed vocally, let's keep it the way it is because it's consistent with `isMultitenant`. We can revisit two names later, when making `isMultitenant` `final`.

PR #214, opened 2016-11-21
Migrate to Guava 20.0.

armiol:
Summary of changes:

* migrated to Guava 20.0;
* code usages have been updated;
* Spine version was bumped to `0.6.4-SNAPSHOT`.

### DIFF
--- a/server/src/test/java/org/spine3/server/storage/AbstractStorageShould.java
+++ b/server/src/test/java/org/spine3/server/storage/AbstractStorageShould.java
@@ -20,18 +20,17 @@
 
 package org.spine3.server.storage;
 
-import com.google.common.base.Throwables;
 import com.google.protobuf.Message;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.spine3.test.Tests;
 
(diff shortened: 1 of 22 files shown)


armiol:
@dmdashenkov PTAL.

server/src/test/java/org/spine3/server/storage/AbstractStorageShould.java
```diff
@@ -73,15 +72,13 @@ public void tearDownAbstractStorageTest() {
 
     /**
      * Closes the storage and propagates an exception if any occurs.
-     *
-     * @see Throwables#propagate(Throwable)
```
dmdashenkov: Maybe we should keep `@see` block with updated API usage?

armiol: `Throwables#propagate` will be removed in the 21.0.

That's why we don't mention it anymore.

armiol:
@dmdashenkov updated a single usage of `new RuntimeException(..)` with `Exceptions.wrapped(..)`.

Please see my answer inline as well.

PTAL.

PR #215, opened 2016-11-21
Aggregate state double-checking

armiol:
Summary of changes:

* ensure the event count (after the last snapshot) is the same during the command dispatching. Do not allow to proceed to the `Aggregate` state update until there are no new events from concurrent actors.
* bump framework version to `0.6.5-SNAPSHOT`.

### DIFF
--- a/server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
+++ b/server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
@@ -21,6 +21,8 @@ package org.spine3.server.aggregate;
 
 import com.google.protobuf.Any;
 import com.google.protobuf.Message;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.spine3.base.Command;
 import org.spine3.base.CommandContext;
 import org.spine3.base.CommandId;
@@ -159,7 +161,8 @@ public abstract class AggregateRepository<I, A extends Aggregate<I, ?, ?>>
     @Nonnull
     @Override
     public A load(I id) throws IllegalStateException {
-        final AggregateEvents aggregateEvents = aggregateStorage().read(id);
+        final AggregateStorage<I> aggregateStorage = aggregateStorage();
+        final AggregateEvents aggregateEvents = aggregateStorage.read(id);
         final Snapshot snapshot = aggregateEvents.getSnapshot();
         final A result = create(id);
         final List<Event> events = aggregateEvents.getEventList();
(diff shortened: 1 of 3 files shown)


armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
```diff
@@ -243,6 +231,55 @@ public void dispatch(Command request) throws IllegalStateException {
         commandStatusService.setOk(commandId);
     }
 
+    @SuppressWarnings("ChainOfInstanceofChecks")        // it's a rare case of handing an exception, so we are OK.
```
alexander-yevsyukov: Can we make it somewhat closer to where it happens?

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
```diff
@@ -243,6 +231,55 @@ public void dispatch(Command request) throws IllegalStateException {
...
+                    commandStatusService.setToFailure(commandId, failure);
+                } else {
+                    commandStatusService.setToError(commandId, Errors.fromThrowable(cause));
+                }
+            }
+
+            eventCountBeforeSave = aggregateStorage.readEventCountAfterLastSnapshot(aggregateId);
+        } while (eventCountBeforeDispatch != eventCountBeforeSave);
+
+        checkNotNull(aggregate);
```
alexander-yevsyukov: Don't we need a message explaining why it might happen?

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
```diff
@@ -243,6 +231,55 @@ public void dispatch(Command request) throws IllegalStateException {
...
+                                   "Repeating the command dispatching.",
+                           command, getAggregateClass(), aggregateId);
+            }
+            eventCountBeforeDispatch = aggregateStorage.readEventCountAfterLastSnapshot(aggregateId);
+            aggregate = load(aggregateId);
+            try {
+                aggregate.dispatch(command, context);
+            } catch (RuntimeException e) {
+                final Throwable cause = e.getCause();
+                if (cause instanceof Exception) {
```
alexander-yevsyukov: This block looks like a typical scenario, which is applicable to PMs too. Don't we have code dup here?

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
armiol: Previously the suppression looked as `//noinspection ...`. It was declared closer to the `instanceof` chain. As we are aiming not only for IDEA codestyle tweaks (`//noinspection` works fine), but also for tools like PMD, we need to use a standard-based suppression mechanism, which is `@SuppressWarnings(..)`.

However, it was not possible just to swap the previously declared `//noinspection` with a descent `@SuppressWarnings(...)` at the same spot due to Java syntax rules. 

That's why I moved the `@SuppressWarnings` to the next best place, which is the method annotation.

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
armiol: As turned out, this check in fact is ambiguous. `aggregate` cannot be null at this point.

Statement removed.

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
armiol: At this point `ProcessManagerRepository` does not have any special error handling in scope of a command dispatching.

As a separate PR it makes sense to introduce such a handling. In addition, the `CommandStatusService` will probably be used for PM repositories.

armiol:
Number of "new" events has been included into the warning message. Comments addressed.

@alexander-yevsyukov PTAL.

PR #217, opened 2016-11-23
Allow `ProcessManager` have no commands handled

armiol:
As long as `ProcessManager`s are designed for inter-communication between `Aggregate`s, some flows may be triggered by a business event, not a command. 

However, the framework did not allow to register dispatchers with no command handler methods.

In scope of this PR, we enable `ProcessManager`s registration even with no command handled. We keep registering those in the bus to keep the `CommandDispatcher` processing consistent for all subclasses of `CommandDispatcher`s.


armiol:
@dmytro-kashcheiev PTAL.

PR #218, opened 2016-11-23
Supply the storage record with a proper `EventContext` instance.

armiol:
Summary of changes:

* The default event context was used for `EventStorage` tests previously.
* `GCD` implementation requires proper event producer set for the `EventStorageRecord`'s event context.
* Proper event context with a valid `producerId` was set to allow `GCD` tests pass.


armiol:
@dmdashenkov PTAL.

PR #220, opened 2016-12-15
Add factory methods, tests and messages for Changes.

alexander-aleksandrov:


client/src/main/proto/spine/base/change.proto
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Definition of a change in a `bool` field.
+message BooleanChange {
+
+    // The value of the field that's changing.
+    bool previous_value = 1;
+
+    // The new value of the field.
+    bool new_value = 2 [(required) = true];
```
alexander-yevsyukov: How can it be missing?

client/src/main/java/org/spine3/base/Changes.java
```diff
@@ -0,0 +1,291 @@
...
+
+    /**
+     * Creates {@link StringChange} object for the passed previous and new field values.
+     *
+     * <p>Passed values cannot be equal.
+     */
+    public static StringChange of(String previousValue, String newValue) {
+        checkNotNull(previousValue, PREVIOUS_VALUE);
+        checkNotNull(newValue, NEW_VALUE);
+        checkArgument(!newValue.isEmpty(), "newValue cannot be empty");
```
alexander-yevsyukov: Please add test for this line and all those that appear yellow (partially tested).

alexander-yevsyukov: Resolved. We have tests. It's a plug-in bug.

client/src/main/proto/spine/base/change.proto
alexander-aleksandrov: Fixed

PR #222, opened 2016-12-15
Thread safety for core singletons.

armiol:
Summary of changes:

- `ConverterRegistry` and `DefaultStateRegistry` made thread-safe.
- `CommandValidator.LogSingleton` renamed to `... Singleton` to reflect its nature.


armiol:
@alexander-yevsyukov PTAL.

PR #226, opened 2016-12-21
Extend `ValueMismatch` with `newValue`

alexander-yevsyukov:
This PR extends the `ValueMismatch` message with `newValue` field that contains value of a field requested in a command requesting data mutation.


client/src/main/java/org/spine3/change/Mismatch.java
```diff
@@ -83,13 +91,15 @@ public static ValueMismatch of(int expected, int actual, int version) {
      *
      * @param expected the value expected by a command
      * @param actual   the value actual in an entity
-     * @param version  the current version of the entity
+     * @param newValue the value from a command, which we wanted to set instead of {@code expected}
+     * @param version  the current version of the entity  @return info on the mismatch
```
alexander-aleksandrov: `@return info on the mismatch ` is used twice.

client/src/main/java/org/spine3/change/Mismatch.java
```diff
@@ -147,30 +163,35 @@ public static ValueMismatch of(boolean expected, boolean actual, int version) {
...
         final ValueMismatch.Builder builder = ValueMismatch.newBuilder();
         if (expected != null) {
-            builder.setExpected(AnyPacker.pack(expected));
+            builder.setExpectedPreviousValue(AnyPacker.pack(expected));
         }
         if (actual != null) {
-            builder.setActual(AnyPacker.pack(actual));
+            builder.setActualPreviousValue(AnyPacker.pack(actual));
+        }
+        if (newValue != null) {
```
alexander-aleksandrov: There is no test for this row. Maybe there should be `set_default_new_value_if_it_was_passed_as_null`.

client/src/main/java/org/spine3/change/Mismatch.java
```diff
@@ -99,13 +109,15 @@ public static ValueMismatch of(long expected, long actual, int version) {
      *
      * @param expected the value expected by a command
      * @param actual   the value actual in an entity
-     * @param version  the current version of the entity
+     * @param newValue the value from a command, which we wanted to set instead of {@code expected}
+     * @param version  the current version of the entity  @return info on the mismatch
```
alexander-aleksandrov: Here the same @return is stated twice.

client/src/main/java/org/spine3/change/Mismatch.java
```diff
@@ -147,30 +163,35 @@ public static ValueMismatch of(boolean expected, boolean actual, int version) {
      *
      * @param expected the value expected by a command, or {@code null} if the command expects not populated field
      * @param actual   the value actual in an entity, or {@code null} if the value is not set
-     * @param version  the current version of the entity
-     * @return info on the mismatch
+     * @param newValue the value from a command, which we wanted to set instead of {@code expected}
+     *@param version  the current version of the entity  @return info on the mismatch
```
alexander-aleksandrov: Probably should be formatted on the next row.

PR #228, opened 2016-12-23
Mismatch improvements

alexander-yevsyukov:

PR #229, opened 2016-12-25
CommandBus improvement

alexander-yevsyukov:
This PR builds on #228. The main idea is to move larger methods into helper classes.

### DIFF
--- a/server/src/main/java/org/spine3/server/command/CommandBus.java
+++ b/server/src/main/java/org/spine3/server/command/CommandBus.java
@@ -21,17 +21,15 @@ package org.spine3.server.command;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Sets;
-import com.google.protobuf.Duration;
 import com.google.protobuf.Message;
-import com.google.protobuf.Timestamp;
 import io.grpc.stub.StreamObserver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.spine3.Internal;
 import org.spine3.base.Command;
 import org.spine3.base.CommandContext;
-import org.spine3.base.CommandContext.Schedule;
 import org.spine3.base.CommandId;
+import org.spine3.base.Error;
 import org.spine3.base.Errors;
 import org.spine3.base.FailureThrowable;
 import org.spine3.base.Response;
@@ -43,28 +41,19 @@ import org.spine3.server.command.error.InvalidCommandException;
 import org.spine3.server.command.error.UnsupportedCommandException;
 import org.spine3.server.type.CommandClass;
 import org.spine3.server.users.CurrentTenant;
-import org.spine3.time.Interval;
 import org.spine3.users.TenantId;
 import org.spine3.util.Environment;
 import org.spine3.validate.ConstraintViolation;
 
 import java.lang.reflect.InvocationTargetException;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.protobuf.util.Timestamps.add;
 import static org.spine3.base.CommandStatus.SCHEDULED;
 import static org.spine3.base.Commands.getId;
 import static org.spine3.base.Commands.getMessage;
 import static org.spine3.base.Commands.isScheduled;
-import static org.spine3.base.Commands.setSchedule;
-import static org.spine3.protobuf.Timestamps.getCurrentTime;
-import static org.spine3.protobuf.Timestamps.isLaterThan;
-import static org.spine3.server.command.error.CommandExpiredException.commandExpiredError;
-import static org.spine3.time.Intervals.between;
-import static org.spine3.time.Intervals.toDuration;
 import static org.spine3.validate.Validate.isDefault;
 
 /**
@@ -89,7 +78,7 @@ public class CommandBus implements AutoCloseable {
 
     private final CommandScheduler scheduler;
 
-    private final CommandRescheduler rescheduler;
+    private final Rescheduler rescheduler;
 
     private final ProblemLog problemLog;
 
@@ -144,7 +133,7 @@ public class CommandBus implements AutoCloseable {
         this.commandStatusService = new CommandStatusService(commandStore);
         this.scheduler = scheduler;
         this.problemLog = problemLog;
-        this.rescheduler = new CommandRescheduler();
+        this.rescheduler = new Rescheduler(this);
         this.isThreadSpawnAllowed = threadSpawnAllowed;
     }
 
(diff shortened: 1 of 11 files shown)


alexander-yevsyukov:
@armiol please pay attention to the commit on injection of `CommandBus` into `CommandScheduler` (a1bab4cc3d51c2c0178c0cb30209164405044356).

I discovered that we don't do this only by analyzing usages in IDEA. We don't have a test that checks that this wiring should be done. I have not added the test. Let's discuss first.

client/src/main/java/org/spine3/change/DoubleMismatch.java
```diff
@@ -0,0 +1,55 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.change;
+
+import static org.spine3.protobuf.Values.pack;
+
+/**
+ * Utility class for working with {@code int} values in {@link ValueMismatch}es.
```
armiol: I think, it should be `{@code double}`.

client/src/main/java/org/spine3/change/DoubleMismatch.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Creates a ValueMismatch instance for a double attribute.
+     *
+     * @param expected the value expected by a command
+     * @param actual   the value actual in an entity
+     * @param newValue the value from a command, which we wanted to set instead of {@code expected}
+     * @param version  the current version of the entity
+     * @return new ValueMismatch instance
+     */
+    public static ValueMismatch of(double expected, double actual, double newValue, int version) {
```
armiol: Do we intentionally avoid `Double` and use `double` instead? What if the expected value is `null`?
If we could use a type instead of a primitive, it should then be possible to have a type hierarchy with a generic type, e.g. 
`NumberMismatch<TYPE extends Number>` 
and `DoubleMismatch extends NumberMismatch<Double>`

In this way we could save up some effort by reusing `unpacking methods`.

The same question applies to every `...Mismatch` except for `boolean`.

client/src/main/java/org/spine3/change/MessageMismatch.java
```diff
@@ -0,0 +1,129 @@
...
+import com.google.protobuf.Any;
+import com.google.protobuf.Message;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.change.Mismatches.checkNotNullOrEqual;
+import static org.spine3.protobuf.AnyPacker.pack;
+import static org.spine3.protobuf.AnyPacker.unpack;
+
+/**
+ * Utility class for working with ValueMismatches for {@code Message} values.
```
armiol: Please wrap the `ValueMismatches` with a `{@link}` to be consistent with the related class-level JavaDocs.

client/src/test/java/org/spine3/change/MessageMismatchShould.java
```diff
@@ -0,0 +1,84 @@
...
+        final ValueMismatch mismatch = expectedDefault(ACTUAL, NEW_VALUE, VERSION);
+
+        assertEquals(DEFAULT_VALUE, unpackExpected(mismatch));
+        assertEquals(ACTUAL, unpackActual(mismatch));
+        assertEquals(NEW_VALUE, unpackNewValue(mismatch));
+        assertEquals(VERSION, mismatch.getVersion());
+    }
+
+    @Test
+    public void creates_instance_for_unexpected_default_value() {
```
armiol: Should be `create_...`.
Since we use `...Should`, `creates` does not go with this verb.

client/src/test/java/org/spine3/change/StringMismatchShould.java
```diff
@@ -0,0 +1,95 @@
...
+        final StringValue newValue = unpack(mismatch.getNewValue());
+
+        assertEquals(EXPECTED, expected.getValue());
+        assertEquals(ACTUAL, actual.getValue());
+        assertEquals(NEW_VALUE, newValue.getValue());
+        assertEquals(VERSION, mismatch.getVersion());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void do_not_accept_same_expected_and_actual() {
```
armiol: Should be `not_accept...`, as the sentence will read `StringMismatch should ... not accept same expected and actual`.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -326,18 +277,36 @@ public void run() {
...
 
-    /** Obtains the instance of the {@link CommandStatusService} associated with this command bus. */
+    /**
+     * Obtains the instance of the {@link CommandStatusService} associated with this command bus.
+     */
     public CommandStatusService getCommandStatusService() {
         return commandStatusService;
     }
 
+    /* package */ boolean isMultitenant() {
```
armiol: As long as we allow building `CommandBus` instances via public API, the `CommandBus.Builder` getters should be `public` as well. 

Otherwise, the users of the API will not be able to tell, what value the instance of `CommandBus.Builder` has.

In more details.

I imagine the `CommandBus.Builder` instance may be build in several code modules — in case of complex dev-staging-beta-producton deployment implementation. And the `Builder` instance may then be passed from one piece of code to another. And it is extremely helpful to understand its state. That's why the getters should be `public`.

armiol:
@alexander-yevsyukov regarding the injection.

I would say we should inject the instance of the `CommandBus` _only_ in case it has not been injected previously by a Spine end-user. Otherwise we will override the rescheduling mechanism the user might set.

In particular, the rescheduling of command in GAE `classic` environment may be tricky — due to the request processing constraints (30 sec per HTTP request). So the best way would be to use a deferred task instead (10 mins to handle the deferred task). 

The custom `CommandBus` instance may serve well — GAE-specific wrapper around the `CommandBus` — being injected into a `rescheduler` by a user.

client/src/main/java/org/spine3/change/DoubleMismatch.java
alexander-yevsyukov: There's a separate PR (#228) on improving mismatches. This PR is based on that. I thought of such an option, but wanted to avoid object wrappers for simplicity. Let's discuss this matter separately.

client/src/main/java/org/spine3/change/MessageMismatch.java
alexander-yevsyukov: I saw how it looks in real Javadoc, and it's too many links if we do so. The link is available in the method signature, so I'll wrap in `{@code }`

server/src/main/java/org/spine3/server/command/CommandBus.java
alexander-yevsyukov: We don't have `setMultitenant()` in the `Builder`.  We have the internal method in `CommandBus` for setting (because it's invoked by `BoundedContext`). And the query method is package access, because we call it from helper classes (in the same package).

alexander-yevsyukov:
@armiol PTAL

client/src/main/java/org/spine3/change/DoubleMismatch.java
armiol: As discussed vocally, the code will be kept simple for now.

Also, as Protobuf values are never `null`, no special handling for nullable expected or actual values is needed.

armiol:
@alexander-yevsyukov LGTM.

PR #231, opened 2016-12-27
Resolve style and error-prone issues found by Codacy and IDEA.

armiol:
Summary of changes:

* Removed redundant `import` statements; adjusted usages of `static import ...` where spotted.
* Avoided addressing a class via FQN when not in need.
* Fixed `switch..case` statements with `default` label where applicable.
* Refactored some parts of the code to avoid duplication.
* Substituted IDEA comment-based warning suppressions with a standard annotation-based suppressions.
* Updated class member declaration order: static fields and constants, fields, constructors, methods.
* Fixed some spelling issues and typos.


armiol:
@alexander-yevsyukov PTAL.

PR #232, opened 2016-12-27
Propagate Entity versions to the Stand

armiol:
Summary of changes:

* Entity state version is now passed to the `Stand` via `StandFunnel`. That allows to store the version properly in the `StandStorage`.
* Bump Spine version to `0.6.13-SNAPSHOT`.


armiol:
@alexander-yevsyukov PTAL.

PR #238, opened 2016-12-28
ProjectionRepository automatic catch-up.

armiol:
Summary of changes:
* Turn on automatic catch-up for `ProjectionRepository` instances the storage is initialized. 
* Make this option configurable by introducing an alternative `ProjectionRepository` constructor.
* Add tests for the behaviour changes.

### DIFF
--- a/server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
+++ b/server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
@@ -152,7 +152,9 @@ public abstract class ProjectionRepository<I, P extends Projection<I, M>, M exte
         setStatus(Status.STORAGE_ASSIGNED);
 
         if(catchUpAfterStorageInit) {
-            log().debug("Storage assigned. Starting catch-up.");
+            if(log().isDebugEnabled()) {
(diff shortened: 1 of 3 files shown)


armiol:
@alexander-yevsyukov PTAL.

armiol:
@alexander-yevsyukov the PR is now complete. Please feel free to review.

server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
```diff
@@ -123,6 +150,13 @@ protected Storage createStorage(StorageFactory factory) {
     public void initStorage(StorageFactory factory) {
         super.initStorage(factory);
         setStatus(Status.STORAGE_ASSIGNED);
+
+        if(catchUpAfterStorageInit) {
+            if(log().isDebugEnabled()) {
```
alexander-yevsyukov: Why do we check the level here? We do not pass parameters to the log.

armiol: I see your point. I have added some parameters to let a developer know which `ProjectionRepository` is catching up.

PR #252, opened 2017-01-04
Allow multiple projection IDs for an event

alexander-yevsyukov:


alexander-yevsyukov:
@armiol this is work in progress. Tests are not added and `addIdSetFunction()`/ `removeIdSetFunction()` methods are not used because of this. Let's discuss before going further.

alexander-yevsyukov:
@armiol 
Instead of overriding the following method...
```java
    protected Set<I> getProjectionIds(Message event, EventContext context) {
        return idSetFunctions.findAndApply(event, context);
    }
```
I would expect that a projection would add custom `IdSetFunction`s in their constructors.

server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
```diff
@@ -174,7 +219,7 @@ public void close() throws Exception {
      */
     @Override
     @Nonnull
-    @SuppressWarnings("RefusedBequest")
+    @SuppressWarnings("MethodDoesntCallSuperMethod")
```
armiol: Please add an explanation.

server/src/test/java/org/spine3/server/projection/ProjectionRepositoryShould.java
```diff
@@ -249,6 +266,48 @@ private void ensureCatchesUpFromEventStorage(ProjectionRepository<ProjectId, Tes
...
+    }
+
+    @Test
+    public void obtain_id_set_function_after_put() {
+        repository.addIdSetFunction(ProjectCreated.class, idSetForCreateProject);
+
+        final Optional<IdSetFunction<ProjectId, ProjectCreated, EventContext>> out = repository.getIdSetFunction(ProjectCreated.class);
+
+        assertTrue(out.isPresent());
+        //noinspection OptionalGetWithoutIsPresent
```
armiol: Please use an annotation-based suppression instead. It's OK, if it is for the whole method, as this is a test codebase.

PR #254, opened 2017-01-05
Defined key comparator into InMemoryAggregateStorage

IlliaShepilov:
Summary of changes:

- Implemented key comparator


IlliaShepilov:
@armiol please take a look at the updates.

PR #256, opened 2017-01-05
Delivering an event to multiple Process Managers

alexander-yevsyukov:
This PR extracts common functionality of having functions for generating multiple IDs into abstract base called `EventDispatchingRepository`.

`ProcessManagerRepository` and `ProjectionRepository` inhering from this class.

### DIFF
--- a/server/src/main/java/org/spine3/server/entity/DefaultIdSetEventFunction.java
+++ b/server/src/main/java/org/spine3/server/entity/DefaultIdSetEventFunction.java
@@ -59,6 +59,11 @@ public class DefaultIdSetEventFunction {
         return new ProducerFromFirstEventMessageField<>();
     }
 
+    /**
+     * The {@code IdSetFunction} that obtains an event producer ID from the context of the event.
+     *
+     * @param <I> the type of entity IDs
+     */
     private static class ProducerFromContext<I> implements IdSetEventFunction<I, Message> {
         @Override
         public Set<I> apply(Message message, EventContext context) {
@@ -67,6 +72,11 @@ public class DefaultIdSetEventFunction {
         }
     }
 
+    /**
+     * The {@code IdSetFunction} that obtains an event producer ID from the context of the event.
+     *
+     * @param <I> the type of entity IDs
+     */
     private static class ProducerFromFirstEventMessageField<I> implements IdSetEventFunction<I, Message> {
 
         private final GetProducerIdFromEvent<I, Message> func = GetProducerIdFromEvent.fromFieldIndex(0);
--- a/server/src/main/java/org/spine3/server/entity/DefaultIdSetEventFunction.java
+++ b/server/src/main/java/org/spine3/server/entity/DefaultIdSetEventFunction.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2016, TeamDev Ltd. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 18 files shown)


server/src/main/java/org/spine3/server/entity/DefaultIdSetEventFunction.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Creates the {@code IdSetEventFunction} that obtains event producer ID from
+     * an {@code EventContext} and returns it as a sole element of the {@code ImmutableSet}.
+     *
+     * @param <I> the type of the IDs managed by the repository
+     * @return new function instance
+     */
+    public static <I> IdSetEventFunction<I, Message> producerFromContext() {
+        return new ProducerFromContext<>();
```
armiol: We may want to have this value constant. In the same way as in Guava implementation in similar cases.

server/src/main/java/org/spine3/server/entity/DefaultIdSetEventFunction.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private static class ProducerFromContext<I> implements IdSetEventFunction<I, Message> {
+        @Override
+        public Set<I> apply(Message message, EventContext context) {
+            final I id = Events.getProducer(context);
+            return ImmutableSet.of(id);
+        }
+    }
+
+    private static class ProducerFromFirstEventMessageField<I> implements IdSetEventFunction<I, Message> {
```
armiol: I would have at least some JavaDocs for this class. 

Surely, I understand that the docs overlap with the `producerFromFirstMessageField` documentation, but maybe you could describe a class purpose/meaning here and then put a reference from the method JavaDoc?

server/src/main/java/org/spine3/server/entity/EventDispatchingRepository.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Abstract base for repositories that deliver events to entities they manage.
+ *
+ * @param <I> the type of IDs of entities
+ * @param <E> the type of entities
+ * @param <S> the type of entity state messages
+ * @author Alexander Yevsyukov
+ */
+public abstract class EventDispatchingRepository<I, E extends Entity<I, S>, S extends Message>
```
armiol: As long as `ProcessManagerRepository` and `ProjectionRepository` are `EventDispatcher`s, I would call this class `EventDispatcherRepository`.

server/src/main/java/org/spine3/server/entity/IdSetFunctionMap.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Helper class for managing {@link IdSetEventFunction}s associated with
+ * a repository that dispatches events to its entities.
+ *
+ * @param <I> the type of the entity IDs of this repository
+ * @author Alexander Yevsyukov
+ */
+@Internal
+public class IdSetFunctionMap<I> {
```
armiol: As long it is not a `Map` per se, I would roll back to the previous name it had, i.e. `IdSetFunctions`.

server/src/main/java/org/spine3/server/entity/EventDispatchingRepository.java
armiol: As discussed vocally, the naming makes sense in terms of the proposed type hierarchy.

So, let's leave it as `EventDispatchingRepository`.

server/src/main/java/org/spine3/server/entity/DefaultIdSetEventFunction.java
alexander-yevsyukov: We cannot because these types are parameterized. Let's live with it, and if we face a performance penalty, we'd revisit this matter.

server/src/main/java/org/spine3/server/entity/EventDispatchingRepository.java
alexander-yevsyukov: As discussed vocally, we need to name it as a repository with is dispatching events, rather than a repository of event dispatchers.

alexander-yevsyukov:
@armiol PTAL

alexander-yevsyukov:
@armiol I've reverted the name back to `IdSetFunctions`. Please also note that in the last commit I'm hiding interfaces and classes for working with event fields. 

The only class that remains visible is `GetTargetIdFromCommand`. It is visible because it's used by several packages. The class is annotated `@Internal`.

PR #260, opened 2017-01-06
Bump Gradle version to 3.3.

armiol:


armiol:
@alexander-yevsyukov PTAL.

PR #261, opened 2017-01-06
Move `EventStore` configuration from the `BoundedContext` to the `EventBus`

armiol:
Before these changes an attempt to create a `BoundedContext` with a custom `EventBus` resulted in a circular dependency like this:

`BoundedContext` -> `EventBus` -> `EventStore` -> `BoundedContext`.

This PR is aimed to break the circle by moving the `EventStore` configuration to the `EventBus`.

@alexander-yevsyukov after the current changes, either

-  a `StorageFactory` (to create an `EventStore` with a default props),
- or a whole custom `EventBus` instance

 ... are required by the `EventBus.Builder`.


armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -346,9 +350,34 @@ public void close() throws Exception {
     /** The {@code Builder} for {@code EventBus}. */
     public static class Builder {
 
+        private static final String EVENT_STORE_CONFIGURED_MESSAGE = "EventStore is already configured.";
+
+        /**
+         * A {@code StorageFactory} for configuring the {@code EventStore} instance for this {@code EventBus}.
+         *
+         * <p>Either a {@code StorageFactory} or an {@code EventStore} are mandatory to create an instance of
+         * {@code EventBus}.
```
alexander-yevsyukov: What happens to this field if we pass only  an`EventStore` instance to Builder?

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -346,9 +350,34 @@ public void close() throws Exception {
...
+         */
+        private StorageFactory storageFactory;
+
+        /**
+         * A {@code EventStore} for storing all the events passed through the {@code EventBus}.
+         *
+         * <p>If not set, a default instance will be created by the builder with the help of {@code storageFactory}/
```
alexander-yevsyukov: Did you intend to have a dot instead of slash?

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -375,7 +404,41 @@ public void close() throws Exception {
...
+         *
+         * <p>This {@code StorageFactory} instance will be used to create an instance of {@code EventStore}
+         * for this {@code EventBus}, <em>if</em> {@code EventStore} was not explicitly set in the builder.
+         *
+         * <p>Either a {@code StorageFactory} or an {@code EventStore} are mandatory to create an {@code EventBus}.
+         *
+         * @see #setEventStore(EventStore)
+         */
+        public Builder setStorageFactory(StorageFactory storageFactory) {
+            checkState(eventStore == null, EVENT_STORE_CONFIGURED_MESSAGE);
```
alexander-yevsyukov: Is the `null` branch tested?

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -375,7 +404,41 @@ public void close() throws Exception {
...
+            this.storageFactory = checkNotNull(storageFactory);
+            return this;
+        }
+
+        @Nullable
```
alexander-yevsyukov: What if we have `Optional` from Guava and then from Java in v2?

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -375,7 +404,41 @@ public void close() throws Exception {
...
+         * <p>This method can be called if neither {@link #setEventStoreStreamExecutor(Executor)}
+         * nor {@link #setStorageFactory(StorageFactory)} were called before.
+         *
+         * <p>Either a {@code StorageFactory} or an {@code EventStore} must be set to create an {@code EventBus}.
+         *
+         * @see #setEventStoreStreamExecutor(Executor)
+         * @see #setStorageFactory(StorageFactory)
+         */
         public Builder setEventStore(EventStore eventStore) {
+            checkState(storageFactory == null, "storageFactory already set.");
```
alexander-yevsyukov: Codecov browser plugin shows that `null` branches are probably not tested. Please check.

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -386,6 +449,30 @@ public EventStore getEventStore() {
...
+         * {@code EventStore} was not explicitly set in the builder.
+         *
+         * <p>If an {@code Executor} is not set in the builder, {@link MoreExecutors#directExecutor()}
+         * will be used.
+         *
+         * @see #setEventStore(EventStore)
+         */
+        @SuppressWarnings("MethodParameterNamingConvention")
+        public Builder setEventStoreStreamExecutor(Executor eventStoreStreamExecutor) {
+            checkState(eventStore == null, EVENT_STORE_CONFIGURED_MESSAGE);
```
alexander-yevsyukov: Codecov browser plugin shows that `null` branches are probably not tested. Please check.

server/src/main/java/org/spine3/server/event/EventBus.java
```diff
@@ -447,7 +534,20 @@ public EventEnricher getEnricher() {
         }
 
         public EventBus build() {
-            checkNotNull(eventStore, "eventStore must be set");
+            checkState(storageFactory != null || eventStore != null,
+                       "Either storageFactory or eventStore must be set to build the EventBus instance");
+
+            if (eventStoreStreamExecutor == null) {
```
alexander-yevsyukov: Codecov browser plugin shows that `null` branches are probably not tested. Please check.

PR #267, opened 2017-01-15
Move storages under corresponding packages

alexander-yevsyukov:
This PR is based on #266. It moves storage classes as well as Protobuf types into packages that use the storages.

As long as these classes and interfaces are of `SPI` level, their API was demoted from `public` to `protected` as much as possible.

Generated classes and interfaces are now live under `storage` sub-packages, and these packages are annotated as `@SPI` via `package-info.java`.


server/src/test/java/org/spine3/server/command/CommandStorageShould.java
```diff
@@ -143,11 +146,11 @@ public void store_command_with_error() {
 
     @Test
     public void store_command_with_error_and_generate_ID_if_needed() {
-        final Command command = Commands.create(Given.CommandMessage.createProject(), CommandContext.getDefaultInstance());
+        final Command command = Commands.create(org.spine3.server.storage.Given.CommandMessage.createProject(), CommandContext.getDefaultInstance());
```
armiol: Please adjust the code to fit 120-symbols-per-line.

server/src/main/java/org/spine3/server/aggregate/AggregatePartRepositoryLookup.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * A helper class for finding and checking {@code AggregatePartRepository}.
+ *
+ * <p>This class is used by {@link AggregateRoot} to find repositories for its parts.
+ *
+ * @param <I> the type of the IDs of the repository to find
+ * @param <S> the type of the state of aggregate parts of managed by the target repository
+ * @author Alexander Yevsyukov
+ */
+class AggregatePartRepositoryLookup<I, S extends Message> {
```
alexander-yevsyukov: I thought that being package-access it doesn't need the annotation. It is internal, though.

server/src/main/java/org/spine3/server/aggregate/AggregatePart.java
```diff
@@ -0,0 +1,55 @@
...
+ * </ul>
+ *
+ * <p>In order to access parts of the aggregate {@link AggregateRoot} should be used.
+ *
+ * @param <I> the type for IDs of this class of aggregates
+ * @param <S> the type of the state held by the aggregate part
+ * @param <B> the type of the aggregate part state builder
+ * @author Alexander Yevsyukov
+ */
+public class AggregatePart<I, S extends Message, B extends Message.Builder> extends Aggregate<I, S, B> {
```
alexander-yevsyukov: It's an omission. I'll fix it.

server/src/main/java/org/spine3/server/aggregate/AggregateRoot.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Obtains a part state by its class.
+     *
+     * @param partStateClass the class of the state of the part
+     * @param <S> the type of the part state
+     * @return the state of the part or a default state if the state was not found
+     */
+    protected <S extends Message> S getPart(Class<S> partStateClass) {
+        final AggregatePartRepository<I, ?> repo = getRepository(partStateClass);
+        //noinspection TestOnlyProblems
```
alexander-yevsyukov: I'll reduce the warning level to Typo instead. It's wrongly reported by IDEA.

server/src/test/java/org/spine3/server/command/CommandStorageShould.java
alexander-yevsyukov: We should use local test data instead. It's a consequence of refactoring. I'll fix it.

server/src/main/java/org/spine3/server/aggregate/AggregateRoot.java
```diff
@@ -0,0 +1,124 @@
...
+        return partState;
+    }
+
+    /**
+     * Obtains a repository for the passed state class.
+     *
+     * @throws IllegalStateException if a repository was not found
+     *                               or the repository does not match expectations of this {@code AggregateRoot}
+     */
+    private <S extends Message, A extends AggregatePart<I, S, ?>> AggregatePartRepository<I, A> getRepository(Class<S> stateClass) {
```
armiol: The line looks too long. Is it possible to split it into two? Perhaps, a carriage return after `<S extends Message, ` would do.

PR #269, opened 2017-01-16
Simplify `valid` option usage

dmdashenkov:
Made the option `boolean` with an adjacent `Message` holding the error message.


dmdashenkov:
@armiol PTAL

armiol:
@alexander-yevsyukov FYI, the `(valid)` option is now simplified to `bool`.

PR #274, opened 2017-01-18
Test object builders

alexander-yevsyukov:


client/src/main/java/org/spine3/base/Identifier.java
```diff
@@ -0,0 +1,322 @@
...
+
+        MESSAGE {
+            @Override
+            <I> boolean matchValue(I id) {
+                return id instanceof Message;
+            }
+
+            @Override
+            boolean matchMessage(Message message) {
+                return !(message instanceof StringValue
```
armiol: This expression worth commenting. 

In particular, it's not clear that this list should be appended with another expression in case the `Type` enum gets another member.

alexander-yevsyukov:
@armoil It is WIP. Want to discuss later today.

Issue #272, opened 2017-01-17
Improve stringification of IDs


alexander-yevsyukov:
Now the code is in `Identifiers`. We need to extract it so that it's easier to understand and test.

alexander-yevsyukov:
Consider having a `Stringifier` interface based on `Function<I, String>`.

alexander-yevsyukov:
Fixed by #275

PR #275, opened 2017-01-18
Have `Stringifier`s separately

alexander-yevsyukov:
This PR introduces `Stringifier` interface and moves string conversions into separate set of classes (from `Identifiers`).


alexander-yevsyukov:
This PR addresses #272

client/src/test/java/org/spine3/base/StringifiersShould.java
```diff
@@ -0,0 +1,232 @@
...
+    public void convert_to_string_message_id_wrapped_in_Any() {
+        final StringValue messageToWrap = newStringValue(TEST_ID);
+        final Any any = AnyPacker.pack(messageToWrap);
+
+        final String result = idToString(any);
+
+        assertEquals(TEST_ID, result);
+    }
+
+    private static final Stringifier<IdWithPrimitiveFields> ID_TO_STRING_CONVERTER = new Stringifier<IdWithPrimitiveFields>() {
```
armiol: Please try to fit into 120 symbols per row.

client/src/main/java/org/spine3/base/StringifierRegistry.java
```diff
@@ -0,0 +1,93 @@
...
+        final Stringifier<?> func = entries.get(valueClass);
+
+        @SuppressWarnings("unchecked") /** The cast is safe as we check the first type when adding.
+            @see #register(Class, Function) */
+        final Stringifier<I> result = (Stringifier<I>) func;
+        return Optional.fromNullable(result);
+    }
+
+    public synchronized <I> boolean hasStringiferFor(Class<I> valueClass) {
+        final boolean contains = entries.containsKey(valueClass) && (entries.get(valueClass) != null);
```
armiol: Codecov reports that one branch out of four is not covered by tests.

client/src/main/java/org/spine3/base/Identifiers.java
```diff
@@ -20,55 +20,33 @@
...
-import static com.google.protobuf.TextFormat.shortDebugString;
-import static java.util.Collections.synchronizedMap;
-import static org.spine3.protobuf.AnyPacker.unpack;
 
 /**
  * This class manages conversion of identifies to/from string.
  *
- * <p>In addition to utility methods for the conversion, it provides {@link ConverterRegistry}
+ * <p>In addition to utility methods for the conversion, it provides {@link StringifierRegistry}
  * which allows to provide custom conversion logic for user-defined types of identifies.
```
armiol: Let's rephrase to avoid `it provides ... which allows to provide`.

client/src/main/java/org/spine3/base/StringifierRegistry.java
alexander-yevsyukov: The null checking is not applicable to us. I'm going to remove it and add test for not finding a key.

client/src/main/java/org/spine3/base/Identifiers.java
alexander-yevsyukov: Well spotted. This text is no longer applicable because the registry is now a separate class.

alexander-yevsyukov:
@armiol PTAL

PR #279, opened 2017-01-19
Handle unexpected default in `MessageMismatch` when changing a value

alexander-yevsyukov:
Until this PR we handled only the case when a command wants to clear a field. This PR adds handling the case when a command expects some value, and wants to change it, but discovers a default value.


client/src/main/java/org/spine3/change/MessageMismatch.java
```diff
@@ -70,6 +70,22 @@ public static ValueMismatch expectedNotDefault(Message expected, int version) {
...
+     * Creates a {@code ValueMismatch} for a command that wanted to <em>change</em> a field value,
+     * but discovered that the field has the default value.
+     *
+     * @param expected the value expected by the command
+     * @param newValue the value the command wanted to set
+     * @param version  the version of the entity in which the mismatch is discovered
+     * @return new {@code ValueMismatch} instance
+     */
+    public static ValueMismatch expectedNotDefault(Message expected, Message newValue, int version) {
+        checkNotNull(expected);
```
armiol: I thought we treat the method parameters as non-nullable due to the annotation in `package-info.java`. If that's still true, I'd think that `checkNotNull` statements should be removed.

If someone passes `null` values to this method, he will get an NPE (or maybe something even worse) for violating an API contract.

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
```diff
@@ -55,16 +55,18 @@
 import static org.spine3.protobuf.Messages.toMessageClass;
 
 /**
- * {@code Enricher} extends information of an event basing on its type and content.
+ * {@code EventEnricher} extends information of an event basing on its type and content.
```
armiol: Have you included the `EventEnricher`-related changes to this PR by a mistake? They just don't seem related to the `Mismatch`es at all.

client/src/main/java/org/spine3/change/MessageMismatch.java
alexander-yevsyukov: We annotate the package to say that the API expects non-nulls. But in the `public` and `protected` API we check for `null`s to prevent accidental errors and their propagation into our code deeper.

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
alexander-yevsyukov: I wanted to have latest `EventEnricher` in  the code from Projects. That's why this PR is based on earlier works.

client/src/main/java/org/spine3/change/MessageMismatch.java
alexander-yevsyukov: As discussed vocally, let's introduce a separate `NullToleranceTest` class that would check `null` parameters for the utility classes.

PR #281, opened 2017-01-23
Other commands in tests

alexander-yevsyukov:


testutil/src/main/java/org/spine3/test/CommandTest.java
```diff
@@ -80,6 +80,33 @@ protected Command createCommand(C commandMessage) {
...
     /**
+     * Creates a command with the passed message.
+     *
+     * <p>Use this method for creating commands of types different than
+     * one which is the subject of the test suite (defined by the generic type {@code <C>}.
+     *
+     * @param commandMessage the message of the command to create
+     * @return new command instance
+     */
+    protected Command createAnotherCommand(Message commandMessage) {
```
armiol: I think it should be `createDifferentCommand`.

`createAnotherCommand` reads more like `create one more command, same as yesterday, same as day before`.

testutil/src/test/java/org/spine3/test/CommandTestShould.java
```diff
@@ -116,7 +122,32 @@ public void create_command_with_custom_Timestamp() {
...
         final Command command = commandTest.createCommand(commandMessage, timestamp);
 
-        assertEquals(timestamp, command.getContext().getTimestamp());
+        assertEquals(timestamp, command.getContext()
+                                       .getTimestamp());
+    }
+
+    @SuppressWarnings("ConstantConditions") // Passing `null` is the purpose of the test.
+    @Test(expected = NullPointerException.class)
+    public void do_not_accept_null_command_on_createOtherCommand() {
```
armiol: `..._createOtherCommand` does not match the original method name, i.e. `commandTest#createAnotherCommand(...)`.

alexander-yevsyukov:
@armiol well spotted! I've renamed. PTAL

PR #284, opened 2017-01-25
Allow dynamic configuration of the `EventEnricher`

armiol:
Before this PR the instance of `EventEnricher` must have been fully initialized by the moment of the `BoundedContext` creation. That made it hard to use the `AggregateRepository` instances in the enrichment functions, since `AggregateRepository` required the `BoundedContext`. There was a circular dependency:

`EventEnricher` (wants) —> `AggregateRepo` (needs) —> `BoundedContext` (requires) —> `EventEnricher`.

In order to break the circle, I made it possible to append the new enrichment functions at runtime.

Summary of changes:

* Allowed to append the enrichment functions at runtime via `boundedContext.getEventBus().addFieldEnrichment(...)`.
* Some API and documentation improvements made to `org.spine3.server.event.enrich.*` classes.
* Spine version increased and now is `0.7.16-SNAPSHOT`.

### DIFF
--- a/server/src/main/java/org/spine3/server/event/enrich/EnrichmentFunction.java
+++ b/server/src/main/java/org/spine3/server/event/enrich/EnrichmentFunction.java
@@ -84,7 +84,15 @@ abstract class EnrichmentFunction<S, T> implements Function<S, T> {
      *
      * <p>During the activation the internal state of the function may be adjusted.
      *
+     * <p>A typical example of such an adjustment would be parsing and validation of the relations
+     * between {@code eventClass} and {@code enrichmentClass} from the corresponding {@code .proto}
+     * definitions. The function internal state in this case is appended with the parsed data, which
+     * is later used at runtime.
+     *
      * <p>After the function is activated successfully, the {@link #isActive()} returns {@code true}.
+     *
+     * <p>If an activation cannot be performed flawlessly, the {@code IllegalStateException}
+     * should be thrown. In this case {@link #isActive()} should return {@code false}.
      *
      * @throws IllegalStateException if the function cannot perform the conversion in its current state
      * or because of the state of its environment
--- a/server/src/main/java/org/spine3/server/event/enrich/EnrichmentFunction.java
+++ b/server/src/main/java/org/spine3/server/event/enrich/EnrichmentFunction.java
@@ -22,6 +22,7 @@ package org.spine3.server.event.enrich;
 
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
+import com.google.common.base.Predicate;
 import org.spine3.base.EventContext;
 import org.spine3.server.event.EventBus;
 
@@ -79,14 +80,37 @@ abstract class EnrichmentFunction<S, T> implements Function<S, T> {
     protected abstract Function<S, T> getFunction();
 
     /**
-     * Validates the instance.
+     * Activates the function.
+     *
+     * <p>During the activation the internal state of the function may be adjusted.
+     *
+     * <p>After the function is activated successfully, the {@link #isActive()} returns {@code true}.
      *
      * @throws IllegalStateException if the function cannot perform the conversion in its current state
      * or because of the state of its environment
      */
-    abstract void validate();
+    abstract void activate();
 
-    //TODO:24-Jan-2017:alex.tymchenko: think of creating `isValid` instead or in addition.
+    /**
+     * Checks whether this instance of {@code EnrichmentFunction} is active
+     * and available to use for the conversion.
+     *
+     * @return {@code true} if the function is eligible for the conversion, {@code false} otherwise.
+     */
+    abstract boolean isActive();
+
+    /**
+     * A helper predicate to filter the active functions only.
+     */
+    static Predicate<EnrichmentFunction<?, ?>> activeOnly() {
+        return new Predicate<EnrichmentFunction<?, ?>>() {
+            @Override
+            public boolean apply(@Nullable EnrichmentFunction<?, ?> input) {
+                checkNotNull(input);
+                return input.isActive();
+            }
+        };
+    }
 
     @Override
     public int hashCode() {
@@ -112,5 +136,17 @@ abstract class EnrichmentFunction<S, T> implements Function<S, T> {
                           .add("eventClass", eventClass)
                           .add("enrichmentClass", enrichmentClass)
                           .toString();
+    }
+
+    /**
+     * Checks whether this instance of {@code EnrichmentFunction} is active.
+     *
+     * <p>Throws {@link IllegalStateException} if the instance is not active.
+     */
+    protected void ensureActive() {
+        if(!isActive()) {
+            throw new IllegalStateException("The given instance of " + getClass() + " is not active at the moment. " +
+                                                    "Please use `#activate()` first.");
+        }
     }
 }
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/spine3/server/event/enrich/EnrichmentFunction.java
```diff
@@ -79,12 +80,37 @@ void setContext(EventContext context) {
     protected abstract Function<S, T> getFunction();
 
     /**
-     * Validates the instance.
+     * Activates the function.
+     *
+     * <p>During the activation the internal state of the function may be adjusted.
```
alexander-yevsyukov: Can you expand on this, please? It's not clear why the state needs to be changed. Perhaps an example would help.

server/src/main/java/org/spine3/server/event/enrich/EnrichmentFunction.java
```diff
@@ -111,4 +137,16 @@ public String toString() {
...
+
+    /**
+     * Checks whether this instance of {@code EnrichmentFunction} is active.
+     *
+     * <p>Throws {@link IllegalStateException} if the instance is not active.
+     */
+    protected void ensureActive() {
+        if(!isActive()) {
+            throw new IllegalStateException("The given instance of " + getClass() + " is not active at the moment. " +
+                                                    "Please use `#activate()` first.");
```
alexander-yevsyukov: Is it a set practice to reference methods via the hash sign in the console output? I though it applies only to Javadocs (and this is so because they follow HTML anchors).

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
```diff
@@ -172,8 +182,10 @@ public Event enrich(Event event) {
         private Action(EventEnricher parent, Message eventMessage, EventContext eventContext) {
             this.eventMessage = eventMessage;
             this.eventContext = eventContext;
-            this.availableFunctions = parent.functions.get(EventClass.of(eventMessage)
-                                                                     .value());
+            final Class<? extends Message> eventClass = EventClass.of(eventMessage)
```
alexander-yevsyukov: Please fix wrapping.

server/src/main/java/org/spine3/server/event/enrich/EventEnricher.java
```diff
@@ -233,17 +245,48 @@ private static void checkEnabled(Event event) {
...
-     * @param eventFieldClass the class of event fields
+     * @param eventFieldClass      the class of event fields
      * @param enrichmentFieldClass the class of enrichment fields
      */
     Optional<EnrichmentFunction<?, ?>> functionFor(Class<?> eventFieldClass,
                                                    Class<?> enrichmentFieldClass) {
         final Optional<EnrichmentFunction<?, ?>> result =
-                FluentIterable.from(functions.values())
-                              .firstMatch(SupportsFieldConversion.of(eventFieldClass, enrichmentFieldClass));
+                from(functions.values())
```
alexander-yevsyukov: Please fix wrapping.

armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/org/spine3/server/event/enrich/FieldEnricher.java
```diff
@@ -59,18 +59,25 @@ private FieldEnricher(Class<S> eventClass, Class<T> enrichmentClass, Function<S,
...
 
     @Override
-    void validate() {
+    void activate() {
         // Do nothing. Field enrichment relies only on the aggregated function.
     }
 
     @Override
+    boolean isActive() {
+        // It is always active, as no special actions are required to activate.
```
alexander-yevsyukov: Why not put it into Javadoc instead?

server/src/main/java/org/spine3/server/event/enrich/FieldEnricher.java
```diff
@@ -59,18 +59,25 @@ private FieldEnricher(Class<S> eventClass, Class<T> enrichmentClass, Function<S,
     }
 
     @Override
-    void validate() {
+    void activate() {
         // Do nothing. Field enrichment relies only on the aggregated function.
```
alexander-yevsyukov: What if we put this into Javadoc?

server/src/main/java/org/spine3/server/event/enrich/ReferenceValidator.java
```diff
@@ -25,21 +25,25 @@
...
-import javax.annotation.Nullable;
 import java.util.List;
 
 import static com.google.protobuf.Descriptors.Descriptor;
 import static com.google.protobuf.Descriptors.FieldDescriptor;
 import static java.lang.String.format;
 
 /**
- * Performs validation checking that all fields annotated in the enrichment message
+ * Performs validation analzing which of fields annotated in the enrichment message
```
alexander-yevsyukov: `analzing` looks like a typo. Please fix.

server/src/main/java/org/spine3/server/event/enrich/ReferenceValidator.java
```diff
@@ -151,11 +157,52 @@ private static IllegalStateException noFieldException(
...
+    private static void logNoFunction(Class<?> sourceFieldClass, Class<?> targetFieldClass) {
+        // Using `DEBUG` level to avoid polluting the `stderr`.
+        if (log().isDebugEnabled()) {
+            log().debug("There is no enrichment function for translating {} into {}",
+                        sourceFieldClass, targetFieldClass);
+        }
+    }
+
+    /**
+     * A wrapper DTO for the validation result.
```
alexander-yevsyukov: It looks like this class is too far from where it's used. Perhaps a name `ValidationResult` would look better?

server/src/test/java/org/spine3/server/event/EventBusShould.java
```diff
@@ -47,21 +50,23 @@
...
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 
-@SuppressWarnings({"InstanceMethodNamingConvention", "ResultOfMethodCallIgnored"})
+@SuppressWarnings({"InstanceMethodNamingConvention", "ResultOfMethodCallIgnored", "ClassWithTooManyMethods"})
```
alexander-yevsyukov: Please consider putting enrichment-related tests into a separate class.

server/src/main/java/org/spine3/server/event/enrich/EnrichmentFunction.java
armiol: An example added.

server/src/main/java/org/spine3/server/event/enrich/ReferenceValidator.java
armiol: Renamed. Though the original idea was to avoid self-repeating: `ReferenceValidator.Result`.

I see your point, since many users will `static import` this class. And the magic is gone.

server/src/test/java/org/spine3/server/event/EventBusShould.java
armiol: The thing is this test code is dealing with an `EventBus` public API. That's why it should belong to the `EventBusShould`, to my mind.

Also, we already have five (!) tests classes devoted to enrichments (see `org.spine3.server.event.enrich.*`). I would rather not create another one.

armiol:
@alexander-yevsyukov the comments addressed. PTAL.

PR #288, opened 2017-01-26
Rename `hasPrivateUtilityConstructor` to `hasPrivateParameterlessCtor`

armiol:
As we may check for the presence of the `private` parameter-less constructors not only in the utility classes, the legacy method name was a bit confusing.

In order to avoid any misuse, the method was renamed.


armiol:
@alexander-yevsyukov PTAL.

PR #291, opened 2017-01-27
Make `StorageFactorySwitch` initialization static

alexander-yevsyukov:
This would improve readability when passing it to `BoundedContext.Builder`.


alexander-yevsyukov:
@armiol, PTAL

PR #294, opened 2017-01-28
Archived and deleted

alexander-yevsyukov:


alexander-yevsyukov:
@armiol, the goal of introducing the `RepositoryView` interface is to be able return view-only instances from a repository like `getArchivedView()` or `getDeletedView()`.

Implementations of the views in `load()` would return mixins that would prohibit mutations. For example, an aggregate returned from a deleted view must not accept commands.

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
```diff
@@ -306,16 +304,36 @@ private A loadAndDispatch(I aggregateId, CommandId commandId, Message command, C
         return aggregate;
     }
 
-    /** Posts passed events to {@link EventBus}. */
-    private void postEvents(Iterable<Event> events) {
-        for (Event event : events) {
-            eventBus.post(event);
+    private void logConcurrentModification(I aggregateId, Message command, int newEventCount) {
+        log().warn("Detected the concurrent modification of {} {}" +
```
armiol: A space symbol is missing after `{} {}`.

server/src/main/java/org/spine3/server/aggregate/storage/Predicates.java
```diff
@@ -0,0 +1,49 @@
...
+            return input == null ||
+                    !(input.getArchived() || input.getDeleted());
+        }
+    };
+
+    private Predicates() {
+        // Prevent instantiation of this utility class.
+    }
+
+    public static Predicate<AggregateStatus> isVisible() {
```
armiol: As long as this is a part of the public API, it needs some JavaDocs.

server/src/main/java/org/spine3/server/entity/RepositoryView.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.spine3.server.entity;
+
+import com.google.common.base.Optional;
+
+import javax.annotation.CheckReturnValue;
+
+/**
+ * A view on a repository.
```
armiol: As a part of the public API, this class needs more details and/or samples.

server/src/main/java/org/spine3/server/storage/AbstractStorage.java
```diff
@@ -95,14 +95,19 @@ protected void checkNotClosed() throws IllegalStateException {
...
+     * @return {@code true} if the storage is open for writing (wasn't closed until now),
+     *         {@code false} otherwise
      * @see #close()
      */
     public boolean isOpen() {
         return this.open;
     }
 
     /**
+     * Verifies if the storage is closed.
```
armiol: `if` should be `that` or `<empty>`.

server/src/main/java/org/spine3/server/storage/AbstractStorage.java
```diff
@@ -95,14 +95,19 @@ protected void checkNotClosed() throws IllegalStateException {
     }
 
     /**
-     * @return {@code true} if the storage is open for writing (wasn't closed until now), {@code false} otherwise
+     * Verifies if the storage is open.
```
armiol: `if` should be `that` or `<empty>`.

server/src/main/java/org/spine3/server/storage/memory/InMemoryAggregateStorage.java
```diff
@@ -60,62 +62,54 @@ protected InMemoryAggregateStorage(boolean multitenant) {
...
+        getStorage().putEventCount(id, eventCount);
     }
 
-    /** Used for sorting by timestamp descending (from newer to older). */
-    private static class AggregateStorageRecordReverseComparator implements Comparator<AggregateStorageRecord>,
-                                                                            Serializable {
-        private static final long serialVersionUID = 0L;
+    @Override
+    protected boolean markArchived(I id) {
+        final AggregateStatus currentStatus = getStorage().getStatus(id);
```
armiol: That could lead to the performance issue.

If the aggregate is NOT archived, we'll be calling the storage twice (which is an expensive operation). And even if it is archived, we'll still call the storage at least once.

I would leave a comment somewhere in the JavaDocs that for a real storage (not `InMemory...` one) they should probably always update the state.

server/src/main/java/org/spine3/server/storage/memory/InMemoryAggregateStorage.java
```diff
@@ -60,62 +62,54 @@ protected InMemoryAggregateStorage(boolean multitenant) {
...
+                                                .build());
+        return true;
+    }
 
-            // To define an order:
-            final int firstHashCode = first.hashCode();
-            final int secondHashCode = second.hashCode();
+    @Override
+    protected boolean markDeleted(I id) {
+        final AggregateStatus currentStatus = getStorage().getStatus(id);
```
armiol: Same as above.

server/src/main/java/org/spine3/server/storage/memory/TenantRecords.java
```diff
@@ -0,0 +1,162 @@
...
+
+import static com.google.common.collect.Maps.newHashMap;
+
+/**
+ * The memory-based storage for {@code EntityStorageRecord} that represents
+ * all storage operations available for data of a single tenant.
+ *
+ * @author Alexander Yevsyukov
+ */
+class TenantRecords<I> implements TenantStorage<I, EntityStorageRecord> {
```
armiol: Do we expect someone using this class from the custom implementations?
It looks like a public API to me.

server/src/main/proto/spine/server/aggregate/aggregate_storage.proto
```diff
@@ -103,3 +103,14 @@ message AggregateEvents {
     // Otherwise, events are full history of the aggregate.
     repeated base.Event event = 2;
 }
+
+// The storage status of an aggregate.
+message AggregateStatus {
+
+    // If `true` the aggregate will not be loaded for regular queries.
+    bool archived = 3;
```
armiol: Is it numbered starting from `3` on purpose?

server/src/main/java/org/spine3/server/storage/Predicates.java
```diff
@@ -0,0 +1,48 @@
...
+import com.google.common.base.Predicate;
+
+import javax.annotation.Nullable;
+
+/**
+ * Collection of predicates for filtering storage records.
+ *
+ * @author Alexander Yevsyukov
+ */
+public class Predicates {
```
armiol: It's weird we have `Predicates` in both `org.spine3.server.storage` and `org.spine3.server.aggregate.storage`.

server/src/main/java/org/spine3/server/storage/memory/TenantRecords.java
alexander-yevsyukov: I think we need to be conservative here. We have `InMemory` classes only for testing purposes. 

We may open this class should we see a need, but I'd keep things hidden as much as possible (during the period from v1 to v2) to reduce support calls.

server/src/main/proto/spine/server/aggregate/aggregate_storage.proto
alexander-yevsyukov: No, it's an omission.

server/src/main/java/org/spine3/server/storage/Predicates.java
alexander-yevsyukov: These classes cover different inputs. We are going to have `isArchived` and `isDeleted` in both of them soon. What would you suggest to make it better?

server/src/main/java/org/spine3/server/storage/memory/InMemoryAggregateStorage.java
alexander-yevsyukov: @armiol, yes I do agree that there may be performance penalty. Here are questions we need to answer:
* What happens if we don't have a record with an ID? Do we throw? Or, do we just log?
* Do we need to handle repeated archive / delete differently from a business point of view?

server/src/main/java/org/spine3/server/entity/RepositoryView.java
alexander-yevsyukov: Added.

alexander-yevsyukov: Strangely GitHub doesn't display 849c4f4f5afb40686f95bd5c0b8c11cb12b1f88e.

alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/aggregate/AggregateRepository.java
```diff
@@ -246,14 +250,28 @@ public void store(A aggregate) {
...
 
     /**
-     * Loads the aggregate by the passed ID, or creates it if the aggregate was not found.
+     * Loads the aggregate by the passed ID.
+     *
+     * <p>If the aggregate is not available in the repository this method returns
+     * a newly created aggregate.
+     *
+     * <p>If the aggregate is “invisible” to regular queries, {@code Optional.absent()}
+     * is returned
```
armiol: A dot is missing.

server/src/main/java/org/spine3/server/entity/Entity.java
```diff
@@ -74,24 +75,57 @@
...
+     *
+     * <p>Results of this method call are:
+     * <ul>
+     *   <li>The state object is set to the value produced by {@link #getDefaultState()}.
+     *   <li>The version number is set to zero.
+     *   <li>The {@link #whenModified} field is set to the system time of the call.
+     *   <li>The {@link #status} field is set to the default instance.
+     * </ul>
+     *
+     * This method cannot be called from within {@code Entity} constructor because
```
armiol: `<p>` is missing.

server/src/main/java/org/spine3/server/entity/Entity.java
```diff
@@ -279,6 +303,52 @@ public Timestamp whenModified() {
...
+     * Sets {@code archived} status flag to the passed value.
+     */
+    protected void setArchived(boolean archived) {
+        this.status = getStatus().toBuilder()
+                                 .setArchived(archived)
+                                 .build();
+    }
+
+    /**
+     * Tests weather the entity is marked as deleted.
```
armiol: `whether`.

server/src/main/java/org/spine3/server/entity/RepositoryView.java
```diff
@@ -27,6 +27,13 @@
 /**
  * A view on a repository.
  *
+ * <p>A {@link Repository} may have a view that provides a different set of entities.
+ * For example, there can be views that return archived or deleted entities (that are
+ * not visible by default).
+ *
+ * <p>The default behaviour of loading entities in a {@code Repository} is also a view,
```
armiol: `The behaviour ... is a view` seems to complex to read. Something is missing in the sentence.

server/src/main/java/org/spine3/server/storage/AbstractStorage.java
```diff
@@ -95,7 +95,7 @@ protected void checkNotClosed() throws IllegalStateException {
     }
 
     /**
-     * Verifies if the storage is open.
+     * Tests weather the storage is open.
```
armiol: `whether`.

server/src/main/java/org/spine3/server/storage/AbstractStorage.java
```diff
@@ -106,7 +106,7 @@ public boolean isOpen() {
     }
 
     /**
-     * Verifies if the storage is closed.
+     * Tests weather the storage is closed.
```
armiol: `whether`.

values/src/main/java/org/spine3/time/Intervals.java
```diff
@@ -55,6 +56,20 @@ public static Interval between(Timestamp start, Timestamp end) {
     }
 
     /**
+     * Verifies in the timestamp is withing the interval.
```
armiol: `Verifies if the timestamp is within the interval`.

PR #297, opened 2017-01-31
Update aggregate testing

dmdashenkov:
Use `Given` utility to create test entities (`Entity`, `Aggregate`, `Projection`, `ProcassManager`, `AggregatePart`).


dmdashenkov:
@alexander-yevsyukov PTAL

server/src/test/java/org/spine3/server/entity/Given.java
```diff
@@ -73,22 +71,27 @@ static CreateProject createProject() {
...
         }
 
         static TestEntity withState() {
-            final Project state = newProject();
-            final int version = 3;
-            final Timestamp whenModified = getCurrentTime();
-            final TestEntity entity = new TestEntity(newUuid());
-            entity.setState(state, version, whenModified);
-            return entity;
+            final TestEntity result = org.spine3.test.Given.entityOfClass(TestEntity.class)
```
alexander-yevsyukov: Can we avoid FQN name here?

I would like to avoid multiple `Given` classes in tests. If we need to extend the API for the main `Given` class, let's discuss.

server/src/test/java/org/spine3/testdata/Sample.java
```diff
@@ -113,12 +175,16 @@ private static void checkClass(Class<? extends Message> clazz) {
         checkNotNull(clazz);
         // Support only generated protobuf messages
         checkArgument(clazz.isAssignableFrom(GeneratedMessageV3.class));
```
alexander-yevsyukov: Assuming that we perform the check, we need to propagate this requirement into method signatures here and above the call chain. Otherwise, there's no way to know it.

alexander-yevsyukov: Why do we need a generated message, BTW?

dmdashenkov:
@alexander-yevsyukov Closing this PR now since there are too many changes in here. Will re-diff as soon as the feature is complete.

PR #298, opened 2017-01-31
Advance version to `0.8.0`.

armiol:


armiol:
@dmdashenkov PTAL.

dmdashenkov:
@armiol I believe we need to set plugin version to a real existing one (`0.7.24`). Now it's dependant on the framework version.

armiol:
@dmdashenkov Done.

PR #313, opened 2017-02-07
Iterating command router

alexander-yevsyukov:


alexander-yevsyukov:
@dmdashenkov FYI: there's overlap in what `Aggregate`, `ProcessManager`, and `CommandHandler` do with generating events. We'll need to address it in a separate PR.

server/src/main/java/org/spine3/server/procman/IteratingCommandRouter.java
```diff
@@ -0,0 +1,113 @@
...
+        return command;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Overrides for return type covariance.
+     */
+    @Override
+    protected IteratingCommandRouter add(Message commandMessage) {
```
dmdashenkov: Maybe, we should try more convenient way to keep the precise type of the `AbstractCommandRouter` subclass. For example, we can add
`T getThis` method, where `<T extends AbstractCommandRouter>` is a type parameter of the class,
 and return its result instead of just using `this`.

dmdashenkov: This is not a unique case, similar overridden methods are also present in `CommandRouter`.

server/src/main/java/org/spine3/server/procman/ProcessManager.java
```diff
@@ -134,15 +123,15 @@ protected CommandBus getCommandBus() {
...
         return Lists.transform(events, new Function<Message, Event>() {
-            @Nullable // return null because an exception won't be propagated in this case
             @Override
             public Event apply(@Nullable Message event) {
                 if (event == null) {
                     return Event.getDefaultInstance();
                 }
-                final EventContext eventContext = createEventContext(cmdContext, event, getState(), whenModified(), getVersion());
+                final EventContext eventContext = createEventContext(event, commandContext
+                );
```
dmdashenkov: Please don't leave `);` along in this line.

server/src/test/java/org/spine3/server/procman/IteratingCommandRouterShould.java
```diff
@@ -0,0 +1,148 @@
...
+    public void routeNext() throws Exception {
+        router.routeFirst();
+
+        final Command command = router.routeNext();
+
+        // Test that 2nd command message is in the next rounded command.
+        assertEquals(messages.get(1), Commands.getMessage(command));
+    }
+
+    private static List<StringValue> unpackAll(List<Any> anyList) {
```
dmdashenkov: Maybe we should add this method to `Any` packer (as a reversed action for `pack(Iterator)`).

server/src/test/java/org/spine3/server/procman/IteratingCommandRouterShould.java
```diff
@@ -0,0 +1,148 @@
...
+        // Test that there's only one produced command and it has correct message.
+        final Command produced = commandRouted.getProduced(0);
+        final StringValue commandMessage = Commands.getMessage(produced);
+        assertEquals(messages.get(0), commandMessage);
+
+        // Test that the produced command context has correct fields.
+        assertTrue(Commands.sameActorAndTenant(sourceContext, produced.getContext()));
+
+        // Test that the event contains messages to follow.
+        assertEquals(messages.size() - 1, commandRouted.getMessageToFollowCount());
```
dmdashenkov: Seems that the test suite has cross-dependant tests: this assertion should fail if `routeNext` is executed before `routeFirst`.

server/src/test/java/org/spine3/server/procman/IteratingCommandRouterShould.java
```diff
@@ -0,0 +1,148 @@
...
+
+        sourceMessage = newStringValue(getClass().getSimpleName());
+        sourceContext = commandFactory.createCommandContext();
+
+        router = new IteratingCommandRouter(commandBus, sourceMessage, sourceContext);
+        router.addAll(messages);
+    }
+
+    @Test
+    public void routeFirst() throws Exception {
```
dmdashenkov: The test method names are not usual for Spine in this test class. Is it on purpose? If so, what about all the other test methods in other classes?

testutil/src/test/java/org/spine3/server/entity/EntityBuilderShould.java
```diff
@@ -84,6 +84,21 @@ public void create_entity() {
     }
 
     @Test
+    public void create_entity_with_default_values() {
+        final int version = 100500;
+        final String expectedState = Values.newStringValue("").getValue();
+        final Timestamp defaultTimestamp = Timestamp.getDefaultInstance();
+        final Entity entity = givenEntity().withVersion(version).build();
```
dmdashenkov: If we try to create an `Entity` with the default values, why do we initialize it with a custom `version`?

server/src/test/java/org/spine3/server/procman/IteratingCommandRouterShould.java
alexander-yevsyukov: This test does not depend on another test. It depends on `routeFirst()` (which it calls) working correctly. It is safe to assume so because we have another test that verifies it. 
If `routeFirst()` breaks this test would break, but we do want to know this too.

alexander-yevsyukov: The confusion should go away after I renamed the test, which tests `routeFirst()`.

server/src/test/java/org/spine3/server/procman/IteratingCommandRouterShould.java
alexander-yevsyukov: We'll on on 2nd occasion for such a need. We may want to have unpacking iterator instead so that we load from storage and un-pack on the fly without loading everything into memory. So, it's a bit more complicated.

gradlew
```diff
@@ -155,13 +155,14 @@ if $cygwin ; then
 fi
 
 # Escape application args
-for s in "${@}" ; do
```
alexander-yevsyukov: It is the fix for the file changes got from `master` and then reverted.

client/src/main/java/org/spine3/protobuf/PackingIterator.java
```diff
@@ -0,0 +1,66 @@
...
+     * of {@code Any} will be returned.
+     *
+     * @return the packed message or default {@code Any}
+     */
+    @Override
+    public Any next() {
+        final Message next = source.next();
+        final Any result = next != null
+                           ? AnyPacker.pack(next)
+                           : Any.getDefaultInstance();
```
alexander-yevsyukov: Protobuf discourages `null`s. That's why we return default `Any` just for double safety. Adding policies is an action in the opposite direction.

testutil/src/test/java/org/spine3/server/entity/EntityBuilderShould.java
alexander-yevsyukov: Fixed in commit 4e0abb8. Not sure why it's not visible here.

client/src/main/java/org/spine3/protobuf/PackingIterator.java
dmdashenkov: I was talking about `Any.getDefaultValue()` vs _skipping_ `null`s.

testutil/src/test/java/org/spine3/server/entity/EntityBuilderShould.java
dmdashenkov: Fixed in #317

alexander-yevsyukov:
Commits that are not visible here:
* 4e0abb8 — Simplify test
* 42da12e — Simplify return type covariance in command routers

@dmdashenkov, PTAL

server/src/main/java/org/spine3/server/procman/AbstractCommandRouter.java
```diff
@@ -133,7 +133,7 @@ protected Command route(Message message) {
         try {
             finishFuture.get();
         } catch (InterruptedException | ExecutionException e) {
-            throw new IllegalArgumentException(e);
+            throw new IllegalStateException(e);
```
dmdashenkov: Is this a possible case? If so, please cover it with a test.

client/src/main/java/org/spine3/protobuf/PackingIterator.java
alexander-yevsyukov: Then we'd have:
1. different order which isn't good by itself
2. we hide the problem deeper because default `Any` may force exploring the `null` input.

PR #314, opened 2017-02-07
Improve `AggregateStorage` testing

dmdashenkov:
- Cover `AggregateStorage`'s methord regarding `EntityStatus`.
 - Update plugin (`tools`) version.

### DIFF
--- a/server/src/test/java/org/spine3/server/aggregate/AggregateStorageShould.java
+++ b/server/src/test/java/org/spine3/server/aggregate/AggregateStorageShould.java
@@ -60,7 +60,7 @@ import static org.spine3.protobuf.Timestamps.getCurrentTime;
  */
 @SuppressWarnings({"InstanceMethodNamingConvention", "ClassWithTooManyMethods"})
 public abstract class AggregateStorageShould
-       extends AbstractStorageShould<ProjectId, AggregateEvents, AggregateStorage<ProjectId>> {
+        extends AbstractStorageShould<ProjectId, AggregateEvents, AggregateStorage<ProjectId>> {
 
     private final ProjectId id = Given.newProjectId();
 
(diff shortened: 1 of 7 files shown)


dmdashenkov:
@alexander-yevsyukov PTAL

server/src/test/java/org/spine3/server/aggregate/AggregateStorageShould.java
```diff
@@ -253,6 +257,129 @@ public void rewrite_event_count_after_last_snapshot() {
...
+        assertNotNull(entityStatus);
+        assertFalse(entityStatus.isPresent());
+    }
+
+    @SuppressWarnings({"OptionalUsedAsFieldOrParameterType", "OptionalGetWithoutIsPresent"})
+    private static void assertArchived(Optional<EntityStatus> entityStatus) {
+        assertStatus(entityStatus, true, false);
+    }
+
+    @SuppressWarnings({"OptionalUsedAsFieldOrParameterType", "OptionalGetWithoutIsPresent"})
```
alexander-yevsyukov: We don't call `get()` inside. Why the warning suppression?

server/src/test/java/org/spine3/server/aggregate/AggregateStorageShould.java
```diff
@@ -253,6 +257,129 @@ public void rewrite_event_count_after_last_snapshot() {
         assertEquals(expectedValue, actualCount);
     }
 
+    @SuppressWarnings("OptionalGetWithoutIsPresent") // Checked in an assertion
+    @Test
+    public void write_entity_status_of_aggregate() {
+        final ProjectId id = Given.newProjectId();
```
alexander-yevsyukov: Why cannot we create the `id` in `setUp()`? We use it almost in every test method.

This would also reduce the need for a specific `Given` class. We have test factory for identifiers. We can call it once in `setUp()` and that would not reduce readability.

dmdashenkov:
@alexander-yevsyukov PTAL

server/src/test/java/org/spine3/server/aggregate/EntityStatusHolderShould.java
```diff
@@ -0,0 +1,179 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests covering the behavior of the {@link AggregateStorage} regarding the {@link EntityStatus}.
+ *
+ * @author Dmytro Dashenkov.
+ */
+public abstract class EntityStatusHolderShould {
```
alexander-yevsyukov: How about the name `AggregateStorageStatusHandlingShould`?

alexander-yevsyukov: It's not about `Entity` we test `AggregateStorage`. The current name is confusing.

dmdashenkov:
@alexander-yevsyukov PTAL

PR #316, opened 2017-02-07
Fix `MessageValidator` logging bug

dmdashenkov:
`MessageValidator` by now logged a warning about the `if_invalid` annotation misusage for all the existing messages except those that do misuse the annotation.


dmdashenkov:
@alexander-yevsyukov PTAL

PR #324, opened 2017-02-09
CommandBus should handle command status.

alexander-aleksandrov:


server/src/test/java/org/spine3/server/command/CommandBusShould.java
```diff
@@ -114,7 +106,7 @@ public void setUp() {
         final InMemoryStorageFactory storageFactory = InMemoryStorageFactory.getInstance();
         commandStore = spy(new CommandStore(storageFactory.createCommandStorage()));
         scheduler = spy(new ExecutorCommandScheduler());
-        log = spy(new Log());
+        final Log log = spy(new Log());
```
alexander-yevsyukov: Do we use this `log` anywhere else? Now as it  became local variable, we lost references to it from other methods.

Why do we `spy` on it then?

alexander-aleksandrov: Actually we don't. I removed it.

PR #325, opened 2017-02-09
Command Bus endpoints

alexander-yevsyukov:
This PR introduces the following changes:

### `CommandEndpoint` interface
The interface is the common based for objects that deliver a command to the place of its handling.
`DispatcherEndpoint` and `HandlerEndpoint` implement this interface.

### `CommandEndpointRegistry` class
This class wraps `DispatcherRegistry` and `HandlerRegistry` and all the work on them (which was previously done by `CommandBus`). `CommandBus` simply delegates registration/un-registration to `CommandEndpointRegistry`.
The endpoint registry is also responsible for caching endpoints. We searched in dispatcher/handler registries several times before.

### AbstractCommandBusTestSuite
... is the abstract base for `CommandBus` tests extracted to avoid duplication of code in `setUp()` and `tearDown()`. It also contains stub class for a `CommandHandler`.

### Tests relocation
Tests on specific areas are migrated from `CommandBusShould` to:
* `CommandEndpointRegistryShould`
* `CommandSchedulingShould`

### Other notable changes
* `CurrentTenant` now return `Optional<TenantId>` instead of `@Nullable TenantId`. 
* Created `TestResponseObserver` for testing `CommandBus` response.
* Simplified `org.spine3.server.command.Given`
* Added tests for `MultitenantStorage`.


armiol:
@alexander-yevsyukov please provide a description for this PR.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -236,43 +237,58 @@ public void unregister(CommandHandler handler) {
...
         if (!filter.handleValidation(command, responseObserver)) {
             return;
         }
+
         if (isScheduled(command)) {
             scheduleAndStore(command, responseObserver);
             return;
         }
+
         if (isMultitenant) {
```
armiol: Should we set the current tenant _before_ handling the command?

I would move the tenant update onto the spot before `if ... { handleUnsupported (..) }`, since even handling the unsupported commands should generally depend on the target tenant.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -236,43 +237,58 @@ public void unregister(CommandHandler handler) {
...
         responseObserver.onCompleted();
     }
 
     /**
-     * Checks if a command is supported by the {@code CommandBus}.
-     *
-     * @param commandClass a class of commands to check
-     * @return {@code true} if there is a {@link CommandDispatcher} or a {@link CommandHandler} registered
-     *      for commands of this type, {@code false} otherwise
+     * Passes a previously scheduled command to corresponding endpoint.
```
armiol: `to the corresponding endpoint`.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -369,10 +358,21 @@ public void setMultitenant(boolean isMultitenant) {
         this.isMultitenant = isMultitenant;
     }
 
+    /**
+     * Closes the instance for further posting of commands.
```
armiol: `Closes the instance, preventing any further posting of commands.`

server/src/main/java/org/spine3/server/command/CommandEndpoint.java
```diff
@@ -0,0 +1,35 @@
...
+ * the place it is handled.
+ *
+ * @author Alexander Yevsyukov
+ */
+interface CommandEndpoint {
+
+    /**
+     * Delivers a command passed in the envelope.
+     */
+    void process(CommandEnvelope commandEnvelope);
```
armiol: Perhaps, `receive`?

It's just that endpoints do not process. They are receiving spots, passing the received objects to someone in charge of processing.

And also if we call it `receive`, then the docs make more sense to me:

``` 
Receives a command packed into an envelope and passes it for the further processing.
```

server/src/main/java/org/spine3/server/command/DispatcherEndpoint.java
```diff
@@ -0,0 +1,40 @@
...
+class DispatcherEndpoint implements CommandEndpoint {
+
+    private final CommandDispatcher dispatcher;
+
+    DispatcherEndpoint(CommandDispatcher dispatcher) {
+        this.dispatcher = dispatcher;
+    }
+
+    @Override
+    public void process(CommandEnvelope commandEnvelope) {
```
armiol: Same as above. I would make it `receive`.

server/src/main/java/org/spine3/server/storage/memory/MultitenantStorage.java
```diff
@@ -67,6 +68,20 @@ S getStorage() {
...
+        final Optional<TenantId> currentTenant = CurrentTenant.get();
+
+        if (!currentTenant.isPresent()) {
+            throw new IllegalStateException("No current tenant found in multitenant execution context.");
+        }
+
+        return currentTenant.get();
+    }
+
     abstract S createSlice();
```
armiol: As I am not able to comment in the line:45, I'll comment here.

We should probably think of concurrency here as we did in this PR for the endpoint registry.
```
  private final Map<TenantId, S> tenantSlices = newHashMap();     //  newConcurrentMap()?
```

server/src/main/java/org/spine3/server/command/CommandBus.java
alexander-yevsyukov: It will be done as a part of the PR #335.

server/src/main/java/org/spine3/server/command/CommandEndpoint.java
alexander-yevsyukov: Right!

server/src/main/java/org/spine3/server/storage/memory/MultitenantStorage.java
alexander-yevsyukov: @armiol, this and the rest above are fixed. PTAL.

PR #335, opened 2017-02-12
Isolate current tenant modifications

alexander-yevsyukov:
Now we set current tenant in `CommandBus` and do not restore it back, which is really fragile. The goal of this PR is to isolate the work with multitenancy at the level of repositories.


alexander-yevsyukov:
Here's the plan:
- [ ] Move `CurrentTenant` to the `entity` package.
- [ ] Make setting and re-setting current tenant `protected` to `Repository`. A repository should get the current state, remember it and restore when done. Presumably, `CurrentTenant` would become a stack.
- [ ] Make `CommandStore` and `EventStore` descend from `Repository`. These stores are not repositories and it's not convenient. We cannot, for example, load a command status.
- [ ] Make `CommandStore` and `EventStore` call tenant management methods derived from the `Repository` class.
- [ ] Make `CurrentTenant` class package-private to the `entity` package.

alexander-yevsyukov:
@armiol, FYI. See above.

alexander-yevsyukov:
I'm closing this PR because it's set to merge into quite an old branch `command-bus-endpoints`. Because of this the difference is huge and cannot be reviewed.

I'll create a new PR that would merge the branch `isolate-current-tenant-modifications` into `master`.

PR #336, opened 2017-02-12
Fix `Commands` null tolerance test

dmdashenkov:
The causes of the failure were following:
 - Not checking the test result (`NullToletranceTest` does not _`fail`_ the test but returns a `boolean` value).
 - `Commands` class could not pass the tests by their design.
 - After one failed method checked `false` was returned. The methods order is undefined though, that's why we could not reproduce the problem in our local environments.
 - Not setting default values for the custom parameter types caused an error when trying to get one.

What's done:
 - Check the test result.
 - Fit `Commands` to our requirements.
 - Add an error message for the missing default value case.

Also, I've turned off the checks for `synthetic` methods (`Smth#access$000`), since we declare to check non-private methods only.

### DIFF
--- a/client/src/test/java/org/spine3/base/CommandsShould.java
+++ b/client/src/test/java/org/spine3/base/CommandsShould.java
@@ -32,6 +32,7 @@ import com.google.protobuf.Timestamp;
 import org.junit.Test;
 import org.spine3.protobuf.AnyPacker;
 import org.spine3.protobuf.Durations;
+import org.spine3.protobuf.Timestamps;
 import org.spine3.test.NullToleranceTest;
 import org.spine3.test.TestCommandFactory;
 import org.spine3.test.commands.TestCommand;
@@ -56,6 +57,8 @@ import static org.spine3.testdata.TestCommandContextFactory.createCommandContext
 
 @SuppressWarnings({"InstanceMethodNamingConvention", "MagicNumber"})
 public class CommandsShould {
+
+    private static final FileDescriptor DEFAULT_FILE_DESCRIPTOR = Any.getDescriptor().getFile();
 
     private final TestCommandFactory commandFactory = TestCommandFactory.newInstance(CommandsShould.class);
     private final StringValue stringValue = newStringValue(newUuid());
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@alexander-yevsyukov PTAL

testutil/src/main/java/org/spine3/test/NullToleranceTest.java
```diff
@@ -140,6 +136,19 @@ public boolean check() {
         return true;
     }
 
+    @SuppressWarnings("OverlyComplexBooleanExpression") //
```
alexander-yevsyukov: Care to document it, or remove the comment at all?

testutil/src/main/java/org/spine3/test/NullToleranceTest.java
```diff
@@ -389,7 +398,11 @@ private Object getDefaultValue(Class<?> type) {
                 }
             }
 
-            checkState(result != null);
+            checkState(
+                    result != null,
+                    String.format(
+                            "Could not generate a default value for %s type. Please, use NullToleranceTest.Builder#addDefaultValue",
```
alexander-yevsyukov: > NullToleranceTest.Builder#addDefaultValue

This syntax is for Javadocs. I would say that a developer in such a situation needs a method name to call.
And I would add `()` at the end.

client/src/test/java/org/spine3/base/CommandsShould.java
```diff
@@ -81,10 +84,13 @@ public void have_private_ctor() {
...
     public void pass_null_tolerance_test() {
-        NullToleranceTest.newBuilder()
-                         .setClass(Commands.class)
-                         .build()
-                         .check();
+        final boolean passed = NullToleranceTest.newBuilder()
+                                                .setClass(Commands.class)
+                                                .addDefaultValue(DEFAULT_FILE_DESCRIPTOR)
+                                                .addDefaultValue(Timestamps.getCurrentTime()) // to fulfil custom validation
+                                                .build()
```
alexander-yevsyukov: What are scenarios for calling `build()`, but calling `check()` never or later?

If `check()` follows `build()` in the majority of cases we can think of, I would eliminate the `build()` to prevent errors.

dmdashenkov: Usage lookup shows that we have 3 cases of not calling `check()` after `build()`, but only 2 of them are in `NullToleranceTestShould`.

dmdashenkov:
@alexander-yevsyukov PTAL again

PR #338, opened 2017-02-14
Right margin wrapping

alexander-aleksandrov:
Since we decided to use 100 columns format, we need to wrap all existing code to new format. 
This is the first bunch of formatted files for Values package. 

New IDEA config file contains a new right margin for 100 columns and also few other wrapping options enabled for better code style.

### DIFF
--- a/testutil/src/main/java/org/spine3/test/Verify.java
+++ b/testutil/src/main/java/org/spine3/test/Verify.java
@@ -75,14 +75,14 @@ public final class Verify extends Assert {
     }
 
     /**
-     * Mangles the stack trace of {@link AssertionError} so that it looks like its been thrown from the line that
-     * called to a custom assertion.
+     * Mangles the stack trace of {@link AssertionError} so that it looks like its been thrown from
+     * the line that called to a custom assertion.
      *
-     * <p>This method behaves identically to {@link #mangledException(AssertionError, int)} and is provided
-     * for convenience for assert methods that only want to pop two stack frames. The only time that you would want to
-     * call the other {@link #mangledException(AssertionError, int)} method is if you have a custom assert
-     * that calls another custom assert i.e. the source line calling the custom asserts is more than two stack frames
-     * away.
+     * <p>This method behaves identically to {@link #mangledException(AssertionError, int)} and is
+     * provided for convenience for assert methods that only want to pop two stack frames. The only
+     * time that you would want to call the other {@link #mangledException(AssertionError, int)}
+     * method is if you have a custom assert that calls another custom assert i.e. the source line
+     * calling the custom asserts is more than two stack frames away.
      *
      * @param e the exception to mangle
      * @see #mangledException(AssertionError, int)
@@ -130,8 +130,8 @@ public final class Verify extends Assert {
      * }
      * </pre>
      *
-     * <p>Without the {@code try ... catch} block around lines 11-13 the stack trace following a test failure
-     * would look a little like:
+     * <p>Without the {@code try ... catch} block around lines 11-13 the stack trace following a
+     * test failure would look a little like:
      *
      * <p><pre>
      * java.lang.AssertionError: ...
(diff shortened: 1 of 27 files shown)


testutil/src/main/java/org/spine3/test/Verify.java
```diff
@@ -390,7 +398,8 @@ public static void assertEmpty(String multimapName, Multimap<?, ?> actualMultima
             assertObjectNotNull(multimapName, actualMultimap);
 
             if (!actualMultimap.isEmpty()) {
-                Assert.fail(multimapName + MSG_SHOULD_BE_EMPTY_ACTUAL_SIZE + actualMultimap.size() + '>');
+                Assert.fail(multimapName + MSG_SHOULD_BE_EMPTY_ACTUAL_SIZE +
```
dmdashenkov: As we are here, please add the static imports for the `Assert` static methods.

testutil/src/test/java/org/spine3/test/VerifyShould.java
```diff
@@ -1125,7 +1125,8 @@ public Object call() {
             }
         };
 
-        Verify.assertThrowsWithCause(runtimeException.getClass(), expectedCause.getClass(), throwsRuntimeException);
+        Verify.assertThrowsWithCause(runtimeException.getClass(), expectedCause.getClass(),
```
dmdashenkov: Same with `Verify`: please add the static imports.

values/src/main/java/org/spine3/net/UrlPrinter.java
```diff
@@ -48,7 +48,8 @@ static String printToString(Url url) {
             return url.getRaw();
         }
 
-        //noinspection StringBufferWithoutInitialCapacity // We don't know the capacity at this point
+        //noinspection StringBufferWithoutInitialCapacity
+        // We don't know the capacity at this point
         final StringBuilder sb = new StringBuilder();
```
dmdashenkov: In Spine we __do not__ use comment based suppression. This can be replaced since applied to a variable declaration.

testutil/src/test/java/org/spine3/test/VerifyShould.java
alexander-aleksandrov: Fixed

testutil/src/main/java/org/spine3/test/Verify.java
alexander-aleksandrov: Fixed

values/src/main/java/org/spine3/net/UrlPrinter.java
alexander-aleksandrov: Fixed

alexander-aleksandrov:
PTAL

values/src/main/java/org/spine3/net/UrlPrinter.java
```diff
@@ -48,7 +48,6 @@ static String printToString(Url url) {
             return url.getRaw();
         }
 
-        //noinspection StringBufferWithoutInitialCapacity
         // We don't know the capacity at this point
         final StringBuilder sb = new StringBuilder();
```
dmdashenkov: Isn't this still an issue? If `IDEA` highlights this e.g. as a _warning_, please suppress with `@Suppress("...")` annotation.

alexander-aleksandrov: It's done for a class

dmdashenkov: Please avoid class level suppressions if possible.

dmdashenkov: It would be convenient to have the suppressions right where they're needed.

alexander-aleksandrov: Done

PR #342, opened 2017-02-15
Changed wrapping to 100 columns for users package.

alexander-aleksandrov:
Since we decided to use 100 columns format, we need to wrap all existing code to new format.
This is the bunch of formatted files for Users package.

### DIFF
--- a/users/src/main/proto/spine/users/user_account.proto
+++ b/users/src/main/proto/spine/users/user_account.proto
@@ -39,7 +39,8 @@ import "spine/users/user_profile.proto";
 
 // Information exposed by an identity provider such as Google Sign-In or Facebook Login.
 //
-// Under Android this message will be populated from instances implementing `firebase.UserInfo` interface.
+// Under Android this message will be populated from instances implementing
+// `firebase.UserInfo` interface.
 //
 message UserInfo {
 
@@ -51,8 +52,8 @@ message UserInfo {
 
     // The domain of the user.
     //
-    // If the authentication provider is "google.com" then this attribute contains Google Apps domain
-    // of the user or "gmail.com" for normal Google account.
+    // If the authentication provider is "google.com" then this attribute contains
+    // Google Apps domain of the user or "gmail.com" for normal Google account.
     //
     // For authentication providers other than "google.com", this field is empty.
     //
@@ -134,7 +135,8 @@ message UserAccount {
 
     // The linked identities of the user obtained from different authentication providers.
     //
-    // For getting Google Account Id from Java, please use `org.spine3.users.UserAccounts.getGoogleUid()`
+    // For getting Google Account Id from Java,
+    // please use `org.spine3.users.UserAccounts.getGoogleUid()`
     //
     repeated UserInfo linked_identity = 20;
 
--- a/users/src/main/proto/spine/users/user_account.proto
+++ b/users/src/main/proto/spine/users/user_account.proto
@@ -53,7 +53,7 @@ message UserInfo {
     // The domain of the user.
     //
     // If the authentication provider is "google.com" then this attribute contains
-    // Google Apps domain of the user or "gmail.com" for normal Google account.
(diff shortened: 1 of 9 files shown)


alexander-aleksandrov:
@dmdashenkov PTAL

users/src/main/proto/spine/users/user_account.proto
```diff
@@ -51,8 +52,8 @@ message UserInfo {
 
     // The domain of the user.
     //
-    // If the authentication provider is "google.com" then this attribute contains Google Apps domain
-    // of the user or "gmail.com" for normal Google account.
+    // If the authentication provider is "google.com" then this attribute contains
+    // Google Apps domain of the user or "gmail.com" for normal Google account.
```
dmdashenkov: I think the name for `Google Apps` is different now. Please check.

users/src/test/java/org/spine3/users/UserAccountsShould.java
```diff
@@ -32,6 +32,9 @@ public void return_absent_for_google_uid_if_no_linked_account_added() {
         assertFalse(UserAccounts.getGoogleUid(UserAccount.getDefaultInstance()).isPresent());
     }
 
+    @SuppressWarnings("ConstantConditions")  /* As the {@code DUNBARS_NUMBER} value may change
+                                                a bit, it's still required to stay in certain
+                                                bounds */
     @Test
     public void declare_Dubars_number() {
         //noinspection MagicNumber
```
dmdashenkov: Please get rid of the comment-based suppression.

users/src/test/java/org/spine3/users/UserAccountsShould.java
```diff
@@ -32,6 +32,9 @@ public void return_absent_for_google_uid_if_no_linked_account_added() {
         assertFalse(UserAccounts.getGoogleUid(UserAccount.getDefaultInstance()).isPresent());
     }
 
+    @SuppressWarnings("ConstantConditions")  /* As the {@code DUNBARS_NUMBER} value may change
+                                                a bit, it's still required to stay in certain
+                                                bounds */
```
dmdashenkov: May it? Dunbar's number __is__ declared as a number _n_ which lies between 100 and 250. Though I don't think this particular `Java` constant is going to change with time unless some extraordinary situations which will still allow us to modify the test.
Anyway, the suppression is caused not by the fact that the Dunbar's number is unstable, but with the fact that we check a value of a constant. I think we need no explanation at all with this suppression since it's obvious.
But it's still worth mentioning why do we use exactly the `[100; 250]` bounds.

dmdashenkov: What about splitting the comment into 2:
 1) Describing the suppression:
`Checking the bounds of a constant value.` or smth. similar.

2) Describing the constant itself (the currently existing comment would work here).

users/src/main/proto/spine/users/user_account.proto
alexander-aleksandrov: You are right it's "G Suite" now. Changed.

users/src/test/java/org/spine3/users/UserAccountsShould.java
alexander-aleksandrov: Done

users/src/test/java/org/spine3/users/UserAccountsShould.java
alexander-aleksandrov: Done

alexander-aleksandrov:
@dmdashenkov PTAL

users/src/test/java/org/spine3/users/UserAccountsShould.java
dmdashenkov: @armiol Does this look appropriate to you?

armiol: @dmdashenkov yes, it does for now.

PR #343, opened 2017-02-16
Entity interfaces and hierarchy

alexander-yevsyukov:
This PR introduces interface `Entity` and abstract base classes that implement these interfaces.

This PR is preparation for:
* introducing `EntityMeta`, which will be based on `AbstractEntity`
* an experiment with buildable `Repository`.


server/src/main/java/org/spine3/server/entity/AbstractEntity.java
```diff
@@ -0,0 +1,277 @@
...
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.protobuf.Timestamps.getCurrentTime;
+
+/**
+ * An abstract base for entities with versions.
+ *
+ * The entity keeps only its latest state and version information associated with this state.
```
armiol: Please insert a `<p>`.

server/src/main/java/org/spine3/server/entity/AbstractEntity.java
```diff
@@ -0,0 +1,277 @@
...
+        return version;
+    }
+
+    /**
+     * Advances the current version by one and records the time of the modification.
+     *
+     * @return new version number
+     */
+    protected int incrementVersion() {
+        ++version;
```
armiol: It seems the `version` would look better as something like `AtomicInteger` to make the increments concurrency-friendly.
Maybe we could even create a `Version` class with synchronized methods. 

Just a thought.

server/src/main/java/org/spine3/server/entity/AbstractEntity.java
```diff
@@ -0,0 +1,277 @@
...
+     * Ensures that the entity is not marked as {@code archived}.
+     *
+     * @throws CannotModifyArchivedEntity if the entity in in the archived status
+     * @see #getStatus()
+     * @see EntityStatus#getArchived()
+     */
+    protected void checkNotArchived() throws CannotModifyArchivedEntity {
+        if (getStatus().getArchived()) {
+            final String idStr = Stringifiers.idToString(getId());
+            throw new CannotModifyArchivedEntity(idStr);
```
armiol: The exception name is questionable. `checkNotArchived` could have been called for any reason at all, not just in case of modification.

server/src/main/java/org/spine3/server/entity/AbstractEntity.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    /**
+     * Ensures that the entity is not marked as {@code deleted}.
+     *
+     * @throws CannotModifyDeletedEntity if the entity is marked as {@code deleted}
+     * @see #getStatus()
+     * @see EntityStatus#getDeleted()
+     */
+    protected void checkNotDeleted() throws CannotModifyDeletedEntity {
```
armiol: Same as above.

server/src/main/java/org/spine3/server/entity/EntityLite.java
```diff
@@ -0,0 +1,145 @@
...
+ *      <li>email address as a couple of local-part and domain
+ *   </ul>
+ *
+ * <p>A state of an entity is defined as a protobuf message.
+ *
+ * @param <I> the type of entity identifier
+ * @param <S> the type of entity state
+ * @author Alexander Yevsyukov
+ */
+public interface EntityLite<I, S extends Message> {
```
armiol: Let's discuss the naming. It's not clear from its description.

server/src/main/java/org/spine3/server/entity/AbstractEntity.java
alexander-yevsyukov: These methods were added as convenience syntax for aggregates that check their status while accepting modifications commands. I guess, they would move to a utility class preserving the names. I'd rather have more specific exceptions than one, which is too general. It would be better for diagnostics.

server/src/main/java/org/spine3/server/entity/EntityLite.java
alexander-yevsyukov: As discussed, we'd have `VersionedEntity` or `EntityWithVersion`, and the main interface called just `Entity`.

This name was after `MessageLite`, but I agree that it may be confusing.

alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/entity/AbstractEntity.java
```diff
@@ -0,0 +1,195 @@
...
+import javax.annotation.Nullable;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.String.format;
+
+/**
+ * Abstract base for lite entities.
```
armiol: Is it still for the `lite` entities?

server/src/main/java/org/spine3/server/entity/DefaultEntityFactory.java
```diff
@@ -0,0 +1,58 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.entity;
+
+import java.lang.reflect.Constructor;
+
+/**
+ * Default factory that creates entities by invoking constructor that
+ * accepts one ID parameter.
```
armiol: Will read better if `accepts a single ID parameter.`

server/src/test/java/org/spine3/server/entity/AbstractVersionableEntityShould.java
```diff
@@ -0,0 +1,56 @@
...
+                                                  .getNumber(), entity.getVersion()
+                                                                      .getTimestamp());
+
+        new EqualsTester().addEqualityGroup(entity, another)
+                          .addEqualityGroup(new AvEntity(42L))
+                          .testEquals();
+    }
+
+    @Test
+    public void have_hashCode() throws Exception {
```
armiol: Why is it empty?

alexander-yevsyukov:
@armiol, PTAL.

armiol:
@alexander-yevsyukov LGTM.

PR #346, opened 2017-02-19
Version and record improvements, and around

alexander-yevsyukov:
This PR introduces the following changes:

* `Version` class backed by Protobuf message `spine.base.Version`.
* Additional checking of version passed to `AbstractVersionedEntity` and `Aggregate`. Now entities cannot accept versions smaller than a number the entity already has. Timestamps are not checked.
* Tests for aggregate behaviour are adjusted accordingly using a new `TimestampProvider` described below.
* A new provider of current time for tests was introduced: “Tests.BackToTheFuture”. It initializes in 30 years ahead from the current moment in time. You can forward back and forth by a number of hours.
* The class `Timestamps` was renamed to `Timestamps2` to avoid name collision with the utility class from the Protobuf library.
* The class `Durations` was renamed to `Durations2` to avoid name collision with the utility class from the Protobuf library.
* Time-related test utilities are moved to `TimeTests` class.
* Deprecates or removes methods in `Timestamps2` that duplicate functionality from `Timestamps`.
* Deprecates or removes methods in `Durations2` that duplication functionality from `Durations`.
* Removes tests that in fact tested `Durations` class, not our code in `Durations2`.

### Breaking changes
* The field `version` in `EventContext` now has type `Version` instead of `int32`
* The field `version` in `AggregateStorageRecord` now has type `Version` instead of `int32`.
* The type `EntityStorageRecord` was renamed to `EntityRecord`. It now has new layout of fields, having `Version` instead of `int32`.
* Methods and types related to `EntityStatus` now are named with `Visibility` (`readVisibility`, `writeVisibility`, etc.).
* `AggregateStorageRecord` was renamed to `AggregateEventRecord` and moved to the `aggregate` package.
* `AggregateEvents` was renamed to `AggregateStateRecord`, it's also now in the `aggregate` package.
* `CommandStorageRecord` was renamed to `CommandRecord`. The proto type layout was changed.
* `EntityField.value` was renamed to `bytes` to better reflect the nature of the content of the field.


alexander-yevsyukov:
@armiol PTAL

server/src/main/java/org/spine3/server/aggregate/AggregateStorage.java
```diff
@@ -62,19 +59,19 @@ protected AggregateStorage(boolean multitenant) {
...
         Snapshot snapshot = null;
 
-        final Iterator<AggregateStorageRecord> historyBackward = historyBackward(aggregateId);
+        final Iterator<AggregateEventRecord> historyBackward = historyBackward(aggregateId);
 
         while (historyBackward.hasNext()
                 && snapshot == null) {
 
-            final AggregateStorageRecord record = historyBackward.next();
+            final AggregateEventRecord record = historyBackward.next();
```
dmdashenkov: May it be just `AggregateRecord`? We store not only events, but also snapshots in this form.

server/src/main/java/org/spine3/server/command/CommandStorage.java
```diff
@@ -102,15 +99,35 @@ protected void store(Command command, Error error) {
         checkNotClosed();
         checkNotDefault(error);
 
+        CommandId id = getOrGenerateCommandId(command);
```
dmdashenkov: `final`?

server/src/main/java/org/spine3/server/entity/AbstractVersionableEntity.java
```diff
@@ -48,9 +49,9 @@
         extends AbstractEntity<I, S>
         implements VersionableEntity<I, S> {
 
-    private final Version version;
+    private Version version;
 
-    private EntityStatus status = EntityStatus.getDefaultInstance();
+    private Visibility visibility = Visibility.getDefaultInstance();
```
dmdashenkov: This is initialized twice (both here and in the `init` method).

server/src/main/java/org/spine3/server/entity/AbstractVersionableEntity.java
```diff
@@ -103,44 +116,81 @@ protected void validate(S state) throws IllegalStateException {
...
 
+    protected void advanceVersion(Version newVersion) {
+        checkNotNull(newVersion);
+        checkIsIncrement(this.getVersion(), newVersion);
+        this.version = newVersion;
+    }
+
+    protected void updateVersion(Version newVersion) {
+        checkNotNull(newVersion);
+        if (this.version.equals(newVersion)) {
```
dmdashenkov: Why calling to `this` so many times in this class?

server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
```diff
@@ -60,12 +60,11 @@ private DefaultEntityStorageConverter(Repository<I, E> repository, FieldMask fie
     protected Tuple<I> doForward(E entity) {
```
dmdashenkov: Usage of `Tuple` with no documentations may lead to <s>brain damage</s> unexpected consequences.

server/src/main/java/org/spine3/server/entity/Predicates.java
```diff
@@ -33,20 +31,20 @@
  */
 public class Predicates {
 
-    private static final Predicate<EntityStatus> isEntityVisible = new Predicate<EntityStatus>() {
+    private static final Predicate<Visibility> isEntityVisible = new Predicate<Visibility>() {
         @Override
-        public boolean apply(@Nullable EntityStatus input) {
+        public boolean apply(@Nullable org.spine3.server.entity.Visibility input) {
```
dmdashenkov: Can we avoid the FQN?

server/src/main/java/org/spine3/server/stand/AggregateQueryProcessor.java
```diff
@@ -136,18 +136,18 @@ public AggregateStateId apply(@Nullable EntityId input) {
...
                                                               : standStorage.readMultiple(stateIds);
-        final ImmutableCollection<EntityStorageRecord> result
+        final ImmutableCollection<EntityRecord> result
                 = FluentIterable.from(bulkReadResults)
-                               .filter(new Predicate<EntityStorageRecord>() {
+                               .filter(new Predicate<EntityRecord>() {
                                    @Override
-                                   public boolean apply(@Nullable EntityStorageRecord input) {
+                                   public boolean apply(@Nullable EntityRecord input) {
                                        return input != null;
```
dmdashenkov: There must be a `Predicate` value for not-null case in `Predicates` class.

server/src/test/java/org/spine3/server/aggregate/AggregateShould.java
```diff
@@ -85,10 +87,13 @@
...
-        aggregate = org.spine3.test.Given.aggregateOfClass(TestAggregate.class)
-                                         .withId(ID)
-                                         .withVersion(100)
-                                         .build();
+        aggregate = newAggregate(ID);
+    }
+
+    private static TestAggregate newAggregate(ProjectId id) {
+        final TestAggregate result = new TestAggregate(id);
+        result.init();
```
dmdashenkov: Why don't we use `Given` here?

server/src/test/java/org/spine3/server/entity/EntityEqualsShould.java
```diff
@@ -96,21 +93,8 @@ public void assure_entities_with_different_states_are_not_equal() {
...
         final TestEntity another = TestEntity.withStateOf(entity);
-        another.setVersion(entity.getVersion()
-                                 .getNumber() + 5, another.whenModified());
+        another.incrementVersion();
 
         assertFalse(entity.equals(another));
     }
-
-    @Test
-    public void assure_entities_with_different_modification_times_are_not_equal() {
```
dmdashenkov: Are those Entities equal now?

testutil/src/main/java/org/spine3/test/TimeTests.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    /**
+     * The time provider that can rewind current time.
+     *
+     * <p>Created in the future, {@linkplain #THIRTY_YEARS_IN_HOURS 30 years} from
+     * the {@link Timestamps2#systemTime() current system time}.
+     */
+    public static class BackToTheFuture implements Timestamps2.Provider {
```
dmdashenkov: Should we still keep our copyright on this class?

testutil/src/test/java/org/spine3/test/TestsShould.java
```diff
@@ -97,15 +87,6 @@ public void do_not_accept_null_UseId_value() {
     }
 
     @Test
-    public void have_frozen_time_provider() {
```
dmdashenkov: Why don't we check this?

server/src/main/proto/spine/server/aggregate/aggregate.proto
```diff
@@ -34,73 +33,80 @@ import "google/protobuf/timestamp.proto";
...
+    // Since events stored are for particular aggregate type, we are not likely to
+    // have a name clash because of not storing fully-qualified names.
+    //
+    // This field is blank if the record stores a snapshot.
+    //
     string event_type = 2;
 
     // String representation of `EventId`.
+    //
+    // When storing a snapshot this field contains the string `Snapshot`.
```
armiol: Maybe `oneof` instead of this strange rule?

server/src/test/java/org/spine3/server/aggregate/Given.java
```diff
@@ -226,17 +225,17 @@ private StorageRecords() {
...
             final Timestamp timestamp2 = add(timestamp1, delta);
             final Timestamp timestamp3 = add(timestamp2, delta);
 
-            final AggregateStorageRecord record1 = StorageRecord.create(timestamp1,
-                                                                        Event.projectCreated(id, createEventContext(timestamp1)));
-            final AggregateStorageRecord record2 = StorageRecord.create(timestamp2,
-                                                                        Event.taskAdded(id, createEventContext(timestamp2)));
-            final AggregateStorageRecord record3 = StorageRecord.create(timestamp3,
-                                                                        Event.projectStarted(id, createEventContext(timestamp3)));
+            final AggregateEventRecord record1 = StorageRecord.create(timestamp1,
```
armiol: Looks like the code of this method doesn't fit 100 columns.

server/src/test/java/org/spine3/testdata/Sample.java
```diff
@@ -65,55 +62,6 @@
     private Sample() {
```
armiol: @dmdashenkov please rename this class. Let's discuss why in person.

testutil/src/test/java/org/spine3/test/TimeTestsShould.java
```diff
@@ -0,0 +1,58 @@
...
+        final Timestamps2.Provider provider =
+                new TimeTests.FrozenMadHatterParty(fiveMinutesAgo);
+
+        assertEquals(fiveMinutesAgo, provider.getCurrentTime());
+    }
+
+    @Test
+    public void return_current_time_in_seconds() {
+        assertNotEquals(0, TimeTests.currentTimeSeconds());
+    }
```
armiol: Do we check the `null` arg values interpretation (i.e. `NPE` thrown) somewhere else? If not, please add such a check to this test.

server/src/test/java/org/spine3/testdata/Sample.java
dmdashenkov: Discussed. Will rename in other PR.

client/src/main/proto/spine/client/query.proto
```diff
@@ -62,7 +69,8 @@ message Query {
 // The result of `Query` processing.
 //
 // Contains the actual processing results and other response attributes.
-// Used as a result of {@code QueryService#Read(Query)} gRPC method call.
+// Used as a result of `QueryService.Read(Query)` gRPC method call.
+//
 message QueryResponse {
 
     // Represents the base part of the response. I.e. whether the `Query` has been acked or not.
```
dmdashenkov: Resolved with @armiol

client/src/main/java/org/spine3/base/Identifiers.java
```diff
@@ -47,14 +47,45 @@ private Identifiers() {
...
+     *   </ul>
+     *
+     * <p>Consider using {@code Message}-based IDs if you want to have typed IDs in your code,
+     * and/or if you need to have IDs with some structure inside.
+     * Examples of such structural IDs are:
+     *   <ul>
+     *      <li>EAN value used in bar codes
+     *      <li>ISBN
+     *      <li>Phone number
+     *      <li>email address as a couple of local-part and domain
```
alexander-yevsyukov: “E-mail” is old style. We prefer “email”.

client/src/main/java/org/spine3/base/Identifiers.java
```diff
@@ -47,14 +47,45 @@ private Identifiers() {
      *
      * @param <I> the type of the ID
      * @param idClass the class of IDs
-     * @throws IllegalArgumentException if the class of IDs is not supported
+     * @throws IllegalArgumentException if the class of IDs is not of supported type
      */
-    public static <I> void checkSupported(Class<I> idClass) {
+    private static <I> void checkSupported(Class<I> idClass) {
```
alexander-yevsyukov: The private static one was left to be used during compile-time model check. But I'm going to move the check from `AbstractEntity` constructor into `Repository.getIdClass()`.

That would optimize the performance (as we won't have to check for id values all the time), and leave only one method.

client/src/main/java/org/spine3/protobuf/Durations2.java
```diff
@@ -82,14 +67,15 @@ public static Duration ofMinutes(long minutes) {
...
         return duration;
     }
 
-    // Methods for brief computations with Durations like
-    //       add(hours(2), minutes(30));
-    /////////////////////////////////////////////////////
+    /*
+     * Methods for brief computations with Durations like
+     *       add(hours(2), minutes(30));
+     ******************************************************/
```
alexander-yevsyukov: We have `fromXxx` methods named after those available from Protobuf.

Our methods that do not have `from` prefix are for calculations similar to that mentioned in the comment. Mixing them with versions with `from` won't read well.

client/src/main/proto/spine/base/event.proto
```diff
@@ -73,18 +80,21 @@ message EventContext {
...
-    int32 version = 5;
+    Version version = 5;
 
     // An event can be enriched with one or more messages. Or, event enrichment can be disabled.
     oneof enrichment_mode {
 
         // One or more enrichments of the event.
-        Enrichments enrichments = 6; // We are having the wrapper type because maps are not allowed in `oneof`s.
+        Enrichments enrichments = 6;
+            // We are having the wrapper type because maps are not allowed in `oneof`s.
```
alexander-yevsyukov: I put the comment on the next line because it 1) doesn't fit into the line 2) it's not meant to be in Javadoc of the generated class 3) I didn't want to break this sentence in two small parts.

client/src/main/proto/spine/base/version.proto
```diff
@@ -19,26 +19,28 @@
...
+message Version {
 
-    // If `true` the aggregate will not be loaded for regular queries.
-    bool archived = 1;
+    // A zero-based version number.
+    //
+    // When numbering versions of entities the zero value is reserved for
+    // the “pre-initialization” version of an entity.
+    //
+    int32 number = 1;
```
alexander-yevsyukov: Well spotted. The initial intent was to use it in the `CommandContext`, which doesn't have to have the target version. But it keeps `int32` now. So, I'll make both `Version` fields required, and we'll see. We can relax it back should we see a need.

client/src/main/proto/spine/base/version.proto
```diff
@@ -19,26 +19,28 @@
...
 
-    // If `true` the aggregate will not be loaded for regular queries.
-    bool deleted = 2;
+    // A moment in time when the new version was created.
+    google.protobuf.Timestamp timestamp = 2;
```
alexander-yevsyukov: We create a new version of the entity.

alexander-yevsyukov: And we definitely create a new instance of `Version` :)

client/src/main/proto/spine/client/query.proto
```diff
@@ -40,13 +40,20 @@ import "spine/client/entities.proto";
 message QueryContext {
 
     ui.Language language = 1;
+
     //TODO:2015-12-16:alexander.yevsyukov: Finish
```
alexander-yevsyukov: No. The comments have what needs to be done in order to make this type definition feature complete for v1.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -293,30 +317,60 @@ private void apply(Iterable<? extends Message> eventMessages, CommandContext com
...
-    private void apply(Message eventOrMsg, CommandContext commandContext)
+    /**
+     * Applies the passed event message or {@code Event} to the aggregate.
+     *
+     * @param eventOrMessage an event message or {@code Event}
+     * @param commandContext the context of the command which handler generated the message or event
+     * @throws InvocationTargetException if the applier method throws the exception
+     * @see #ensureEventMessage(Message)
+     */
+    private void apply(Message eventOrMessage, CommandContext commandContext)
```
alexander-yevsyukov: This is so to support import of events. It does look strange, but it should be OK for a private method.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -349,15 +399,18 @@ void restore(Snapshot snapshot) {
         // See if we're in the state update cycle.
         final B builder = this.builder;
 
-        // If the call to restore() is made during a reply
-        // (because the snapshot event was encountered)
-        // use the currently initialized builder.
+        /* If the call to restore() is made during a replay
+           (because the snapshot event was encountered)
```
alexander-yevsyukov: Not quite. It's stored along with other events. And some people call it a special kind of event. But I agree it may be confusing.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -405,4 +455,15 @@ private IllegalStateException missingEventApplier(Class<? extends Message> event
...
     }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Overrides to expose the method to the package.
+     */
+    @Override
+    @VisibleForTesting
+    protected int versionNumber() {
```
alexander-yevsyukov: Because we inherit it from the parent class.

server/src/main/java/org/spine3/server/aggregate/AggregateStorage.java
alexander-yevsyukov: Making a snapshot is kind of internal optimization event in the history of an aggregate. This message does not keep an aggregate state (unlike `EntityRecord`).

server/src/main/java/org/spine3/server/entity/AbstractVersionableEntity.java
alexander-yevsyukov: Right. Because `init()` may not be called and we want to have a valid field in between.

server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
alexander-yevsyukov: See `EntityStorageConverter.Tuple`. It is documented. I'll add to more references in the class headers.

server/src/test/java/org/spine3/server/aggregate/AggregateShould.java
alexander-yevsyukov: Because we don't need to init specific state.

server/src/main/java/org/spine3/server/stand/AggregateQueryProcessor.java
alexander-yevsyukov: I don't get it. Could you expand on this?

server/src/test/java/org/spine3/server/entity/EntityEqualsShould.java
alexander-yevsyukov: They are not. It's treated by `Version` now.

testutil/src/main/java/org/spine3/test/TimeTests.java
alexander-yevsyukov: :) I don't think we violate copyright here. I selected such a name deliberately because:
1. It prevents getting such a class into production.
2. The class behaviour is expected from the movie.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -248,27 +249,50 @@ private void invokeApplier(Message eventMessage) throws InvocationTargetExceptio
...
-                    setVersion(context.getVersion(), context.getTimestamp());
+                    applyEvent(message);
+                    final Version newVersion = context.getVersion();
+                    advanceVersion(newVersion);
                 } catch (InvocationTargetException e) {
                     throw wrappedCause(e);
                 }
             }
         } finally {
+            /*
```
alexander-yevsyukov: This comment block is actually for the `try/finally` block, not the method call.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -248,27 +249,50 @@ private void invokeApplier(Message eventMessage) throws InvocationTargetExceptio
     /**
      * Applies passed events.
      *
-     * <p>The events passed to this method is the aggregate data loaded
-     * by a repository and passed to the aggregate so that it restores its state.
+     * <p>The events passed to this method is the aggregate data (which may include
```
alexander-yevsyukov: Fixed. Since the comment is on the next line, it's not shown as outdated.

server/src/main/proto/spine/server/aggregate/aggregate.proto
alexander-yevsyukov: Let's discuss together with @dmdashenkov. This and the above field look like kind of optimization for storage, which we dropped already for other record types.

client/src/test/java/org/spine3/base/VersionsShould.java
```diff
@@ -0,0 +1,43 @@
...
+        Tests.hasPrivateParameterlessCtor(Versions.class);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void check_version_increment() {
+        Versions.checkIsIncrement(
+                Tests.newVersionWithNumber(2),
+                Tests.newVersionWithNumber(1)
+        );
+    }
```
alexander-yevsyukov: Null-check added.

alexander-yevsyukov:
@armiol, @dmdashenkov PTAL at the updates and let's discuss outstanding issues.

client/src/main/proto/spine/base/event.proto
dmdashenkov: I was talking about the comment tabulation. It's strange.

server/src/main/java/org/spine3/server/entity/Repository.java
```diff
@@ -88,7 +90,33 @@ protected BoundedContext getBoundedContext() {
...
-        return Classes.getGenericParameterType(getClass(), ID_CLASS_GENERIC_INDEX);
+        final Class<I> idClass = Classes.getGenericParameterType(getClass(),
+                                                                 ID_CLASS_GENERIC_INDEX);
+        checkIdClass(idClass);
+
+        return idClass;
+    }
+
+    /**
+     *  Check that this class of identifiers is supported by the framework.
```
armiol: An extra space before `Check` should probably be removed.

server/src/main/java/org/spine3/server/entity/Repository.java
```diff
@@ -88,7 +90,33 @@ protected BoundedContext getBoundedContext() {
...
+     *
+     * <p>The type of entity identifiers ({@code <I>}) cannot be bound because
+     * it can be {@code Long}, {@code String}, {@code Integer}, and class implementing
+     * {@code Message}.
+     *
+     * <p>We perform the check to to detect possible programming error
+     * in declarations of entity and repository classes <em>Integer</em> we have
```
armiol: I don't get what `<em>Integer</em>` is doing in the middle of this sentence. Perhaps, the sentence is missing something.

testutil/src/main/java/org/spine3/test/EventTests.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private EventTests() {
+        // Prevent instantiation of this utility class.
+    }
+
+    public static EventContext newEventContext() {
+        return newEventContext(getCurrentTime());
+    }
+
+    public static EventContext newEventContext(Timestamp time) {
```
armiol: Please check the `time` for non-null. Seems, there is no `null`-tolerance checking test for this class.

server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
dmdashenkov: Is the content of the `Tuple` always defined (e.g. ID -> `EntityRecord`)? If so, let's rename it to make more understandable.

server/src/main/java/org/spine3/server/stand/AggregateQueryProcessor.java
dmdashenkov: Let's use `Predicates.<?>notNull()` instaed of declaring own `Predicate`.

server/src/main/java/org/spine3/server/entity/Repository.java
alexander-yevsyukov: It should have been _until_.

alexander-yevsyukov:
@armiol PTAL

PR #349, opened 2017-02-22
Prepare to FailureBus

alexander-yevsyukov:
This PR:
* Removes `EventStorageRecord` in favor of passing `Event` to storage.
* Introduces `MessageEnvelope` interface and abstract base class.
* Introduces `Enrichment` type.
* Adds validation attributes to `Event`.
* `CommandEnvelope` moved to the `base` package.

### Breaking changes:
* `CommandClass` and `EventClass` were moved to  the`base` package.
* `EventContext` now uses `Enrichment`.
* `Failure` now has `context` and `enrichment` fields.


alexander-yevsyukov:
@dmdashenkov you may want to have a look as this PR affects storages too.

dmdashenkov:
@alexander-yevsyukov Why do we want to delegate the `Event` and the `EventStorage` validation to the implementations instead of having them in the parent class?

client/src/main/java/org/spine3/base/MessageEnvelope.java
```diff
@@ -18,11 +18,20 @@
...
+ *
+ * @param <T> the type of the object that wraps a message
+ * @author Alex Tymchenko
+ * @author Alexander Yevsyukov
  */
+public interface MessageEnvelope<T> {
 
-@ParametersAreNonnullByDefault
-package org.spine3.server.type;
+    T getOuterObject();
```
armiol: The Javadocs are missing for the methods. Do you want me to pick this up?

client/src/test/java/org/spine3/base/FailuresShould.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.junit.Test;
+import org.spine3.test.Tests;
+
+import static org.junit.Assert.assertFalse;
+
+/**
+ * @author Alexander Yevsyukov
+ */
+public class FailuresShould {
```
armiol: Do we test `Failures` for null-tolerance somewhere else?

client/src/main/java/org/spine3/base/MessageEnvelope.java
alexander-yevsyukov: I'll add it myself.

client/src/test/java/org/spine3/base/FailuresShould.java
alexander-yevsyukov: We don't. The reason I didn't add the check because now there are no methods with arguments in this utility class. But I'll add the check anyway so that we don't return to this matter in the future.

alexander-yevsyukov:
@dmdashenkov we want to validate `Event` at the part that generates them or maybe in `EventBus`. It's too late to validate them in the storage.

PR #355, opened 2017-02-24
Rename `Sample` class

dmdashenkov:
Renamed class `Sample` into [`Exemplum`](https://en.wikipedia.org/wiki/Exemplum).

### DIFF
--- a/server/src/test/java/org/spine3/testdata/Exemplum.java
+++ b/server/src/test/java/org/spine3/testdata/Exemplum.java
@@ -57,9 +57,9 @@ import static java.lang.String.format;
  *
  * @author Dmytro Dashenkov
  */
-public class Sample {
+public class Exemplum {
 
-    private Sample() {
+    private Exemplum() {
     }
 
     public static Event eventBy(Message producerId, Class<? extends Message> eventClass) {
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ buildscript {
         guavaVersion = '20.0'
         protobufGradlePluginVerison = '0.8.0'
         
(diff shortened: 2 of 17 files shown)


dmdashenkov:
@armiol PTAL

server/src/test/java/org/spine3/testdata/Exemplum.java
```diff
@@ -52,14 +52,14 @@
...
 /**
- * Utility for creating simple stubs for generated messages, DTOs (like {@link Event} and {@link Command}),
- * storage objects and else.
+ * Utility for creating simple stubs for generated messages, DTOs (like {@link Event} and
+ * {@link Command}), storage objects and else.
  *
  * @author Dmytro Dashenkov
  */
-public class Sample {
+public class Exemplum {
```
armiol: Isn't it `GeneratedMessages`?

dmdashenkov: The idea was to have human-readable invocations to this class methods:
```
Sample.messageOfType(Foo.class);
Sample.builderForType(Bar.class);
```
The name `Exemplum` is similar to `Example`, though not so abstract, so it could replace `Sample`.
Though the name `GeneratedMessages` says just "utility class for generating messages", not telling the main purpose of this class - generating some __sample__ non-empty data.

I'm not sure if we want to cut that descriptiveness in the name of one more utility name.

dmdashenkov:
@armiol It's becoming too hard to maintain this branch up to date with `master`. Let's first discuss the name and then update it.
Closing this PR for now.

PR #357, opened 2017-02-24
Method predicate improvements

alexander-yevsyukov:
This PR improves hierarchy of `MethodPredicate` classes. Namely:

- `MethodPredicate` is introduced as a generic predicate for methods
- `HandlerMethodPredicate` is introduced as a specific filter for handler methods
- Context class and annotation class are passed as constructor parameters

### DIFF
--- a/server/src/main/java/org/spine3/server/aggregate/EventApplierMethod.java
+++ b/server/src/main/java/org/spine3/server/aggregate/EventApplierMethod.java
@@ -24,6 +24,7 @@ import com.google.common.base.Predicate;
 import com.google.protobuf.Empty;
 import com.google.protobuf.Message;
 import org.spine3.server.reflect.HandlerMethod;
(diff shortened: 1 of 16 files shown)


server/src/main/java/org/spine3/server/aggregate/EventApplierMethod.java
```diff
@@ -49,10 +50,18 @@
     }
 
     /**
-     * Invokes {@link HandlerMethod#invoke(Object, Message, Message)} passing the default message as the context parameter
-     * as event appliers do not have this parameter.
+     * Invokes the applier method.
+     *
+     * <p>The method {@linkplain HandlerMethod#invoke(Object, Message, Message) delegates}
+     * the invocation passing {@linkplain Empty#getDefaultInstance() empty message}
```
armiol: Should probably be `passing an {@linkplain ... empty message}`.

PR #358, opened 2017-02-24
CommandHandler is not an entity

alexander-yevsyukov:
This PR:
* Extracts all code shared by objects that handle commands as static methods in `CommandHandlingMethod`.
* Makes `CommandHandler` an interface with abstract implementation in `AbstractCommandHandler`, which is no longer an entity.
* Isolates code required for working with handling methods under the `reflect` package.

Some more work is expected in following PR(s) for hiding event-related reflection-based code.


alexander-yevsyukov:
@armiol PTAL

server/src/main/java/org/spine3/server/aggregate/AggregateRoot.java
```diff
@@ -63,6 +65,34 @@ protected AggregateRoot(BoundedContext boundedContext, I id) {
     }
 
     /**
+     * Creates a new {@code AggregateRoot} entity and sets it to the default state.
+     *
+     * @param id        the ID of the {@code AggregatePart} is managed by {@code AggregateRoot}
```
armiol: `is` seems unnecessary.

server/src/main/java/org/spine3/server/reflect/CommandHandlerMethod.java
```diff
@@ -67,11 +116,99 @@ public CommandHandlerMethod(Method method) {
...
+
+    /**
+     * Creates new {@code CommandContext} with passed parameters.
+     *
+     * @param producerId the ID of an object which produced the event
+     * @param version optional version of the object
+     * @param commandContext the context of the command handling of which produced the event
+     * @return new {@code CommandContext}
+     */
+    public static EventContext createEventContext(Any producerId,
```
armiol: As long as it is `public`, we should ensure both `producerId` and `commandContext` aren't `null`s.

I also assume we don't have a nullability check for the `static` methods of this class in tests.

server/src/main/java/org/spine3/server/reflect/CommandHandlerMethod.java
```diff
@@ -67,11 +116,99 @@ public CommandHandlerMethod(Method method) {
...
+                                                         .setTimestamp(timestamp)
+                                                         .setCommandContext(commandContext)
+                                                         .setProducerId(producerId);
+        if (version != null) {
+            builder.setVersion(version);
+        }
+        return builder.build();
+    }
+
+    public static List<Event> toEvents(final Any producerId,
```
armiol: Same as above.

server/src/main/java/org/spine3/server/reflect/EventApplierMethod.java
```diff
@@ -35,20 +36,48 @@
...
-    EventApplierMethod(Method method) {
+    private EventApplierMethod(Method method) {
         super(method);
     }
 
+    static EventApplierMethod from(Method method) {
+        return new EventApplierMethod(method);
+    }
+
+    public static EventApplierMethod forEventMessage(Class<? extends Aggregate> cls,
```
armiol: Let's ensure neither `cls` nor `eventMessage` is `null`.

server/src/main/java/org/spine3/server/reflect/EventSubscriberMethod.java
```diff
@@ -36,34 +41,77 @@
...
         super(method);
     }
 
     /**
+     * Obtains the method for handling the event in the passed class.
+     *
+     * @throws IllegalStateException if the passed class does not have an event handling method
+     *                               for the class of the passed message
+     */
+    public static EventSubscriberMethod forMessage(Class<?> cls, Message eventMessage) {
```
armiol: Let's ensure the parameters aren't `null`s.

alexander-yevsyukov:
@armiol PTAL

PR #363, opened 2017-03-01
Fix reporting the code coverage to Codacy

armiol:
As long as `jpm4j` seems to be no longer supported, the suggested way to submit the coverage reports from Travis CI to Codacy became invalid.

This PR is aimed to switch from using the `jpm` tool to a pre-build Codacy reporter JAR.


armiol:
@dmdashenkov PTAL.

PR #365, opened 2017-03-01
`Projection.TypeInfo` and tests code cleanup

alexander-yevsyukov:
This PR:
* Adds `Projection.TypeInfo`
* Enforces checking of parameterless ctor via `Tests.assertHasParameterlessCtor()`
* Removes redundant warning suppression in tests.
* Add test coverage for `UserAccounts` and `TypeInfo` classes.

### DIFF
--- a/server/src/main/java/org/spine3/server/reflect/Classes.java
+++ b/server/src/main/java/org/spine3/server/reflect/Classes.java
@@ -29,6 +29,8 @@ import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 /**
  * Utilities for working with classes.
  *
@@ -52,7 +54,7 @@ public class Classes {
      *     <li>The generic parameter must not be a type variable.
      * </ol>
      *
-     * @param clazz the class to check
+     * @param cls the class to check
      * @param paramNumber a zero-based index of the generic parameter in the class declaration
      * @param <T> the generic type
      * @return the class reference for the generic type
@@ -60,11 +62,13 @@ public class Classes {
      *                            parameter of the expected class
      */
     @CheckReturnValue
-    public static <T> Class<T> getGenericParameterType(Class<?> clazz, int paramNumber) {
+    public static <T> Class<T> getGenericParameterType(Class<?> cls, int paramNumber) {
+        checkNotNull(cls);
+
         // We cast here as we assume that the superclasses of the classes
         // we operate with are parametrized too.
         final ParameterizedType genericSuperclass =
-                (ParameterizedType) clazz.getGenericSuperclass();
+                (ParameterizedType) cls.getGenericSuperclass();
 
         final Type[] typeArguments = genericSuperclass.getActualTypeArguments();
         final Type typeArgument = typeArguments[paramNumber];
@@ -104,17 +108,20 @@ public class Classes {
      *
      * <p>The method must match {@code getFieldName} notation, have no argument to be found.
      *
-     * @param clazz     class containing the getter method
+     * @param cls       class containing the getter method
      * @param fieldName field to find a getter for
      * @return {@link Method} instance reflecting the getter method
      * @throws RuntimeException upon reflective failure
      */
-    public static Method getGetterForField(Class<?> clazz, String fieldName)
+    public static Method getGetterForField(Class<?> cls, String fieldName)
             throws NoSuchMethodException {
+        checkNotNull(cls);
+        checkNotNull(fieldName);
+
         @SuppressWarnings("DuplicateStringLiteralInspection")
         final String fieldGetterName = "get" + fieldName.substring(0, 1)
                                                         .toUpperCase() + fieldName.substring(1);
-        final Method fieldGetter = clazz.getMethod(fieldGetterName);
+        final Method fieldGetter = cls.getMethod(fieldGetterName);
         return fieldGetter;
     }
 }
--- a/server/src/test/java/org/spine3/server/aggregate/AggregatePartShould.java
+++ b/server/src/test/java/org/spine3/server/aggregate/AggregatePartShould.java
@@ -91,7 +91,7 @@ public class AggregatePartShould {
     }
 
     @Test
-    public void have_TypeInfo() {
+    public void have_TypeInfo_utility_class() {
         assertHasPrivateParameterlessCtor(AggregatePart.TypeInfo.class);
     }
 
--- a/server/src/test/java/org/spine3/server/aggregate/AggregatePartShould.java
+++ b/server/src/test/java/org/spine3/server/aggregate/AggregatePartShould.java
@@ -32,6 +32,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.spine3.base.Identifiers.newUuid;
 import static org.spine3.server.aggregate.AggregatePart.create;
 import static org.spine3.server.aggregate.AggregatePart.getConstructor;
+import static org.spine3.test.Tests.assertHasPrivateParameterlessCtor;
 
 /**
  * @author Illia Shepilov
@@ -87,6 +88,11 @@ public class AggregatePartShould {
         final Constructor<AnAggregatePart> constructor =
                 getConstructor(AnAggregatePart.class);
         assertNotNull(constructor);
+    }
+
+    @Test
+    public void have_TypeInfo() {
+        assertHasPrivateParameterlessCtor(AggregatePart.TypeInfo.class);
     }
 
     /*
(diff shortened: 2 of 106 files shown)


alexander-yevsyukov:
@armiol, PTAL

alexander-yevsyukov:
@dmdashenkov PTAL

server/src/main/java/org/spine3/server/reflect/Classes.java
```diff
@@ -52,19 +54,21 @@ private Classes() {
      *     <li>The generic parameter must not be a type variable.
      * </ol>
      *
-     * @param clazz the class to check
+     * @param cls the class to check
```
dmdashenkov: Why is `cls` better than `clazz`?

server/src/test/java/org/spine3/server/aggregate/AggregatePartShould.java
```diff
@@ -89,6 +90,11 @@ public void obtain_aggregate_part_constructor() {
         assertNotNull(constructor);
     }
 
+    @Test
+    public void have_TypeInfo() {
+        assertHasPrivateParameterlessCtor(AggregatePart.TypeInfo.class);
```
dmdashenkov: It feels there is a mismatch between the test case name and its content. Why don't we split this into a private ctor test and call to `TypeInfo` test, that will make sure it's present?

dmdashenkov: Or we can just use `Class.forName` to make sure the class exists.

server/src/test/java/org/spine3/server/aggregate/AggregateShould.java
```diff
@@ -586,6 +590,10 @@ public void do_not_allow_getting_state_builder_from_outside_the_event_applier()
         new TestAggregateWithIdInteger(100).getBuilder();
     }
 
+    @Test
+    public void have_TypeInfo() {
+        assertHasPrivateParameterlessCtor(Aggregate.TypeInfo.class);
```
dmdashenkov: The same as above.

server/src/test/java/org/spine3/server/entity/EntityShould.java
```diff
@@ -299,6 +300,11 @@ public void create_and_initialize_entity_instance() {
         assertFalse(entity.isDeleted());
     }
 
+    @Test
+    public void have_TypeInfo() {
+        assertHasPrivateParameterlessCtor(Entity.TypeInfo.class);
```
dmdashenkov: The same as above.

server/src/test/java/org/spine3/server/entity/GetTargetIdFromCommandShould.java
```diff
@@ -44,6 +43,7 @@ public void return_empty_Optional_if_fail_to_get_ID_from_command_message_without
         assertFalse(id.isPresent());
     }
 
+    @SuppressWarnings("OptionalGetWithoutIsPresent")
```
dmdashenkov: Please comment this suppression.

server/src/test/java/org/spine3/server/procman/ProcessManagerShould.java
```diff
@@ -323,6 +327,11 @@ CommandRouted handle(StartProject command, CommandContext context) {
         }
     }
 
+    @Test
+    public void have_TypeInfo() {
+        assertHasPrivateParameterlessCtor(ProcessManager.TypeInfo.class);
```
dmdashenkov: The same as cases above.

server/src/test/java/org/spine3/server/projection/ProjectionShould.java
```diff
@@ -73,13 +73,19 @@ public void throw_exception_if_no_handler_for_event() {
...
+                Projection.TypeInfo.getEventClasses(TestProjection.class);
 
         assertEquals(TestProjection.HANDLING_EVENT_COUNT, classes.size());
         assertTrue(classes.contains(EventClass.of(StringValue.class)));
         assertTrue(classes.contains(EventClass.of(Int32Value.class)));
     }
 
+    @Test
+    public void have_type_info_utility_class() {
+        assertHasPrivateParameterlessCtor(Projection.TypeInfo.class);
```
dmdashenkov: The same case again. BTW, this test case has another name which may be more appropriate for all the test cases like this.

server/src/test/java/org/spine3/server/projection/ProjectionRepositoryShould.java
```diff
@@ -496,31 +497,32 @@ private static TaskAdded taskAdded() {
...
 
-        protected ManualCatchupProjectionRepository(BoundedContext boundedContext, Duration catchUpMaxDuration) {
+        protected ManualCatchupProjectionRepository(BoundedContext boundedContext,
+                                                    Duration catchUpMaxDuration) {
             super(boundedContext, false, catchUpMaxDuration);
         }
 
+        @SuppressWarnings("unused")
         @Subscribe
         public void apply(ProjectCreated event, EventContext eventContext) {
```
dmdashenkov: Maybe we should finally add an IDEA config to avoid suppressing handler methods as `unused`?

testutil/src/main/java/org/spine3/test/Tests.java
```diff
@@ -64,12 +67,24 @@ private Tests() {
...
      *     {@literal @}Test
      *     public void have_private_utility_ctor() {
-     *         assertTrue(hasPrivateParameterlessCtor(MyUtility.class));
+     *         assertHasPrivateParameterlessCtor(MyUtility.class));
      *     }
      * </pre>
-     * @return true if the class has private parameter-less constructor
      */
-    public static boolean hasPrivateParameterlessCtor(Class<?> targetClass) {
+    public static void assertHasPrivateParameterlessCtor(Class<?> targetClass) {
```
dmdashenkov: For methods like this we have `Verify` class. Let's move `assertHasPrivateParameterlessCtor` there.

dmdashenkov: BTW, `Verify extends Assert`, so _JUnit v4.x_ is not going to leave us in the nearest future.

users/src/main/java/org/spine3/users/UserAccounts.java
```diff
@@ -20,28 +20,42 @@
...
      * <p>This value should be used for calculating the cache size for storing
      * {@code UserAccount} instances.
      *
      * @see <a href="https://en.wikipedia.org/wiki/Dunbar%27s_number">Dunbar's number</a>
      */
-    public static final int DUNBARS_NUMBER = 150;
-
-    private UserAccounts() {}
+    public static int getDunbarsNumber() {
+        return DUNBARS_NUMBER;
```
dmdashenkov: Why do we need a getter for a constant?

users/src/test/java/org/spine3/users/UserAccountsShould.java
```diff
@@ -36,6 +46,22 @@ public void return_absent_for_google_uid_if_no_linked_account_added() {
...
+    @SuppressWarnings("OptionalGetWithoutIsPresent")
+    @Test
+    public void obtain_user_google_uid() {
+        final String uid = newUuid(); // In reality Google's user ID has different formant.
+
+        final UserAccount userAccount = UserAccount.newBuilder()
+                                             .addLinkedIdentity(
+                                                     UserInfo.newBuilder()
+                                                             .setProviderId(GOOGLE_AUTH_PROVIDER_ID)
+                                                             .setUid(uid))
```
dmdashenkov: Please, make the code block more readable.

users/src/main/java/org/spine3/users/UserAccounts.java
alexander-yevsyukov: This is a security thing. A constant can be modified. See how Guava does this.

testutil/src/main/java/org/spine3/test/Tests.java
alexander-yevsyukov: We may drop the whole Verify then.

server/src/test/java/org/spine3/server/projection/ProjectionRepositoryShould.java
alexander-yevsyukov: This is not about the method. This is about parameters.

server/src/main/java/org/spine3/server/reflect/Classes.java
alexander-yevsyukov: Shorter. This is the way Guava does.

server/src/test/java/org/spine3/server/aggregate/AggregatePartShould.java
alexander-yevsyukov: `TypeInfo` is the utility class. In the method name we declare that type information should be obtained via such class. In the body of the method we cover the utility constructor.

alexander-yevsyukov:
@dmdashenkov PTAL

PR #366, opened 2017-03-02
Avoid dependency on JUnit4 from `Tests`

alexander-yevsyukov:
This PR avoids dependency on JUnit4  recently introduced in the `Tests` class.


alexander-yevsyukov:
@dmdashenkov PTAL

PR #368, opened 2017-03-02
Disallow `dispatchForTest()` in production

alexander-yevsyukov:
This PR:
* Prevents calling `dispatchForTest()` in the production mode.
* Adds concurrency support in `Dispatcher` registry.


alexander-yevsyukov:
@dmdashenkov PTAL

PR #369, opened 2017-03-02
Avoid dependency on JUnit4 from `Tests`

alexander-yevsyukov:
This PR avoids dependency on JUnit4 recently introduced in the `Tests` class

### DIFF
--- a/testutil/src/main/java/org/spine3/test/Tests.java
+++ b/testutil/src/main/java/org/spine3/test/Tests.java
@@ -40,8 +40,6 @@ import java.lang.reflect.Modifier;
 import java.util.List;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 
 /**
  * Utilities for testing.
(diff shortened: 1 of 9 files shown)


alexander-yevsyukov:
@dmdashenkov PTAL

testutil/src/test/java/org/spine3/test/TestsShould.java
```diff
@@ -97,6 +97,32 @@ public void return_empty_StreamObserver() {
...
+    }
+
+    @SuppressWarnings("ConstantConditions") // The call with `false` should always fail.
+    @Test(expected = AssertionError.class)
+    public void have_own_assertTrue() {
+        // This should pass.
+        Tests.assertTrue(true);
+
+        // This should fail.
+        Tests.assertTrue(false);
```
dmdashenkov: This test case in error-prone. Both the first and the second statement may cause an `AssertionError`. 
Try using `try-catch` instead. This may help us to determine that the `AssetionError` is thrown from the second statement.

testutil/src/main/java/org/spine3/test/Tests.java
```diff
@@ -55,6 +53,33 @@ private Tests() {
...
+     */
+    @VisibleForTesting
+    static void assertEquals(boolean expected, boolean actual) {
+        if (expected != actual) {
+            throw new AssertionError();
+        }
+    }
+
+    /**
+     * Asserts that a condition is true. If it isn't it throws an
```
dmdashenkov: Please add a comma.

alexander-yevsyukov:
@dmdashenkov PTAL

PR #370, opened 2017-03-02
Isolate current tenant modifications

alexander-yevsyukov:
This PR improves the way we work with multi-tenancy.

### Tenant-Aware operations
A tenant-aware operation (see `TenantAware` class) is an operation which requires `TenantId` present in the execution context.  This PR isolates tenant-aware operations under the `server.tenant` package making such “transparent” to most of the code of the framework. 

Setting and obtaining the current tenant ID is performed by package-private class 
`CurrentTenant`. 

### Proto changes
* Changed order and numbers in `CommandContext` proto. No further changes are expected.
* `QueryId`, `QueryContext` are introduced, and are now parts of `Query`.
* `CommandValidationError` enum got new item `TENANT_INAPPLICABLE`.
 
### New and changed API
* `TenantAwareOperation` — a `Runnable` which works in a tenant context.
* `TenantAwareFunction` — a `Function` with one parameter.
* `TenantAwareFunction0` — parameterless function.
* `TenantFunction` — the function that accepts the current `TenantId`.
* `CommandOperation`, `EventOperation`, and `QueryOperation` are abstract classes for tenant-aware operations that are performed in response to corresponding messages that hold a `TenantId` in their contexts.
* `TenantIndex` — the interface for registering new `TenantId`s.
* `AllTenantOperation` is the operation that requires `TenantIndex` for its execution. Catching up a projection repository for all tenants is an example of such an operation.
* `TenantRepository` — the abstract base for repository-based implementations of `TenantIndex`, with the default implementation `DefaultTenantRepository`.
* `StorageFactorySwitch` now has single- and multi-tenant implementations. Initializing is now done via the instance method (`init()`). Previously it was done via the static method with the same name.
* `CommandStore` now accepts `TenantIndex` as the second parameter of the constructor.

### Other notable changes
* Multi-tenancy of a `CommandBus` is now set in `Builder` and cannot be modified.
* `CommandBus` now allows to have `CommandBusFilter`s.
* `CommandStatusService` was transformed into corresponding methods of `CommandStore`.
* `Rescheduler` became a part of `CommandScheduler`.
* `CommandScheduler` is now one of default `CommandBusFilter`s.
* Command validation is now performed using `CommandValidationFilter`.
* Handling a dead command is now performed by `DeadCommandFilter`.
* `BoundedContext.Builder` now can accept a custom `TenantIndex`. If not set, it would create the default implementation.
* Repositories now perform tenant-aware operations on their storages.
* Getters in some builders are got the `get` prefix back.
* `AggregateRepository.snapshotTrigger` property is made `protected`. Derived classes can expose it, if needed.
* `RepositoryView.load()` was renamed to `find()` to better reflect possible negative outcome of the method.
* `RecordBasedRepository.loadOrCreate()` was renamed to `findOrCreate()`.
* `Repository.getStorage()` now returns `Storage` instead of `AutoClosable`.
* The `Storage` interface got method `isOpen()`.
* `Storage.close()` no longer throws an `Exception`. Implementations may define, if needed.
* `IntegrationEventBus` was removed. Integration events are to be delivered via `BoundedContext.notify()`.
* `ProcessManager`'s `dispatchEvent()` no longer throws an `InvocationTargetException`. If it occurs it would be propagated inside `ISE` with the root cause.
* `Stand` is now multi-tenancy aware. Further testing will be done in a separate PR.
* `StorageFactory` got method `toSingleTenant()` which allows to convert a multi-tenant implementation into single-tenant. This is required for creating storages for single-tenant repositories (like default implementation of `TenantIndex`) in a multi-tenant application.

EOF


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/command/CommandScheduler.java
```diff
@@ -20,32 +20,83 @@
...
+    @Nullable
     private CommandBus commandBus;
 
+    @Nullable
+    private Rescheduler rescheduler;
+
+    protected CommandScheduler() {
+    }
+
+    public void setCommandBus(CommandBus commandBus) {
```
armiol: Looks like its only usage is in the `CommandBus.Builder`, so let's hide it from the public API.

server/src/main/java/org/spine3/server/command/CommandScheduler.java
```diff
@@ -20,32 +20,83 @@
...
+        this.commandBus = commandBus;
+        this.rescheduler = new Rescheduler(commandBus);
+    }
+
+    Rescheduler rescheduler() {
```
armiol: This method can be `private`.

server/src/main/java/org/spine3/server/command/CommandStore.java
```diff
@@ -107,92 +153,164 @@ public void storeWithError(Command command, CommandException exception) {
...
+        };
+        op.execute();
     }
 
     /**
-     * Sets the status of the command to {@link CommandStatus#OK}
-     *
-     * @param commandId an ID of the command
-     * @throws IllegalStateException if the storage is closed
+     * Updates the status of the command with the passed error.
```
armiol: Please format the Javadoc section.

server/src/main/java/org/spine3/server/command/CommandValidationFilter.java
```diff
@@ -0,0 +1,105 @@
...
+
+    CommandValidationFilter(CommandBus commandBus) {
+        this.commandBus = commandBus;
+    }
+
+    /**
+     * Returns {@code true} if a command is valid, {@code false} otherwise.
+     */
+    @Override
+    public boolean accept(CommandEnvelope envelope, StreamObserver<Response> responseObserver) {
```
armiol: Let's try to simplify this method by extracting some bits as methods. At the moment there are three `return` conditions with a lot of `if` statements.

server/src/main/java/org/spine3/server/command/FilterChain.java
```diff
@@ -0,0 +1,103 @@
...
+
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * The chain of {@code CommandBusFilter}s that {@link CommandBus} applies before
+ * {@linkplain CommandBus#post(Command, StreamObserver) posting} a command.
+ *
+ * @author Alexander Yevsyukov
+ */
+class FilterChain implements CommandBusFilter {
```
armiol: Looks `final`.

server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
```diff
@@ -337,20 +338,42 @@ void store(Collection<P> projections) {
      * Updates projections from the event stream obtained from {@code EventStore}.
      */
     public void catchUp() {
```
armiol: While the implementation looks logical enough, it will probably take ages to complete in an AppEngine application.

Let's discuss the example I have. It's easier to talk about it, than describe in a written form.

server/src/main/java/org/spine3/server/tenant/AllTenantOperation.java
```diff
@@ -0,0 +1,60 @@
...
+public abstract class AllTenantOperation implements Runnable {
+
+    private final TenantIndex tenantIndex;
+
+    /**
+     * Creates a new instance of an operation that will be {@linkplain #execute()} executed}
+     * for all tenants in the passed index.
+     */
+    protected AllTenantOperation(TenantIndex tenantIndex) {
+        this.tenantIndex = tenantIndex;
```
armiol: Since this is `public` class, I'd check for `null` here explicitly — just like we do in our public `static` methods.

server/src/main/java/org/spine3/server/tenant/CommandOperation.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.spine3.base.Command;
+import org.spine3.base.CommandId;
+
+/**
+ * A tenant-aware operation performed in response to a command or in relation to a command.
+ *
+ * @author Alexander Yevsyukov
+ */
+public abstract class CommandOperation extends TenantAwareOperation {
```
armiol: Not sure, but this class (at the very least) and maybe the whole package looks `@Internal`.

The more I see, the more I think about introducing such a "restriction" for the external framework users.

server/src/main/java/org/spine3/server/tenant/TenantAware.java
```diff
@@ -0,0 +1,99 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.validate.Validate.isDefault;
+
+/**
+ * An abstract base for operations performed in a tenant context.
+ *
+ * @author Alexander Yevsyukov
+ */
+class TenantAware {
```
armiol: It's not `abstract` as it should be according to its docs.

server/src/main/java/org/spine3/server/tenant/TenantAwareFunction0.java
```diff
@@ -0,0 +1,55 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * A tenant-aware function that takes no parameters.
+ *
+ * @param <T> the type of the result returned by the function
+ * @author Alexander Yevsyukov
+ */
+public abstract class TenantAwareFunction0<T> extends TenantAwareFunction<Empty, T> {
```
armiol: In which case it's just a `TenantAwareRunnable` with the abstract `run()` instead of `apply()`.

armiol:
@alexander-yevsyukov another general comment is that everything is very well hidden and separated :)  It's unclear how we would implement cross-tenant logic, that we have discussed previously.

server/src/main/java/org/spine3/server/command/CommandBus.java
```diff
@@ -219,129 +180,68 @@ public void post(Command command, StreamObserver<Response> responseObserver) {
...
+        commandStore().store(command);
         responseObserver.onNext(Responses.ok());
-        doPost(commandEnvelope, dispatcher.get());
+
+        doPost(commandEnvelope);
         responseObserver.onCompleted();
     }
 
+    /**
+     * Does nothing because commands for which are no registered dispatchers
```
alexander-yevsyukov: Added.

server/src/main/java/org/spine3/server/command/CommandScheduler.java
alexander-yevsyukov: It's going to be `protected`. We need this method to inject the Command Bus into scheduler when the bus is built.

server/src/main/java/org/spine3/server/command/CommandValidationFilter.java
alexander-yevsyukov: Done

alexander-yevsyukov:
@armiol, as discussed vocally, we'd verify the ease of multi-tenancy support when implementing new GAE and JDBC storages. That may affect our API.

PTAL at my recent fixes made according to your comments.

armiol:
@alexander-yevsyukov LGTM.

PR #371, opened 2017-03-06
Repository views

alexander-yevsyukov:
This PR introduces:
* `DefaultRecordBasedRepository`, which is `RecordBasedRepository` with default entity construction and conversion behaviour.
* Index iterators for storages
* Filtering iterator for repositories

### DIFF
--- a/server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
+++ b/server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
@@ -64,7 +64,7 @@ class DefaultEntityStorageConverter<I, E extends AbstractEntity<I, S>, S extends
                             .setState(stateAny);
 
         if (entity instanceof AbstractVersionableEntity) {
-            AbstractVersionableEntity versionable = (AbstractVersionableEntity) entity;
+            final AbstractVersionableEntity versionable = (AbstractVersionableEntity) entity;
             builder.setVersion(versionable.getVersion())
                    .setVisibility(versionable.getVisibility());
         }
@@ -86,7 +86,7 @@ class DefaultEntityStorageConverter<I, E extends AbstractEntity<I, S>, S extends
         final EntityRecord record = tuple.getState();
         if (entity != null) {
             if (entity instanceof AbstractVersionableEntity) {
-                AbstractVersionableEntity versionable = (AbstractVersionableEntity) entity;
+                final AbstractVersionableEntity versionable = (AbstractVersionableEntity) entity;
                 versionable.setState(state, record.getVersion());
                 versionable.setVisibility(record.getVisibility());
             } else {
--- a/server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
+++ b/server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
@@ -23,7 +23,6 @@ package org.spine3.server.entity;
 import com.google.protobuf.Any;
 import com.google.protobuf.FieldMask;
 import com.google.protobuf.Message;
-import org.spine3.base.Version;
 import org.spine3.protobuf.TypeUrl;
 
 import static org.spine3.protobuf.AnyPacker.pack;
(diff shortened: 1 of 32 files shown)


server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
```diff
@@ -60,12 +59,15 @@ private DefaultEntityStorageConverter(Repository<I, E> repository, FieldMask fie
...
-        final Version version = entity.getVersion();
         final EntityRecord.Builder builder =
                 EntityRecord.newBuilder()
-                            .setState(stateAny)
-                            .setVersion(version)
-                            .setVisibility(entity.getVisibility());
+                            .setState(stateAny);
+
+        if (entity instanceof AbstractVersionableEntity) {
+            AbstractVersionableEntity versionable = (AbstractVersionableEntity) entity;
```
armiol: Should probably be declared `final`.

server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
```diff
@@ -83,8 +85,13 @@ protected E doBackward(Tuple<I> tuple) {
 
         final EntityRecord record = tuple.getState();
         if (entity != null) {
-            entity.setState(state, record.getVersion());
-            entity.setVisibility(record.getVisibility());
+            if (entity instanceof AbstractVersionableEntity) {
+                AbstractVersionableEntity versionable = (AbstractVersionableEntity) entity;
```
armiol: Should probably be declared `final`.

BTW, looks like an IDEA `instanceof` template should be edited :)

server/src/main/java/org/spine3/server/entity/Repository.java
```diff
@@ -294,4 +307,44 @@ public int getIndex() {
             return result;
         }
     }
+
+    /**
+     * An iterator of all entities from the storage.
+     */
+    private static class EntityIterator<I, E extends Entity<I, ?>>
```
armiol: This implementation is suitable only for `in-memory` storages. Should we say that in the documentation and recommend to override it via redefining an `#iterator`?

server/src/main/java/org/spine3/server/entity/Repository.java
```diff
@@ -294,4 +307,44 @@ public int getIndex() {
...
+
+        @Override
+        public boolean hasNext() {
+            final boolean result = index.hasNext();
+            return result;
+        }
+
+        @Override
+        public E next() {
+            I id = index.next();
```
armiol: Let's make it `final`.

server/src/main/java/org/spine3/server/entity/Repository.java
```diff
@@ -294,4 +307,44 @@ public int getIndex() {
             return result;
         }
     }
+
+    /**
+     * An iterator of all entities from the storage.
```
armiol: As this `Iterator` implementation does not support element removal, let's mention that in docs.

server/src/main/java/org/spine3/server/storage/memory/TenantStorage.java
```diff
@@ -30,6 +32,8 @@
  */
 interface TenantStorage<I, R> {
 
+    Iterator<I> index();
```
armiol: The method-level Javadocs are missing for this interface.

server/src/main/java/org/spine3/server/storage/Storage.java
```diff
@@ -42,4 +46,29 @@
...
+
+    /**
+     * Returns an iterator over identifiers of records in the storage.
+     */
+    Iterator<I> index();
+
+    /**
+     * Reads a record from the storage by the passed ID.
+     *
+     * @param id the ID of the record to load
```
armiol: Probably, should be `to read` to keeps the term usage consistent.

server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
alexander-yevsyukov: > BTW, looks like an IDEA instanceof template should be edited :)
You're right. Can we propagate it to the whole team automatically?

server/src/main/java/org/spine3/server/entity/Repository.java
alexander-yevsyukov: Here we rely on having an index of identifiers from underlying storage. It can be implemented for any storage, not only in-memory.

alexander-yevsyukov: The idea is to use iterator over IDs instead of `loadAll()` methods in `RecordBasedRepository`.

alexander-yevsyukov:
@armiol PTAL

server/src/main/java/org/spine3/server/entity/DefaultRecordBasedRepository.java
```diff
@@ -0,0 +1,64 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.entity;
+
+import com.google.protobuf.Message;
+import org.spine3.server.BoundedContext;
+
+/**
+ * Implementation of {@link RecordBasedRepository} that manges entities
```
armiol: Should be `manages`.

PR #372, opened 2017-03-06
Lifecycle flags for entities

alexander-yevsyukov:
This PR:
* Removes `markArchived()` and `markDeleted()` methods in repositories and storages.
* Renames `Visibility` proto type into `LifecycleFlags`.
* Introduces `EntityWithLifecycle` interface.
* Introduces `StorageWithLifecycleFlags` interface.
* Renames `VisibilityField` into `LifecycleFlagField`.


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/entity/VisibleEntity.java
```diff
@@ -18,27 +18,24 @@
...
-package org.spine3.change;
+package org.spine3.server.entity;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
+import com.google.protobuf.Message;
 
 /**
- * Routines common for all mismatches.
+ * An entity which has visibility.
```
armiol: Can we reference some definition of `visibility`? Or something that would allow readers to catch up on the subject.

server/src/main/java/org/spine3/server/storage/RecordStorage.java
```diff
@@ -115,22 +118,36 @@ public void write(Map<I, EntityRecord> records) {
...
+    public void writeVisibility(I id, Visibility visibility) {
+        final Optional<EntityRecord> optional = read(id);
+        if (optional.isPresent()) {
+            final EntityRecord record = optional.get();
+            final EntityRecord updated = record.toBuilder()
+                                               .setVisibility(visibility)
+                                               .build();
+            write(id, updated);
+        } else {
+            // The AggregateStateId is a special case, which is not handled by the Identifier class.
```
armiol: `Identifiers`?

server/src/main/java/org/spine3/server/entity/VisibleEntity.java
```diff
@@ -18,27 +18,24 @@
...
-
-    static void checkNotNullOrEqual(Object expected, Object actual) {
-        checkNotNull(expected);
-        checkNotNull(actual);
-        checkArgument(!expected.equals(actual), ERR_CANNOT_BE_EQUAL);
-    }
+    /**
+     * Verifies whether visibility of the entity changed since its initialization.
+     */
+    boolean visibilityChanged();
```
armiol: That's an odd interface method. Seems more like a utility method of an `abstract class`-level.

server/src/main/java/org/spine3/server/entity/VisibleEntity.java
```diff
@@ -18,27 +18,24 @@
...
-        // Prevent instantiations.
-    }
-
-    static void checkNotNullOrEqual(Object expected, Object actual) {
-        checkNotNull(expected);
-        checkNotNull(actual);
-        checkArgument(!expected.equals(actual), ERR_CANNOT_BE_EQUAL);
-    }
+    /**
+     * Verifies whether visibility of the entity changed since its initialization.
```
armiol: Should probably be `Tells whether ...`, since if fact there is verification done per se, i.e. no exceptions thrown.

server/src/main/java/org/spine3/server/storage/StorageWithVisibility.java
```diff
@@ -0,0 +1,53 @@
...
+ * @author Alexander Yevsyukov
+ */
+public interface StorageWithVisibility<I, R extends Message> extends Storage<I, R> {
+
+    /**
+     * Reads the visibility status for the entity with the passed ID.
+     *
+     * <p>This method returns {@code Optional.absent()} if none of the
+     * flags of visibility flags were set before. This means that
+     * the entity is visible to the regular queries.
```
armiol: While it is perfectly clear to me, `the regular queries` may cause some questions. What are they? Are there irregular queries? Do the irregular queries act differently in the past? Etc.

server/src/main/java/org/spine3/server/storage/RecordStorage.java
alexander-yevsyukov: The code that creates the string form of an ID is actually in the `Identifier.Type` enum. 

The `AggregateStateId` being a plain Java class (not one of primitive wrappers that we support, and not a `Message`) doesn't fall into the `case` in `Identifier.toString()`, which causes the error.

I didn't want to extend the `Identifier` for this particular case, hence the comment.

server/src/main/java/org/spine3/server/storage/StorageWithVisibility.java
alexander-yevsyukov: I'll remove the last sentence. It repeats what is already said in `Visibility`.

alexander-yevsyukov:
@armiol, PTAL.

alexander-yevsyukov:
@armiol, PTAL. Specific repository views will be in an separate PR.

server/src/main/java/org/spine3/server/storage/StorageWithLifecycleFlags.java
```diff
@@ -18,27 +18,35 @@
...
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
+import com.google.common.base.Optional;
+import com.google.protobuf.Message;
+import org.spine3.server.entity.LifecycleFlags;
 
 /**
- * Routines common for all mismatches.
+ * A storage that allows to update lifecycle flags of entities.
```
armiol: Let's reference the `LifecycleFlags` via `{@link...}` here as well.

server/src/main/java/org/spine3/server/storage/memory/InMemoryStandStorage.java
```diff
@@ -66,10 +66,13 @@ public static Builder newBuilder() {
...
+    public ImmutableCollection<EntityRecord> readAllByType(final TypeUrl type,
+                                                           FieldMask fieldMask) {
         final Map<AggregateStateId, EntityRecord> allRecords = readAll(fieldMask);
-        final Map<AggregateStateId, EntityRecord> resultMap = Maps.filterKeys(allRecords, new Predicate<AggregateStateId>() {
-            @Override
+        final Map<AggregateStateId, EntityRecord> resultMap = Maps.filterKeys(
+                allRecords,
+                new Predicate<AggregateStateId>() {
+                    @Override
             public boolean apply(@Nullable AggregateStateId stateId) {
```
armiol: Looks unformatted. Maybe, just a GitHub bug.

server/src/main/proto/spine/server/entity/entity.proto
```diff
@@ -48,19 +48,28 @@ message EntityRecord {
...
+    LifecycleFlags lifecycle_flags = 4;
 }
 
-// The visibility flags of an entity.
-//
-// If one of the flags is set to `true` the entity becomes “invisible” to regular queries.
-//
-message Visibility {
+// The lifecycle flags of an entity.
+message LifecycleFlags {
```
armiol: It's a pity we have to duplicate `LifecycleFlags` proto message and `LifecycleFlagField` class structure.

Can we avoid it somehow? ATM, we'll have to append the same changes to these two classes if any new flags are introduced.

alexander-yevsyukov: It would require having a general mechanism of transforming a message into a set of storage fields. And detecting simple cases like just `boolean` flags like this. Kinda Hibernate :) I'd say, it should be a separate PR anyway.

alexander-yevsyukov:
@armiol, PTAL.

armiol:
@alexander-yevsyukov as discussed vocally, the storage mechanism will be updated and refactored.
So LGTM for now.

PR #373, opened 2017-03-07
`InternetDomains` and `EmailAddresses` utility classes

alexander-yevsyukov:
This PR introduces `InternetDomains` and `EmailAddresses` utility classes.


alexander-yevsyukov:
It is expected that further enhancement will be made when we start working with importing user data.

See recommendations from a StackOverflow [answer on the topic](http://stackoverflow.com/questions/11809631/fully-qualified-domain-name-validation).

alexander-yevsyukov:
Note that [patterns from the Android SDK](https://developer.android.com/reference/android/util/Patterns.html) are now deprecated.

client/src/main/java/org/spine3/net/EmailAddresses.java
```diff
@@ -0,0 +1,74 @@
...
+
+package org.spine3.net;
+
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Utility class for working with {@link EmailAddress}es.
```
armiol: Should be ` with {@linkplain EmailAddress email addresses}.`

client/src/main/java/org/spine3/net/InternetDomains.java
```diff
@@ -0,0 +1,118 @@
...
+
+package org.spine3.net;
+
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Utility class for working with {@link InternetDomain}s.
```
armiol: Should be something like ` with {@linkplain InternetDomain internet domains}.`

client/src/main/java/org/spine3/net/InternetDomains.java
```diff
@@ -0,0 +1,118 @@
...
+            + "|v[aceginu]"
+            + "|w[fs]"
+            + "|(xn\\-\\-0zwm56d|xn\\-\\-11b5bs3a9aj6g|xn\\-\\-80akhbyknj4f|xn\\-\\-9t4b11yi5a|xn\\-\\-deba0ad|xn\\-\\-g6w251d|xn\\-\\-hgbk6aj7f53bba|xn\\-\\-hlcj6aya9esc7a|xn\\-\\-jxalpdlp|xn\\-\\-kgbechtv|xn\\-\\-zckzah)"
+            + "|y[etu]"
+            + "|z[amw])";
+
+    /**
+     * Regular expression pattern to match all IANA top-level domains.
+     */
+    private static final Pattern TOP_LEVEL_DOMAIN = Pattern.compile(TOP_LEVEL_DOMAIN_STR);
```
armiol: I would create an enum for the `Pattern`-typed constants, as they can be easily grouped. Perhaps, the same for the string regexps.

client/src/main/java/org/spine3/net/EmailAddresses.java
alexander-yevsyukov: The idea of the current syntax is to tell that this is a utility class for working with instances of `EmailAddress`.

client/src/main/java/org/spine3/net/InternetDomains.java
alexander-yevsyukov: Same as above.

client/src/main/java/org/spine3/net/InternetDomains.java
alexander-yevsyukov: I'm going to have a separate `Patterns.java` file (partially copied from Android) with the constant. Some of them are quite inter-related and keeping them that way would be easier.

alexander-yevsyukov:
Reported the duplicated string issue to Codacy. Assuming we have explicit warning suppression, it looks like their bug.

client/src/main/java/org/spine3/net/EmailAddresses.java
armiol: @alexander-yevsyukov in this case I think 
```
{@link EmailAddress EmailAddresses}.
```
would look better, being rendered as HTML.

alexander-yevsyukov:
That would reference another class :)

PR #374, opened 2017-03-07
Prepare validation kit

IlliaShepilov:
Summary of changes:

- Created the `ValidatingBuilder` interface.
- Created the `AbstractValidatingBuilder` class.
- Created `ListStringifier`.
- Created the tests for the `ListStringifier` class.

The usage of the `ListStringifier`:
```java
// Stringifier creation.
final Stringifier<List<Integer>> listStringifier = Stringifiers.listStringifier();
```

```java
// The registration of the stringifier.
final Stringifier<List<Integer>> listStringifer = Stringifiers.listStringifier();
final Type type = Types.listTypeOf(Integer.class);
StringifierRegistry.getInstance().register(listStringifier, type);

// Obtain already registered stringifier.
final Stringifier<List<Integer>> listStringifier = StringifierRegistry.getInstance().getStringifier(type);

// Convert to string.
final List<Integer> listToConvert = newArrayList(1, 2, 3);

// The result is: \"1\",\"2\",\"3\".
final String convertedString = listStringifer.toString(listToConvert);

// Convert from string.
final String stringToConvert = ...
final List<Integer> convertedList = listStringifier.fromString(stringToConvert);
```


IlliaShepilov:
@armiol PTAL.

client/src/main/java/org/spine3/base/PluralKey.java
```diff
@@ -0,0 +1,70 @@
...
+package org.spine3.base;
+
+import com.google.common.base.Objects;
+
+/**
+ * A value object, serves as key of the {@code StringifierRegistry} warehouse.
+ *
+ * <p>It used when is needed to identify object by the few classes.
+ *    For example:
+ *         {@code
```
armiol: Use `<pre>` for such pre-formatted pieces of code.

client/src/main/java/org/spine3/base/RegistryKey.java
```diff
@@ -0,0 +1,31 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.base;
+
+/**
+ * A marker interface, serves as the type identifier
+ * for the keys into {@code StringRegistry} warehouse.
```
armiol: `... the keys into {@code StringRegistry} warehouse.` does not make much sense. Please rephrase.

server/src/main/java/org/spine3/server/validate/StringifierNotFoundException.java
```diff
@@ -0,0 +1,35 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.validate;
+
+/**
+ * Signals that {@code StringifierRegistry} does not contain
```
armiol: Please find another word instead of `Signals`. Or rephrase the sentence.

client/src/test/java/org/spine3/base/StringifiersShould.java
```diff
@@ -303,6 +309,49 @@ public void convert_string_to_event_id() {
     }
 
     @Test
+    public void convert_string_to_list_of_strings() {
+        final String stringToConvert = "1,2,3,4,5";
+        final List<String> convertedList =
+                new Stringifiers.ListStringifier<>(String.class).reverse()
+                                                                .convert(stringToConvert);
+        assertNotNull(convertedList);
+        assertEquals(5, convertedList.size());
```
armiol: That is not enough.

client/src/main/java/org/spine3/base/Stringifiers.java
```diff
@@ -206,4 +212,90 @@ protected CommandId doBackward(String s) {
...
+    protected static class ListStringifier<T> extends Stringifier<List<T>> {
+
+        private final Class<T> listGenericClass;
+
+        /**
+         * The delimiter for the passed elements in the String representation. Comma by default.
+         */
+        private String delimiter = ",";
+
+        public ListStringifier(Class<T> listGeneric) {
```
armiol: The parameter should be named better. Right now it's a bit of slang.

client/src/main/java/org/spine3/base/Stringifiers.java
```diff
@@ -206,4 +212,90 @@ protected CommandId doBackward(String s) {
...
+     * @param <T> the type of the elements in the list.
+     */
+    protected static class ListStringifier<T> extends Stringifier<List<T>> {
+
+        private final Class<T> listGenericClass;
+
+        /**
+         * The delimiter for the passed elements in the String representation. Comma by default.
+         */
+        private String delimiter = ",";
```
armiol: That should be `final` and set to a default value if no custom delimiter is passed.

IlliaShepilov:
@armiol PTAL.
Also I created subpackage for the related to the stringifiers functionality and I did it in the separate PR https://github.com/SpineEventEngine/core-java/pull/378

server/src/main/java/org/spine3/server/validate/AbstractValidatingBuilder.java
```diff
@@ -74,10 +87,19 @@
         throw new StringifierNotFoundException(exMessage);
     }
 
-    public <V> void validate(FieldDescriptor descriptor, V fieldValue, String valueName)
+    /**
+     * Validates the field according to the protoBuf declaration.
```
armiol: What is `protoBuf`?

server/src/main/java/org/spine3/server/validate/AbstractValidatingBuilder.java
```diff
@@ -74,10 +87,19 @@
...
+     *
+     * @param descriptor the {@code FieldDescriptor} of the field
+     * @param fieldValue the value of the field
+     * @param fieldName  the name of the field
+     * @param <V>        the type of the field value
+     * @throws ConstraintViolationThrowable if the validating value contains constraint violations
```
armiol: `if there are some constraint violations`.

client/src/main/java/org/spine3/base/RegistryKey.java
```diff
@@ -21,8 +21,8 @@
 package org.spine3.base;
 
 /**
- * A marker interface, serves as the type identifier
- * for the keys into {@code StringRegistry} warehouse.
+ * A marker interface, serves as the type identifier for the keys
+ * of the {@code StringRegistry} warehouse to save and retrieve the {@code Stringifier} instances.
```
armiol: In IT `warehouse` usually means something much bigger than `StringRegistry`. Please find another word.

client/src/main/java/org/spine3/base/Stringifiers.java
```diff
@@ -242,16 +242,20 @@ protected Integer doBackward(String s) {
      */
     protected static class ListStringifier<T> extends Stringifier<List<T>> {
 
+        private static final String DEFAULT_DELIMITER = ",";
+
         private final Class<T> listGenericClass;
 
         /**
-         * The delimiter for the passed elements in the String representation. Comma by default.
+         * The delimiter for the passed elements in the String representation,
```
armiol: `{@code String}`.

client/src/main/java/org/spine3/base/Stringifiers.java
```diff
@@ -242,16 +242,20 @@ protected Integer doBackward(String s) {
...
 
-        public ListStringifier(Class<T> listGeneric) {
+        public ListStringifier(Class<T> listGenericClass) {
             super();
-            this.listGenericClass = listGeneric;
+            this.delimiter = DEFAULT_DELIMITER;
+            this.listGenericClass = listGenericClass;
         }
 
         /**
```
armiol: Please create a valid documentation for the 
`public ListStringifier(Class<T> listGenericClass, String delimiter)` constructor.

client/src/test/java/org/spine3/base/StringifiersShould.java
```diff
@@ -337,11 +343,16 @@ public void convert_list_of_integers_to_string() {
...
-                                                                        .convert(stringToConvert);
-        assertNotNull(convertedList);
-        assertEquals(5, convertedList.size());
+        final String delimiter = "\\|";
+        final List<Integer> actualList =
+                new Stringifiers.ListStringifier<>(Integer.class,
+                                                   delimiter).reverse()
+                                                             .convert(stringToConvert);
+        assertNotNull(actualList);
+        assertEquals(5, actualList.size());
```
armiol: Let's remove this line. It does nothing (since there are more checks), but adds another magic constant.

client/src/test/java/org/spine3/base/StringifiersShould.java
```diff
@@ -352,6 +363,14 @@ public void emit_exception_when_list_type_does_not_have_appropriate_stringifier(
     }
 
     @Test
+    public void convert_integer_to_string() {
+        final Integer integerToConvert = 1;
+        final String convertedValue =
+                new Stringifiers.IntegerStringifier().convert(integerToConvert);
+        assertEquals("1", convertedValue);
```
armiol: Can you actually use the `integerToConvert` to create an expected value?

client/src/test/java/org/spine3/base/StringifiersShould.java
```diff
@@ -311,11 +314,14 @@ public void convert_string_to_event_id() {
...
     public void convert_string_to_list_of_strings() {
         final String stringToConvert = "1,2,3,4,5";
-        final List<String> convertedList =
+        final List<String> actualList =
                 new Stringifiers.ListStringifier<>(String.class).reverse()
                                                                 .convert(stringToConvert);
-        assertNotNull(convertedList);
-        assertEquals(5, convertedList.size());
+        assertNotNull(actualList);
+        assertEquals(5, actualList.size());
```
armiol: Same as below.

IlliaShepilov:
@armiol PTAL.

armiol:
@IlliaShepilov one thing left is an exposure of `Identifier` to `public`. Please discuss this change with @alexander-yevsyukov.

Except for that, LGTM.

IlliaShepilov:
@armiol PTAL.

IlliaShepilov:
@armiol PTAL.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,154 @@
...
+            final T convertedValue = Stringifiers.convert(unquote(element), listGenericClass);
+            result.add(convertedValue);
+        }
+        return result;
+    }
+
+    private static void checkElement(CharSequence element) {
+        final boolean isQuoted = isQuotedString(element);
+        if (!isQuoted) {
+            throw newIllegalArgumentException("Illegal format of the element");
```
armiol: Please include the original argument into the exception text.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
```
armiol: This line should be removed.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,154 @@
...
+    private static String createElementDelimiterPattern(char delimiter) {
+        return Pattern.compile("(?<!\\\\)\\\\\\" + delimiter)
+                      .pattern();
+    }
+
+    @Override
+    protected String toString(List<T> list) {
+        final StringBuilder stringBuilder = new StringBuilder(0);
+        for (T element : list) {
+            final char quote = '"';
```
armiol: That is fragile. 

The knowledge about quoting the each element is both here and in `org.spine3.base.Stringifiers.isQuotedString`. Making any change to either of these places will lead to a hardly-traceable fail.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,154 @@
...
+            stringBuilder.append(quote)
```
armiol: I would make something like
```
// Encloses each element of the list into quotes.
protected static class QuotedListItem<T> {
    private final T object;
    private final ... genericlClass;
    // private final char quoteSymbol = "\"".   --> if any customization of the "quote" symbol is required.

    // private ctor and static factory method named `of`.
    
    public String toString() {
        return quote(Stringifiers.convert(object, genericClass))
    }
}

// ... and then in this method:
 
final List<QuotedListItem<T>> quotedItems = Lists.transform(list, new Function<T, QuotedListItem<T>> {
    public QuotedListItem<T> apply(... T input) {
          return QuotedListItem.of(input);
    }
})
return Joiner.on(delimiter).join(quotedItems);

File: client/src/main/java/org/spine3/base/ListStringifier.java [hunk 19]
```diff
@@ -0,0 +1,154 @@
...
+                         .append(delimiter);
+        }
+        final int length = stringBuilder.length();
+        final String result = stringBuilder.substring(0, length - 1);
+        return result;
+    }
+
+    @Override
+    protected List<T> fromString(String s) {
+        final String escapedString = escaper.escape(s);
```
armiol: Same as above, but using Guava's `Splitter` instead of `Joiner` and extending the same `QuotedListItem` to produce an object of `T` via a method called like `T parse(String)`.

In this way you'll have the knowledge of quoting inside the `QuotedListIem` only.

client/src/test/java/org/spine3/base/ListStringifierShould.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Test(expected = MissingStringifierException.class)
+    public void emit_exception_when_list_type_does_not_have_appropriate_stringifier() {
+        final String stringToConvert = "\"{value:123456}\"";
+        new ListStringifier<>(Task.class).reverse()
+                                         .convert(stringToConvert);
+    }
+
+    private static String unquote(CharSequence stringToUnescape) {
+        return Pattern.compile("\"")
```
armiol: Again, somehow here you know exactly which symbol is used for quotes. That's the third place.

server/src/test/java/org/spine3/server/validate/AbstractValidatingBuilderShould.java
```diff
@@ -0,0 +1,110 @@
...
+    @Test(expected = ConstraintViolationThrowable.class)
+    public void throw_exception_when_field_contains_constraint_violations()
+            throws ConstraintViolationThrowable {
+        final FieldDescriptor descriptor = PatternStringFieldValue.getDescriptor()
+                                                                  .getFields()
+                                                                  .get(0);
+        validatingBuilder.validate(descriptor, "incorrectEmail", "email");
+    }
+
+    private static class ConcreteValidatingBuilder extends AbstractValidatingBuilder<Task> {
```
armiol: Please rename it so that you don't reference any [construction material stuff](https://en.wikipedia.org/wiki/Concrete).

IlliaShepilov:
@armiol PTAL.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,201 @@
...
+        final List<QuotedListItem<T>> quotedItems = Lists.transform(list, function);
+        final String result = Joiner.on(delimiter)
+                                    .join(quotedItems);
+        return result;
+    }
+
+    @Override
+    protected List<T> fromString(String s) {
+        final String escapedString = escaper.escape(s);
+        final Splitter splitter = Splitter.onPattern(elementDelimiterPattern);
```
armiol: Looks like this instance of `splitter` can also be created once and forever for this class — as long as `elementDelimiterPattern` is a `final` field.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,201 @@
...
+        final List<String> elements = newArrayList(splitter.split(escapedString));
+        final Function<String, T> function = new Function<String, T>() {
+            @SuppressWarnings("ConstantConditions")
+            // It should be processed by the stringifier.
+            @Nullable
+            @Override
+            public T apply(@Nullable String input) {
+                return parse(input, listGenericClass);
```
armiol: I would add `checkNotNull(input)` instead of suppressing a warning.

client/src/main/java/org/spine3/base/MapStringifier.java
```diff
@@ -133,111 +137,136 @@ private static String createKeyValuePattern() {
...
-    private static Escaper createEscaper(char charToEscape) {
-        final String escapedChar = "\\" + charToEscape;
-        final Escaper result = Escapers.builder()
-                                       .addEscape('\"', "\\\"")
-                                       .addEscape(charToEscape, escapedChar)
-                                       .build();
-        return result;
+        @Override
+        public String toString() {
+            final String keuToQuote = isString(keyClass) ?
```
armiol: Please fix a typo.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,201 @@
...
+        private static void checkElement(CharSequence element) {
+            final boolean isQuoted = isQuotedString(element);
+            if (!isQuoted) {
+                throw newIllegalArgumentException("Illegal format of the element: " + element);
+            }
+        }
+
+        @Override
+        public String toString() {
+            if (isString(genericClass)) {
```
armiol: It seems like this is a responsibility of `Stringifiers.toString(object, genericClass)`.

You could save yourselves some `if`s here and in the `MapStringifier` class as well.

client/src/main/java/org/spine3/base/Stringifiers.java
```diff
@@ -188,4 +222,92 @@ protected Integer fromString(String s) {
                        .convert(s);
         }
     }
+
+    abstract static class QuotedItem{
```
armiol: Please add Javadocs.

client/src/main/java/org/spine3/validate/ConstraintViolationThrowable.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.spine3.validate;
+
+import java.util.List;
+
+import static java.util.Collections.unmodifiableList;
+
+/**
+ * Signals that the validated value contains the constraint violations.
```
armiol: `...violates the constraints set for it.`

IlliaShepilov:
@armiol PTAL.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,172 @@
...
+            public T apply(@Nullable String input) {
+                checkNotNull(input);
+                return Stringifiers.convert(input, listGenericClass);
+            }
+        };
+        final List<String> unquotedList = newArrayList(Lists.transform(elements, unquoteFunction));
+        final List<T> result = newArrayList(Lists.transform(unquotedList, convertFunction));
+        return result;
+    }
+
```
armiol: Please remove an empty line.

client/src/main/java/org/spine3/base/ListStringifier.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    protected List<T> fromString(String s) {
+        final String escapedString = escaper.escape(s);
+        final List<String> elements = newArrayList(splitter.split(escapedString));
+        final Function<String, String> unquoteFunction = new Function<String, String>() {
+            @Nullable
+            @Override
+            public String apply(@Nullable String input) {
+                checkNotNull(input);
+                return QuotedItem.unquote(input);
```
armiol: What if we had `QuotedItem.converter(String)` and `Stringifiers.converter(input, listGenericClass)` (both package-access)? The first would return a converter from quoted string to unquoted (forward) and from unquoted to quoted (backward). The result of the second would be a converter from `String` to `T` and backwards. 

Then you could chain the converters and avoid functions at all.

armiol: @IlliaShepilov Please see [that](https://google.github.io/guava/releases/17.0/api/docs/com/google/common/base/Converter.html#andThen(com.google.common.base.Converter)) on chaining.

IlliaShepilov:
@armiol PTAL.

client/src/main/java/org/spine3/base/ItemQuoter.java
```diff
@@ -0,0 +1,101 @@
...
+
+import com.google.common.base.Converter;
+
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * Serves as enclosure and disclosure for the elements into and from quotes.
```
armiol: `Encloses and discloses the {@code String} objects with quotes.`

client/src/main/java/org/spine3/base/ItemQuoter.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Checks that the {@code CharSequence} contains escaped quotes.
+     *
+     * @param stringToCheck the sequence of chars to check
+     * @return {@code true} if sequence contains further
+     * and prior escaped quotes, {@code false} otherwise
+     */
+    static boolean isQuotedString(CharSequence stringToCheck) {
```
armiol: It's weird that this method isn't `private`.

client/src/main/java/org/spine3/base/Stringifiers.java
```diff
@@ -154,9 +158,79 @@ private Stringifiers() {
...
+                                       .addEscape('\"', "\\\"")
+                                       .addEscape(charToEscape, escapedChar)
+                                       .build();
+        return result;
+    }
+
+    /**
+     * The {@code Stringifier} for the {@code String} values.
+     */
+    private static class StringStringifier extends Stringifier<String> {
```
armiol: That probably should be `NoopStringifier` which description says it does nothing.

Same with the static `Stringifiers.stringStringifier()` method.

IlliaShepilov:
@armiol PTAL.

PR #376, opened 2017-03-08
Get rid of `Tuple`

alexander-yevsyukov:
This PR utilizes recently introduced `EntityRecord.entity_id` field for symmetry of conversion between an `Entity` and `EntityRecord`.

As the result we no longer need the `Tuple` class that hold an ID together with an `EntityRecord`.

This PR is based on #372.


alexander-yevsyukov:
@armiol, PTAL.

PR #377, opened 2017-03-08
Protect aggregate state

alexander-yevsyukov:
This PR makes the `Aggregate` class reject calls to `setState()` from the derived aggregate classes.

### DIFF
--- a/server/src/main/java/org/spine3/server/aggregate/Aggregate.java
+++ b/server/src/main/java/org/spine3/server/aggregate/Aggregate.java
@@ -25,6 +25,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.protobuf.Any;
 import com.google.protobuf.Message;
+import org.spine3.Internal;
 import org.spine3.base.CommandClass;
 import org.spine3.base.CommandContext;
 import org.spine3.base.Event;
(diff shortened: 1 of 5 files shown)


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -390,7 +413,30 @@ void restore(Snapshot snapshot) {
...
+     * {@code AggregateBuilder} test utility class from the {@code testutil} module.
+     */
+    @VisibleForTesting
+    void injectState(S stateToRestore, Version versionFromSnapshot) {
+        try {
+            @SuppressWarnings("unchecked")
+                // The cast is safe as we checked the type on the construction.
+            final B fakeBuilder = (B) getState().newBuilderForType();
+            this.builder = fakeBuilder;
             setState(stateToRestore, versionFromSnapshot);
```
armiol: If we are making the builder mandatory, then we could use it as well as an additional protection level.
E.g. what if we either 
* accept the `Builder` instances instead of `S`;
* or create the `Builder` instances manually from the `stateToRestore` passed with `injectState` method;

and then call something like `validate` on this builder before setting the state. Thus we'll protect from passing an invalid state, such as ignored Spine-custom field bounds (`required`, `min`/`max` etc).

alexander-yevsyukov: `injectState()` is called only once from the `restore(Snapshot)` method. We should have a valid `Snapshot` at this stage. That's why we “inject”.

But I agree that we need yet more protection of an `Aggregate` as we discussed. So, I wanted to discuss having default validation in the `AbstractVersionableEntity.validate()`, which currently does nothing. That would solve the problem of passing invalid state to entities or building an invalid state in an aggregate applier.

PR #378, opened 2017-03-09
Stringifiers subpackage

IlliaShepilov:
Summary of changes:

- Create subpackage for the classes related to the `stringifier` classes.


client/src/test/java/org/spine3/convert/StringifiersShould.java
```diff
@@ -367,7 +369,7 @@ public void convert_integer_to_string() {
         final Integer integerToConvert = 1;
         final String convertedValue =
                 new Stringifiers.IntegerStringifier().convert(integerToConvert);
```
alexander-yevsyukov: @IlliaShepilov,  can we use `Ints.stringConverter()` instead? There's also `Longs.stringConverter()`.

PR #384, opened 2017-03-13
Sort out packages for type-related classes

alexander-yevsyukov:
This PR rearranges classes that are related to working with type information. Namely:

* `envelope` package is introduced
* Classes related to `KnownTypes` are moved to the `type` package.
* `TypeName` is restored as a value type.
* The `Queries` utility class is moved to the `client` package into which query-related classes are generated.
* Moves annotation types under the `annotations` package.


alexander-yevsyukov:
@dmdashenkov, PTAL.

client/src/main/java/org/spine3/type/TypeName.java
```diff
@@ -0,0 +1,125 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * Utility class for obtaining a type name from a {@code Message}.
+ *
+ * @author Alexander Yevsyukov
+ */
+public class TypeName extends StringTypeValue {
```
alexander-yevsyukov: Because we now already need a first class value type (e.g. we have a need for a simple type name). Later we'd need to have type names for other things like displaying all known types, etc.

alexander-yevsyukov:
@dmdashenkov, PTAL.

testutil/src/main/java/org/spine3/test/TestEventFactory.java
```diff
@@ -0,0 +1,62 @@
...
+import org.spine3.base.Event;
+import org.spine3.base.EventContext;
+import org.spine3.base.Events;
+import org.spine3.protobuf.AnyPacker;
+import org.spine3.server.reflect.CommandHandlerMethod;
+
+import static org.spine3.protobuf.Values.newStringValue;
+
+/**
+ * A factory for creating tests instances of {@link org.spine3.base.Event Event}s.
```
dmdashenkov: If we already have an alternative text, why don't we say `{@link org.spine3.base.Event Events}`?

testutil/src/main/java/org/spine3/test/TestEventFactory.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @author Alexander Yevsyukov
+ */
+@Internal
+@VisibleForTesting
+public class TestEventFactory {
```
dmdashenkov: Why is this in `main` module, not in `test`?

alexander-yevsyukov: See the path above. It's under `testutil/`

client/src/main/java/org/spine3/type/TypeUrl.java
```diff
@@ -231,15 +231,8 @@ private static String getTypeUrlPrefix(GenericDescriptor descriptor) {
...
-        checkNotNull(this);
-        final ClassName className = KnownTypes.getClassName(this);
-        try {
-            @SuppressWarnings("unchecked") // the client considers this message is of this class
-            final Class<T> result = (Class<T>) Class.forName(className.value());
-            return result;
-        } catch (ClassNotFoundException e) {
-            throw new UnknownTypeException(getTypeName(), e);
-        }
+        final TypeUrl typeUrl = this;
```
dmdashenkov: Strange expression. Why do we need an alias reference for `this`?

client/src/test/java/org/spine3/type/TypeUrlShould.java
```diff
@@ -191,4 +195,23 @@ public void do_not_accept_empty_name() {
     public void do_not_accept_malformed_type_URL() {
         TypeUrl.of("prefix/prefix/type.Name");
     }
+
+    @Test(expected = UnknownTypeException.class)
+    public void throw_exception_for_unknown_Java_class() {
+        TypeUrl.of("unknown/JavaClass")
+               .toMessageClass();
```
dmdashenkov: Which of the calls exactly throws the `Exception`? It might be `TypeUrl.of`.

server/src/main/java/org/spine3/server/entity/FieldMasks.java
```diff
@@ -186,7 +186,8 @@ public static FieldMask maskOf(Descriptors.Descriptor typeDescriptor, int... fie
         return result;
     }
 
-    @SuppressWarnings("unchecked")      // We assume that {@code KnownTypes#getClassName(TypeUrl) works properly.
+    @SuppressWarnings("unchecked")
+        // We assume that {@code KnownTypes#getClassName(TypeUrl) works properly.
```
dmdashenkov: The alignment is strange. Can we move the comment to the start of the line?

alexander-yevsyukov: This comment is nested to the warning suppression.

PR #385, opened 2017-03-14
More `type` improvements

alexander-yevsyukov:
This PR adds naming and responsibility improvements for classes in the `type` package.

It also:
* Introduces the `Field` class in the `server.reflect` package.
* Exposes unpacking function from `AnyPacker`.
* Migrates from `NullToleranceTest` to `NullPointerTester`


alexander-yevsyukov:
@dmdashenkov, PTAL.

client/src/main/java/org/spine3/type/KnownTypes.java
```diff
@@ -216,7 +216,7 @@ public boolean apply(@Nullable TypeUrl input) {
         final TypeUrl typeUrl = getTypeUrl(typeName);
         checkArgument(typeUrl != null, "Given type name is invalid");
 
-        final Class<?> cls = getClass(typeUrl);
+        final Class<?> cls = getJavaClass(typeUrl);
```
dmdashenkov: Isn't it `typeUrl.getJavaClass()`?

client/src/main/java/org/spine3/type/KnownTypes.java
```diff
@@ -245,8 +245,7 @@ public boolean apply(@Nullable TypeUrl input) {
      *
      * @throws UnknownTypeException if there is no Java class for the passed type URL
      */
-    @VisibleForTesting
-    public static <T extends Message> Class<T> getClass(TypeUrl typeUrl) {
+    static <T extends Message> Class<T> getJavaClass(TypeUrl typeUrl) throws UnknownTypeException {
```
dmdashenkov: Why don't we move all this logic to `TypeUrl` letting it manage its form by itself?

client/src/main/java/org/spine3/type/TypeUrl.java
```diff
@@ -116,35 +125,45 @@ public static TypeUrl from(EnumDescriptor descriptor) {
...
         }
 
-        final String prefix = parts[0].trim();
+        final String prefix = parts[0];
         if (prefix.isEmpty()) {
             throw malformedTypeUrl(typeUrl);
         }
 
-        final String typeName = parts[1].trim();
+        final String typeName = parts[1];
```
dmdashenkov: Why don't we `trim` the parts anymore? "` `" part also indicates a malformed URL.

dmdashenkov: Maybe we could check for space characters before parsing.

client/src/main/java/org/spine3/type/TypeUrl.java
```diff
@@ -230,8 +257,25 @@ private static String getTypeUrlPrefix(GenericDescriptor descriptor) {
...
+     * Obtains the type descriptor for the type of this URL.
+     *
+     * @return {@code Descriptor} or {@code Optional.absent()} if this type URL
+     *         represents an outer class
+     */
+    public Optional<Descriptor> getTypeDescriptor() {
+        final Class<? extends Message> clazz = getJavaClass();
+        final GenericDescriptor descriptor = getClassDescriptor(clazz);
+        // Skip outer class descriptors.
+        if (descriptor instanceof Descriptor) {
```
dmdashenkov: - Outer class descriptors are represented with `FileDescriptor` type.
 - There are concrete type descriptors that are not represented with `Descriptor` type (e.g. `enum`s' descriptors are `EnumDescriptor extends GenericDescriptor`).

If we want to avoid __outer class__ descriptors, we should check `!(descriptor instanceof FileDescriptor)` instead.

server/src/main/java/org/spine3/server/reflect/Field.java
```diff
@@ -0,0 +1,186 @@
...
+     *                               The root cause will be available from the thrown instance.
+     */
+    public Message getValue(Message object) {
+        final Message fieldValue;
+        final Message result;
+        try {
+            fieldValue = (Message) getter.invoke(object);
+
+            result = fieldValue instanceof Any
+                     ? AnyPacker.unpack((Any) fieldValue)
```
dmdashenkov: Please check what happens if the value is `Any.getDefaultInstance()`. If I remember it right, the `AnyPacker` will throw an exception in that case.

server/src/main/java/org/spine3/server/stand/SubscriptionRegistry.java
```diff
@@ -68,7 +69,8 @@ synchronized void activate(Subscription subscription, Stand.EntityUpdateCallback
     synchronized Subscription addSubscription(Target target) {
         final String subscriptionId = Identifiers.newUuid();
         final String typeAsString = target.getType();
-        final TypeUrl type = TypeUrl.of(typeAsString);
+        final TypeUrl type = TypeName.of(typeAsString)
+                                     .toUrl();
```
dmdashenkov: Is there any difference between `TypeName.of(smth).toUrl()` and `TypeUrl,parse(smth)`? If not, I suppose we should use the second option here.

server/src/main/java/org/spine3/server/storage/memory/InMemoryStandStorage.java
```diff
@@ -130,13 +130,13 @@ public boolean delete(AggregateStateId id) {
...
-        final TypeUrl recordType = TypeUrl.of(record.getState()
-                                                    .getTypeUrl());
+        final TypeUrl recordType = TypeUrl.parse(record.getState()
+                                                       .getTypeUrl());
         final TypeUrl recordTypeFromId = id.getStateType();
-        checkState(
-                recordTypeFromId.equals(recordType),
-                String.format(TYPE_URL_MISMATCH_MESSAGE_PATTERN, recordType, recordTypeFromId));
-
+        if (!recordTypeFromId.equals(recordType)) {
```
dmdashenkov: What is wrong with `checkState`?

client/src/main/java/org/spine3/type/TypeUrl.java
```diff
@@ -42,12 +43,13 @@
...
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.List;
 import java.util.Objects;
-import java.util.regex.Pattern;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Throwables.getRootCause;
+import static com.google.protobuf.Internal.getDefaultInstance;
```
dmdashenkov: Is it OK to use something named `not.our.package.Internal`?

server/src/main/java/org/spine3/server/stand/SubscriptionRegistry.java
alexander-yevsyukov: We had this ambiguity before. A type name is not a type URL. Here we convert to a URL via internal knowledge of all known types.

server/src/main/java/org/spine3/server/storage/memory/InMemoryStandStorage.java
alexander-yevsyukov: It could be via `checkState()` but we still have the question on why we do it.

client/src/main/java/org/spine3/type/KnownTypes.java
alexander-yevsyukov: Because it's `KnownTypes` that keep the map to a class name.

alexander-yevsyukov:
@dmdashenkov, PTAL.

client/src/main/java/org/spine3/type/TypeUrl.java
```diff
@@ -38,10 +39,13 @@
 import org.spine3.envelope.MessageEnvelope;
 import org.spine3.type.error.UnknownTypeException;
 
+import java.util.List;
 import java.util.Objects;
-import java.util.regex.Pattern;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.protobuf.Internal.getDefaultInstance;
```
dmdashenkov: Can we avoid using classes matching `not.out.package.Internal`?

alexander-yevsyukov: I'm glad you spotted. I specifically put it into the `import static` section to highlight the fact. I am not aware of any means that can give us the features we need from that `Internal` class.

PR #388, opened 2017-03-16
Right margin wrapping

alexander-aleksandrov:
This PR contains the changes in wrapping for 100 column right margin.


alexander-aleksandrov:
@dmytro-grankin PTAL

client/src/main/java/org/spine3/base/Commands.java
```diff
@@ -269,7 +270,8 @@ public int compare(Command o1, Command o2) {
      * Checks if the file is for commands.
      *
      * @param file a descriptor of a {@code .proto} file to check
-     * @return {@code true} if the file name ends with the {@link #FILE_NAME_SUFFIX}, {@code false} otherwise
+     * @return {@code true} if the file name ends with the {@link #FILE_NAME_SUFFIX},
+     *                      {@code false} otherwise
```
dmytro-grankin: Possible, it's nice formatting for a Javadoc, but `idea`'s `reformat code` will reformat a Javadoc with such formatting. We should avoid this here and in similar cases.

client/src/main/java/org/spine3/util/Math.java
```diff
@@ -20,6 +20,8 @@
 
 package org.spine3.util;
 
+import static java.lang.String.*;
```
dmytro-grankin: Do not `static` use imports like this, specify a particular import. For example, use `java.lang.String.format`. Do it in all similar cases.

client/src/test/java/org/spine3/testdata/TestCommandContextFactory.java
```diff
@@ -60,7 +60,8 @@ public static CommandContext createCommandContext(UserId userId, CommandId comma
                                                              .setActor(userId)
                                                              .setTimestamp(when)
                                                              .setZoneOffset(UTC)
-                                                             .setTenantId(TenantId.newBuilder().setValue(newUuid()));
+                                                             .setTenantId(TenantId.newBuilder()
+                                                                                  .setValue(newUuid()));
```
dmytro-grankin: This row is larger than 100 symbols.

server/src/main/java/org/spine3/server/storage/BulkStorageOperationsMixin.java
```diff
@@ -40,12 +41,13 @@
      *
      * <p>The size of {@link Iterable} returned is always the same as the size of given IDs.
      *
-     * <p>In case there is no record for a particular ID, {@code null} will be present in the result instead.
-     * In this way {@code readMultiple()} callees are able to track the absence of a certain element by comparing
-     * the input IDs and resulting {@code Iterable}.
+     * <p>In case there is no record for a particular ID, {@code null} will be present
+     * in the result instead. In this way {@code readMultiple()} calls are able to track
```
dmytro-grankin: Is really `callees` should be replaced by `calls`?

alexander-aleksandrov: No, my mistake fixed

client/src/main/java/org/spine3/base/Commands.java
alexander-aleksandrov: Done

client/src/test/java/org/spine3/testdata/TestCommandContextFactory.java
alexander-aleksandrov: Done

alexander-aleksandrov:
@dmytro-grankin PTAL I've added few fixes. Please take a notice that we should not change links that are longer than 100 columns.

client/src/test/java/org/spine3/protobuf/MessageFieldShould.java
```diff
@@ -28,6 +28,7 @@
 import static com.google.protobuf.Descriptors.FieldDescriptor.JavaType;
 import static org.junit.Assert.assertEquals;
 import static org.spine3.base.Identifiers.newUuid;
+import static org.spine3.protobuf.MessageField.*;
```
dmytro-grankin: Optimize imports for this file, then `...MessageField.*` import will disappear.

PR #389, opened 2017-03-16
Improve stringifiers

alexander-yevsyukov:
This PR improves working with stringifiers. Namely:
* It's possible now to convert an non-generic object without passing its type information.
* For generic objects `toString(Object, Type)` should be  used.
* `Type` is used for passing type information instead of `TypeToken`.
* Stringifiers for well-known types are moved to corresponding utility classes.

Also:
* Work with Json is moved into the `json` package.


alexander-yevsyukov:
@dmdashenkov, PTAL.

alexander-yevsyukov:
@IlliaShepilov, FYI. Have a look too when you're back.

client/src/main/java/org/spine3/base/StringifierRegistry.java
```diff
@@ -56,47 +57,60 @@ private StringifierRegistry() {
...
 
-    public <I extends Message> void register(TypeToken<I> valueClass, Stringifier<I> converter) {
-        checkNotNull(valueClass);
+    static <T> Stringifier<T> getStringifier(Type typeOfT) {
+        final Optional<Stringifier<T>> stringifierOptional = getInstance().get(typeOfT);
+
+        if (!stringifierOptional.isPresent()) {
+            final String errMsg =
+                    format("No stringifier registered for the type: %s", typeOfT);
+            throw new MissingStringifierException(errMsg);
```
dmdashenkov: This branch seems to be not covered.

client/src/main/java/org/spine3/protobuf/Timestamps2.java
```diff
@@ -222,4 +253,79 @@ public static Date convertToDate(TimestampOrBuilder timestamp) {
...
+        private static String toWebSafe(String str) {
+            final String result = PATTERN_COLON.matcher(str)
+                                               .replaceAll("-");
+            return result;
+        }
+
+        /**
+         * Converts the passed web-safe timestamp representation to the RFC 3339 date string format.
+         */
+        @SuppressWarnings("MagicNumber") // are fixed positions in the timestamp string
```
dmdashenkov: Please extract those as constants with appropriate names.

PR #390, opened 2017-03-17
Bump version for fqn checker fix

alexander-aleksandrov:
We needed to fix hidden files checking in FQN checker tool.


alexander-aleksandrov:
@dmdashenkov PTAL

PR #391, opened 2017-03-20
Storage Fields basic support

dmdashenkov:
Introducing __Spine Storage Fields__ — a convenient way to store custom _indexable_ fields of an `Entity` in a Spine Storage.

In this PR:
 - Basic implementation of the Storage Fields retrieval mechanism
 - `v1` API of the Storage Fields:
   - All the values are retrieved from the `Entity` getters.
   - A getter may have `@javax.annotation.Nullable` annotation to show that the field may accept `null`s. By default all the Storage Fields are __non-null__.
   - Supported field types are: 
     - `Message`
     - `Integer`
     - `Long`
     - `String`
     - `Boolean`
     - `Float`
     - `Double`
   - To iterate over the `StorageFields` object regarding the fields' types use `StorageFieldTraverser`.

##### These methods represent storage fields:
```
--UserGroupAggregate.java--

// note primitive autoboxing
public int getUsersCount() {
    return getState().getUserCount();
} 

--PaymentProcessManager.java--

@Nullable
public String getPaymentSystemName() {
    if (getState().hasThirdpartySystem()) {
        return getState().getThirdPartySystem().getName();
    } else {
        return null;
    }
}
```

##### And these methods are __not__ valid storage fields:
```
--UserAggregate.java--

// non-public methods may not be storage fields
private double getHourRate() { ... }

// methods of unsupported types may not represent storage fields
public Object isEmployed() { ... }

// only methods starting with "get" or "is" may be considered as storage fields
public boolean hasChildren() { ... }

// getter methods must not accept arguments
public User getStateOf(UserAggregate other) { ... } 

// only instance methods are considered as Storage Fields
public static Integer isNew(UserAggregate aggregate) { ... }
```
The examples above will not be included when generating the `StorageFields` object.

##### Faulty storage fields:
The example below will throw an `Exception` when trying to generate the storage fields.
```
--EmploeeProjection.java--

// method should be annotated as @Nullable to return a null value
public Message getAddress() {
    return null;
}
```

Also note, that the `Entity` class which the storage fields are generated for should be `public`. If it's not, the framework user will see a warning message in the log and the `StorageFields` object will never be generated.


dmdashenkov:
@armiol, PTAL.

client/src/test/java/org/spine3/base/IdentifiersShould.java
```diff
@@ -59,7 +59,7 @@ public void have_private_constructor() {
     @Test(expected = IllegalArgumentException.class)
     public void throw_exception_when_object_of_unsupported_class_passed() {
         //noinspection UnnecessaryBoxing
```
armiol: Let's use an annotation-based suppression instead. Even though it's in the test code.

client/src/test/java/org/spine3/base/IdentifierShould.java
```diff
@@ -26,11 +26,11 @@
 import com.google.protobuf.UInt32Value;
 import com.google.protobuf.UInt64Value;
 import org.junit.Test;
-import org.spine3.base.Identifier.Type;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.spine3.base.Identifier.*;
```
armiol: I suspect we don't have wildcard imports in our codebase, do we?
Codacy should've caught that, but it does not check the test codebase.

Please address.

client/src/test/java/org/spine3/base/IdentifierShould.java
```diff
@@ -40,18 +40,18 @@
 
     @Test
     public void getDefaultValue_by_class_id() {
-        assertEquals(0L, Identifier.getDefaultValue(Long.class).longValue());
-        assertEquals(0, Identifier.getDefaultValue(Integer.class).intValue());
-        assertEquals("", Identifier.getDefaultValue(String.class));
-        assertEquals(Timestamp.getDefaultInstance(), Identifier.getDefaultValue(Timestamp.class));
+        assertEquals(0L, getDefaultValue(Long.class).longValue());
```
armiol: I'd have separate tests for each type. Right now, if something is broken, it's more difficult to trace the reason.

Same applies to the test method below.

server/src/main/java/org/spine3/server/entity/AbstractVersionableEntity.java
```diff
@@ -274,7 +274,8 @@ public LifecycleFlags getLifecycleFlags() {
      *
      * @return {@code true} if the entity is archived, {@code false} otherwise
      */
-    protected boolean isArchived() {
+    @Override
+    public boolean isArchived() {
```
armiol: Did we consider declaring these methods (which became `public` as discussed) as `final`?

Let's discuss. My idea is to prevent the user from overriding the implementation of the "internal" fields, that we expose as a part of framework.

server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
```diff
@@ -20,9 +20,11 @@
 
 package org.spine3.server.entity;
 
+import com.google.common.base.Optional;
 import com.google.protobuf.Any;
 import com.google.protobuf.FieldMask;
 import com.google.protobuf.Message;
+import org.spine3.server.entity.storagefield.StorageFieldsExtractor;
```
armiol: I wouldn't call the package `storagefield`, but `storage` and then `.field` if it is really needed.

server/src/main/java/org/spine3/server/entity/storagefield/StorageFieldTraverser.java
```diff
@@ -0,0 +1,131 @@
...
+ * of the entry.
+ *
+ * <p>This class has a set of methods {@code hitSmth}, where {@code Smth} is the type of the entry.
+ * To handle the types, implement those methods. They take a {@code String} param containing
+ * the name of the storage field and a {@linkplain Nullable nullable} value of the field.
+ *
+ * @author Dmytro Dashenkov.
+ */
+@SPI
+public abstract class StorageFieldTraverser {
```
armiol: The purpose of this class is not clear, especially taking into account that it's marked `@SPI`. 

Let's discuss when the framework users ever need to define some custom strategy.

server/src/main/java/org/spine3/server/entity/storagefield/StorageFieldTraverser.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Iterates over the given {@link StorageFields}.
+     *
+     * <p>The order of iteration is following:
```
armiol: `is as follows:`

server/src/main/java/org/spine3/server/entity/storagefield/StorageFieldTraverser.java
```diff
@@ -0,0 +1,131 @@
...
+
+    protected abstract void hitString(String fieldName, @Nullable String value);
+
+    protected abstract void hitBoolean(String fieldName, @Nullable Boolean value);
+
+    protected abstract void hitFloat(String fieldName, @Nullable Float value);
+
+    protected abstract void hitDouble(String fieldName, @Nullable Double value);
+
+    private void traverseMessages(StorageFields fields) {
```
armiol: This class looks like it should be a number of `Visitor`s instead. 

I don't think the current implementation concept is any good. Let's discuss.

server/src/main/java/org/spine3/server/entity/storagefield/StorageFieldsDecomposer.java
```diff
@@ -0,0 +1,161 @@
...
+            case UNRECOGNIZED:
+            default:
+                throw unsupportedType(name, fieldClass);
+
+        }
+    }
+
+    private static IllegalArgumentException unsupportedType(String fieldName, Class cls) {
+        throw new IllegalArgumentException(format(
+                "Cannot add field %s : %s to StorageFields.",
```
armiol: English articles, please.

server/src/main/java/org/spine3/server/reflect/Getter.java
```diff
@@ -0,0 +1,116 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.String.format;
+
+/**
+ * A representation of a Java-style getter method.
```
armiol: I think, it should be `JavaBean-style`.

server/src/main/java/org/spine3/server/reflect/Getter.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * <p>Aggregates a {@link Method} and some meta information on it.
```
armiol: `and its metadata`.

server/src/main/proto/spine/server/entity/entity.proto
```diff
@@ -72,3 +75,47 @@ message LifecycleFlags {
...
+//
+// These fields should be searchable in an effiefient way.
+//
+message StorageFields {
+    option (SPI_type) = true;
+
+    // The ID of the entity.
+    google.protobuf.Any entity_id = 1;
+
+    // A map of Message filed values.
```
armiol: `filed` should be fixed everywhere.

server/src/test/java/org/spine3/server/entity/storagefield/StorageFieldTraverserShould.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Test
+    public void traverse_over_storage_fields() {
+        final String intKey = "Int_value";
+        final String longKey = "Long_value";
+        final String stringKey = "String_value";
+        final String boolKey = "Bool_value";
+        final String floatKey = "Float_value";
+        final String doubleKey = "Double_value";
+        final String message1Key = "Message_value";
```
armiol: Please don't ever use `message1` and `message2` as variable names or variable name parts.

server/src/test/java/org/spine3/server/entity/storagefield/StorageFieldTraverserShould.java
```diff
@@ -0,0 +1,121 @@
...
+        verify(traverser).hitInteger(eq(intKey), eq(intValue));
+        verify(traverser).hitLong(eq(longKey), eq(longValue));
+        verify(traverser).hitString(eq(stringKey), eq(stringValue));
+        verify(traverser).hitBoolean(eq(boolKey), eq(boolValue));
+        verify(traverser).hitFloat(eq(floatKey), eq(floatValue));
+        verify(traverser).hitDouble(eq(doubleKey), eq(doubleValue));
+        verify(traverser, times(2)).hitMessage(anyString(), eq(messageValue));
+    }
+
+    private static class Traverser extends StorageFieldTraverser {
```
armiol: Please rename it, so that it's clear. And also please provide some docs for it.

server/src/test/java/org/spine3/server/entity/storagefield/StorageFieldTraverserShould.java
```diff
@@ -0,0 +1,121 @@
...
+
+        @Override
+        protected void hitMessage(String fieldName, @Nullable Message value) {
+
```
armiol: Please remove the empty lines at all, or have a proper comment in each of them.

server/src/test/java/org/spine3/server/entity/storagefield/StorageFieldTraverserShould.java
```diff
@@ -0,0 +1,121 @@
...
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+/**
+ * @author Dmytro Dashenkov.
+ */
+public class StorageFieldTraverserShould {
```
armiol: The issue with the `...TraverserShould` and `...DecomposerShould` is that the single test method covers the whole functionality of the related classes.

Let's rethink the design of the original classes, so that the framework-provided type handlers and user-provided type handlers can be tested separately one by one.

server/src/test/java/org/spine3/server/entity/storagefield/StorageFieldsExtractorShould.java
```diff
@@ -0,0 +1,203 @@
...
+        public Float getFloatNull() {
+            return null;
+        }
+
+        public Message getSomeMessage() {
+            return someMessage;
+        }
+
+        int getSomeNonPublicMethod() {
+            throw new AssertionError("getSomeNonPublicMethod invoked");
```
armiol: Let's use `IllegalStateException` instead for all similar cases.

server/src/main/proto/spine/server/entity/entity.proto
```diff
@@ -72,3 +75,47 @@ message LifecycleFlags {
     //
     bool deleted = 2;
 }
+
+// Custom fields to store in the storage.
+//
+// These fields should be searchable in an effiefient way.
```
armiol: This comment is not enough. 

Let's expand it by explaining why we store the fields in this way (e.g. per type etc).

server/src/test/java/org/spine3/server/entity/storagefield/StorageFieldsExtractorShould.java
dmdashenkov: This is a test code.
I've chosen the `AssertionError` to make it clear we __do not__ handle is anywhere in the code.

dmdashenkov:
Closing this PR due to it's out of date.

PR #392, opened 2017-03-20
Exceptions belong to packages

alexander-yevsyukov:
This PR moves exception classes — that used to be under `error` sub-packages — to the level of packages that raise those exceptions. It would improve visibility of these exception classes.

In addition this PR:
* Removes `client.proto` file, which is not used.
* Removes redundant `Logging` utility class.


alexander-yevsyukov:
@alexander-aleksandrov, PTAL.

PR #393, opened 2017-03-21
Entity state validation

alexander-yevsyukov:
This PR simplifies validation state of entities. It is now possible to validate a new state at the level of `AbstractEntity` (previously it was available from `AbstractVersionableEntity`).

Also custom validation can be implemented via overriding `isValid()` method, which returns `boolean`.

The method `setState()` was renamed to `updateState()` to better reflect the nature of the call and highlight the fact of validation.

### DIFF
--- a/server/src/main/java/org/spine3/server/entity/AbstractEntity.java
+++ b/server/src/main/java/org/spine3/server/entity/AbstractEntity.java
@@ -21,6 +21,7 @@
 package org.spine3.server.entity;
 
 import com.google.protobuf.Message;
+import org.spine3.base.Stringifiers;
 import org.spine3.protobuf.Messages;
 
 import javax.annotation.CheckReturnValue;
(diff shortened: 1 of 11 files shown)


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/entity/AbstractEntity.java
```diff
@@ -181,6 +181,49 @@ private static IllegalStateException noSuchConstructor(String entityClass, Strin
...
+    protected void updateState(S state) {
+        validate(state);
+        injectState(state);
+    }
+
+    /**
+     * Verifies whether the new state is valid.
+     *
+     * <p>Default implementation does nothing always returning {@code true}.
+     * Derived classes may provide custom validation logic overriding this method.
```
armiol: Let's insert `by`, so that it will become `Derived classes may provide custom validation logic by overriding this method.`

server/src/main/java/org/spine3/server/entity/AbstractEntity.java
```diff
@@ -181,6 +181,49 @@ private static IllegalStateException noSuchConstructor(String entityClass, Strin
...
+     *
+     * @param newState a state object to replace the current state
+     * @return {@code true} if the new state object is valid, {@code false} otherwise
+     */
+    @SuppressWarnings("unused") // OK for this default implementation
```
armiol: Looks like a redundant suppression. This method seems to be called at line:220.

alexander-yevsyukov: This is about the parameter, which is not used.

alexander-yevsyukov: But removing it doesn't cause a warning, so I did.

alexander-yevsyukov:
@armiol, PTAL.

armiol:
@alexander-yevsyukov LGTM.

PR #394, opened 2017-03-21
CommandStore as a Repository

alexander-yevsyukov:
This PR implements `CommandStore` as a `Repository`.

Breaking changes:
* The `Repository` now does not have `BoundedContext` constructor argument. `AggregatePartRepository`, `ProcessManagerRepository`, and `ProjectionRepository` still have such a constructor argument.
* `CommandStore` creates its storage with the help of `StorageFactory` for creating `CommandStore` as a repository.
* Because of the above, `StorageFactory` does not produce `CommandStorage` any more.
* Because of the above, `InMemoryCommandStorage` class was also removed.


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/command/CommandRecords.java
```diff
@@ -0,0 +1,127 @@
...
+     * <p>{@code targetId} and {@code targetIdType} are set to empty strings if
+     * the command is not for an entity.
+     *
+     * @param command
+     *            a command to convert to a record. This includes instances of faulty commands.
+     *            An example of such a fault is missing command ID.
+     * @param status
+     *            a command status to set in the record
+     * @param generatedCommandId
+     *            a command ID to used because the passed command does not have own ID.
```
armiol: Should be `... to be used...` or ` ... to use ...`.

server/src/main/java/org/spine3/server/command/CommandRecords.java
```diff
@@ -0,0 +1,127 @@
...
+     *            If the command has own ID this parameter is {@code null}.
```
armiol: Please insert a comma after `has own ID`.

server/src/main/java/org/spine3/server/command/CommandRecords.java
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * <p>We don't have a command ID in the passed command.
+     * But need an ID to store the error in the record associated
+     * with this command. So, the ID will be generated.
+     *
+     * <p>We pass this ID to the record, so that it has an identity.
+     * But this ID does not belong to the command.
+     *
+     * <p>Therefore, commands without ID can be found by records
+     * where `command.context.command_id` field is empty.
```
armiol: Should we use `{@code command.context.command_id}` instead of backquotes?

server/src/main/java/org/spine3/server/projection/ProjectionRepository.java
```diff
@@ -64,43 +64,23 @@
...
+    private final BoundedContext boundedContext;
 
-        /** The repository is getting events from EventStore and builds projections. */
-        CATCHING_UP,
+    /** An instance of {@link StandFunnel} to be informed about state updates */
+    private final StandFunnel standFunnel;
 
-        /** The repository completed the catch-up process. */
-        ONLINE,
+    /** If {@code true} the projection will {@link #catchUp()} after initialization. */
```
armiol: I would add an alternative text to this `{@link #catchUp()}`.

PR #395, opened 2017-03-21
EventStorage as a repository

alexander-yevsyukov:
This PR implements storing events in a repository (instead of having a separate kind of a storage).

Breaking changes:
* The notion of abstract event storage is removed. `StorageFactory` no longer produces `EventStorage`.
* The `EventStorage` class is package-access to `org.spine3.server.event`.


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/event/EventEntity.java
```diff
@@ -0,0 +1,70 @@
...
+            final Event event1 = e1.getState();
+            final Event event2 = e2.getState();
+            final int result = Events.eventComparator()
+                                     .compare(event1, event2);
+            return result;
+        }
+    };
+
+    /**
+     * {@inheritDoc}
```
armiol: `@inheritDoc` [cannot be used](https://github.com/SpineEventEngine/core-java/wiki/Javadoc#do-not-use-inheritdoc-in-a-constructor-documentation) for constructors.

PR #396, opened 2017-03-22
Improve testing

dmdashenkov:
- Testing of some `Storage`s improved
 - Fixed transitive dependency on `protobuf-lite` issue

_Note:_ these changes improve the quality of the tests, not the coverage.


dmdashenkov:
@armiol, PTAL.

server/src/test/java/org/spine3/server/storage/AbstractStorageShould.java
```diff
@@ -152,6 +165,42 @@ public void rewrite_record_if_write_by_the_same_id() {
...
+        }
+
+        final Iterator<I> index = storage.index();
+        final Collection indexValues = Sets.newHashSet(index);
+
+        assertEquals(ids.size(), indexValues.size());
+        Verify.assertContainsAll(indexValues, ids.toArray());
+    }
+
+    @Test
```
armiol: Please use `expected` to match exceptions instead of catching and ignoring those.

Also, it's weird that we don't know for sure, which exception is expected. For me it should be either `UnsupportedOperationException` OR `IllegalStateException`, but not both in this particular use case.

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
```diff
@@ -194,6 +195,39 @@ public void write_record_bulk() {
         close(storage);
     }
 
+    @Test
+    public void rewrite_records_in_bulk() {
+        final RecordStorage<I> storage = getStorage();
+        final I id1 = newId();
```
armiol: Could we use a proper `for` statement instead of manual duplication of entries? If we use `for`, we could adjust the number of records to test easily.

server/src/test/java/org/spine3/server/storage/AbstractStorageShould.java
dmdashenkov: In this case, we rely on a Guava Collection implementation of `Iterator`. In one particular case  (in `InMemoryEventStorage`) we use a collection which throws `IllegalStateException`.
This is the reason why we don't use `expected` - it does not support multiple `Exception` types.
I don't like this either.

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
dmdashenkov: The point here was to show that some records are changed and some are not. I thought using a Collection here would make it more complicated since it would be less obvious what happens to the data.

dmdashenkov:
@armiol, please see my comments.

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
armiol: It is going to be obvious enough. Please do.

server/src/test/java/org/spine3/server/storage/AbstractStorageShould.java
armiol: Please have a comment explaining it then.

armiol:
@dmdashenkov please see my replies.

server/src/test/java/org/spine3/server/storage/AbstractStorageShould.java
dmdashenkov: Done

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
dmdashenkov: Done

dmdashenkov:
@armiol, PTAL.

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
```diff
@@ -197,27 +197,22 @@ public void write_record_bulk() {
 
     @Test
     public void rewrite_records_in_bulk() {
+        final int recordsCount = 3;
```
armiol: `recordCount`.

PR #398, opened 2017-03-23
Entity `Columns`

dmdashenkov:
Introducing __Columns__ — a convenient way to store custom _indexable_ fields of an `Entity` in a Spine Storage.

In this PR:
 - Basic implementation of the Columns retrieval mechanism
 - `v1.1` API of the Columns:
   - All the values are retrieved from the `Entity` getters.
   - A getter may have `@javax.annotation.Nullable` annotation to show that the field may accept `null`s. By default all the Columns are __non-null__.
   - A Column may have any type.
   - To handle the fields when writing a record, use the `ColumnType` interface methods.
   - To override the default storage types, pass an instance of `ColumnTypeRegistry` to the `StorageFactory`.

##### These methods represent the Columns:
```
--UserGroupAggregate.java--

public int getUserCount() {
    return getState().getUserCount();
} 

--PaymentProcessManager.java--

@Nullable
public Date getPaymentSystemName() {
    if (getState().hasExpirationDate()) {
        return new Date();
    } else {
        return null;
    }
}
```

##### And these methods are __not__ valid Columns:
```
--UserAggregate.java--

// non-public methods may not represent Columns
private double getHourRate() { ... }

// only methods starting with "get" or "is" may be considered Columns
public boolean hasChildren() { ... }

// getter methods must not accept arguments
public User getStateOf(UserAggregate other) { ... } 

// only instance methods are considered Columns
public static Integer isNew(UserAggregate aggregate) { ... }
```
The examples above will not be included when generating the `StorageFields` object.

##### Faulty Columns:
The example below will throw an `Exception` when trying to generate the Columns.
```
--EmployeeProjection.java--

// method should be annotated as @Nullable to return a null value
public Message getAddress() {
    return null;
}
```

Also note, that `Entity` classes which the Columns are generated for, should be `public`. If it's not, the framework user will see a warning message in the log and the `StorageFields` object will never be generated.

##### ColumnType example:
```
--JdbcTypes.java--

class VarcharDateType implements ColumnType<Date, String, PreparedStatement, Integer> {
    @Override
    public String convert(Date fieldValue) {
        return MY_DATE_FORMAT.format(fieldValue);
    }

    @Override
    public void set(PreparedStatement storageRecord, String value, Integer columnIdentifier) {
        storageRecord.setString(columnIdentifier, value);
    }
}
``` 

And then in your `BoundedContext`:
```
final JdbcColumnTypeRegistry registry = JdbcColumnTypeRegistry.newBuilder()
                                                      .put(Date.class, new VarcharDateType())
                                                      .build();

JdbcStorageFactory.newBuilder()
                  // ...
                  .setColumnTypeRegistry(registry)
                  .build();
```

_Note:_ `ColumnTypeRegistry` supports by superclass `ColumnType` resolving, meaning, that if a precise type of a Storage Field is not registered, but its superclass is, the superclass's mapping will be used.

For example, you may register all the protobuf `Message`s to be mapped to a single  `ColumnType` by passing next params to the registry: `put(AbstractMessage.class, someColumnOfAbstractMessage)`. 

In the example above, if we store a Storage Field of unregistered type, which `extends AbstractMessage`, this field will be mapped via that `someColumnOfAbstractMessage` column.

_Note:_ indirect registry of a `ColumnType` by an instance of `Class` which corresponds to an `interface` is not supported, that's why registering a `ColumnType` for `com.google.protobuf.Message` is redundant, since this type will never be used.


dmdashenkov:
@armiol, PTAL.
_Note:_ the work is still in progress.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordEnvelope.java
```diff
@@ -0,0 +1,88 @@
...
+
+import javax.annotation.Nullable;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * @author Dmytro Dashenkov
+ */
+public class EntityRecordEnvelope {
```
armiol: No idea why you need that at all.

Envelope is something deliverable. Most likely, deliverable via `Bus`. Not sure why this concept is used here.

server/src/main/java/org/spine3/server/storage/WritableStorage.java
```diff
@@ -0,0 +1,38 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.storage;
+
+/**
+ * @author Dmytro Dashenkov
+ */
+public interface WritableStorage<I, R> {
```
armiol: Looks weird. I thought all `Storage`s were writable.

dmdashenkov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * A registry of type conversion strategies for the Storage Fields.
+ *
+ * @author Dmytro Dashenkov
+ */
+public class ColumnTypeRegistry {
+
+    private static final ColumnTypeRegistry EMPTY_INSTANCE = newBuilder().build();
+
+    private final Map<Class, ColumnType> columnTypeMap;
```
armiol: It's weird that you declare a `Map` here, while `EntityRecordWithStorageFields` declares an `ImmutableMap` under the similar circumstances.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A packed value of an {@link EntityRecord} with its {@linkplain StorageFields storage fields}.
+ *
+ * @author Dmytro Dashenkov
+ */
+public class EntityRecordWithStorageFields {
```
armiol: This class looks `final`.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,115 @@
...
+import org.spine3.server.entity.EntityRecord;
+import org.spine3.server.entity.storage.reflect.Column;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A packed value of an {@link EntityRecord} with its {@linkplain StorageFields storage fields}.
```
armiol: It's not `packed` per se. It's a value object, representing the record along with the additional storage fields.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,115 @@
...
+    @Nullable
+    private final ImmutableMap<String, Column.MemoizedValue<?>> storageFields;
+
+    /**
+     * Creates a new instance of the {@code EntityRecordWithStorageFields}.
+     *
+     * @param record        {@link EntityRecord} to pack
+     * @param storageFields {@linkplain StorageFields storage fields map} to pack
+     */
+    public EntityRecordWithStorageFields(EntityRecord record,
```
armiol: I'd make it `private` and create a factory method instead — to comply with the construction approach we have for value objects.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,115 @@
...
+     * <p>An object created with this constructor will always return {@code false} on
+     * {@link #hasStorageFields()}.
+     *
+     * @param record {@link EntityRecord} to pack
+     * @see #hasStorageFields()
+     */
+    @SuppressWarnings("ConstantConditions") // null value for the storage fields map
+    public EntityRecordWithStorageFields(EntityRecord record) {
+        this.record = checkNotNull(record);
+        this.storageFields = null;
```
armiol: I'd make it just empty.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @SuppressWarnings("ReturnOfCollectionOrArrayField") // Immutable structure
+    public Map<String, Column.MemoizedValue<?>> getStorageFields() {
+        return storageFields == null
+               ? StorageFields.empty()
+               : storageFields;
+    }
+
+    /**
+     * Shows whether there were any storage fields passed in the object initialization or not.
```
armiol: `Determines whether there are any storage fields for this record.`

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * <p>If returns {@code false}, the {@linkplain StorageFields storage field} are not considered
+     * in the storage and
```
armiol: And what? (c)

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Multimaps.synchronizedListMultimap;
+
+/**
+ * A utility for generating the Storage fields {@linkplain Map}.
```
armiol: Please make the mentions of  `Storage field`(s) consistent in terms of case of the letters.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * <p>All the methods of the passed {@link Entity} that fit
+ * <a href="http://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/">the Java Bean getter spec</a>
+ * are considered as Storage Fields and indexed respectively.
+ *
+ * <p>When passing an instance of already known {@link Entity} type, the methods are retrieved form
```
armiol: `from`.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Multimaps.synchronizedListMultimap;
+
+/**
+ * A utility for generating the Storage fields {@linkplain Map}.
+ *
+ * <p>All the methods of the passed {@link Entity} that fit
+ * <a href="http://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/">the Java Bean getter spec</a>
```
armiol: Please follow the link and fix its anchor text.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+ * are considered as Storage Fields and indexed respectively.
```
armiol: Remove `as`.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * <p>When passing an instance of already known {@link Entity} type, the methods are retrieved form
+ * a cache and are not updated.
+ *
+ * <p>There are several excluded methods, which are never indexed and are <b>NOT</b> considered as
```
armiol: Same as above. Please review the similar cases.

armiol: Just bold is enough.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+ * <ul>
+ *     <li>{@link Object#getClass()}
+ *     <li>{@link Entity#getId()}
+ *     <li>{@link Entity#getState()}
+ *     <li>{@link Entity#getDefaultState()}
+ *     <li>{@link org.spine3.server.entity.EntityWithLifecycle#getLifecycleFlags() EntityWithLifecycle#getLifecycleFlags()}
+ *     <li>{@link org.spine3.server.aggregate.Aggregate#getBuilder() Aggregate#getBuilder()}
+ * </ul>
+ *
+ * <p>Note: if creating a getter method with a name which intersects with one of these methods'
```
armiol: `method names`.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+ * names, your getter method will also <b>NOT</b> be considered as a Storage Field.
```
armiol: Again, bold is enough.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+     * A one to many container of the {@link Class} to {@link Column} relations.
+     *
+     * <p>This container is mutable and thread safe.
+     */
+    private static final Multimap<Class<? extends Entity>, Column<?>> knownEntityProperties =
+            synchronizedListMultimap(
+                    LinkedListMultimap.<Class<? extends Entity>, Column<?>>create());
+
+    private StorageFields() {
+    }
```
armiol: Please have a comment in this empty c'tor body.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,201 @@
...
+     * @return a {@link Map} of the Storage Field names to their
+     * {@linkplain Column.MemoizedValue memoized values}.
+     * @see Column.MemoizedValue
+     */
+    public static <E extends Entity<?, ?>> Map<String, Column.MemoizedValue<?>> from(E entity) {
+        checkNotNull(entity);
+        final Class<? extends Entity> entityType = entity.getClass();
+        final int modifiers = entityType.getModifiers();
+        if (!Modifier.isPublic(modifiers)) {
+            log().warn(NON_PUBLIC_CLASS_WARNING);
```
armiol: `NON_PUBLIC_CLASS_WARNING` is a template, not a ready-to-go `String`.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
armiol: I don't think we should call those `indexed`. As it is a Storage domain, we should be very careful with this word.

server/src/main/java/org/spine3/server/entity/storage/reflect/Column.java
```diff
@@ -0,0 +1,207 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.entity.storage.reflect;
```
armiol: The hierarchy is too deep already. I doubt that even `org.spine3.server.entity.storage` makes sense. `org.spine3.server.entity.storage.reflect` is both too much and incorrect.

Perhaps you could review the visibility of this class and its methods. And also mark it either `@SPI` or `@Internal`.

server/src/main/java/org/spine3/server/entity/storage/reflect/Column.java
```diff
@@ -0,0 +1,207 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.String.format;
+
+/**
+ * A reflective representation of a Storage Field.
```
armiol: Please expand this description. It's way too short and thus unclear.

server/src/main/java/org/spine3/server/entity/storage/reflect/Column.java
```diff
@@ -0,0 +1,207 @@
...
+        return getter.equals(column.getter);
+    }
+
+    @Override
+    public int hashCode() {
+        return getter.hashCode();
+    }
+
+    /**
+     * A memoized in a point in time value of the Storage Field.
```
armiol: `A value of a Storage Field saved at some point in time.`

BTW, I'd add something else, since that's too short — basically you just repeated the class name (i.e. `Column.MemoizedValue`.

Also, since you are using it from other classes, I'd probably move it to an upper level.

server/src/main/java/org/spine3/server/entity/storage/reflect/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+@ParametersAreNonnullByDefault
+package org.spine3.server.entity.storage.reflect;
```
armiol: That's the problem with the long package names — it's hard to give a reasonable description :)

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
```diff
@@ -172,25 +187,48 @@ public void delete_record() {
...
         }
-        storage.write(expected);
+        storage.write(initial);
 
         final Collection<EntityRecord> actual = newLinkedList(
-                storage.readMultiple(expected.keySet())
+                storage.readMultiple(initial.keySet())
         );
+        final Collection<EntityRecord> expected =
+                Collections2.transform(initial.values(),
```
armiol: Can we please keep 100 symbols wide in tests as well?

server/src/main/java/org/spine3/server/entity/storage/ColumnType.java
```diff
@@ -0,0 +1,62 @@
...
+package org.spine3.server.entity.storage;
+
+import org.spine3.annotations.SPI;
+
+/**
+ * An interface for handling type conversion for the storage fields.
+ *
+ * @param <J> the Java type represented by the column
+ * @param <S> the "store as" type of the column
+ * @param <R> the type of the record in the database, which holds a single tuple of data and which
```
dmdashenkov: @armiol, could you please explain the difference between `tuple` and `cortege`?

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
dmdashenkov: The point was to make it `null`. An empty map is a valid argument here (i.e. custom `Entity` impl with no getters). But in some cases, we just don't need to write the Storage Fields (or we simply can't). This cases should be handled by the `Storage` respectively.
See `hasStorageFields` method.

server/src/main/java/org/spine3/server/entity/storage/reflect/Column.java
dmdashenkov: There is a such a precedent: `org.spine3.server.storage.datastore` and simmilar packages.

dmdashenkov: Though I agree it's pretty strange. I'll remove the package and let this class (which is a single citizen of the package) be situated in a higher package.

dmdashenkov:
@armiol, PTAL again.

dmdashenkov:
@armiol, PTAL again.

dmdashenkov:
@armiol 
I've finished all the changes I wanted to apply according to the experience of migrating to Storage Field in `gae-java`. Please have a look.

dmdashenkov:
@armiol 
UPD: please also take a look at the updated PR description (the last section describing the `ColumnTypeRegistry`).

armiol:
@dmdashenkov please fix the PR description:
```
public int getUsersCount() {     // should be `getUserCount()` as below.
    return getState().getUserCount();
} 
```

armiol:
@dmdashenkov again, the PR description issue:
```
// only instance methods are considered as Storage Fields   <--- remove `as`.
```

dmdashenkov:
@armiol, it seems OK now. PTAL.

server/src/main/java/org/spine3/server/entity/storage/ColumnType.java
```diff
@@ -0,0 +1,62 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.entity.storage;
+
+import org.spine3.annotations.SPI;
+
+/**
+ * An interface for handling type conversion for the Storage Fields.
+ *
```
armiol: Some details are really required.

Please describe the flow. See `CommandBus`/`EventBus` or `Subscribe` annotation for examples.

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+         * Create a mapping between the Java class and the database
+         * {@linkplain ColumnType column type}.
+         *
+         * @param javaType   the Java class to map the value from
+         * @param columnType the the database {@linkplain ColumnType column type} to map value to;
+         *                   must be an instance of {@code <C>}
+         * @param <J>        the Java type
+         * @return self for call chaining
+         */
+        public <J> Builder<C> put(Class<J> javaType, ColumnType<J, ?, ?, ?> columnType) {
```
armiol: Let's check for `null`s at least. Or maybe something else as well. 

The aim is to ensure the `ColumnTypeRegistry` is built correctly, or fail fast.

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * A registry of type conversion strategies for the Storage Fields.
+ *
+ * @param <C> storage-specific implementation of the {@link ColumnType}
+ * @author Dmytro Dashenkov
+ */
+public class ColumnTypeRegistry<C extends ColumnType> {
```
armiol: Looks `final`.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,136 @@
...
+    @SuppressWarnings("ConstantConditions") // null value for the Storage Fields map
+    private EntityRecordWithStorageFields(EntityRecord record) {
+        this.record = checkNotNull(record);
+        this.storageFields = null;
+    }
+
+    /**
+     * Creates a new instance of the {@code EntityRecordWithStorageFields}.
+     */
+    public static EntityRecordWithStorageFields newInstance(
```
armiol: Maybe, call these methods `of` for brevity and also to conform the contract we use to follow in such cases?

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,136 @@
...
+ * A value of {@link EntityRecord} associated with its {@linkplain StorageFields Storage Fields}.
+ *
+ * @author Dmytro Dashenkov
+ */
+public final class EntityRecordWithStorageFields {
+
+    private final EntityRecord record;
+
+    @Nullable
+    private final ImmutableMap<String, Column.MemoizedValue<?>> storageFields;
```
armiol: It's easier and clearer to have just an empty instance instead of `@Nullable` field. 

And you'll get rid of a warning suppression in the constructor.

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithStorageFields.java
```diff
@@ -0,0 +1,136 @@
...
+import com.google.common.collect.ImmutableMap;
+import org.spine3.server.entity.EntityRecord;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A value of {@link EntityRecord} associated with its {@linkplain StorageFields Storage Fields}.
```
armiol: It's weird that you reference the utility class.

server/src/main/java/org/spine3/server/entity/storage/SimpleColumnType.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A base for implementing {@link ColumnType} interface regardless the type conversion.
+ *
+ * @param <T> the type of the storage field
+ * @param <R> the type of the record in the database, which holds a single cortege of data and
+ *            is consumed by the database upon write
+ * @param <C> the type of the column identifier in the {@code R}
+ * @author Dmytro Dashenkov
+ */
+public abstract class SimpleColumnType<T, R, C> implements ColumnType<T, T, R, C> {
```
armiol: Looks `@SPI`. Same as its interface, by the way.

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,211 @@
...
+        return getter.equals(column.getter);
+    }
+
+    @Override
+    public int hashCode() {
+        return getter.hashCode();
+    }
+
+    /**
+     * A value of a Storage Field saved at some point in time.
```
armiol: This mysterious description is missing something.

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,211 @@
...
+     *
+     * <p>The class associates the Storage Field value with its metadata - {@linkplain Column}.
```
armiol: Replace `-` with `i.e.` or rephrase the sentence.

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,211 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.String.format;
+
+/**
+ * A reflective representation of a Storage Field.
```
armiol: The word `reflective` is not good enough to describe the idea behind this class.

Please remove `reflective` from here, but add much more content below to explain the intention.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,203 @@
...
+public class StorageFields {
+
+    private static final String NON_PUBLIC_CLASS_WARNING =
+            "Passed entity class %s is not public. Storage fields won't be extracted.";
+
+    /**
+     * An immutable {@link java.util.Set set} of excluded methods which fit the getter pattern but
+     * are not Storage Fields.
+     */
+    private static final ImmutableSet<String> EXCLUDED_METHODS =
```
armiol: Instead, I'd have a nested enum (with a descent description for each of the items) and an API telling whether some method is excluded.

server/src/main/java/org/spine3/server/entity/storage/StorageFields.java
```diff
@@ -0,0 +1,203 @@
...
+        if (knownEntityProperties.containsKey(entityType)) {
+            return;
+        }
+        addToCaches(entityType);
+    }
+
+    /**
+     * Caches the {@linkplain Entity entity type} for further Storage Fields retrieving.
+     */
+    private static void addToCaches(Class<? extends Entity> entityType) {
```
armiol: `addToCache`.

server/src/main/java/org/spine3/server/storage/RecordStorage.java
```diff
@@ -92,26 +93,43 @@ protected RecordStorage(boolean multitenant) {
     }
 
     /**
-     * {@inheritDoc}
+     * Writes a record and its
+     * {@linkplain org.spine3.server.entity.storage.StorageFields Storage Fields} into the storage.
```
armiol: One thought that I have is why we keep calling that `StorageFields`, while we don't have anything named like that in our API? Maybe we should call it `Entity Columns` and reference the `Column` class?

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
dmdashenkov: Done.

In fact, the main constraint for the `ColumnTypeRegistry` is matching the `class` object to the `ColumnType` object which receives the same type as the input.
Current signature guarantees that the matching is accurate.

dmdashenkov:
@armiol, PTAL again.

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,235 @@
...
+ * {@link javax.annotation.Nullable javax.annotation.Nullable}. Otherwise, the Column is considered
+ * non-null.
+ *
+ * <p>If a non-null getter method returns {@code null} when trying to get the value of a Column,
+ * a {@linkplain RuntimeException} is thrown. See {@link #isNullable()}.
+ *
+ * @see ColumnType
+ * @author Dmytro Dashenkov
+ */
+public class Column<T> {
```
armiol: Some of the code samples that you have listed in the PR description would look nice here as well.

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,235 @@
...
+        return result;
+    }
+
+    /**
+     * Retrieves the name of the property which represents the Column.
+     *
+     * <p>For example, if the getter method has name "isArchivedOrDeleted", the returned value is
+     * "archivedOrDeleted".
+     *
+     * @return the name of the property reflected by this object
```
armiol: Did you mean `exposed by this object`?

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,235 @@
...
+                           getName(),
+                           source),
+                    e);
+        }
+    }
+
+    /**
+     * Retrieves the Column value from the given {@link Entity}.
+     *
+     * <p>The value is wrapped into a special container, which bears information about the Field's
```
armiol: `which bears the field metadata.`

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,235 @@
...
+     * metadata.
+     *
+     * @param source the {@link Entity} to get the Fields from
```
armiol: Please avoid capitalizing `Fields` everywhere. That's just a field.

You can refer to it as to an "Entity property" or "Entity field".

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
```diff
@@ -0,0 +1,136 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * A registry of type conversion strategies for the {@link Column Columns}.
+ *
+ * @param <C> storage-specific implementation of the {@link ColumnType}
+ * @author Dmytro Dashenkov
+ */
+public final class ColumnTypeRegistry<C extends ColumnType> {
```
armiol: A code sample is required here as well.

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private ColumnTypeRegistry(Builder<C> builder) {
+        this.columnTypeMap = ImmutableMap.copyOf(builder.columnTypeMap);
+    }
+
+    /**
+     * @return an empty registry, used for
+     * {@linkplain org.spine3.server.storage.memory.InMemoryStorageFactory in-memory storages} etc.
+     */
+    public static <C extends ColumnType> ColumnTypeRegistry<C> empty() {
```
armiol: The more I look at this method, the more I think it belongs to the `InMemoryStorageFactory` under the `emptyColumnRegistry` name.

It's just a stranger to the public API.

server/src/main/java/org/spine3/server/entity/storage/Column.java
dmdashenkov: Done

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
dmdashenkov: Done

dmdashenkov:
@armiol, PTAL again.

server/src/test/java/org/spine3/server/entity/storage/reflect/ColumnShould.java
```diff
@@ -0,0 +1,196 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.entity.storage.reflect;
```
armiol: The package is wrong. 

Please review the packages of the test classes you created within this PR. They should correspond their targets.

Once this is fixed, at least `Column.MemoizedValue` can become package-visible.

server/src/main/java/org/spine3/server/entity/storage/ColumnType.java
```diff
@@ -0,0 +1,93 @@
...
+     * is stored.
+     *
+     * <p>Common example is converting
+     * {@link com.google.protobuf.Timestamp com.google.protobuf.Timestamp} into
+     * {@link java.util.Date java.util.Date}.
+     *
+     * @param fieldValue the {@link Column Column} of the initial type
+     * @return the {@link Column Column} of the "store as" type
+     */
+    S convertColumnValue(J fieldValue);
```
armiol: IDEA says this method is not used.

I would expect at least some tests to call it.

server/src/main/java/org/spine3/server/entity/storage/ColumnType.java
```diff
@@ -0,0 +1,93 @@
...
+     * Set the {@link Column Column} value to the database record type.
+     *
+     * <p>Common example is setting a value to
+     * a {@link java.sql.PreparedStatement PreparedStatement} instance into a determined position.
+     *
+     * @param storageRecord    the database record
+     * @param value            the value to store
+     * @param columnIdentifier the identifier of the column, e.g. its index
+     */
+    void setColumnValue(R storageRecord, S value, C columnIdentifier);
```
armiol: Same as above.

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
```diff
@@ -0,0 +1,156 @@
...
+        return Collections.unmodifiableMap(columnTypeMap);
+    }
+
+    public static <C extends ColumnType> Builder<C> newBuilder() {
+        return new Builder<>();
+    }
+
+    public static <C extends ColumnType> Builder<C> newBuilder(ColumnTypeRegistry<C> src) {
+        final Builder<C> builder = new Builder<>();
+        builder.columnTypeMap.putAll(src.columnTypeMap);
```
armiol: Let's use `Builder.put` calls instead to reuse the validation.

Even if the validation code is expanded with some more sophisticated checks, we'll always be okay.

server/src/main/java/org/spine3/server/entity/RecordBasedRepository.java
```diff
@@ -284,9 +287,12 @@ protected E findOrCreate(I id) {
     /**
      * Converts the passed entity into the record.
      */
-    protected EntityRecord toRecord(E entity) {
+    protected EntityRecordWithColumns toRecord(E entity) {
         final EntityRecord entityRecord = entityConverter().convert(entity);
```
armiol: `entityRecord` can be `null`. Detected by IDEA, please pay attention.

server/src/main/java/org/spine3/server/entity/RecordBasedRepository.java
```diff
@@ -284,9 +287,12 @@ protected E findOrCreate(I id) {
...
-        return entityRecord;
+        final Map<String, Column.MemoizedValue<?>> storageFields = Columns.from(entity);
```
armiol: I don't like exposing `Column.MemoizedValue` out of its package.

Can we create a factory method in the  `EntityWithColumns`, which produces `EntityRecordWithColumns` instance out of `EntityRecord`?

server/src/main/java/org/spine3/server/entity/RecordBasedRepository.java
dmdashenkov: The point is that the Guava `Converter` has all its `public` methods `@Nullable`.
I'll add `checkNotNull`.

server/src/main/java/org/spine3/server/entity/storage/ColumnType.java
dmdashenkov: Done

server/src/main/java/org/spine3/server/entity/storage/ColumnType.java
dmdashenkov: Done

dmdashenkov:
@armiol, PTAL again.
The primary change - `ColumnRecords` utility helps to deliver the Columns to the database-specific container.

server/src/main/java/org/spine3/server/entity/storage/ColumnRecords.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private ColumnRecords() {
+        // Prevent initialization of the utility class
+    }
+
+    /**
+     * Feeds the given {@link EntityRecordWithColumns} to the given record representation.
+     *
+     * <p>This method deals with the {@linkplain Column columns} only.
+     * You should handle {@link org.spine3.server.entity.EntityRecord EntityRecord} separately when
```
armiol: Please use the passive voice instead.

server/src/main/java/org/spine3/server/entity/storage/ColumnRecords.java
```diff
@@ -0,0 +1,101 @@
...
+        checkNotNull(recordWithColumns);
+        checkNotNull(columnTypeRegistry);
+        checkArgument(recordWithColumns.hasColumns(),
+                      "Passed record has no Entity Columns.");
+
+        for (Map.Entry<String, Column.MemoizedValue<?>> column : recordWithColumns.getColumnValues()
+                                                                                  .entrySet()) {
+            final I columnIdentifier = mapColumnIdentifier.apply(column.getKey());
+            checkNotNull(columnIdentifier);
+            @SuppressWarnings("unchecked")
```
armiol: Some comments are required for the suppressions.

server/src/main/java/org/spine3/server/entity/storage/ColumnType.java
```diff
@@ -88,6 +93,19 @@
...
      * @param value            the value to store
      * @param columnIdentifier the identifier of the column, e.g. its index
+     * @see #setNull(Object, Object)
      */
     void setColumnValue(R storageRecord, S value, C columnIdentifier);
+
+    /**
+     * Sets the {@code null} value to the {@linkplain Column}.
+     *
+     * <p>This method is called when the {@linkplain Column} value desired for storing turn out
```
armiol:
```
Used if the actual value of the {@linkplain Column}, defined for storing, is {@code null}.
```

server/src/main/java/org/spine3/server/entity/storage/EntityRecordWithColumns.java
```diff
@@ -35,6 +38,17 @@
  */
 public final class EntityRecordWithColumns {
 
+    private static final Maps.EntryTransformer<String, Column.MemoizedValue<?>, Column<?>>
+    columnTransformer = new Maps.EntryTransformer<String, Column.MemoizedValue<?>, Column<?>>() {
```
armiol: Can we extract this code into a static method instead?

Maybe having an internal class would be even better. The aim is 

a) to get rid of ` Maps.EntryTransformer<String, Column.MemoizedValue<?>, Column<?>>` definition, which is really long;
b) to avoid long field initialization, that is hard to read.

server/src/main/java/org/spine3/server/entity/storage/ColumnRecords.java
```diff
@@ -0,0 +1,101 @@
...
+            final Column.MemoizedValue<Object> columnValue =
```
armiol: Let's use static import for `MemoizedValue`.

armiol: Let's use static import for `MemoizedValue`.

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -0,0 +1,296 @@
...
+ *         public Message getAddress() {
+ *             return null;
+ *         }
+ *     </pre>
+ * </code>
+ *
+ * @see ColumnType
+ * @author Dmytro Dashenkov
+ */
+public class Column<T> {
```
armiol: I am not sure, but it looks `final`.

dmdashenkov: It's effectively `final` (having a single and only `private` constructor).

dmdashenkov: Though, I'll add the modifier to make it explicit.

dmdashenkov:
@armiol, let's keep `Column` effectively `final` without marking it with the modifier. This is required for testing: __Mockito__ isn't able to _mock_ a `final` class.

dmdashenkov:
@armiol, PTAL again.

PR #399, opened 2017-03-23
Fix the package info description of `org.spine3.server.entity`

armiol:
Before this PR the description of the package was irrelevant to its contents.


armiol:
@alexander-yevsyukov PTAL.

PR #402, opened 2017-03-28
Use `StorageFactorySwitch` instead of `InMemoryStorageFactory.getInstance()`

armiol:
Before this PR both test and production code used `InMemoryStorageFactory.getInstance()` explicitly. That made impossible to use an alternative storage mechanism in tests in case of a need.

This request aims to migrate to `StorageFactorySwitch.getInstance().get()` call instead. That makes `StorageFactorySwitch` a single configuration spot, allowing to specify another `StorageFactory` implementation.


armiol:
@alexander-yevsyukov PTAL.

armiol:
@alexander-yevsyukov I'd merge those with `master` once #370 is merged as well.

PR #403, opened 2017-03-28
Right margin wrapping server module first part

alexander-aleksandrov:
This PR contains the changes in wrapping for 100 column right margin. Unfortunately, it is not possible to state exact packages, since all changes was made through inspection results.

PR #404, opened 2017-03-28
Change wrapping in server module - last part.

alexander-aleksandrov:
This PR contains the changes in wrapping for 100 column right margin.


server/src/main/java/org/spine3/server/validate/StringFieldValidator.java
```diff
@@ -48,8 +48,8 @@
      * @param descriptor    a descriptor of the field to validate
      * @param fieldValues   values to validate
      * @param rootFieldPath a path to the root field (if present)
-     * @param strict        if {@code true} the validator would assume that the field is required even
-     *                      if the corresponding option is not set
+     * @param strict        if {@code true} the validator would assume that the field is required
+     *                      even if the corresponding option is not set
```
alexander-yevsyukov: When you put `even` on the next line, it makes the first line look complete. It does not encourage further reading, and the reader may miss an important piece of information we put on the following line. This pattern repeats in many wrapping of yours. I would put `if` on the previous line instead.

alexander-aleksandrov: Done

alexander-aleksandrov:
The work has been done and merged in #403 PR.

PR #416, opened 2017-03-31
Improve package structure and Event factories

alexander-yevsyukov:
This PR improves the package structure as follows:
* The `command` package was split into `command`, `commandbus`, and `commandstore`
* Classes from `server.validate` were moved to `validate` in the client API.
* Creation of events now is server-side code. 
   * `EventFactory` is a configurable class which can generate a sequence of events in response to a command.
   * `ImportEventFactory` is a utility class for creating import events. It will be transformed into a factory in a following PR.

Other changes:
* `AbstractValidatingBuilder` conversion method was renamed to just `convert()` with parameter order similar to Gson and other converters.
* Direct creation of `EventId` and `EventContext` was deprecated. The deprecations will be addressed in a following PR.


alexander-yevsyukov:
@IlliaShepilov, consider merging with this branch in your validator-related works.

alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/command/EventFactory.java
```diff
@@ -0,0 +1,100 @@
...
+                                  .setMessage(packed)
+                                  .setContext(context)
+                                  .build();
+        return result;
+    }
+
+    /**
+     * Creates new {@code CommandContext} with passed parameters.
+     *
+     * @param producerId the ID of an object which produced the event
```
armiol: The params section looks unformatted.

alexander-yevsyukov:
@armiol, please have a look at new `EventFactory` API.

server/src/main/java/org/spine3/server/command/EventIdSequence.java
```diff
@@ -0,0 +1,127 @@
...
+
+import static org.spine3.base.Identifiers.idToString;
+
+/**
+ * Generates a sequence of identifiers of events produced in response to a command.
+ *
+ * @author Alexander Yevsyukov
+ * @see EventId
+ */
+class EventIdSequence {
```
armiol: Looks `final`.

server/src/main/java/org/spine3/server/command/EventIdSequence.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * The length of the sequence number suffix.
+     */
+    private final int suffixLength;
+
+    /**
+     * The number of the next ID in the sequence.
+     */
+    private int count = 0;
```
armiol: Just in case, I'd use `AtomicInteger` instead. 

I realise our current use cases operate with `Event` ID generation in a synchronous manner. However, in future that may easily change.

Perfomance-wise it IS going to affect us; however, [`Strings.padStart` method](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Strings.java#L88) already takes much more CPU and memory, than increments based upon `AtomicInteger` will.

server/src/main/java/org/spine3/server/command/ImportEventFactory.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.protobuf.Any;
+import com.google.protobuf.Message;
+import org.spine3.base.Event;
+import org.spine3.base.EventContext;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.protobuf.Timestamps2.getCurrentTime;
+
+/**
+ * The factory for creating import events.
```
armiol: If I understand correctly, we used to call those `integration events`. 

Is this the same thing?

server/src/main/java/org/spine3/server/command/ImportEventFactory.java
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * <p>The method does not set {@code CommandContext} because there was no command.
+     *
+     * <p>The {@code version} attribute is not populated either.
+     * It is the responsibility of the target aggregate to populate the missing fields.
+     *
+     * @param producerId the ID of the producer which generates the event packed into {@code Any}
+     * @return new instance of {@code EventContext} for the imported event
+     */
+    public static EventContext createEventContext(Any producerId) {
```
armiol: I wonder how would our tenant-aware `EventOperation`'s go with imported events. That would be difficult, since `tenantId` is a property of `CommandContext`,

server/src/test/java/org/spine3/server/command/ImportEventFactoryShould.java
```diff
@@ -0,0 +1,59 @@
...
+import org.spine3.protobuf.AnyPacker;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.spine3.protobuf.AnyPacker.unpack;
+import static org.spine3.protobuf.Values.newStringValue;
+import static org.spine3.test.Tests.newUuidValue;
+import static org.spine3.validate.Validate.isNotDefault;
+
+public class ImportEventFactoryShould {
```
armiol: Please have an `@author` tag, as we agreed recently.

server/src/main/java/org/spine3/server/command/ImportEventFactory.java
alexander-yevsyukov: As discussed vocally, now these are different things that might converge in the near future. We created this functionality for updating aggregates from data coming from external data sources.

server/src/main/java/org/spine3/server/command/ImportEventFactory.java
alexander-yevsyukov: For now this class is just a holder of factory methods that are likely to go away as we re-think importing events and working with integration events.

alexander-yevsyukov:
@armiol, PTAL.

server/src/test/java/org/spine3/server/command/EventFactoryShould.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.google.protobuf.Empty;
+import org.junit.Test;
+import org.spine3.base.CommandContext;
+import org.spine3.client.CommandFactory;
+import org.spine3.test.TestCommandFactory;
+
+import static org.spine3.test.Tests.newUuidValue;
+
+public class EventFactoryShould {
```
armiol: Please add an `@author` tag.

PR #417, opened 2017-04-04
Add static import on demand for all `String.format()` cases.

alexander-aleksandrov:
Our  [Java Code Style](https://github.com/SpineEventEngine/core-java/wiki/Java-Code-Style#fqns-in-javadoc) states that we should use static import for String.format() cases.


alexander-aleksandrov:
@dmytro-grankin PTAL

PR #419, opened 2017-04-05
AggregateStateId Stringifier

dmdashenkov:
Added a `Stringifier` for `AggregateStateId`.
For converting the ID it uses the `Stringifier` of the underlying ID type, i.e. if an `Aggregate` has a `Message` type, a corresponding `Stringifier` should be registered.

The behavior of the `Stringifier` may be overridden by the framework user (as it is for any other `Stringifier`).


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL.
Since the default Stringifier is already implemented, let's see if this changes may be applied now.

server/src/main/java/org/spine3/server/stand/AggregateStateId.java
```diff
@@ -83,4 +111,133 @@ public String toString() {
                           .add("stateType", stateType)
                           .toString();
     }
+
+    private static class AggregateStateIdStringifier extends Stringifier<AggregateStateId> {
```
armiol: Looks as a package-private class outside the `AggregateStateId`.

server/src/main/java/org/spine3/server/stand/AggregateStateId.java
```diff
@@ -37,19 +51,33 @@
...
+        this.aggregateId = checkNotNull(aggregateId);
+        this.stateType = checkNotNull(stateType);
     }
 
     @CheckReturnValue
     public static <I> AggregateStateId of(I aggregateId, TypeUrl stateType) {
         return new AggregateStateId<>(aggregateId, stateType);
     }
 
+    public static Stringifier<AggregateStateId> stringifier() {
```
armiol: I don't understand why you need this method at all.

server/src/main/java/org/spine3/server/stand/AggregateStateId.java
```diff
@@ -37,19 +51,33 @@
  * @author Alex Tymchenko
  */
 public final class AggregateStateId<I> {
+
+    private static final AggregateStateIdStringifier stringifier =
+            new AggregateStateIdStringifier();
+
+    static {
+        StringifierRegistry.getInstance()
```
armiol: This way of registration is dangerous. Since it's a `static` initialization block, it may never be executed. I would put it somewhere else.

server/src/main/java/org/spine3/server/stand/AggregateStateId.java
dmdashenkov: Intuitively a user will search for a `Stringifier` for `AggregateStateId` within the class itself since of other types we declare the `Stringifier`s in the classes themselves or in the related utility classes (e.g. `Events.stringifier()`).

server/src/main/java/org/spine3/server/stand/AggregateStateId.java
dmdashenkov: Can we call `Stringifiers.fromString` without invoking the `static` initialization block?

dmdashenkov:
@armiol, PTAL again.

server/src/main/java/org/spine3/server/stand/AggregateStateId.java
dmdashenkov: Fixed as discussed verbally.

server/src/main/java/org/spine3/server/stand/AggregateStateId.java
dmdashenkov: Extracted into `StandStorage` as discussed verbally.

server/src/main/java/org/spine3/server/stand/AggregateStateIdStringifier.java
```diff
@@ -0,0 +1,181 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.base.Identifiers.checkSupported;
+import static org.spine3.util.Exceptions.wrappedCause;
+
+/**
+ * A {@link Stringifier} for the {@link AggregateStateId}.
+ *
+ * <p>The protocol of strigification is as follows:
+ * {@code {@linkplain TypeUrl}_of_the_state_type}-{@code ID_TYPE_NAME_or_SHORT_REPR_FOR_PRIMITIVES}-{@code generic_ID_stringifier_output}
```
armiol: Can we please make it more readable? A short sample usage instead of this line would do.

server/src/main/java/org/spine3/server/stand/AggregateStateIdStringifier.java
```diff
@@ -0,0 +1,181 @@
...
+
+    private static Class idTypeFromString(String idTypeString) {
+        final Class type = IdType.getTypeFrom(idTypeString);
+        return type;
+    }
+
+    /**
+     * An {@code enum} of allowed ID types.
+     */
+    private enum IdType {
```
armiol: As discussed vocally, let's avoid duplicating the enum of supported ID types (see `Identifier.Type`).

dmdashenkov:
@armiol, PTAL.

server/src/test/java/org/spine3/server/stand/AggregateStateIdStringifierShould.java
```diff
@@ -0,0 +1,186 @@
...
+
+        assertNotNull(id);
+        assertEquals(ANY_TYPE_URL, id.getStateType());
+        assertEquals(messageId, id.getAggregateId());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void fail_to_convert_invalid_string() {
+        final String invalidId = "I'm invalid!";
+        stringifier()
```
armiol: Please format.

server/src/main/java/org/spine3/server/stand/AggregateStateIdStringifier.java
```diff
@@ -0,0 +1,122 @@
...
+                                                                   .append(DIVIDER)
+                                                                   .append(idTypeString)
+                                                                   .append(DIVIDER)
+                                                                   .append(genericIdString)
+                                                                   .toString();
+        return result;
+    }
+
+    @Override
+    protected AggregateStateId fromString(String s) {
```
armiol: Not all of the negative cases seem to be tested. Please re-check.

dmdashenkov: Fixed.

dmdashenkov:
@armiol, PTAL again.

PR #424, opened 2017-04-07
Fix Javadoc error.

dmdashenkov:
Removed a link to a nonexisting method from the Javadoc.


dmdashenkov:
@armiol, PTAL.
Maybe we should add Javadoc generation to the CI build script?
There are some other older errors when running `./gradlew javadoc`.

PR #436, opened 2017-04-18
Stringifiers for well known types

alexander-yevsyukov:
### Notable changes
* Stringifiers were moved into the `string` package.
* Classes from the `time` package into the `client` module for further separation of this code into the Spine.Time library.
* Added boolean stringifier.
* Added stringifiers for date/time data structures from the `time` package.
* `ExperimentalApi` annotation is introduced.

### Changes in Spine Time
* The `Interval` type and `Intervals` utility class are annotated as experimental.
* Added `toString()` and `parse() methods for all data types except `Interval`.
* Constants and time providers were moved into the `time.Time` class.
* Math methods `plus/minus` were renamed to `add/subtract`.
* `changes.Time` was renamed to `changes.TimeChanges`.

### Other changes
* Added `Exceptions.illegalArgumentWithCauseOf()`.
* `Exceptions.wrappedCause()` is renamed to `illegalStateWithCauseOf()`.
* Utility classes made `final`.


alexander-yevsyukov:
@armiol, there's one test that's failing at _only_ at the Travis env. (as we discussed over the phone call tonight). The test will be turned on back once we have ability to get the test log.

Meanwhile, PTAL.

client/src/main/java/org/spine3/time/Formats.java
```diff
@@ -0,0 +1,211 @@
...
+            return format("%1$09d", nanos);
+        }
+    }
+
+    /**
+     * Appends the string with zone offset info to the passed string builder.
+     */
+    static void appendZoneOffset(StringBuilder builder, ZoneOffset offset) {
+        if (offset.getAmountSeconds() == 0) {
+            builder.append(UTC_ZONE_SIGN);
```
armiol: This line (or `OffsetDateTimes.create()` method) seems to be a reason of the failing tests. 

The thing is that in the original `ZoneOffset` the ID is cleared (see `AbstractZonedTimeTest.convert_values_at_current_time_zone()`). So its representation in `now()`has empty "offset":
```
Now is: date { year: 2017 month: APRIL day: 24 } time { hours: 12 minutes: 11 seconds: 18 millis: 197 } offset { } 
```

But in this method we always append `Z` if the "seconds" amount is zero. So in its string representation we get this:
```
String is: 2017-04-24T12:11:18.197Z 
```
Parsed, it looks like this:
```
Parsed is: date { year: 2017 month: APRIL day: 24 } time { hours: 12 minutes: 11 seconds: 18 millis: 197 } offset { id: "UTC" } 
```

We need to make the behaviour consistent. The best fix would be to modify `OffsetDateTimes.create()` to transform the given `ZoneOffset` into `UTC`, if it is empty.

client/src/main/java/org/spine3/time/OffsetDateTimes.java
```diff
@@ -0,0 +1,372 @@
...
+     */
+    public static OffsetDateTime of(LocalDate date, LocalTime time, ZoneOffset offset) {
+        return create(date, time, offset);
+    }
+
+    private static OffsetDateTime create(LocalDate date, LocalTime time, ZoneOffset offset) {
+        return OffsetDateTime.newBuilder()
+                             .setDate(date)
+                             .setTime(time)
+                             .setOffset(offset)
```
armiol: See my comment above. I don't think that's what we want to set right away, but rather analyze and adjust before setting the `offset` value.

client/src/main/java/org/spine3/time/Time.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Tests whether the passed year is a leap one.
+     *
+     * @return {@code true} for a leap year, {@code false} otherwise
+     */
+    @SuppressWarnings("MagicNumber") // The number is part of leap year calc.
+    public static boolean isLeapYear(int year) {
+        return year % 4 == 0 && year % 100 != 0 || year % 400 == 0;
```
armiol: I'd use a [solution,](http://stackoverflow.com/questions/1021324/java-code-for-calculating-leap-year/1021373#1021373) based upon JDK.

client/src/main/java/org/spine3/time/Formats.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private Formats() {
+        // Prevent instantiation of this utility class.
+    }
+
+    /**
+     * Creates format for local date.
+     */
+    private static DateFormat createDateFormat(TimeZone timeZone) {
+        final SimpleDateFormat sdf = createDigitOnlyFormat(DATE_FORMAT);
```
alexander-yevsyukov: We cannot reuse `DateFormat`s between threads. That's why they are accessed via `ThreadLocal`s.

client/src/main/java/org/spine3/time/Time.java
alexander-yevsyukov: Assuming we use this method only for validation of input parameters, I'd rather use a faster version. `Calendar.getInstance()` from the suggested solution is a bit heavy for the task.

alexander-yevsyukov:
@armiol, PTAL.

armiol:
@alexander-yevsyukov LGTM.

PR #437, opened 2017-04-19
Validation of Stand input

armiol:
This PR brings the automatic validation of actor requests sent to `Stand` instances.

Each request `Message` is validated in two steps.
* It has to conform the constraints set in its `Message` definition.
* It has to be supported by the `Stand` instance. In particular:
  * the `TypeUrl` of `Query.getTarget()` must be exposed for reading;
  * the `TypeUrl` of `Topic.getTarget()` must be exposed for reading;
  * the `Subscription` must be present in the subscription registry for it to activate or cancel.

In addition, the public API of `Stand` has been modified to send responses via  `StreamObserver`s passed as method arguments:
* `Stand.subscribe(Topic, StreamObserver<Subscription>)`;
*  `Stand.activate(Subscription, EntityUpdateCallback, StreamObserver<Response>)`;
* `Stand.cancel(Subscription, StreamObserver<Response>)`;

The validation errors are passed to the `StreamObserver.onError()` callback in a unified way.


armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/org/spine3/server/stand/QueryValidator.java
```diff
@@ -0,0 +1,81 @@
...
+            return Optional.absent();
+        }
+
+        return Optional.of(missingInRegistry(getTypeOf(target)));
+    }
+
+    private static RequestNotSupported<Query> missingInRegistry(TypeUrl topicTargetType) {
+        final String errorMessage = format("The query target type is not supported: %s",
+                                           topicTargetType.getTypeName());
+        return new RequestNotSupported<Query>(
```
alexander-yevsyukov: Please see if we can fit into one line here and below.

PR #438, opened 2017-04-19
Expose `FailureBus` from the `BoundedContext` instance 

armiol:
Before this PR it was very hard to register custom `FailureDispatcher`s in an application, since the `FailureBus` instance was private to `CommandBus`.

As of these changes it became possible as simple as this:
```
final BoundedContext bc = ...;

bc.getFailureBus().register(myCustomFailureDispatcher);
```

Also, the `BoundedContext` now exposes all types of `Bus`es, i.e. `EventBus`, `CommandBus`, and now `FailureBus`, bringing more consistency into API.


armiol:
@alexander-yevsyukov PTAL.

Issue #400, opened 2017-03-27
Entity Columns for Aggregates


dmdashenkov:
Currently, [`Entity Columns feature`](https://github.com/SpineEventEngine/core-java/pull/398) is available only for `Projection`s. It's required to add support for `Aggregate`s (in `StandStorage`).

For more info, see `Stand.update()`.

dmdashenkov:
This should be implemented at least partially for `archived` and `deleted` columns, which are present effectively at any entity type.

This change should eliminate double-faced nature of the `RecordStorage` API: currently, it has two ways of writing a record, with and without entity columns.

armiol:
Addressed in `2.0` snapshots.

PR #440, opened 2017-04-24
Entity Column querying

dmdashenkov:
Support for the Entity Column queries to a `RecordBasedRepository`.

With this PR, a new class is intoduced: __`EntityQuery`__. This is a DTO, which is transferred from a `RecordBasedRepository` to the underlying `RecordStorage`.

The `EntityQuery` is a Java `Object`-based representation of the `EntityFilters` message. It contains the main filters for an `EntityRecordWithColumns`:
 - `EntityIdFilter` with a list of requested IDs
 - A collection of `FieldFilters` for matching the Entity Columns.

Example:
```java
// --CustomerProjection.java--

public PersonName getContactName() {
    return getState().getContactPerson().getName();
}
```
Then in your client code:
```java
// --Retrieve the customer who is presented by the given person--

// Get the query parameters (e.g. from UI)
final PersonName contactNameFromUi = getCustomerContactPersonName();

// Construct the Query `instance` 
final Query customerQuery = myRequestFactory.query()
                                            .select(Customer.class)
                                            .withMask("id", "address", "domain")
                                            .where(eq("contactName", contactNameFromUi))
                                            .build();
// Get the client stub of the `QueryService`
final QueryService queryService = getMyServerApi();
// Execute the request
queryService.read(query, myCallback);
```
In the example above all the entities retrieved from the request match the given query.

_Note:_ the `StandStorage` [does not support](https://github.com/SpineEventEngine/core-java/issues/400) the querying by Entity Columns feature yet.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/entity/storage/ColumnTypeRegistry.java
```diff
@@ -117,7 +117,9 @@ public C get(Column<?> field) {
 
     @SuppressWarnings("unchecked")
         // Unchecked copying from the src instance
-        // Never leads to a failure, since checke while writing into the instance itself
+        // Never leads to a failure, since checked while writing into the instan
+    //
+    // ce itself
```
armiol: TODO: fix.

server/src/main/java/org/spine3/server/entity/storage/Columns.java
```diff
@@ -118,11 +118,43 @@ private Columns() {
...
+    /**
+     * Generates the {@link Column Column} values considering the passed
+     * {@linkplain Entity entity type} indexed.
      *
      * @param entityType indexed type of the {@link Entity}
-     * @param entity the object which to take the values from
+     * @param entity     the object which to take the values from
      * @return a {@link Map} of the {@link Column Columns}
      */
     private static Map<String, Column.MemoizedValue<?>> getStorageFields(
```
armiol: BTW, this method should be renamed according to its definition above. And we don't have `Storage Fields` anymore.

server/src/main/java/org/spine3/server/entity/storage/EntityQuery.java
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * <p>The query contains the acceptable values of the record IDs and the
+ * {@linkplain Column Entity Columns}.
+ *
+ * <p>A storage may ignore the query or throw an exception if it's specified (see
+ * {@link org.spine3.server.stand.StandStorage StandSotrage}). By default, a
+ * {@link org.spine3.server.storage.RecordStorage RecordStorage} supports the Entity queries.
+ *
+ * <p>If the {@linkplain EntityQuery#getIdFilter()} accepted IDs list} is empty, all the IDs are
+ * considered as queried.
```
armiol: Please remove `as`.

server/src/main/java/org/spine3/server/entity/storage/EntityQuery.java
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * <p>Acts like a DTO between the
+ * {@linkplain org.spine3.server.entity.RecordBasedRepository repository} and the
+ * {@link org.spine3.server.storage.RecordStorage RecordStorage}.
+ *
+ * <p>The query contains the acceptable values of the record IDs and the
+ * {@linkplain Column Entity Columns}.
+ *
+ * <p>A storage may ignore the query or throw an exception if it's specified (see
+ * {@link org.spine3.server.stand.StandStorage StandSotrage}). By default, a
```
armiol: Please remove ` a`.

server/src/main/java/org/spine3/server/entity/storage/EntityQuery.java
```diff
@@ -0,0 +1,127 @@
...
+ * the record is considered <b>not matching</b>.
+ *
+ * <p>If both the {@linkplain EntityQuery#getIdFilter()} accepted IDs list} and
+ * {@linkplain EntityQuery#getParameters() query parameters} are empty, all the records are
+ * considered matching.
+ *
+ * @author Dmytro Dashenkov
+ * @see EntityRecordWithColumns
+ */
+public final class EntityQuery {
```
armiol: Looks `@Internal`.

server/src/main/java/org/spine3/server/entity/storage/EntityQueries.java
```diff
@@ -0,0 +1,76 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A factory for generating the {@link EntityQuery} instances.
+ *
+ * @author Dmytro Dashenkov
+ * @see EntityQuery
+ */
+public final class EntityQueries {
```
armiol: Looks `@Internal`.

server/src/main/java/org/spine3/server/entity/storage/EntityQueries.java
```diff
@@ -0,0 +1,76 @@
...
+import org.spine3.client.EntityFilters;
+import org.spine3.client.EntityIdFilter;
+import org.spine3.server.entity.Entity;
+
+import java.util.Collection;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A factory for generating the {@link EntityQuery} instances.
```
armiol: I think, that's just a `utility class for working with {@link EntityQuery} instances`.

server/src/main/java/org/spine3/server/entity/storage/EntityQueryMatcher.java
```diff
@@ -0,0 +1,111 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Collections2.transform;
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * A {@link Predicate} on the {@link EntityRecordWithColumns} matching it upon the given
+ * {@link EntityQuery}.
+ *
+ * <p>See {@link EntityQuery} for the matching contract
```
armiol: Please format with `@see`. And pay attention to the dots at the end: in this case it is required, but in `@see` case you'll have to deal with it.

server/src/main/java/org/spine3/server/entity/storage/EntityQueryMatcher.java
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @author Dmytro Dashenkov
+ */
+@Internal
+public final class EntityQueryMatcher implements Predicate<EntityRecordWithColumns> {
```
armiol: That doesn't look `final`, as you would want to extend it in `GAE`-based implementation. Let's discuss if needed.

server/src/main/java/org/spine3/server/entity/storage/ProtoToJavaMapper.java
```diff
@@ -0,0 +1,246 @@
...
+ *     <li>Java primitives - the passed {@link Any} is unpacked into one of types {@code Int32Value,
+ *     Int64Value, UInt32Value, UInt64Value, FloatValue, DoubleValue, BoolValue, StringValue,
+ *     BytesValue} and then into the corresponding Java type, either a primitive value, or
+ *     {@code String} of {@link ByteString}.
+ * </ul>
+ *
+ * @author Dmytro Dashenkov
+ */
+// TODO:2017-04-21:dmytro.dashenkov: Maybe we should move this to client code to make this util reusable.
+final class ProtoToJavaMapper {
```
armiol: That belongs to `client` / `org.spine3.protobuf`. And then it should probably be `@Internal` and get used only from `core-java` and `tools` projects.

server/src/main/java/org/spine3/server/storage/memory/EntityRecordUnPacker.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * The {@link Function} converting the {@link EntityRecordWithColumns} into {@link EntityRecord} by
+ * calling {@link EntityRecordWithColumns#getRecord()}.
+ *
+ * <p>The {@code null} values are mapped into {@code null} values of type {@code EntityRecord}.
+ *
+ * @author Dmytro Dashenkov
+ */
+enum EntityRecordUnPacker implements Function<EntityRecordWithColumns, EntityRecord> {
```
armiol: `...Unpacker`.

server/src/test/java/org/spine3/server/entity/TestEntityWithStringColumn.java
```diff
@@ -0,0 +1,28 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.entity;
+
+/**
+ * @author Dmytro Dashenkov
+ */
+public interface TestEntityWithStringColumn {
```
armiol: Let's discuss whether we could hide it from `public` (move it to `Given` — which is ideal — or remove `public`) and write some docs for it.

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
```diff
@@ -407,5 +484,13 @@ public Timestamp getNow() {
         public Project getCounterState() {
             return getState();
         }
+
+        public int getInjectableState() {
+            return injectableState;
+        }
+
+        private void injectInjectableState(int state) {
```
armiol: This isn't good. 

I don't really understand why you would need some `int` field anyway. Couldn't you reuse the `Project`-typed state and use `updateState()` if you really need to do that?

armiol:
@dmdashenkov please fix the typos in the PR description.

Also, please add some code sample: a simple `Entity` with a custom column defined, a sample Query to it and the expected results.

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
dmdashenkov: Let's keep this field. We should consider both `Message` and proto-primitive fields in the queries. Using this `Entity`, we test both via the `version` and this field.

server/src/main/java/org/spine3/server/entity/storage/EntityQuery.java
dmdashenkov: Actually, it is not internal. The updated `RecordStorage` API accepts this type as a parameter for the by-field read.

server/src/main/java/org/spine3/server/entity/storage/EntityQueryMatcher.java
dmdashenkov: I hope to avoid using this predicate, as well as any other `Predicate`, in the `GAE` Storage implementation. The point is to perform the queries using the database "native" means.

And if we would consider filtering the records in the memory in other Storage impls., we would probably have a single behavior for all of them.

Thinking of the problems which may appear while trying to query the `GAE` Datastore, I assume that we _may_ want to use the in-memory filtering, but then we'll try to optimize it as mush as possible and there would be no sense in extending this particular `Predicate` implementation.

server/src/test/java/org/spine3/server/entity/TestEntityWithStringColumn.java
dmdashenkov: I'll add some documentation but we can neither move it to a `Given` class nor make is `package-private`.

dmdashenkov:
@armiol, PTAL again.

dmdashenkov:
@armiol, PTAL.

client/src/main/java/org/spine3/client/ActorRequestFactory.java
```diff
@@ -245,21 +302,348 @@ public Query allWithMask(Class<? extends Message> entityClass, String... maskPat
...
+                      .append(valueSeparator);
+                }
+            }
+            sb.append(");");
+            return sb.toString();
+        }
+    }
+
+    /**
+     * A parameter of a {@link Query} to the read side.
```
armiol: Please remove `to the read side`.

client/src/main/java/org/spine3/client/ActorRequestFactory.java
```diff
@@ -245,21 +302,348 @@ public Query allWithMask(Class<? extends Message> entityClass, String... maskPat
...
+     *
+     * <p>This class may be considered as a filter for the query. An instance contains the name of
```
armiol: Please read [this](https://www.englishforums.com/English/ToBeConsideredAs/bbcmjr/post.htm), as that's a common mistake in the JavaDocs you create.

client/src/main/java/org/spine3/client/Targets.java
```diff
@@ -56,7 +59,46 @@ public static Target someOf(Class<? extends Message> entityClass,
...
+     * @param columnFilters the Entity Columns filters serving as the predicate of the constructed
+     *                      {@link Target}
+     * @return the instance of {@code Target} assembled according to the parameters.
+     */
+    public static Target someOf(Class<? extends Message> entityClass,
+                                Set<? extends Message> ids,
+                                Map<String, Any> columnFilters) {
+        checkNotNull(entityClass);
+        checkNotNull(ids);
+        checkNotNull(columnFilters);
```
armiol: This line is reported to be untested.

client/src/main/java/org/spine3/client/Targets.java
```diff
@@ -56,7 +59,46 @@ public static Target someOf(Class<? extends Message> entityClass,
...
+
+        final Target result = composeTarget(entityClass, ids, columnFilters);
```
armiol: This expression is reported to be uncovered with tests as well.

client/src/main/java/org/spine3/client/Targets.java
```diff
@@ -56,7 +59,46 @@ public static Target someOf(Class<? extends Message> entityClass,
...
+     *
+     * @param entityClass   the class of a target entity
+     * @param columnFilters the Entity Columns filters serving as the predicate of the constructed
+     *                      {@link Target}
+     * @return the instance of {@code Target} assembled according to the parameters.
+     */
+    public static Target someOf(Class<? extends Message> entityClass,
+                                Map<String, Any> columnFilters) {
+        checkNotNull(entityClass);
+        checkNotNull(columnFilters);
```
armiol: Same as above.

client/src/main/java/org/spine3/protobuf/ProtoJavaMapper.java
```diff
@@ -0,0 +1,366 @@
...
+ *         a primitive value, or {@code String} of {@link ByteString}.
+ * </ul>
+ *
+ * @author Dmytro Dashenkov
+ */
+@Internal
+public final class ProtoJavaMapper {
+
+    private ProtoJavaMapper() {
+        // Prevent utility class initialization
```
armiol: Dot at the end is missing.

client/src/main/java/org/spine3/protobuf/ProtoJavaMapper.java
```diff
@@ -0,0 +1,366 @@
...
+/**
+ * A utility for converting the {@linkplain Message Protobuf Messages} (in form of {@link Any}) into
+ * arbitrary {@linkplain Object Java Objects} and vice versa.
+ *
+ * <p>Currently, the supported types are:
+ * <ul>
+ *     <li>{@link Message} - converted via {@link AnyPacker};
+ *     <li>Java primitives - the passed {@link Any} is unpacked into one of the types
+ *         {@code Int32Value, Int64Value, UInt32Value, UInt64Value, FloatValue, DoubleValue,
+ *         BoolValue, StringValue, BytesValue} and then into the corresponding Java type, either
```
armiol: `...and then transformed...`

client/src/main/java/org/spine3/protobuf/ProtoJavaMapper.java
```diff
@@ -0,0 +1,366 @@
...
+    /**
+     * Performs the {@link Object} to {@link Any} mapping.
+     *
+     * @param value the {@link Object} value to convert
+     * @param <T>   the converted object type
+     * @return the packed value
+     */
+    public static <T> Any map(T value) {
+        checkNotNull(value);
+        @SuppressWarnings("unchecked") // Must be checked in runtime
```
armiol: `at runtime`. Please review the similar cases for the same issue.

client/src/main/java/org/spine3/protobuf/ProtoJavaMapper.java
```diff
@@ -0,0 +1,366 @@
...
+            checkArgument(converter != null,
+                          "Could not find a wrapper type for %s.",
+                          cls.getCanonicalName());
+            final Message result = converter.reverse()
+                                            .convert(input);
+            return result;
+        }
+    }
+
+    abstract static class PrimitiveHandler<M extends Message, T> extends Converter<M, T> {
```
armiol: Requires documentation.

Also, it is declared as `package-private`, but I don't see why it cannot be `private`. Perhaps, `@VisibleForTesting` is missing?

client/src/main/proto/spine/net/url.proto
```diff
@@ -61,6 +61,8 @@ message Url {
 
         // Most often used application layer protocols.
         enum Schema {
+            // TODO:2017-04-20:dmytro.dashenkov: Check the enum values.
```
armiol: Please be more specific in this `TODO` item, or resolve it before merging.

server/src/test/java/org/spine3/server/storage/RecordStorageShould.java
```diff
@@ -357,6 +362,69 @@ public void write_record_with_columns() {
...
+
+        // 2 of 3 have required values
+        matchingEntity.setStatus(requiredValue);
+        wrongEntity1.setStatus(requiredValue);
+        wrongEntity2.setStatus(Project.Status.CANCELLED);
+
+        // Change internal Entity state
+        wrongEntity1.getCounter();
+
+        // After the mutation above the single matching record is record under the ID `idMatching`
```
armiol: `record is record` needs to be fixed.

Also, the ordering should be `under the 'idMatching' ID`.

client/src/main/java/org/spine3/client/ActorRequestFactory.java
dmdashenkov: "to consider as" [has](https://ell.stackexchange.com/questions/11226/is-considered-always-followed-by-as) a bit different meaning. 
Usually, this __is__ a mistake, but not in this case.

client/src/main/java/org/spine3/client/ActorRequestFactory.java
```diff
@@ -245,21 +302,348 @@ public Query allWithMask(Class<? extends Message> entityClass, String... maskPat
...
+     * any of its overloads.
+     *
+     * <p>Usage example:
+     * <pre>
+     *     {@code
+     *     final Query query = factory().query()
+     *                                  .select(Customer.class)
+     *                                  .fields("name", "address", "email")
+     *                                  .whereIdIn(getWestCostCustomersIds())
+     *                                  .where({@link QueryParameter#eq(String, Object) eq}("type", "permanent"),
```
dmdashenkov: This works fine, but we still cannot build Javadoc due to multiple errors (18).

dmdashenkov:
@armiol, PTAL again.

dmdashenkov:
@armiol, PTAL again.

client/src/main/java/org/spine3/protobuf/ProtoJavaMapper.java
```diff
@@ -0,0 +1,394 @@
...
+ *     <li>Java primitives - the passed {@link Any} is unpacked into one of the types
+ *         {@code Int32Value, Int64Value, UInt32Value, UInt64Value, FloatValue, DoubleValue,
+ *         BoolValue, StringValue, BytesValue} and then transformed into the corresponding Java
+ *         type, either a primitive value, or {@code String} of {@link ByteString}.
+ * </ul>
+ *
+ * @author Dmytro Dashenkov
+ */
+@Internal
+public final class ProtoJavaMapper {
```
armiol: As long as this class belongs to `org.spine3.protobuf`, it should be named `TypeConverter`. Another reason is to avoid "map", which isn't quite what we do. The documentation should be updated accordingly.

client/src/main/java/org/spine3/client/QueryBuilder.java
```diff
@@ -0,0 +1,307 @@
...
+ *                                  .where({@link QueryParameter#eq eq}("type", "permanent"),
+ *                                         eq("discountPercent", 10),
+ *                                         eq("companySize", Company.Size.SMALL))
+ *                                  .build();
+ *     }
+ * </pre>
+ *
+ * @see QueryFactory#select(Class) for the intialization
+ */
+public final class QueryBuilder {
```
dmdashenkov: Done (see `QueryFactoryShould`).

client/src/test/java/org/spine3/client/QueryParameterShould.java
```diff
@@ -0,0 +1,70 @@
...
+
+import static java.lang.String.*;
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.Assert.assertThat;
+import static org.spine3.client.QueryParameter.eq;
+
+/**
+ * @author Dmytro Dashenkov
+ */
+public class QueryParameterShould {
```
dmdashenkov: Done.

dmdashenkov:
@armiol, PTAL.

client/src/main/java/org/spine3/client/QueryBuilder.java
```diff
@@ -0,0 +1,309 @@
...
+
+/**
+ * A builder for the {@link Query} instances.
+ *
+ * <p>The API of this class is inspired by the SQL syntax.
+ *
+ * <p>Calling any of the methods is optional. Call {@link #build() build()} to retrieve
+ * the instance of {@link Query}.
+ *
+ * <p>Calling of any of the builder methods overrides the previous call of the given method of
```
armiol: Please rewrite this sentence. It has two flaws: 

* `of any of` (which is a grammar issue);
* it is totally unclear, but gives a __very__ short and precise definition.

Let's fix the grammar and become closer to people. Several sentences, examples etc — anything that helps explaining the same idea, but much simpler, will do.

client/src/main/java/org/spine3/client/QueryParameter.java
```diff
@@ -0,0 +1,107 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.protobuf.AnyPacker.unpack;
+
+/**
+ * A parameter of a {@link Query}.
+ *
+ * <p>This class may be considered a filter for the query. An instance contains the name of
+ * the Entity Column to filter by and the value of the Column.
+ *
+ * <p>The supported types for querying are {@link Message} and Protobuf primitives.
```
armiol: `{@linkplain Message Message instances}`.

It also would be good to have an actual link to the list of Proto primitives that they have and we support.

client/src/main/java/org/spine3/client/TopicFactory.java
```diff
@@ -0,0 +1,95 @@
...
+import org.spine3.annotations.Internal;
+import org.spine3.base.ActorContext;
+
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.spine3.client.Targets.composeTarget;
+
+/**
+ * Public API for creating {@link Topic} instances, using the {@code ActorRequestFactory}
```
armiol: Let's add `@see ActorRequestFactory.topic()` here and similar `@see ...` statements in `CommandFactory` and `QueryFactory`.

client/src/main/java/org/spine3/client/CommandFactory.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Public API for creating {@link Command} instances, using the {@code ActorRequestFactory}
+ * configuration.
+ */
+public final class CommandFactory {
+
+    private final ActorRequestFactory actorRequestFactory;
+
+    CommandFactory(ActorRequestFactory actorRequestFactory) {
```
armiol: It's weird, that `QueryFactory` and `TopicFactory` take `ActorContext`, and this one takes `ActorRequestFactory` as a param.

Let's make it consistent.

client/src/main/java/org/spine3/client/QueryParameter.java
dmdashenkov: It's not `instances`, but the `types`.

Done for the second part of your comment.

client/src/main/java/org/spine3/client/TopicFactory.java
dmdashenkov: Done.

dmdashenkov:
@armiol, PTAL again.

server/src/test/java/org/spine3/server/entity/TestEntityWithStringColumn.java
```diff
@@ -0,0 +1,34 @@
...
+package org.spine3.server.entity;
+
+/**
+ * The contract for the test {@linkplain Entity entities} which serve for testing the subclasses of
+ * {@link RecordBasedRepository}.
+ *
+ * @author Dmytro Dashenkov
+ * @see RecordBasedRepositoryShould
+ */
+public interface TestEntityWithStringColumn {
```
dmdashenkov: It should be `public` since it's implemented in the test `Entity` types used e.g. for `ProcessManager` testing, which involves another package.

client/src/main/java/org/spine3/client/QueryBuilder.java
```diff
@@ -0,0 +1,325 @@
...
+ * </pre>
+ *
+ * @see QueryFactory#select(Class)
+ */
+public final class QueryBuilder {
+
+    private final QueryFactory queryFactory;
+    private final Class<? extends Message> targetType;
+
+    // All the optional fields are initialized only when and if set
```
dmdashenkov: I thought we've agreed not to use Javadoc blocks in the plain comments.

client/src/main/java/org/spine3/client/QueryBuilder.java
```diff
@@ -0,0 +1,325 @@
...
+        this.targetType = checkNotNull(targetType);
+        this.queryFactory = checkNotNull(queryFactory);
+    }
+
+    /**
+     * Sets the ID predicate to the {@linkplain Query}.
+     *
+     * <p>Though it's not prohibited at compile-time, please make sure to pass instances of the
+     * same type to the argument of this method. Moreover, the instances must be of the type of
+     * the query target type identifier. This method or any of its overload do not check these
```
dmdashenkov: I don't see how a type parameter could help here. A user can anyway call `byId(Collection<Object>)` and there is no adequate way to check the ID types.

I think we may rely on the user here since this restriction corresponds to common sense understanding of how do the queries work.

server/src/main/java/org/spine3/server/stand/StandStorage.java
```diff
@@ -69,4 +72,24 @@ protected StandStorage(boolean multitenant) {
      */
     public abstract ImmutableCollection<EntityRecord> readAllByType(TypeUrl type,
                                                                     FieldMask fieldMask);
+
+    /**
+     * The {@code StandStorage} is specified not to satisfy the {@link EntityQuery}.
```
dmdashenkov: As stated, the `readAll(EntityQuery, FieldMask)` operation is not supported in the `StandStorage` since there are no Entity Columns in `StandStorage`(#400). Instead, this method calls `readAll(FieldMask)`. 

That's why the `{@inheritDoc}` is omitted in this case (since we totally override the behavior).

As soon as `StandStorage` supports Entity Columns, this particular issue will also be resolved.

client/src/main/java/org/spine3/client/QueryFactory.java
```diff
@@ -0,0 +1,192 @@
...
+                                             .addAllPaths(Arrays.asList(maskPaths))
+                                             .build();
+        final Query result = composeQuery(entityClass, ids, null, fieldMask);
+        return result;
+    }
+
+    /**
+     * Creates a {@link Query} to read certain entity states by IDs.
+     *
+     * <p>Allows to specify a set of identifiers to be used during the {@code Query} processing.
```
dmdashenkov: Done for this particular case. When talking about the results of the query we're currently building, let's prefer Future Simple if possible.

server/src/main/java/org/spine3/server/entity/storage/EntityQueries.java
```diff
@@ -0,0 +1,94 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A utility class for working with {@link EntityQuery} instances.
+ *
+ * @author Dmytro Dashenkov
+ * @see EntityQuery
+ */
+@Internal
+public final class EntityQueries {
```
dmdashenkov: The only production code user of this class is `RecordBasedRepository`. We can move the class into the same package as it's user, but, as for me, it's rather connected to the `entity.storage`, not to the `entity` itself.

dmdashenkov:
@armiol, please see my answers and PTAL at the code changes again.

server/src/main/java/org/spine3/server/stand/StandStorage.java
armiol: As discussed vocally, let's modify this Javadoc section, so that it has the general description, the equivalence statement you have below. And only then this thought, but explained in much more words, so that other people understand what it means.

server/src/main/java/org/spine3/server/entity/storage/EntityQueries.java
armiol: `it's user` should be `its user`. That's a common mistake in English world, please pay attention.

Let's leave it where it is now.

PR #446, opened 2017-04-26
Rename ID value fields

alexander-yevsyukov:
This PR renames the fields named `uuid` into `value` to reflect the idea of having type-based prefixes.

The PR also renames generation methods to `generateId()` and move them into corresponding utility classes where appropriate.


alexander-yevsyukov:
@armiol, PTAL.

PR #450, opened 2017-04-27
Command and Event owns their IDs

alexander-yevsyukov:
This PR:
* removes the `CommandContext.command_id` property, and introduces `Command.id` property instead.
* removes `EventContext.event_id` property, and introduces `Event.id` property instead.
* uses `CommandEnvelope` (instead of a couple of command message with context) for dispatching a command in a handling entity.


alexander-yevsyukov:
@armiol, PTAL.

alexander-yevsyukov:
@armiol, PTAL at the event ID related changes.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -303,15 +303,15 @@ void play(AggregateStateRecord aggregateStateRecord) {
      * Applies event messages.
      *
      * @param eventMessages the event message to apply
-     * @param commandContext the context of the command, execution of which produces
+     * @param envelope the context of the command, execution of which produces
```
armiol: The description of this parameter does not match.

server/src/main/java/org/spine3/server/aggregate/Aggregate.java
```diff
@@ -321,14 +321,14 @@ private void apply(Iterable<? extends Message> eventMessages, CommandContext com
      * Applies the passed event message or {@code Event} to the aggregate.
      *
      * @param eventMessages  the event message to apply
-     * @param commandContext the context of the command which handler generated the message or event
+     * @param envelope the context of the command which handler generated the message or event
```
armiol: Same as above.

PR #452, opened 2017-04-27
Failure owns ID

alexander-yevsyukov:
This PR moves the id of a failure into the `Failure` type.


alexander-yevsyukov:
@armiol, PTAL. This one is easy.

PR #453, opened 2017-04-28
Test factory fixes

dmdashenkov:
Fixed an issue in the `TestEventFactory` which was generating events on the same `CommandId` (`"NULL"`).


dmdashenkov:
@armiol, PTAL.

PR #455, opened 2017-04-28
Command scheduling as a system property

alexander-yevsyukov:
This PR:
 * Moves the `scheduling_time` field into `Command.SystemProperties`.
 * Marks the whole type `SystemProperties` as `internal_type` (which implies that all the fields are `internal`).
 * Marks the `Command.system_properties` field as `internal`.


alexander-yevsyukov:
@armiol, PTAL.

PR #456, opened 2017-04-30
Bump the framework version to 0.9.0.

armiol:
As long as this PR is merged to `master`, we should create a tag with the same version number. Later, this tag is used as an anchor for the GitHub release.


armiol:
@alexander-yevsyukov PTAL.

PR #457, opened 2017-05-03
Projection Repo Catch Up performance

dmdashenkov:
The Catch-Up process in the `ProjectionRepository` uses `Repository.iterator`.

Currently, `Repository.iterator` uses `Storage.index` to get the instances of desired entities. The issue is that the Storage index is, in fact, an `Iterator` of the Storage IDs. When the Repository iterates over the index it first reads all the entities from the Storage and transforms them into the ID and then iterates over these IDs and fetches each entity separately.

The fix is following: the `RecordBasedRepository` overrides the `iterator` method implementation by using the advantage of `RecordStorage.readAll` operation.

This fix should get version `0.8.41-SNAPSHOT`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ buildscript {
         guavaVersion = '20.0'
         protobufGradlePluginVerison = '0.8.0'
         
-        spineVersion = '0.8.40-SNAPSHOT'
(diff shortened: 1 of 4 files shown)


dmdashenkov:
@armiol, PTAL.

Issue #458, opened 2017-05-04
Remove excessive logging of Storage Column extraction


armiol:
At the moment there are several built-in entities, which are processed during the Storage Field extraction. As they are not `public`, the extraction is never successful for those. However, the log always contains the entries such as
```
org.spine3.server.entity.storage.Columns from: Passed entity class org.spine3.server.commandstore.Entity is not public. Storage fields won't be extracted. (Columns.java:109)
...
org.spine3.server.entity.storage.Columns from: Passed entity class org.spine3.server.event.EventEntity is not public. Storage fields won't be extracted. (Columns.java:109)
```

Such a logging is irritating and not helpful to the users of the framework. Therefore it should be removed.

PR #460, opened 2017-05-04
Fix excessive logging while Column extraction

dmdashenkov:
This PR fixes the issue #458.

Now the Spine internal `Entity` implementations that are left non-public (`org.spine3.server.commandstore.Entity` and `org.spine3.server.event.EventEntity`) are not logged when trying to extract the `Column`s from.

All other Spine internal Entities are known to be `public`.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL again.

dmdashenkov:
@armiol, PTAL again.

PR #463, opened 2017-05-05
Entity visibility option

alexander-yevsyukov:
This PR:
* Merges all option-related proto files into `options.proto`. Options are generated into the package `options` (instead of `annotations`).
* Changes the option numbers to be in the range `73812-75000`. 
* Adds `EntityOption`.


alexander-yevsyukov:
@armiol, PTAL.

PR #465, opened 2017-05-06
Singular for package names

alexander-yevsyukov:
This PR renames `options` package into `option` and `annotations` into `annotation` to be inline with singular in `javax.annotation`.


alexander-yevsyukov:
@armiol, PTAL. I have not increased the version number because no usages are expected in the timeframe passed since `0.9.2-SNAPSHOT` introduction.

PR #466, opened 2017-05-06
`EventPlayingEntity`, transactions and `ValidatingBuilder`-based approach of event applying

armiol:
This PR brings massive changes to the `Entity` API.

**`EventPlayingEntity` and playing Events**

All the `Entity` implementations, which are able to handle events, are now descendants of `EventPlayingEntity`. The newly introduced API allows to play events in a unified fashion.

**State modification is now available via `ValidatingBuilder`**

In order to provide framework users with an ability to trace potential validation issues that may take place during the `EventPlayingEntity` state modification, the `getBuilder()` now returns a `ValidatingBuilder` instance. 

Its setter methods perform on-the-fly validation, raising constraint violation exceptions right away, allowing proper handling before the state is written to the storage.

**Transactions**

The direct state update, which was previously available to `Projection`s, `Aggregate`s and `ProcessManager`s via numerous methods (such as `updateState` and `incrementState`), is NOT available anymore. The designed way to modify the entity is via `getBuilder().set...` calls. 

In addition, automatic version management is introduced. For each applied `Event` or event `Message` the version is automatically incremented. Direct version modification is not exposed as public API anymore.

Each transaction holds the pending modifications made to entity state, version and lifecycle flags until it is committed. In case of any exception during the transaction propagation, the transaction is automatically rolled back, and no changes are applied to an actual entity. Only upon a successful commit the actual entity state and attributes are updated.

The transactions are automatically started and committed for the following entities:

* `Aggregate`: a transaction is started along with the event applying phase and committed after events are applied. For a list of events (a product of a command) a single transaction is created.
* `Projection`: a transaction is started and then committed for each event handled.
* `ProcessManager`: a transaction is started and committed for each command handled (so it's possible to modify the PM state in the command handler method). Also, the transaction is started and committed for each event handled.

For the framework core developers an internal API is available to start the transaction:

* `AggregateTransaction.start(...)` and `AggregateTransaction.startWith(...)`;
* `ProjectionTransaction.start(...)` and `ProjectionTransaction.startWith(...)`;
* `ProcManTransaction.start(...)` and `ProcManTransaction.startWith(...)`.

(whereas `startWith(...)` methods allows to inject the initial transaction state, which is handy for testing)

Also, `TransactionListener` concept is introduced for those starting transactions in order to be able to follow the transaction propagation stages.

See `org.spine3.server.entity.Transaction` and `org.spine3.server.entity.TransactionListener` for more details.

**Writing `Projection` and `ProcessManager` to storage**

In case there were no modifications made to the `Projection` or `ProcessManager` in scope of a transaction, no update of the record storage is done.

**`ValidatingBuilder` generation**

The generation of the validating builders is now enabled for all the Messages in the project. This is required in order to provide the builder implementations to use along with `EventPlayingEntity`.

**`ConstraintViolationThrowable` is now `RuntimeException`**

To avoid a confusion from the framework users, the `ConstraintViolationThrowable` is no longer a checked exception. However, it has NOT been renamed yet — to simplify this PR. A related issue is created: #482.

IMPORTANT: The projects built on top of Spine must enable the builder generation via 

```
// Spine model compiler in `gradle.build`.
modelCompiler {
    generateValidatingBuilders = true
}
```

The framework version has been increased to `0.9.6-SNAPSHOT`.


client/src/main/java/org/spine3/validate/ValidatingBuilder.java
```diff
@@ -40,16 +52,117 @@
...
+            return result;
+        }
+
+        // as the method names are the same, but methods are different.
+        @SuppressWarnings("DuplicateStringLiteralInspection")
+        public static Method getNewBuilderMethod(Class<? extends ValidatingBuilder<?, ?>> cls) {
+            try {
+                return cls.getMethod("newBuilder");
+            } catch (NoSuchMethodException e) {
+                throw illegalArgumentWithCauseOf(e);
```
armiol: This line is actually tested, though codecov thinks it's not. See `ValidatingBuilderShould.throw_exception_if_accessing_newBuilder_method_for_improperly_declared_builder`.

client/src/main/java/org/spine3/validate/ValidatingBuilders.java
```diff
@@ -0,0 +1,300 @@
...
+            final Method newBuilderMethod =
+                    ValidatingBuilder.TypeInfo.getNewBuilderMethod(builderClass);
+            final Object raw = newBuilderMethod.invoke(null);
+
+            // By convention, `newBuilder()` always returns instances of `B`.
+            @SuppressWarnings("unchecked")
+            final B builder = (B) raw;
+            return builder;
+        } catch (Exception e) {
+            throw illegalStateWithCauseOf(e);
```
armiol: This line is tested as well, while codecov still marks it untested. See `ValidatingBuildersShould.propagate_exceptions_occurred_in_newBuilder_as_ISE`.

armiol:
@alexander-yevsyukov PTAL.

client/src/main/java/org/spine3/validate/AbstractValidatingBuilder.java
```diff
@@ -67,9 +112,77 @@
...
+    @Internal
+    public T internalBuild() {
+        @SuppressWarnings("unchecked")  // OK, as real types of `B`
+                                        // are always generated to be compatible with `T`.
+        final T result = (T) getMessageBuilder().build();
+        return result;
+    }
+
+    private B createBuilder() {
+
```
alexander-yevsyukov: Consider removing this empty line.

server/src/test/java/org/spine3/server/entity/TransactionShould.java
```diff
@@ -0,0 +1,356 @@
...
+        final TransactionListener<I, E, S ,B> listener = mock(TransactionListener.class);
+        final E entity = createEntity();
+        final Transaction<I, E, S ,B> tx = createTxWithListener(entity, listener);
+        final Event event = createEvent(createEventMessage());
+
+        verifyZeroInteractions(listener);
+        applyEvent(tx, event);
+
+        verify(listener).onAfterPhase(argThat(matchesSuccessfulPhaseFor(event)));
+
```
alexander-yevsyukov: Please remove the empty line here.

client/src/main/java/org/spine3/validate/ValidatingBuilder.java
```diff
@@ -40,16 +52,117 @@
...
+     * previously) is cleared.
      */
-    T build();
+    @Internal
+    void clear();
+
+    /**
+     * Enumeration of generic type parameters of this interface.
+     */
+    enum GenericParameter {
```
armiol: I have surely seen that interface. But it's in `server` and I didn't want to move it to `client`, as it just did not belong to `client`.

client/src/main/java/org/spine3/validate/AbstractValidatingBuilder.java
armiol: Done.

server/src/main/java/org/spine3/server/aggregate/AggregateTransaction.java
```diff
@@ -0,0 +1,105 @@
...
+ * @param <B> the type of a {@code ValidatingBuilder} for the aggregate state
+ * @author Alex Tymchenko
+ */
+class AggregateTransaction<I,
+                           S extends Message,
+                           B extends ValidatingBuilder<S, ? extends Message.Builder>>
+        extends Transaction<I, Aggregate<I, S, B>, S, B> {
+
+    @VisibleForTesting
+    AggregateTransaction(Aggregate<I, S, B> entity) {
```
armiol: Done. Also done for `ProjectionTransaction` and `ProcManTransaction`.

server/src/main/java/org/spine3/server/aggregate/AggregateTransaction.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    protected void invokeApplier(Aggregate entity,
+                                 Message eventMessage,
+                                 @Nullable EventContext ignored) throws InvocationTargetException {
+        entity.invokeApplier(eventMessage);
+    }
+
+    @SuppressWarnings("RedundantMethodOverride") // overrides to expose to this package.
```
armiol: Done.

server/src/main/java/org/spine3/server/entity/DefaultEntityStorageConverter.java
```diff
@@ -86,9 +86,15 @@ protected E doBackward(EntityRecord entityRecord) {
         final E entity = repository.create(id);
 
         if (entity != null) {
+
             if (entity instanceof AbstractVersionableEntity) {
                 final AbstractVersionableEntity versionable = (AbstractVersionableEntity) entity;
-                versionable.updateState(state, entityRecord.getVersion());
+                if(versionable instanceof EventPlayingEntity) {
```
armiol: Done.

server/src/main/java/org/spine3/server/entity/Transaction.java
```diff
@@ -0,0 +1,515 @@
...
+
+                throw invalidStateException;
+            } catch (RuntimeException genericException) {
+                rollback(genericException);
+                throw illegalStateWithCauseOf(genericException);
+            } finally {
+                releaseTx();
+            }
+        } else {
+
```
armiol: Done.

server/src/main/java/org/spine3/server/entity/TransactionListener.java
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * A common contract for the {@linkplain Transaction transaction} listeners.
+ *
+ * <p>Provides an ability to add callbacks to the transaction execution stages.
+ *
+ * @author Alex Tymchenko
+ */
+@Internal
+public interface TransactionListener<I,
+                                    E extends EventPlayingEntity<I, S, B>,
```
armiol: Done.

server/src/main/java/org/spine3/server/projection/ProjectionTransaction.java
```diff
@@ -0,0 +1,118 @@
...
+        super(entity, listener);
+    }
+
+    @VisibleForTesting
+    ProjectionTransaction(Projection<I, M, B> entity, M state, Version version) {
+        super(entity, state, version);
+    }
+
+    @Override
+    protected void invokeApplier(Projection entity,
```
armiol: One option I can see is `invokeHandler`, which would have the similar flaw.

The idea of this method is _just_ call the event applier/handler method, so it cannot be called `apply`, because nothing is actually applied in the descendants. 

I will think of another name.

server/src/test/java/org/spine3/server/entity/TransactionShould.java
armiol: Removed.

server/src/main/java/org/spine3/server/projection/ProjectionTransaction.java
armiol: Renamed to `dispatch`.

client/src/main/java/org/spine3/validate/ValidatingBuilders.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    /**
+     * Validating builder for {@linkplain StringValue} messages.
+     */
+    public static final class StringValueValidatingBuilder
```
armiol: As discussed vocally, the builders will be renamed with `VBuilder` suffix (see #484).

client/src/main/java/org/spine3/validate/ValidatingBuilder.java
alexander-yevsyukov: Please move `GenetricTypeIndex` to the client side. This is already one case when we need it, and there may be other as we develop UI-related technology.

armiol: Moved to `client` module (the new package is `org.spine3.reflect`) and used in this case.

armiol:
@alexander-yevsyukov Issues are now addressed. PTAL again.

PR #467, opened 2017-05-10
Entity visibility control

alexander-yevsyukov:
This PR introduces `VisiblityGuard` which allows to get an access to a registered repository depending on the visibility state of entities this repository manages. 

`BoundedContext` now is able to return type names of entity states by their visibility.

In addition:
* `Repository.isOpen()` made public.

### DIFF
--- a/client/src/main/java/org/spine3/option/EntityOptions.java
+++ b/client/src/main/java/org/spine3/option/EntityOptions.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2017, TeamDev Ltd. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.option;
+
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.Message;
+import org.spine3.annotation.Internal;
+import org.spine3.option.EntityOption.Visibility;
+import org.spine3.type.TypeName;
+
+/**
+ * Utilities for working with {@link EntityOption}s.
+ *
+ * @author Alexander Yevsyukov
+ */
+@Internal
+public class EntityOptions {
+
+    private EntityOptions() {
+        // Prevent instantiation of this utility class.
+    }
+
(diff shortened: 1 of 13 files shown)


alexander-yevsyukov:
@armiol, PTAL.

client/src/main/java/org/spine3/option/EntityOptions.java
```diff
@@ -0,0 +1,51 @@
...
+ * @author Alexander Yevsyukov
+ */
+@Internal
+public class EntityOptions {
+
+    private EntityOptions() {
+        // Prevent instantiation of this utility class.
+    }
+
+    public static Visibility getVisibility(Class<? extends Message> stateClass) {
```
armiol: Let's check the parameter for `null` and test this case via `NullPointerTester`.

server/src/main/java/org/spine3/server/BoundedContext.java
```diff
@@ -316,39 +292,55 @@ public void notify(IntegrationEvent integrationEvent,
...
      * @param aggregateStateClass the class of the aggregate state
      * @return repository instance or empty {@code Optional} if not found
      */
     public Optional<? extends AggregateRepository<?, ?>> getAggregateRepository(
             Class<? extends Message> aggregateStateClass) {
-        final AggregateRepository<?, ?> result = aggregateRepositories.get(aggregateStateClass);
-        return Optional.fromNullable(result);
+        // See if there is a repository for this state at all.
+        if (!guard.hasRepository(aggregateStateClass)) {
+            throw newIllegalStateException("No repository found for the aggregate state class %s",
```
armiol: This branch is reported to be untested.

server/src/main/java/org/spine3/server/entity/VisibilityGuard.java
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * @param stateClass the class of the state of entities managed by the repository
+     * @return the repository wrapped into {@code Optional} or {@code Optional#absent()} if the
+     * entity state is {@linkplain Visibility#NONE not visible}
+     * @throws IllegalArgumentException if the repository for the passed state class was not
+     *                                  {@linkplain #register(Repository) registered} with the guard
+     *                                  prior to this call, or if all repositories were
+     *                                  {@linkplain #shutDownRepositories() shut down}
+     */
+    public Optional<Repository> getRepository(Class<? extends Message> stateClass) {
```
armiol: As long we we perform the `checkNotNull(...)` in `getEntityTypes(visibility)` below, I think we should do the same for `stateClass` as well.

server/src/main/java/org/spine3/server/entity/VisibilityGuard.java
```diff
@@ -0,0 +1,182 @@
...
+ * A registry of repositories that controls access to them depending on the visibility of
+ * corresponding entity states.
+ *
+ * @author Alexander Yevsyukov
+ * @see EntityOptions
+ */
+@Internal
+public final class VisibilityGuard {
+
+    private final Map<Class, RepositoryAccess> repositories = newHashMap();
```
armiol: I would make the bounds more strict: `Map<Class<? extends Message>, RepositoryAccess>`.

server/src/main/java/org/spine3/server/entity/VisibilityGuard.java
```diff
@@ -0,0 +1,182 @@
...
+     * Registers the passed repository with the guard.
+     */
+    public void register(Repository repository) {
+        checkNotNull(repository);
+        final Class stateClass = repository.getEntityStateClass();
+        checkNotAlreadyRegistered(stateClass);
+        repositories.put(stateClass, new RepositoryAccess(repository));
+    }
+
+    private void checkNotAlreadyRegistered(Class stateClass) {
```
armiol: Let's check this parameter for `null` in order to be consistent. 

See more below on the same matter.

server/src/test/java/org/spine3/server/entity/VisibilityGuardShould.java
```diff
@@ -0,0 +1,164 @@
...
+public class VisibilityGuardShould {
+
+    private VisibilityGuard guard;
+    private List<Repository> repositories;
+    private BoundedContext boundedContext;
+
+    @Before
+    public void setUp() {
+        boundedContext = BoundedContext.newBuilder()
+                                                            .build();
```
armiol: Please fix the wrapping.

server/src/main/java/org/spine3/server/BoundedContext.java
alexander-yevsyukov: Seems to be a bug of Codecov. There's a test for it. Please see `BoundedContextShould.throw_ISE_when_no_repository_registered()`.

client/src/main/java/org/spine3/option/EntityOptions.java
alexander-yevsyukov: Done

server/src/main/java/org/spine3/server/entity/VisibilityGuard.java
alexander-yevsyukov: Done

alexander-yevsyukov:
@armiol, PTAL.

PR #468, opened 2017-05-10
Add Right margin plugin to build script file

alexander-aleksandrov:
Update version of tools projects and add right margin plugin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ buildscript {
         //TODO:2016-12-12:alexander.yevsyukov: Advance the plug-in version together with all
         // the components and change the version of the dependency below to
         // `spineVersion` defined above.
-        spineToolsVersion = '0.9.2-SNAPSHOT'
+        spineToolsVersion = '0.9.4-SNAPSHOT'
 
         // Defines option for the `spine-model-compiler` plugin.
         // Indicates whether the generation of the validating builders is enabled.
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -38,7 +38,7 @@ buildscript {
         //TODO:2016-12-12:alexander.yevsyukov: Advance the plug-in version together with all
         // the components and change the version of the dependency below to
         // `spineVersion` defined above.
-        spineToolsVersion = '0.9.1-SNAPSHOT'
+        spineToolsVersion = '0.9.2-SNAPSHOT'
```
dmdashenkov: It seems like this version number is already taken.
Try `0.9.4-SNAPSHOT`.

build.gradle
```diff
@@ -140,10 +140,16 @@ subprojects {
...
-
             threshold = 0
             reportType = "warn"
         }
+
+        // Right margin line checker.
+        //
+        // Finds and reports about the lines of a code with length above threshold.
+        rightMarginWrappingChecker {
+            threshold = 100
```
dmdashenkov: Why do we still call this parameter `threshold`? Let's make it consistent with the tools internal naming.

alexander-aleksandrov: Done [here](https://github.com/SpineEventEngine/tools/pull/88)

alexander-aleksandrov: We decided on a meeting to name this parameter `maxTextWidth`

alexander-aleksandrov:
@dmdashenkov PTAL

PR #469, opened 2017-05-11
Enable the clean-up of artifacts stored at Google Cloud Storage

armiol:
This PR enables the clean-up of artifacts during Travis build via new  `gcs-plugin`.

From now all credentials used during Travis build are stored in `credentials.tar.enc`, which is decrypted during the build process.


armiol:
@dmytro-grankin PTAL.

PR #475, opened 2017-05-16
Fix the right margin plugin output in a build log.

alexander-aleksandrov:
We need to fix the right margin plugin so that it should log the result only once per project without duplicating for every module.


alexander-aleksandrov:
@dmdashenkov PTAL

PR #477, opened 2017-05-17
Javadoc generation

alexander-aleksandrov:
Fix all Javadoc errors and ensure that we have generated Javadocs.


alexander-aleksandrov:
@dmdashenkov PTAL

server/src/main/java/org/spine3/server/entity/storage/Column.java
```diff
@@ -42,8 +42,8 @@
  *
  *
  * <p>These methods represent the Columns:
- * <code>
- *      <pre>
+ * <pre>
+ *      <code>
```
dmdashenkov: Please prefer `{@code ... }` to `<code> ... </code>` here and in other similar places.

alexander-aleksandrov:
@dmdashenkov PTAL

testutil/src/main/java/org/spine3/test/Verify.java
```diff
@@ -1508,7 +1508,9 @@ public static void assertThrows(
      * <p>
      * e.g.
      * <pre>
-     * Verify.assertThrowsWithCause(RuntimeException.class, IOException.class, new Callable<Void>()
+     * Verify.assertThrowsWithCause(RuntimeException.class,
+     *                              IOException.class,
+     *                              new Callable&lt;Void&gt;()
```
dmdashenkov: Please fix the code formatting in this snippet regarding the Spine code style conventions.

PR #478, opened 2017-05-18
Serializable Columns

dmdashenkov:
Since we need to pass the Entity Columns values and queries between several instances of JVM, we need to make `Column` class and the aggregating types `Serializable`.

Also, in this PR we revoke the type parameter from the `Column` class since it created no compile-time checks, but only `"unchecked"` warnings.


dmdashenkov:
@armiol, PTAL.

PR #480, opened 2017-05-19
Catch Up performance hotfix

dmdashenkov:


dmdashenkov:
@armiol, PTAL.

server/src/main/java/org/spine3/server/event/EventStorage.java
```diff
@@ -57,11 +66,31 @@ public Event apply(@Nullable EventEntity input) {
                 }
             };
 
+    @Override
+    protected org.spine3.server.storage.EventStorage createStorage(StorageFactory factory) {
```
armiol: Let's avoid FQNs as much as we can.

server/src/main/java/org/spine3/server/event/EventStorage.java
```diff
@@ -57,11 +66,31 @@ public Event apply(@Nullable EventEntity input) {
...
     Iterator<Event> iterator(EventStreamQuery query) {
-        final Iterator<EventEntity> filtered = iterator(createEntityFilter(query));
-        final List<EventEntity> entities = Lists.newArrayList(filtered);
-        Collections.sort(entities, EventEntity.comparator());
-        final Iterator<Event> result = Iterators.transform(entities.iterator(), getEventFunc());
+        final org.spine3.server.storage.EventStorage storage = recordStorage();
+        final Map<EventId, EntityRecord> records = storage.readAll(query);
+        final Collection<EventEntity> entities = transform(records.entrySet(),
+                                                           storageRecordToEntity());
+        // TODO:2017-05-19:dmytro.dashenkov: Remove after the Entity Column approach is implemented.
```
armiol: @dmdashenkov please create an issue in `core-java` right after this PR is merged.

server/src/test/java/org/spine3/server/event/EventStoreShould.java
```diff
@@ -0,0 +1,147 @@
...
+    private static Event projectCreated(Timestamp timestamp) {
+        final ProjectCreated msg = Sample.messageOfType(ProjectCreated.class);
+        return eventFactory.createEvent(msg, null, timestamp);
+    }
+
+    @SuppressWarnings("BusyWait")
+    private static void await(AtomicBoolean done) {
+        while (!done.get()) {
+            try {
+                Thread.sleep(200);
```
armiol: Let's eliminate it, even in this PR.

server/src/main/java/org/spine3/server/event/EventStorage.java
dmdashenkov: We should for sure but we cannot avoid usage of FQN if the qualified class has the same short name as the one that uses it.

dmdashenkov:
@armiol, PTAL again.

PR #493, opened 2017-05-30
Better generic utils

alexander-yevsyukov:
This PR improves the way of obtaining values of generic type arguments. Before this PR we relied on a fact that immediate ancestor of the passed class has generic parameters. This did not allow for building type hierarchies neither for entities nor for repositories.

Now to obtain a type argument value for a class it is needed to pass the class and its ancestor which has a generic parameter of the interest.


alexander-yevsyukov:
@armiol, PTAL.

PR #497, opened 2017-06-02
Change domain name to `io.spine` 

alexander-aleksandrov:
On this step, the names will be changed in all cases except referencing to validating builders


alexander-aleksandrov:
@armiol PTAL

build.gradle
```diff
@@ -22,7 +22,7 @@ import groovy.time.TimeCategory
 buildscript {
     repositories {
         mavenCentral()
-        mavenLocal()
+//        mavenLocal()
```
armiol: @alexander-aleksandrov FYI. I have fixed that already in this branch (see my recent commits).

build.gradle
```diff
@@ -35,12 +35,12 @@ buildscript {
...
         protobufGradlePluginVerison = '0.8.0'
         
-        spineVersion = '0.9.14-SNAPSHOT'
+        spineVersion = '0.9.14.1-SNAPSHOT'
         
         //TODO:2016-12-12:alexander.yevsyukov: Advance the plug-in version together with all
         // the components and change the version of the dependency below to
         // `spineVersion` defined above.
-        spineToolsVersion = '0.9.5-SNAPSHOT'
+        spineToolsVersion = '0.9.6-SNAPSHOT'
```
armiol: @alexander-aleksandrov FYI. I have updated the version already. See my recent commits to this branch.

armiol:
@dmdashenkov 

> why is the difference between the added and deleted code lines count so big (~4K of lines)?

It's because we keep `ValidatingBuilder` hierarchy in both `io.spine ...` and `org.spine3 ...`  version. There are about 25 classes (interface and descendants) in `org.spine3` that were not renamed, but copied. I think, they easliy make 4K lines altogether. 

Such a trick is required for `tools` plugins migration. The duplicated `org.spine3` classes are going to be deleted once `tools` is fully migrated.

armiol:
@dmdashenkov BTW, I have resolved the issues. PTAL again.

.idea/inspectionProfiles/Project_Default_copy.xml
```diff
@@ -0,0 +1,1350 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default copy" />
```
armiol: @alexander-aleksandrov I have removed this file, as it's a duplicate.

PR #499, opened 2017-06-04
Improve repository registration

alexander-yevsyukov:
This PR improves registration of `Repositories` in a `BoundedContext`. This PR makes each repository has a link to the `BoundedContext` to which it belongs. The `BoundedContext` sets this relationship when the repository is registered.

After the repository is registered the `BoundedContext` calls `Repository.onRegistered()` allows the repository to perform post registration work specific to the type of the repository.

Breaking changes:
 * `BoundedContext` is no longer sent to the constructor of types descending from `Repository`.
 * Calling `Repository.getStorage()` returns a storage initialized for the repository. If it wasn't done, `IllegalStateException` will be thrown.

The number of changed files in this PR is that big because of the optimized imports followed the recent root package migration.


alexander-yevsyukov:
@armiol, PTAL.

server/src/test/java/io/spine/server/aggregate/given/AggregateRootTestEnv.java
```diff
@@ -0,0 +1,124 @@
...
+    private AggregateRootTestEnv() {
+    }
+
+    public static class AnAggregateRoot extends AggregateRoot<String> {
+        protected AnAggregateRoot(BoundedContext boundedContext, String id) {
+            super(boundedContext, id);
+        }
+    }
+
+    @SuppressWarnings("TypeMayBeWeakened") // Exact message classes without OrBuilder are needed.
```
armiol: I thought this was just a minor warning according to our inspection profile. If that's still true, I wouldn't bother suppressing it.

We are always going to get warnings like this one, since we use Protobuf messages. In 99% of time those can be weakened to `...MessageOrBuilder`. IDEA uses every chance to remind us about that, making the rest of "TypeMayBeWeakened" warnings invisible for us.

server/src/test/java/io/spine/server/bc/given/BoundedContextTestEnv.java
```diff
@@ -0,0 +1,236 @@
...
+
+        public ProjectPmRepo() {
+            super();
+        }
+    }
+
+    public static class ProjectReport
+            extends Projection<ProjectId, Empty, EmptyValidatingBuilder> {
+
+        @SuppressWarnings("PublicConstructorInNonPublicClass")
```
armiol: The class is `public` now. 

In general we should eventually run IDEA's own inspections. Among other things it points to the redundant warning suppressions.

server/src/test/java/io/spine/server/aggregate/given/AggregateRootTestEnv.java
alexander-yevsyukov: That was an old suppression. I've fixed it. Let's discuss the policy change as to this and other warnings in person.

PR #500, opened 2017-06-05
`Identifier` in public API

alexander-yevsyukov:
This PR makes the `Identifier` class public so that it is possible to know an ID type from outside the `io.spine.base` package. This is a preparation step for providing deterministic serializers for Beam support.

Since the `Identifier` is now a public class, the code of the`Identifiers` utility class was moved into it. `idToAny()` and `idFromAny()` methods were renamed to `pack()`/`unpack()`.

In addition this PR:
 * Makes `TypeUrl` a serializable class
 * Unifies the value of `serialVersionUID` to be `0L` thought the code
 * Adds some test coverage

The Spine version is advanced to `0.9.17-SNAPSHOT`.


alexander-yevsyukov:
@armiol , PTAL.

PR #504, opened 2017-06-07
Introduce `ValidationException`

armiol:
This PR brings `ValidationException` class as a future replacement for `ConstraintViolationThrowable`.

The framework version has been increased to `0.9.18.1-SNAPSHOT`, which states for an intermediate version, not suitable for immediate use by other Spine libraries. This version is only required for the migration of `tools` library.

In scope of the following PRs `tools` repo is migrated to the `ValidationException`. After that `core-java` will be migrated to the updated `tools` version, eliminating `ConstraintViolationThrowable`.

### DIFF
--- a/client/src/main/java/io/spine/validate/ValidationException.java
+++ b/client/src/main/java/io/spine/validate/ValidationException.java
@@ -79,8 +79,7 @@ public class ValidationException extends RuntimeException {
         @Override
         public String apply(@Nullable ConstraintViolation input) {
             checkNotNull(input);
-            return Stringifiers.toString(
-                    input);
+            return Stringifiers.toString(input);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


armiol:
@alexander-yevsyukov PTAL.

client/src/main/java/io/spine/validate/ValidationException.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * A function, transforming a {@linkplain ConstraintViolation constraint violation}
+     * into a {@code String}.
+     */
+    private static final class ToStringFn implements Function<ConstraintViolation, String> {
+        @Nullable
+        @Override
+        public String apply(@Nullable ConstraintViolation input) {
+            checkNotNull(input);
+            return Stringifiers.toString(
```
alexander-yevsyukov: Please fix wrapping here.

PR #507, opened 2017-06-09
Fix `AggregateEventRecord` test data creation.

dmdashenkov:
Now any test `AggregateEventRecord` created with `io.spine.server.aggregate.Given.StorageRecord` will have a `KindCase`. By default it is `EVENT`.


dmdashenkov:
@armiol, PTAL.

PR #509, opened 2017-06-12
Storage Iterators

dmdashenkov:
Instead of storing the values retrieved from the Storage into a `Collection` or `Map`, which may cause some memory issues, now we get an `Iterator`.

The in-memory iterators are straightforward, but those produced by the "real" `Storage` implementations should be lazy and fetch the data only when it's really needed.

Minding that, some operations of the results of the storage read are changed to correspond the lazy processing fashion.

Also, in this PR the `OperatorEvaluator` becomes and `enum`.


dmdashenkov:
@alexander-yevsyukov, PTAL.

dmdashenkov:
@alexander-yevsyukov, PTAL.

client/src/main/java/io/spine/util/BigIterators.java
```diff
@@ -0,0 +1,279 @@
...
+         * may be moved or some elements may be {@linkplain Iterator#remove() removed}). That's why
+         * all the subsequent calls throw {@link IllegalStateException}.
+         *
+         * @throws IllegalStateException if called more than once
+         */
+        @Nonnull
+        @Override
+        public Iterator<E> iterator() {
+            if (used) {
+                throw new IllegalStateException("Reusing one-time Iterable.");
```
alexander-yevsyukov: As discussed, this is fragile. Let's discuss vocally.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -347,13 +320,13 @@ private E toEntity(EntityRecord record) {
      *
      * @return new instance of the transforming function
      */
-    protected final Function<Map.Entry<I, EntityRecord>, E> storageRecordToEntity() {
-        return new Function<Map.Entry<I, EntityRecord>, E>() {
+    protected final Function<EntityRecord, E> storageRecordToEntity() {
```
alexander-yevsyukov: Please change it in the following way:
 * Have the conversion function which is created on a repository initialization. 
 * Make it `Serializable`. It will be needed for integration with Beam.
 * Pass the instance of the function into `transform()`.
 * Call the function from the `toEntity()` method.

server/src/main/java/io/spine/server/event/ERepository.java
```diff
@@ -74,14 +77,13 @@ public Event apply(@Nullable EEntity input) {
...
         final Predicate<EEntity> detailedLookupFilter = createEntityFilter(query);
-        final Iterator<EEntity> entityIterator = FluentIterable.from(entities)
-                                                                   .filter(detailedLookupFilter)
-                                                                   .toSortedList(comparator())
-                                                                   .iterator();
-        final Iterator<Event> result = Iterators.transform(entityIterator, getEventFunc());
+        final Iterator<EEntity> filtered = filter(entities, detailedLookupFilter);
+        final List<EEntity> mutableList = newArrayList(filtered);
+        sort(mutableList, comparator());
+        final Iterator<Event> result = Iterators.transform(mutableList.iterator(), getEventFunc());
```
alexander-yevsyukov: Let's standardize on invoking `transform()`. Other parts of the code static import it. It would be better to have this line read:

```java
final Iterator<Event> result = transform(eventEntities, toEvent());
```
... which also assumes renaming `getEventFunc()` to `toEvent()`.

dmdashenkov:
@alexander-yevsyukov, PTAL.
As discussed, I've left the `EntityStorageConverter` not `Serializable` for now.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -249,7 +251,8 @@ protected E findOrCreate(I id) {
         final EntityQuery<I> entityQuery = EntityQueries.from(filters, getEntityClass());
         final EntityQuery<I> completeQuery = toCompleteQuery(entityQuery);
         final Iterator<EntityRecord> records = recordStorage().readAll(completeQuery, fieldMask);
-        final Iterator<E> result = transform(records, storageRecordToEntity());
+        final Function<EntityRecord, E> transformer = entityConverter().reverse();
```
alexander-yevsyukov: Pease rename the variable `transformer` into the kind of transformation. Do not suggest `OptimusPrime` or something :) Tell what the transformation does.

server/src/main/java/io/spine/server/event/ERepository.java
```diff
@@ -83,7 +83,7 @@ public Event apply(@Nullable EEntity input) {
         final Iterator<EEntity> filtered = filter(entities, detailedLookupFilter);
         final List<EEntity> mutableList = newArrayList(filtered);
         sort(mutableList, comparator());
-        final Iterator<Event> result = Iterators.transform(mutableList.iterator(), getEventFunc());
+        final Iterator<Event> result = transform(mutableList.iterator(), getEvent());
```
alexander-yevsyukov: Please extract the variable `eventEntities` or something like that. `mutableList` is not descriptive.

dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #515, opened 2017-06-16
Extract `Users` library

alexander-yevsyukov:
This PR:
* Moves `UserId` and `TenantId` into the `spine/base` package.
* Removes the `users` module which code was moved into the Users library. 

Spine version advances to `0.9.25-SNAPSHOT`


alexander-yevsyukov:
@armiol, PTAL.

PR #517, opened 2017-06-18
Extract `base` sub-project. Stage 1 of 3.

alexander-yevsyukov:
This PR is the first step of extracting Spine Base subproject. The Base subproject would provide:
 * options for annotating proto types with validation attributes
 * code generation from annotated proto definitions
 * utilities and types for working with the generated code.

This PR separates just modules. The following step would be to rename `io.spine.base` into `core`.


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -102,6 +102,24 @@ protected FieldValidator(FieldDescriptor descr,
     }
 
     /**
+     * Checks if the file is for commands.
+     *
+     * @param file a descriptor of a {@code .proto} file to check
+     * @return {@code true} if the file name ends with {@code "commands"},
+     * {@code false} otherwise
+     */
+    static boolean isCommandsFile(FileDescriptor file) {
```
armiol: I don't understand why this method is here. I.e.  `FieldValidator.isCommandsFile(..)` looks really weird.

PR #520, opened 2017-06-19
Make direct `Delivery` strategies non-singletons

armiol:
Before this PR there were some side effects from having the `...Delivery.directDelivery()` returning singletons for each of delivery subclasses. 

In particular, several bounded contexts created within a single JVM could share the same delivery strategy (since it was a singleton) and thus reference the same consumer provider.

This PR addresses this issue, implementing the "direct" delivery by creating a new instance each time instead of sharing the singleton instance.

The framework version is increased to `0.9.26-SNAPSHOT`.


armiol:
@alexander-yevsyukov PTAL.

PR #523, opened 2017-06-21
Updated FieldValidator with Map handling

IuriiSergiichuk:
Previously `FieldValidator` wasn't aware of possible `Map` values passed to validate, that's why generated `ValidatingBuilders` were not able to validate `map` values.

This PR fixes `ClassCastException` that was thrown by `MessageFieldValidator.isTimestamp()` when `AbstractValidatingBuilder` was trying to `validate` values, that are going to be passed to `map` type field, e.g. for such message:
```proto
message MapRequiredMsgFieldValue {
    map<string, google.protobuf.StringValue> value = 1;
}
```

Also, I've added several unit tests to be sure, that validators could be created for protobuf fields with `reapeated` and `map` types.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ buildscript {
 
         spineRepository = 'http://maven.teamdev.com/repository/spine'
         spineSnapshotsRepository = 'http://maven.teamdev.com/repository/spine-snapshots'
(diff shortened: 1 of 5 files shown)


dmdashenkov:
As outdated.

PR #524, opened 2017-06-22
Publish `core`

alexander-yevsyukov:
This PR includes the `core` module into publishing. The version is not bumped since it just fixed the publishing problem of the version increased by the previous PR.


alexander-yevsyukov:
@armiol, PTAL.

PR #526, opened 2017-06-23
Move deployment-related code into a separate repository

alexander-yevsyukov:
This PR removes the code of `server-deployment` module which code was moved into the `deployment` repository.


alexander-yevsyukov:
@armiol, PTAL.

PR #527, opened 2017-06-23
Introduce `ThrowableMessage` and use it instead of `FailureThrowable` where possible

armiol:
This PR introduces a `ThrowableMessage` as a more generic abstraction for the `Throwables`, having a complex state, than `FailureThrowable`s we had before.

This exception is a base routine, which is later going to be used in Failure generator plugin (part of a `base/model-compiler`.

In the following PRs, the `FailureThrowable` will be phased out and removed from the source code base. It is impossible to do now, since it is still used in the generated code.

Also, `.gitignore` is optimized.

The framework version is increased to `0.9.29-SNAPSHOT`.

### DIFF
--- a/.codacy.yaml
+++ b/.codacy.yaml
@@ -2,6 +2,7 @@
 exclude_paths:
 - "client/src/main/java/io/spine/net/Patterns.java"
 - "base/src/test/**/*.*"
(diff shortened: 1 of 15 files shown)


armiol:
@alexander-yevsyukov PTAL.

PR #528, opened 2017-06-23
Remove `FailureThrowable`

armiol:
This PR completes a migration from `FailureThrowable` to `ThrowableMessage`.

Summary of changes:
* The new model compiler, which is based upon `ThrowableMessage`, is now used.
* `FailureThrowable` and its test class are  removed. The references are updated according to `ThrowableMessage` API.
* As a minor change, an excessive test code dependency onto `testutil-server` is removed from the `base` module. That is required to make its planned migration to another repository easier.

The version has been increased to `0.9.30-SNAPSHOT`.


armiol:
@alexander-yevsyukov PTAL.

PR #530, opened 2017-06-26
Move change-related classes to `core`

alexander-yevsyukov:
This PR moves classes and proto types related to `Change` objects into the `core` module.

The `Preconditions` class used for verifying arguments of change creation is no longer a part of public API.


alexander-yevsyukov:
@armiol, PTAL.

PR #535, opened 2017-07-06
Move `GivenCommandContext` to `testutil-core`

armiol:
In scope of this PR the `GivenCommandContext` test utility is moved to `testutil-core`. 

Thus, all the `Given...` utilities become located in the same `io.spine.core.given` package of `testutil-core` module. It allows the rest of Spine libraries to depend on `testutil` modules rather than on `spine-core`.

The framework version has been increased to `0.9.37-SNAPSHOT`.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -25,7 +25,7 @@
  *  as we want to manage the versions in a single source.
  */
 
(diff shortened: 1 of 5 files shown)


armiol:
@alexander-yevsyukov PTAL.

PR #537, opened 2017-07-10
Move import of `options.proto`

alexander-aleksandrov:
Move
```protobuf
import "spine/options.proto";
``` 
at the top of the import section in proto files.


alexander-aleksandrov:
@armiol PTAL

PR #540, opened 2017-07-18
AggregateStateId clean up

dmdashenkov:
In this PR:
 - the duplicate `static` initializer in `AggregateStateId` is removed;
 - the `AggregateStateIdStringifier` became a singleton.

Since no actual change in the behavior is or public API is performed, we may leave the framework version unchanged.


dmdashenkov:
@armiol, PTAL.

PR #544, opened 2017-07-24
Uni- and multi-cast dispatchers

alexander-yevsyukov:
This PR allows dispatchers to return IDs of entities to which a message was dispatched.

Also, `IsSent` was renamed to `Ack`.

### DIFF
--- a/server/src/main/java/io/spine/server/aggregate/AggregateMessageEndpoint.java
+++ b/server/src/main/java/io/spine/server/aggregate/AggregateMessageEndpoint.java
@@ -24,7 +24,7 @@ import com.google.protobuf.Message;
 import io.spine.core.Event;
 import io.spine.core.MessageEnvelope;
 import io.spine.server.entity.LifecycleFlags;
-import io.spine.server.tenant.TenantAwareOperation;
+import io.spine.server.tenant.TenantAwareFunction0;
 
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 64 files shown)


alexander-yevsyukov:
@armiol, I decided to leave the ability to return of entity IDs from dispatchers primary because of the clearer notion of command dispatching via `UnicastDispatcher` interface. This also opens ability for easier tracing of multicast dispatching, as we'll be able to compare expected IDs with actually dispatched.

alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/io/spine/server/aggregate/AggregateMessageEndpoint.java
```diff
@@ -51,17 +53,26 @@
...
-        final Set<I> targets = getTargets();
-        for (I target : targets) {
-            dispatchTo(target);
+    @SuppressWarnings("unchecked")
+    R dispatch() {
+        final R targets = getTargets();
+        if (targets instanceof Set) {
+            final Set<I> set = (Set<I>) targets;
+            for (I id : set) { //TODO:2017-07-21:alexander.yevsyukov: Can we run this in parallel?
+                //TODO:2017-07-21:alexander.yevsyukov: What if we dispatch events to multiple aggregates
```
armiol: Then we should remember the issue and continue dispatching. And handle the exceptions later.

To be addressed outside of this PR.

server/src/main/java/io/spine/server/bus/MulticastDispatcher.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Dispatches a message to several entities.
+ *
+ * @param <C> the type of dispatched messages
+ * @param <E> the type of envelopes for dispatched objects that contain messages
+ * @param <I> the type of IDs of entities to which messages are dispatched
```
armiol: So if an event is being dispatched to several entities of different types (e.g. Projections, PMs etc), we'll need to create several `MulticastDispatcher`s — as it won't be possible to set a single type for `<I>`.

Perhaps, that's worth mentioning. I.e. that this kind of dispatcher dispatches a message to several entities of a single type.

PR #545, opened 2017-07-24
Remove `IdConverter`

alexander-yevsyukov:
This PR eliminates the hierarchy of `IdConverter` in favor of using `Identifier`.


alexander-yevsyukov:
@armiol, PTAL.

PR #549, opened 2017-07-27
Allow error callback in `MessageDispatcher`s

alexander-yevsyukov:
This PR introduces `MessageDispatcher.onError()` and makes implementations to call the method when dispatching causes an exception.

Other notable changes:
  * `Environment` class ”moved” to the `base` project.
  * `causedByRejection()` utility method was added to `Commands`.
  * Testing utility class`AggregateCommandDispatcher` was renamed to `AggregateMessageDispatcher` and got method for dispatching `EventEnvelope`.
  * `EventSubscriber` now catches dispatching errors and passes them to `onError()`.


alexander-yevsyukov:
@armiol, I can eliminate one TODO after migrating to `base` version based on [this PR](https://github.com/SpineEventEngine/base/pull/33).

alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/io/spine/server/entity/Repository.java
```diff
@@ -376,6 +387,34 @@ protected void checkNotClosed() {
...
+
+    /**
+     * Logs error caused by a message processing into the {@linkplain #log() repository log}.
+     *
+     * <p>The formatted message has the following parameters:
+     * <ol>
+     *     <li>The name of the message class.
+     *     <li>The message ID.
+     * </ol>
+     * @param msgFormat the format of the message
```
armiol: Please allow an empty line above the @param section.

PR #554, opened 2017-08-04
Reacting instead of subscribing

alexander-yevsyukov:
This PR makes Process Managers react on events and rejections rather than subscribe to them. It also adds the ability for Aggregates to react on rejections.

Notable changes:
  * `ActorMessageEnvelope` was introduced to allow wrapping messages without `ActorContext`.
  * Added `EntityMessageEndpoint` to unify work with message delivery by repositories.
  * Allowed returning `Iterable` in addition to `List` from aggregates and process managers.
  * Now it is possible to return `Empty` from a reacting method, which would be treated as empty `List`, and no event would be stored.
  * `AggregateRepostitory.find()` now follows the contract of `RepositoryView` and does not enforce creation of an aggregate, if it wasn't stored before.


alexander-yevsyukov:
@armiol, PTAL.

core/src/main/java/io/spine/core/ActorMessageEnvelope.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package io.spine.core;
+
+import com.google.protobuf.Message;
+
+/**
+ * A common interface for messages sent by an actor.
+ *
+ * @param <I> the the of the message id
```
armiol: `the type of the message ID`

server/src/main/java/io/spine/server/aggregate/AggregateRejectionEndpoint.java
```diff
@@ -0,0 +1,68 @@
...
+
+package io.spine.server.aggregate;
+
+import com.google.protobuf.Message;
+import io.spine.core.RejectionEnvelope;
+
+import java.util.List;
+import java.util.Set;
+
+class AggregateRejectionEndpoint<I, A extends Aggregate<I, ?, ?>>
```
armiol: This class is missing documentation.

server/src/test/java/io/spine/server/procman/given/ProcessManagerRepositoryTestEnv.java
```diff
@@ -169,19 +177,66 @@ CommandRouted handle(PmStartProject command, CommandContext context) {
...
         public String getIdString() {
             return getId().toString();
         }
     }
+
+    public static class GivenCommandMessage {
+
+        public static final ProjectId ID = Sample.messageOfType(ProjectId.class);
+
+        /** Prevents instantiation on this utility class. */
```
armiol: Weird Javadoc block formatting.

alexander-yevsyukov: We have this formatting for all the utility constructors. This way it's compact.

alexander-yevsyukov:
@armiol, PTAL.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -167,6 +179,18 @@ public static TenantId getTenantId(Event event) {
...
     /**
+     * Creates an empty {@link Iterable} over the messages of the type {@code <M>}.
+     *
+     * <p>This method is useful for returning empty result from reacting methods.
+     *
+     * @param <M> the type of messages
+     * @return empty {@link Iterable}
+     */
+    public static <M> Iterable<M> nothing() {
+        return ImmutableList.of();
```
armiol: This line is reported to be untested. Which is weird.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -104,10 +114,14 @@
...
     private Set<CommandClass> commandClasses;
 
-    /** The set of event classes to which aggregates of this repository {@linkplain React react}. */
+    /** The set of event classes on which aggregates {@linkplain React react}. */
     @Nullable
     private Set<EventClass> reactedEventClasses;
 
+    /** The set of rejection classes on which aggregates {@linkplain React react}. */
+    @Nullable
+    private Set<RejectionClass> reactedRejectionClasses;
```
armiol: If that's an expected set of classes, not the actual (i.e. this field is not a statement of a fact, but rather a declaration of what's expected), then I would name it like `expectedRejectionReactions` or `declaredRejectionReactions`. Similarly, the set of event-related reactions should be renamed as well.

The bottom line is that I'd like to avoid the feeling I get from this kind of naming. Right now it looks like a piece of history.

server/src/main/java/io/spine/server/procman/PmCommandEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+    protected void onError(CommandEnvelope envelope, RuntimeException exception) {
+        repository().onError(envelope, exception);
+        throw exception;
+    }
+
+    /**
+     * Throws {@link IllegalStateException} with the message containing details of
+     * the process manager and the command in response to which empty set of event messages
+     * was generated.
+     * @throws IllegalStateException always
```
armiol: An empty line seems to be missing before `@throws`.

server/src/main/java/io/spine/server/procman/PmEndpoint.java
```diff
@@ -0,0 +1,78 @@
...
+
+import io.spine.core.ActorMessageEnvelope;
+import io.spine.core.Event;
+import io.spine.server.entity.EntityMessageEndpoint;
+import io.spine.server.entity.Repository;
+
+import java.util.List;
+
+/**
+ * Common base message for endpoints of Process Managers.
```
armiol: Perhaps, `Common base for endpoints of Process Managers.`

server/src/main/java/io/spine/server/procman/ProcessManager.java
```diff
@@ -121,43 +122,47 @@ protected B getBuilder() {
...
-     * Dispatches an event to the event subscriber method of the process manager.
+     * Dispatches an event to the event reactor method of the process manager.
      *
-     * @param event the envelope with the event
+     * @param  event the envelope with the event
+     * @return a list of produced events or an empty list if the process manager does not
+     *         produce new events because of the passed event
      */
-    void dispatchEvent(EventEnvelope event)  {
+    List<Event> dispatchEvent(EventEnvelope event)  {
```
armiol: In future the returning value of this method should probably become `Iterable`. `List` is too much of possibilities (`add`, `set`, `remove` etc).

Not to address right now though.

server/src/test/java/io/spine/server/aggregate/AggregateRepositoryShould.java
```diff
@@ -379,33 +386,90 @@ public void now_allow_anemic_aggregates() {
...
+                                           .build()
+        );
+        commandBus.post(startProject, observer);
+
+        for (ProjectId childProject : childProjects) {
+            final Optional<RejectionReactingAggregate> optional = repository.find(childProject);
+            assertTrue(optional.isPresent());
+
+            // Check that all the aggregates:
+            // 1. got Rejections.AggCannotStartArchivedProject
```
armiol: Let's add a `;` at the end.  And for the p.2 as well :)

server/src/test/java/io/spine/server/aggregate/given/AggregateRepositoryTestEnv.java
```diff
@@ -416,9 +431,9 @@ public RuntimeException getLastException() {
     }
 
     /**
-     * An aggregate class which neither handle commands nor react on events.
+     * An aggregate class which neither handle commands nor react on events or rejections.
```
armiol: Should probably be `... which neither handles commands nor reacts on events or rejections.`

server/src/test/java/io/spine/server/aggregate/given/AggregateRepositoryTestEnv.java
```diff
@@ -471,30 +489,134 @@ private ReactingAggregate(ProjectId id) {
...
         }
     }
+
+    /*
+     * Test environment for testing aggregates reacting on rejections.
+     *******************************************************************/
+
+    /**
+     * The aggregate with the state of {@linkplain AggProjectCreated creation event}
+     * wrapped into Any. The event keeps the list of child project IDs, which we use
```
armiol: `{@code Any}`.

server/src/test/proto/spine/test/aggregate/rejections.proto
```diff
@@ -0,0 +1,36 @@
...
+option (type_url_prefix) = "type.spine.io";
+option java_package="io.spine.test.aggregate.rejection";
+option java_generate_equals_and_hash = true;
+
+import "spine/test/aggregate/project.proto";
+
+message AggCannotStartArchivedProject {
+    ProjectId project_id = 1;
+    repeated ProjectId child_project_id = 2;
+
```
armiol: Please remove an empty line.

server/src/main/java/io/spine/server/procman/ProcessManager.java
alexander-yevsyukov: Being package-private is quite safe for now, but I agree: let's to return to this matter in a while.

PR #555, opened 2017-08-08
Introduce `RejectionSubscriber` and remove `RejectionReactor`

armiol:
This PR brings `RejectionSubscriber` back. 

It serves as a base class for those non-`Entity` classes wishing to register as dispatchers of certain Rejections.

At the same time `RejectionReactor` has been removed, as its contract (allowing to return events in response to a rejection) makes no sense outside of `Entity` context.

The framework version has been increased to `0.9.49-SNAPSHOT`.


armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/io/spine/server/reflect/RejectionHandlerMethod.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    static IllegalStateException missingRejectionHandler(
+            Class<?> cls, Class<? extends Message> rejectionClass) {
+        final String msg = format(
+                "Missing handler for rejection class %s in the class %s",
+                rejectionClass, cls
+        );
+        return new IllegalStateException(msg);
```
alexander-yevsyukov: Using `newIllegalStateException()` which can format a string would make the above code more compact.

armiol: Adjusted. Thanks for the tip!

PR #560, opened 2017-08-16
Use `ConsumerID` to pick target envelope consumer

armiol:
Before this PR the identification of consumers in scope of `Delivery` was built upon the consumer `Class<...>`.

However, as `AggregateRepository`uses delegates (such as `DelegatingEventDispatcher`) to register itself, the class of consumer cannot be used as an identity — as long as several different `AggregateRepository` classes have the consumers with the same class (e.g. `DelegatingEventDispatcher.class`).

This PR introduces `ConsumerId` to identify the target consumer instead of `Class<?>`. The instances of `ConsumerId ` are is built taking into account the real class name. In case of `Delegating...Dispatcher`, an FQN of the real delegate is used.

The framework version is increased to `0.9.51-SNAPSHOT`.

### DIFF
--- a/server/src/main/java/io/spine/server/delivery/Consumers.java
+++ b/server/src/main/java/io/spine/server/delivery/Consumers.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2017, TeamDev Ltd. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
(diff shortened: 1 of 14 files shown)


armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/io/spine/server/delivery/Consumers.java
```diff
@@ -0,0 +1,64 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A utility class for working with delivery consumers.
+ *
+ * @author Alex Tymchenko
+ */
+public final class Consumers {
+
+    private Consumers() {
```
alexander-yevsyukov: Please add Javadoc.

PR #561, opened 2017-08-16
Update `base` version

dmdashenkov:
This PR brings updated `base` module with the `(is)` and `(every_is)` options feature.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -51,4 +51,3 @@ ext {
 
     protobufGradlePluginVerison = '0.8.3'
 }
-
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@armiol, PTAL.

ext.gradle
```diff
@@ -25,14 +25,14 @@
  *  as we want to manage the versions in a single source.
  */
 
-def final SPINE_VERSION = '0.9.50-SNAPSHOT'
+def final SPINE_VERSION = '0.9.51-SNAPSHOT'
```
armiol: This version is already used  (and published). Let's discuss.

PR #564, opened 2017-08-18
Back to UUID-based `EventId`s and smarter `EntityClass`

alexander-yevsyukov:
This PR removes generation of events based on source message ID. The idea of having sequential event identifiers based on original command ID aimed to help finding original commands. But this turned to be a wrong practice for generating events in reacting methods in a distributed environment.

In order to have a unified IDs for all the events it was decided to revert to UUID-based IDs of events. The name of the `EventId` field remains to be `value` in order to be compatible with products already based on Spine.

Also this PR makes `EntityClass` aware of ID class, state class, state type, and entity constructor. Further improvements would be done in a following PR.


alexander-yevsyukov:
@armiol, PTAL.

PR #565, opened 2017-08-21
Smarter entity classes

alexander-yevsyukov:
This PR continues extending Model classes and migrating the code to using their new abilities. In particular:
  * `EntityClass` now can create entities.
  * `AggregatePartClass` is introduced.


alexander-yevsyukov:
@armiol, PTAL.

armiol:
@alexander-yevsyukov Could you please add some description to the PR? It's handy to trace the changes in future.

server/src/main/java/io/spine/server/model/Model.java
```diff
@@ -83,14 +86,35 @@ void clear() {
...
-            modelClass = AggregateClass.of(cls);
+            modelClass = new AggregateClass<>(cls);
             checkDuplicates((CommandHandlingClass) modelClass);
             classes.put(cls, modelClass);
         }
         return (AggregateClass<?>) modelClass;
     }
 
     /**
+     * Obtains an instance of aggregate part class information.'
```
armiol: The last symbol looks a typo.

server/src/main/java/io/spine/server/model/Model.java
```diff
@@ -83,14 +86,35 @@ void clear() {
...
+     *
+     * <p>If the passed class was not added to the model before, it would be added as the result of
```
armiol: I would say
```
If the passed class is not added to the model before, it will be added as the result of this method call.
```
instead. It's a bit simpler to comprehend, as to me.

server/src/main/java/io/spine/server/model/Model.java
```diff
@@ -83,14 +86,35 @@ void clear() {
...
+     * this method call.
+     *
+     * @throws DuplicateCommandHandlerError if there is the aggregate part class handles one or more
```
armiol: I would say the same, but simpler:

```
@throws DuplicateCommandHandlerError if the given aggregate part class handles one or commands which are already known to the model as handled by another class.
```

server/src/main/java/io/spine/server/entity/AbstractEntity.java
```diff
@@ -131,66 +156,26 @@ private S createDefaultState() {
...
-        checkNotNull(idClass);
-
-        try {
-            @SuppressWarnings("JavaReflectionMemberAccess") // Required in the Entity definition.
-            final Constructor<E> result = entityClass.getDeclaredConstructor(idClass);
-            result.setAccessible(true);
-            return result;
-        } catch (NoSuchMethodException ignored) {
-            throw noSuchConstructor(entityClass.getName(), idClass.getName());
+        //TODO:2017-08-21:alexander.yevsyukov: Use the below code to overcome the issue in the test
```
alexander-yevsyukov: Here's the related issue: #567

PR #566, opened 2017-08-22
Fix concurrency issue in Bus

dmdashenkov:
At this moment we have the following issue in `Bus`:
 - the filter chain is initialized lazily;
 - the queue which represents the chain is mutable and not synchronized (internally or externally).

This may (and does) cause unexpected behavior (for instance, `null` `BusFiler`s in the chain).

The fix is in making the defensive copies of the chain in `createFilterChain()` method. By doing so, we prevent the chain to be mutated by multiple threads with no extra synchronization is required.


dmdashenkov:
@armiol, PTAL.

server/src/test/java/io/spine/server/event/EventBusShould.java
```diff
@@ -347,6 +365,47 @@ public void create_validator_once() {
...
+    public void store_filters_regarding_possible_concurrent_modifications() throws InterruptedException {
+        final Thread[] threads = new Thread[50];
+        // "Random" more or less valid Event.
+        final Event event = Event.newBuilder()
+                                 .setId(EventId.newBuilder().setValue("123-1"))
+                                 .setMessage(pack(Int32Value.newBuilder()
+                                                            .setValue(42)
+                                                            .build()))
+                                 .build();
+        // Catch not-easily reproducible bugs.
```
armiol: `non-easily`.

server/src/test/java/io/spine/server/event/EventBusShould.java
```diff
@@ -347,6 +365,47 @@ public void create_validator_once() {
         assertSame(validator, eventBus.getValidator());
     }
 
+    @SuppressWarnings("MethodWithMultipleLoops") // OK for such test case.
+    @Ignore // This test is used only to diagnose EventBus malfunctions in concurrent environment.
+            // It's too long to execute this test per each build, so we leave it as is for now.
+            // Please see build log to find out if there were some errors during the test execution.
+    @Test
+    public void store_filters_regarding_possible_concurrent_modifications() throws InterruptedException {
+        final Thread[] threads = new Thread[50];
```
armiol: Isn't it easier to use `Executors. ...` utility instead? 

Having an array of `Thread`s is so 2002.

server/src/main/java/io/spine/server/bus/Bus.java
```diff
@@ -349,7 +350,7 @@ private void doPost(Iterable<E> envelopes, StreamObserver<Ack> observer) {
          * Creates a new instance of the bus builder.
          */
         protected AbstractBuilder() {
-            this.filters = newLinkedList();
+            this.filters = new ConcurrentLinkedDeque<>();
```
armiol: Something is wrong here. Let's discuss.

dmdashenkov:
@armiol, PTAL.

server/src/test/java/io/spine/server/event/EventBusShould.java
```diff
@@ -347,6 +354,43 @@ public void create_validator_once() {
...
+    @SuppressWarnings("MethodWithMultipleLoops") // OK for such test case.
+    @Ignore // This test is used only to diagnose EventBus malfunctions in concurrent environment.
+            // It's too long to execute this test per each build, so we leave it as is for now.
+            // Please see build log to find out if there were some errors during the test execution.
+    @Test
+    public void store_filters_regarding_possible_concurrent_modifications()
+            throws InterruptedException {
+        final int threadCount = 50;
+
+        // "Random" more or less valid Event.
```
armiol: Please have a single text explaining the test scenario. ATM it's totally impossible to understand the idea.

Issue #570, opened 2017-08-22
Improve `ProjectionRepository` tests


alexander-yevsyukov:
Try the following test mutation:

Remove the line:
```java
     store(projection);
```
from `ProjectionEndpoint.deliverNowTo()`. Test should pass, and the only side effect would be missing coverage reported by CodeCov for the method `ProjectionRepository.getStand()`. This is true because the method is not called by `ProjectionEndpoint` because of the removal of the line. We should have at least one failing test instead.

alexander-yevsyukov:
@armiol, FYI.

armiol:
@alexander-yevsyukov addressed in #571.

armiol:
Resolved in #571.

PR #571, opened 2017-08-23
`ProjectionEndpoint` test improvements

armiol:
This PR adds a test for `ProjectionEndpoint` delivery strategy.

Also #570 is addressed: `ProjectionEndpoint.deliverNowTo(..)` is ensured to store the projection instance after an event is successfully delivered.

As a minor addition, it fixes a minor code duplication — all `...Endpoint` classes now use own static `of` method to create new instances instead of explicit `new ...Endpoint<>` calls.

The framework version left intact.


armiol:
@alexander-yevsyukov PTAL.

PR #572, opened 2017-08-23
Reject in command routing

alexander-yevsyukov:
This PR adds a test of correct processing of a command which is rejected in a command routing function.

Also an ability to get a command status by command instance was introduced. `CommandStore` was also made visible via a `CommandBus` instance.

### DIFF
--- a/server/src/test/java/io/spine/server/route/given/switchman/Switchman.java
+++ b/server/src/test/java/io/spine/server/route/given/switchman/Switchman.java
@@ -29,7 +29,7 @@ import io.spine.server.route.given.switchman.event.SwitchPositionConfirmed;
 import io.spine.validate.UInt32ValueVBuilder;
 
 /**
- * The aggregate that handles commands send to a swtichman.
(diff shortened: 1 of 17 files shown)


alexander-yevsyukov:
@armiol, PTAL.

server/src/test/java/io/spine/server/route/given/switchman/Switchman.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.protobuf.UInt32Value;
+import io.spine.server.aggregate.Aggregate;
+import io.spine.server.aggregate.Apply;
+import io.spine.server.command.Assign;
+import io.spine.server.route.given.switchman.command.SetSwitch;
+import io.spine.server.route.given.switchman.event.SwitchPositionConfirmed;
+import io.spine.validate.UInt32ValueVBuilder;
+
+/**
+ * The aggregate that handles commands send to a swtichman.
```
armiol: `to a switchman.`

PR #573, opened 2017-08-25
Write scripts for updating of common files for projects

dmytro-grankin:
This PR adds scripts (`.sh` and `.bat` version), that allow to update files from a remote source (currently from `core-java` ). Also `IDEA` configuration was slightly updated.

The script creates a new branch and pushes the updated files to this branch. If there are no changes, the script does nothing.

For the details see [wiki](https://github.com/SpineEventEngine/core-java/wiki/IntelliJ-IDEA-Configuration#automatic-configuration-updates) page.


dmytro-grankin:
@armiol PTAL.

PR #577, opened 2017-08-31
Allow to specify entity column names via the annotation

dmytro-grankin:
This PR brings an opportunity to specify a name for an entity column as the `name` property of the entity column annotation. The name will be used as a name for storing in a storage and does not affect the querying.

Since the change, there are possibilities to specify the invalid names. To notify a user about invalid columns, they are validated after a repository registration.

Entity column names cannot be changed in subclasses, i.e. a getter for an entity column should be annotated with the column annotation only once. If a subclass annotates an entity column getter with the column annotation, an exception will be raised.

Other changes:
- The `javax.persistence.Column` was replaced by the new Spine annotation with the same name.
- `Column` was renamed to `EntityColumn`.
- Code style settings was fixed (ternary operator wrapping).

Wrapping before:
```java
final String nameForStore = nameFromAnnotation.isPresent()
       ? nameFromAnnotation.get()
       : nameForQuery;
```

After:
```java
final String nameForStore = nameFromAnnotation.isPresent()
                            ? nameFromAnnotation.get()
                            : nameForQuery;
```


dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

server/src/test/java/io/spine/server/entity/VisibilityGuardShould.java
```diff
@@ -174,4 +182,27 @@ private HiddenRepository() {
             super();
         }
     }
+
+    private static class EntityWithInvalidColumns extends AbstractVersionableEntity<String, Any> {
```
armiol: Please migrate to `...TestEnv`-based approach. See `server` module source code for numerous examples.

dmytro-grankin:
@armiol PTAL.

armiol:
@dmytro-grankin please update the framework version to the next one after you merge with the latest changes.

PR #578, opened 2017-09-01
Externalize `OperatorEvaluator`

dmdashenkov:
In this PR we externalize `OperatorEvaluator` class to make it available in `datastore` Storage implementation.


dmdashenkov:
@armiol, PTAL.

PR #581, opened 2017-09-19
Add cleaning of `EntityContext` to `EEntity`

dmytro-grankin:
This PR fixes work of `EventStore`. Before, `EventStore` stored redundant data such as enrichments.

Such a behavior could cause performance issues, especially after addition of `React` annotation.
Because `EventContext.origin` may contain another `EventContext` and so on and so on. And every `EventContext` may contain information about `Enrichment` of an event.

So from now, `EEntity` doesn't contain data about the enrichment of the event (which the entity represents), enrichment of the origin (if the origin is `EventContext` or `RejectionContext`) and nested origins.


server/src/test/java/io/spine/server/event/EventStoreShould.java
```diff
@@ -210,21 +221,102 @@ public void fail_to_store_events_of_different_tenants_in_a_single_operation() {
...
+        final Event event = projectCreated(Time.getCurrentTime());
+        final Event enriched = event.toBuilder()
+                                    .setContext(event.getContext()
+                                                     .toBuilder()
+                                                     .setRejectionContext(origin))
+                                    .build();
+        eventStore.append(enriched);
+        final MemoizingObserver<Event> observer = memoizingObserver();
+        eventStore.read(EventStreamQuery.getDefaultInstance(), observer);
+        final RejectionContext modifiedOrigin = observer.responses()
```
alexander-yevsyukov: What did you mean by `modifiedOrigin`? This question applies to this variable and similar below.

server/src/main/java/io/spine/server/event/EEntity.java
```diff
@@ -124,4 +129,37 @@ public String getType() {
...
+     *
+     * @param event the event to compact
+     * @return the compacted event
+     */
+    private static Event compact(Event event) {
+        final EventContext context = event.getContext();
+        final EventContext.Builder resultContext = context.toBuilder()
+                                                          .clearEnrichment();
+        final EventContext originEventContext = context.getEventContext();
+        if (!isDefault(originEventContext)) {
```
alexander-yevsyukov: Please use case detection API from Protobuf.

server/src/main/java/io/spine/server/event/EEntity.java
```diff
@@ -162,7 +162,8 @@ private static Event compact(Event event) {
                 // Does nothing.
                 break;
             case ORIGIN_NOT_SET:
-                throw newIllegalStateException("EventContext.origin should be set.");
+                // Does nothing because the origins is `null`.
+                break;
             default:
                 throw newIllegalStateException("Not all of `OriginCase` values were handled.");
```
alexander-yevsyukov: Have the message which says something like “Unsupported origin case encountered: %s”.

server/src/main/java/io/spine/server/event/EEntity.java
```diff
@@ -162,7 +162,8 @@ private static Event compact(Event event) {
                 // Does nothing.
                 break;
             case ORIGIN_NOT_SET:
-                throw newIllegalStateException("EventContext.origin should be set.");
+                // Does nothing because the origins is `null`.
```
alexander-yevsyukov: Protofub doesn't have nulls. This comment is misleading.

PR #582, opened 2017-09-20
Prevent irrelevant messages appearing in Travis build logs

dmytro-grankin:
This PR prevents appearing of warnings, which are not important during Travis build, but pollute the log (Javadoc and code style warnings).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -159,7 +159,9 @@ subprojects {
     }
 
     // Specific setup for a Travis build,
-    // which prevents appearing of irrelevant messages in build logs.
+    // which prevents appearing of warning messages in build logs.
+    //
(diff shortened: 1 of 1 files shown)


dmytro-grankin:
@alexander-yevsyukov PTAL.

build.gradle
```diff
@@ -156,6 +158,23 @@ subprojects {
         }
     }
 
+    // Specific setup for a Travis build,
+    // which prevents appearing of irrelevant messages in build logs.
```
alexander-yevsyukov: Please change `irrelevant` to `warning`. Please also add the following line:

`It is expected that warnings are viewed and analyzed in the local build logs.`

PR #587, opened 2017-09-26
Publish a separate artifact with the Proto definitions

dmdashenkov:
From now we publish the `.proto` definitions of `client` module in a separate artifact.

The artifact includes all the Protos from the `client` module and its dependencies.

To depend on the artifact:
```groovy
dependencies {
    compile "io.spine:spine-client:$version@proto"
}
```

The artifact itself is a JAR file. It contains all the `.proto` definitions preserving their relative path, i.e. if the file `foo.proto` is declared in the project dependency under `mydependency/foo/foo.proto`, it can be found in the archive under exact `mydependency/foo/foo.proto` path.

There are no files except the `.proto` files in the archive.

_Note:_ the archive may also contain empty directories.

To enable other modules to publish such artifacts, apply `script/assemble-proto.gradle` to the module Gradle project.

### DIFF
--- a/scripts/assemble-proto.gradle
+++ b/scripts/assemble-proto.gradle
@@ -69,7 +69,7 @@ buildscript {
  */
 Collection<File> collectProto() {
     final def dependencies = configurations.compile.files
-    final def jarFiles = dependencies.collect { it.name }
+    final def jarFiles = dependencies.collect { JarFileName.ofFile(it) }
     final def result = new HashSet<>()
     for (final File jarFile in dependencies) {
         if (jarFile.name.endsWith(".jar")) {
@@ -89,11 +89,11 @@ Collection<File> collectProto() {
  * Returns the root directory containing a Proto package.
  *
  * @param member the member File of the Proto package
- * @param jars the full listing of the project JAR dependencies
+ * @param jarNames the full listing of the project JAR dependencies
  */
-File getProtoRoot(File member, Collection<String> jars) {
+File getProtoRoot(final File member, final Collection<JarFileName> jarNames) {
     File pkg = member
-    while (!jars.contains(jarName(pkg.parentFile))) {
+    while (!jarNames.contains(jarName(pkg.parentFile))) {
         pkg = pkg.parentFile
     }
     return pkg.parentFile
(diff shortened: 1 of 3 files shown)


dmdashenkov:
@armiol, PTAL.

scripts/assemble-proto.gradle
```diff
@@ -0,0 +1,140 @@
...
+ * A predicate checking if the given {@link File} is a {@code .proto}
+ */
+enum IsProtoFile implements Predicate<File> {
+
+    PREDICATE
+
+    @Override
+    boolean apply(final File input) {
+        return (input?.isDirectory() && input.list().length != 0) ||
+                (input?.isFile() && input?.name?.endsWith(".proto"))
```
armiol: Please extract `.proto` to increase its visibility.

scripts/assemble-proto.gradle
```diff
@@ -0,0 +1,140 @@
...
+
+            artifact assembleProto {
+                classifier = "proto"
+            }
+        }
+    }
+}
+
+/**
+ * A predicate checking if the given {@link File} is a {@code .proto}
```
armiol: Please check that you have dots when ending each sentence.

scripts/assemble-proto.gradle
```diff
@@ -0,0 +1,140 @@
...
+    }
+    result.addAll(sourceSets.main.proto.srcDirs)
+    return result
+}
+
+/**
+ * Returns the root directory containing a Proto package.
+ *
+ * @param member the member File of the Proto package
+ * @param jars the full listing of the project JAR dependencies
```
armiol: Having those as strings is fragile. At least describe and check the format. It's of course better to have a value object instead of a `String`.

Also, I wouldn't call it jars. They are jarPaths or something like that (depending on the value object name).

scripts/assemble-proto.gradle
```diff
@@ -0,0 +1,140 @@
...
+        mavenLocal()
+    }
+
+    dependencies {
+        classpath "com.google.guava:guava:$guavaVersion"
+    }
+}
+
+/**
+ * Collects all the directories from current project and its dependencies (including zip tree
```
armiol: Please have a preface explaining how to use this file. Just like you did in the PR description.

dmdashenkov: Done.

PR #588, opened 2017-09-27
Default state registry thread safety fix

dmitry-ganzha:
When using _DefaultStateRegistry_ in a multithreaded environment, the following problem occurred in this code:
```java
final DefaultStateRegistry registry = DefaultStateRegistry.getInstance();
if (!registry.contains(entityClass)) {
    final S state = createDefaultState();
    registry.put(entityClass, state);
}
```
Method put code:
```java
void put(Class<? extends Entity> entityClass, Message state) {
     if (contains(entityClass)) {
         throw newIllegalArgumentException("This class is registered already: %s", entityClass);
     }
     defaultStates.put(entityClass, state);
}
```
Actual result: Unpredictable result, can throw exception(This class is registered already) from **DefaultStateRegistry#put**.
Expected result: Don't throw exception and put state only **once** into _defaultStates_.

To solve this problem, it was decided to replace the put method with putOrGet and make it thread safe as well as other methods(**DefaultStateRegistry#get**, **DefaultStateRegistry#contains**), which inserts the value only once, and if the value already exists, it returns the already existing value.

Method **DefaultStateRegistry#putOrGet** code:
```java
Message putOrGet(Class<? extends Entity> entityClass, Message state) {
     lock.writeLock().lock();
     try {
         if (!contains(entityClass)) {
             defaultStates.put(entityClass, state);
         }
         final Message result = get(entityClass);
         return result;
     } finally {
         lock.writeLock().unlock();
     }
}
```

Example how to use new **DefaultStateRegistry#putOrGet** instead of
**DefaultStateRegistry#put** :

Before:
```java
public S getDefaultState() {
     final Class<? extends Entity> entityClass = getClass();
     final DefaultStateRegistry registry = DefaultStateRegistry.getInstance();
     if (!registry.contains(entityClass)) {
         final S state = createDefaultState();
         registry.put(entityClass, state);
     }
     @SuppressWarnings("unchecked")
     // cast is safe because this type of messages is saved to the map
     final S defaultState = (S) registry.get(entityClass);
     return defaultState;
}
```

After:
```java
public S getDefaultState() {
     final Class<? extends Entity> entityClass = getClass();
     final DefaultStateRegistry registry = DefaultStateRegistry.getInstance();
     final S state = createDefaultState();
     @SuppressWarnings("unchecked")
     // cast is safe because this type of messages is saved to the map
     final S defaultState = (S) registry.putOrGet(entityClass, state);
     return defaultState;
}
```

### DIFF
--- a/server/src/test/java/io/spine/server/entity/DefaultStateRegistryShould.java
+++ b/server/src/test/java/io/spine/server/entity/DefaultStateRegistryShould.java
@@ -20,6 +20,7 @@
 
 package io.spine.server.entity;
 
+import com.google.common.collect.Lists;
 import com.google.protobuf.Timestamp;
 import org.junit.Before;
 import org.junit.Test;
@@ -32,9 +33,11 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
+import static com.google.common.collect.Lists.newArrayListWithExpectedSize;
 import static com.google.common.collect.Maps.newConcurrentMap;
 import static io.spine.Identifier.newUuid;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
@@ -47,6 +50,7 @@ import static org.mockito.Mockito.verify;
 public class DefaultStateRegistryShould {
 
     private static final String DEFAULT_STATES_FIELD_NAME = "defaultStates";
+
     private DefaultStateRegistry registry;
     private Map<Object, Object> spyMap;
 
@@ -54,19 +58,13 @@ public class DefaultStateRegistryShould {
     public void setUp() {
         spyMap = spy(newConcurrentMap());
         registry = DefaultStateRegistry.getInstance();
-        try {
-            final Field defaultStates = registry.getClass()
-                                          .getDeclaredField(DEFAULT_STATES_FIELD_NAME);
-            defaultStates.setAccessible(true);
-            defaultStates.set(registry, spyMap);
-        } catch (NoSuchFieldException | IllegalAccessException ignored) {
-        }
+        injectField(registry, DEFAULT_STATES_FIELD_NAME, spyMap);
     }
 
     @Test
     public void verify_put_invoked_one_time_when_invoke_get_default_state_in_multithreaded_environment() {
         final int numberOfEntities = 1000;
-        final Collection<Callable<Object>> tasks = new ArrayList<>();
+        final Collection<Callable<Object>> tasks = newArrayListWithExpectedSize(numberOfEntities);
         for (int i = 0; i < numberOfEntities; i++) {
             tasks.add(Executors.callable(new Runnable() {
                 @Override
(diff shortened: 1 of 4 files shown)


server/src/test/java/io/spine/server/entity/DefaultStateRegistryShould.java
```diff
@@ -24,22 +24,94 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
 /**
  * @author Alexander Yevsyukov
+ * @author Dmitry Ganzha
  */
 public class DefaultStateRegistryShould {
 
+    private static final String DEFAULT_STATES_FIELD_NAME = "defaultStates";
```
armiol: Please have an empty line after the `static final` member declaration.

server/src/test/java/io/spine/server/entity/DefaultStateRegistryShould.java
```diff
@@ -24,22 +24,94 @@
...
     private DefaultStateRegistry registry;
+    private Map<Object, Object> spyMap;
 
     @Before
     public void setUp() {
-        registry = new DefaultStateRegistry();
+        spyMap = spy(newConcurrentMap());
+        registry = DefaultStateRegistry.getInstance();
+        try {
```
armiol: Please extract this `try {} catch {}` into a method. E.g. `injectField(Object target, String fieldName, Object injectableValue)`. 

In this way the code will be more self-documenting than now.

server/src/test/java/io/spine/server/entity/DefaultStateRegistryShould.java
```diff
@@ -24,22 +24,94 @@
...
+            defaultStates.setAccessible(true);
+            defaultStates.set(registry, spyMap);
+        } catch (NoSuchFieldException | IllegalAccessException ignored) {
+        }
+    }
+
+    @Test
+    public void verify_put_invoked_one_time_when_invoke_get_default_state_in_multithreaded_environment() {
+        final int numberOfEntities = 1000;
+        final Collection<Callable<Object>> tasks = new ArrayList<>();
```
armiol: I think that a better way to express your intention would be to use Guava's API:
```java
... = Lists.newArrayListWithExpectedSize(numberOfEntities);
``` 
(and use static import for this method).

Also, maybe a collection of `Callable<Void>` would look better. I'm not sure on the latter though.

server/src/test/java/io/spine/server/entity/DefaultStateRegistryShould.java
```diff
@@ -24,22 +24,94 @@
...
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void do_not_accept_same_entity_class_twice() {
-        registry.put(TimerSnapshot.class, Timestamp.getDefaultInstance());
-        registry.put(TimerSnapshot.class, Timestamp.getDefaultInstance());
+    @Test
+    public void verify_put_invoked_one_time_when_invoke_put_or_get_in_multithreaded_environment() {
+        final int numberOfEntities = 1000;
+        final Collection<Callable<Object>> tasks = new ArrayList<>();
```
armiol: Same comments as above apply.

server/src/test/java/io/spine/server/entity/DefaultStateRegistryShould.java
```diff
@@ -24,22 +24,94 @@
...
-        registry = new DefaultStateRegistry();
+        spyMap = spy(newConcurrentMap());
+        registry = DefaultStateRegistry.getInstance();
+        try {
+            final Field defaultStates = registry.getClass()
+                                          .getDeclaredField(DEFAULT_STATES_FIELD_NAME);
+            defaultStates.setAccessible(true);
+            defaultStates.set(registry, spyMap);
+        } catch (NoSuchFieldException | IllegalAccessException ignored) {
+        }
```
armiol: I suppose that in this case we should fail the test, not just ignore the exception.

PR #590, opened 2017-09-28
Exclude empty directories from proto artifact

dmdashenkov:
From now the `assemble-proto` Gradle plugin does not include empty directories to the generated JAR artifact.

### DIFF
--- a/scripts/assemble-proto.gradle
+++ b/scripts/assemble-proto.gradle
@@ -135,7 +135,7 @@ task assembleProto(type: Jar) {
  *
  * @param candidate the {@link File} to check
  * @return {@code true} if the {@code candidate} {@linkplain #isProtoFile is a Protobuf file} or
- * a directory containing at least one Protobuf file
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@armiol, PTAL.

scripts/assemble-proto.gradle
```diff
@@ -125,11 +124,45 @@ JarFileName jarName(final File jar) {
...
+
+/**
+ * Checks if the given abstract pathname represents either a {@code .proto} file, or a directory
+ * containing proto files.
+ *
+ * <p>If {@code candidate} is a directory, scans its children recursively.
+ *
+ * @param candidate the {@link File} to check
+ * @return {@code true} if the {@code candidate} {@linkplain #isProtoFile is a Protobuf file} or
+ * a directory containing at least one Protobuf file
```
armiol: This line does not seem to be properly indented.

PR #591, opened 2017-09-29
Fix inconsistency of storing commands in `CommandStore` in case of concurrent access

dmitry-ganzha:
**Issue Summary**

At the moment `MultitenantStorage` is used as an abstract base for in-memory storages, that Spine provides for a multitenant bounded contexts. However, even if the context is single-tenant, the same abstract base class is used.

The issue is that a single-tenant `BoundedContext`, which is being accessed from outside from within several threads, fails to use this storage properly. In particular, sending several commands simultaneously leads to data loss on a `CommandStore` level.

In scope of such a scenario, `MultitenantStorage.getStorage()` is called to obtain the data slice for a single tenant configured. First call to this storage leads to creation of a storage slice for a single tenant. All subsequent calls to `getStorage()` read the previously created slice from tenant-ID-to-slice map and re-use the stored value.

**Expected Result** 

`MultitenantStorage.getStorage()` returns the same result in every thread.

**Actual Result**

`MultitenantStorage.getStorage()` creates and returns different instances of storage slices — one per calling thread. In this way only the last slice created and returned by `getStorage()` is actually used for further execution. This leads to data corruption.

**Proposed Solution**

Within this PR the `getStorage()` method is made thread-safe.

Additionally, similar concurrency-related issues were found in the related `TenantRecords` storage slice implementation. In particular, a `HashMap` was used to manage the data in multithreaded environment. This issue has also been addressed.

The framework version has been updated to `0.9.71-SNAPSHOT`.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -25,7 +25,7 @@
  *  as we want to manage the versions in a single source.
  */
  
-def final SPINE_VERSION = '0.9.70-SNAPSHOT'
+def final SPINE_VERSION = '0.9.71-SNAPSHOT'
 
 ext {
     // The version of the modules in this project.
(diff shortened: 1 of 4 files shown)


dmitry-ganzha:
@armiol PTAL

PR #594, opened 2017-10-05
Check default state on registering Repository in BoundedContext

dmitry-ganzha:
This PR adds an additional check that default state exists, when registering the `Repository` in the `BoundedContext`. Therefore, if during the registration of the repository the default state for entity is not found, an exception will be thrown.

PR changes add possibility to get default state of entity by entity class via `Model`. This implies, that Spine will have single point to retrieve default state of entity. To get the default state, use the following method `Model.getDefaultState(Class<? extends Entity> cls)`.

PR introduces changes to `AbstractEntity`, which used the DefaultStateRegistry mechanics to get the default state, the approach described above is used now.

The framework version has been updated to 0.9.73-SNAPSHOT.

### DIFF
--- a/server/src/main/java/io/spine/server/BoundedContext.java
+++ b/server/src/main/java/io/spine/server/BoundedContext.java
@@ -66,6 +66,7 @@ import static io.spine.validate.Validate.checkNotEmptyOrBlank;
  *
  * @author Alexander Yevsyukov
  * @author Mikhail Melnik
+ * @author Dmitry Ganzha
  */
 public final class BoundedContext
         extends IntegrationEventSubscriberGrpc.IntegrationEventSubscriberImplBase
@@ -254,7 +255,7 @@ public final class BoundedContext
      * <p>If the repository does not have a storage assigned, it will be initialized
      * using the {@code StorageFactory} associated with this bounded context.
      *
-     * <p>Checks whether there is a default state for passed entity or aggregate type.
+     * <p>Checks whether there is a default state for entity or aggregate type of the repository.
      *
      * @param repository the repository to register
      * @param <I>        the type of IDs used in the repository
--- a/server/src/main/java/io/spine/server/BoundedContext.java
+++ b/server/src/main/java/io/spine/server/BoundedContext.java
@@ -40,6 +40,7 @@ import io.spine.server.event.EventFactory;
 import io.spine.server.integration.IntegrationBus;
 import io.spine.server.integration.IntegrationEvent;
 import io.spine.server.integration.grpc.IntegrationEventSubscriberGrpc;
+import io.spine.server.model.Model;
 import io.spine.server.rejection.RejectionBus;
 import io.spine.server.stand.Stand;
 import io.spine.server.stand.StandStorage;
@@ -253,6 +254,8 @@ public final class BoundedContext
      * <p>If the repository does not have a storage assigned, it will be initialized
      * using the {@code StorageFactory} associated with this bounded context.
      *
+     * <p>Checks whether there is a default state for passed entity or aggregate type.
+     *
      * @param repository the repository to register
      * @param <I>        the type of IDs used in the repository
      * @param <E>        the type of entities or aggregates
@@ -260,6 +263,10 @@ public final class BoundedContext
      */
     public <I, E extends Entity<I, ?>> void register(Repository<I, E> repository) {
         checkNotNull(repository);
+        final Message defaultState = Model.getInstance()
+                                          .getDefaultState(repository.getEntityClass());
+        checkNotNull(defaultState);
+
         repository.setBoundedContext(this);
         guard.register(repository);
         repository.onRegistered();
--- a/server/src/main/java/io/spine/server/BoundedContext.java
+++ b/server/src/main/java/io/spine/server/BoundedContext.java
@@ -255,7 +255,7 @@ public final class BoundedContext
      * <p>If the repository does not have a storage assigned, it will be initialized
      * using the {@code StorageFactory} associated with this bounded context.
      *
-     * <p>Checks whether there is a default state for entity or aggregate type of the repository.
+     * <p>Checks whether there is a default state for entity type.
      *
      * @param repository the repository to register
      * @param <I>        the type of IDs used in the repository
--- a/server/src/main/java/io/spine/server/entity/AbstractEntity.java
+++ b/server/src/main/java/io/spine/server/entity/AbstractEntity.java
@@ -22,7 +22,6 @@ package io.spine.server.entity;
 
 import com.google.common.base.MoreObjects;
 import com.google.protobuf.Message;
-import io.spine.protobuf.Messages;
 import io.spine.server.model.Model;
 import io.spine.string.Stringifiers;
 import io.spine.validate.ConstraintViolation;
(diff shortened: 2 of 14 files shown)


dmitry-ganzha:
@armiol PTAL

server/src/main/java/io/spine/server/entity/EntityClass.java
```diff
@@ -98,7 +98,7 @@ private static ModelError noSuchConstructor(String entityClass, String idClass)
     /**
      * Retrieves the state class of the passed entity class.
      */
-    static <S extends Message> Class<S> getStateClass(Class<? extends Entity> entityClass) {
+    public static <S extends Message> Class<S> getStateClass(Class<? extends Entity> entityClass) {
```
armiol: This change requires some discussion.

server/src/test/java/io/spine/server/model/ModelShould.java
```diff
@@ -73,6 +76,12 @@ public void not_allow_more_than_one_command_duplication() {
         }
     }
 
+    @Test
+    public void get_default_state_should_not_return_null() {
```
armiol: The naming pattern is different. Let's discuss.

server/src/main/java/io/spine/server/entity/AbstractEntity.java
```diff
@@ -135,42 +134,10 @@ void setState(S newState) {
...
-        return result;
-    }
-
-    /**
-     * Obtains the class of the entity state.
-     */
-    private Class<S> getStateClass() {
-        final Class<? extends AbstractEntity> entityClass = getClass();
-
-        //TODO:2017-08-21:alexander.yevsyukov: Use the below code to overcome the issue in the test
```
armiol: Let's discuss if this workaround isn't required anymore. Just to be sure.

armiol:
@dmitry-ganzha 

> Also PR adds to additional check that default state exists, when registering the Repository in the BoundedContext. If the check fails, bootstrap of application will be interrupted and an exception will be thrown.

I'm not sure I understand how we test this behavior. Let's discuss this as well.

dmitry-ganzha:
@armiol PTAL

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -253,13 +255,19 @@ public TenantIndex getTenantIndex() {
      * <p>If the repository does not have a storage assigned, it will be initialized
      * using the {@code StorageFactory} associated with this bounded context.
      *
+     * <p>Checks whether there is a default state for entity or aggregate type of the repository.
```
armiol: As long as aggregates are also entities, let's remove `or aggregate` from this sentence.
`of the repository` is also obvious, as repositories always manage entities. So please remove this portion as well.

server/src/test/java/io/spine/server/model/ModelShould.java
```diff
@@ -73,6 +76,12 @@ public void not_allow_more_than_one_command_duplication() {
         }
     }
 
+    @Test
+    public void return_default_state() {
```
armiol: Should probably be `return_default_state_for_entity_class()`.

server/src/test/java/io/spine/server/model/ModelShould.java
```diff
@@ -73,6 +76,12 @@ public void not_allow_more_than_one_command_duplication() {
...
+        final Message defaultState = model.getDefaultState(MAggregate.class);
+        assertNotNull("Default state should not be null.", defaultState);
```
armiol: `"Default state cannot be null for the entity class."`

server/src/test/java/io/spine/server/bc/BoundedContextShould.java
```diff
@@ -337,4 +351,53 @@ public void do_not_expose_invisible_aggregate() {
...
+        } catch (NoSuchFieldException e) {
+            fail("Field " + fieldName + " should exist.");
+        } catch (IllegalAccessException e) {
+            fail(e.getMessage());
+        }
+        assertNotNull(result);
+        return result;
+    }
+
+    private void injectField(Object target, String fieldName, Object injectableValue) {
```
armiol: `injectableValue` --> `valueToInject`. It feels a bit different and will suit better in this case.

server/src/test/java/io/spine/server/bc/BoundedContextShould.java
```diff
@@ -337,4 +351,53 @@ public void do_not_expose_invisible_aggregate() {
         assertFalse(boundedContext.findRepository(SecretProject.class)
                                   .isPresent());
     }
+
+    @Test(expected = NullPointerException.class)
+    public void throw_NPE_when_default_state_is_null() {
```
armiol: Perhaps this test does not belong here. Or it should be named differently. My clue is that the following sentence does not make much sense:

"BoundedContext should ... throw NPE when default state is null".

Why throw? Which default state? etc. 

I'd suggest to rename the test method and probably also add some Javadocs to it to explain what we are testing.

dmitry-ganzha:
@armiol PTAL

server/src/main/java/io/spine/server/entity/EntityClass.java
```diff
@@ -97,8 +98,13 @@ private static ModelError noSuchConstructor(String entityClass, String idClass)
 
     /**
      * Retrieves the state class of the passed entity class.
+     *
+     * <p>Method is public because another way to get state class is create {@code EntityClass}
```
armiol: This is too much information for a user who just needs to understand what this method is for.

This is enough:
```
<p>Though this method is {@code public}, it is <em>not</em> considered a part of the public API. It is used internally by other framework routines and not designed for efficient execution by Spine users.
```

server/src/test/java/io/spine/server/bc/BoundedContextShould.java
```diff
@@ -337,4 +350,61 @@ public void do_not_expose_invisible_aggregate() {
         assertFalse(boundedContext.findRepository(SecretProject.class)
                                   .isPresent());
     }
+
+    /**
+     * In the process of registering {@link Repository} in the {@link BoundedContext},
```
armiol:
```
This test checks, whether {@code BoundedContext} properly handles the issues upon repository registration.

<p>In particular, we intentionally break  an interaction between {@code Model} and a {@code BoundedContext} on attempt to ensure there is an entity default state present for the given repository instance. 

<p>The expected behavior of {@code BoundedContext} instance is to fail fast in case such a default state is absent. 

<p>In real-life this use case can never happen given the current implementation of {@code Model} and {@code BoundedContext}. However, previously such an issue was caught. Therefore this test case ensures it's never happening again.
```

server/src/test/java/io/spine/server/bc/BoundedContextShould.java
```diff
@@ -337,4 +350,61 @@ public void do_not_expose_invisible_aggregate() {
...
+     * method {@link io.spine.server.model.Model#getDefaultState(Class) getDefaultState} is called,
+     * which gets the default state and check is performed that the default state is not null,
+     * if it is null an exception will be thrown.
+     *
+     * <p>In this test, it is checked whether the mechanism described above works correctly.
+     */
+    @Test(expected = NullPointerException.class)
+    public void throw_NPE_when_register_repository_and_default_state_is_null() {
```
armiol: `...when_registering_repository...`.

PR #595, opened 2017-10-06
Versioning policy update: # of `Event`s emitted --> # of version increments. Even if state is NOT modified

armiol:
In the lifecycle of `Aggregate` handling a command is always followed by emitting one or more events. In turn they are fed to the `Aggregate`'s event appliers and modify its state. All event appliers are called in scope of a transaction.

When a transaction is committed, there is a check of the state modifications made. If there weren't any, no increment of version is performed.

However, there may be a case when some events are emitted after handling a command, but applying them does not change the `Aggregate` state — perhaps, because the `Aggregate` instance is already in the desired state. In this case events get `N + 1`, `N + 2`, ... version numbers — where `N` is an initial version of this `Aggregate`.  However, an `Aggregate` is left with `N`. 

The problem is that next time we load this `Aggregate` instance from the events, it will get `N + 2` version — as the `Aggregate` version always follows the version of its source event, when loading from `EventStore`. So there's a collision of behavior.

This PR fixes this issue.

From now on emitting any event will lead to a version increment, has the instance been modified or not.

The framework version has been increased  to `0.9.74-SNAPSHOT`.

### DIFF
--- a/server/src/test/java/io/spine/server/aggregate/AggregateShould.java
+++ b/server/src/test/java/io/spine/server/aggregate/AggregateShould.java
@@ -22,6 +22,7 @@ package io.spine.server.aggregate;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.google.protobuf.Message;
 import com.google.protobuf.Timestamp;
 import io.spine.Identifier;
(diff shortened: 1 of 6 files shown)


armiol:
@alexander-yevsyukov PTAL.

server/src/test/java/io/spine/server/aggregate/AggregateShould.java
```diff
@@ -659,4 +705,50 @@ private void dispatchCommands(Command... commands) {
...
+            return event;
+        }
+
+        @Assign
+        List<Message> handle(AggCancelProject cmd, CommandContext ctx) {
+            final AggProjectPaused firstPaused = projectPaused(cmd.getProjectId());
+            final AggProjectCancelled thenCancelled = projectCancelled(cmd.getProjectId());
+            return Lists.<Message>newArrayList(firstPaused, thenCancelled);
+        }
+
```
alexander-yevsyukov: Please remove excessive line.

PR #598, opened 2017-10-11
Improve tests for `AggregateStorage`

dmytro-grankin:
This PR improves testing of `AggregateStorage`.

The following test cases were added:
* History reading until the most recent snapshot.
* Sorting of event records by version.
* Sorting of event records by version rather then by timestamp. Behavior of in-memory implementation was fixed according to this scenario.

The version was increased to `0.9.76-SNAPSHOT`.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -25,7 +25,7 @@
  *  as we want to manage the versions in a single source.
  */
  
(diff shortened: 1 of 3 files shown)


dmytro-grankin:
@armiol PTAL.

PR #600, opened 2017-10-12
Ensure that `Projection`s and `ProcessManager`s declare at least one message handler during repository registration

armiol:
Before this PR it was possible to register `ProjectionRepository` and `ProcessManagerRepository` instances  in `BoundedContext`, even if the managed entity did not declare any handler methods (i.e. nothing marked with `@Subscribe` or `@React`).

At the same time such a check was already implemented for `AggregateRepository`.

This changeset addresses this behavior inconsistency and prevents Spine users from such a mistake.

Now, mandatory checks for the proper entity declaration are implemented. They are performed upon a repository registration in the `BoundedContext`. If an entity has no message handler methods, an `IllegalStateException` is raised — similar to the behavior implemented for `AggregateRepository`.

The version of the framework has been set to `0.9.77-SNAPSHOT`.

### DIFF
--- a/server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
+++ b/server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
@@ -30,6 +30,8 @@ import io.spine.core.EventEnvelope;
 import io.spine.core.RejectionClass;
 import io.spine.core.RejectionEnvelope;
 import io.spine.server.BoundedContext;
+import io.spine.server.bus.Bus;
+import io.spine.server.bus.MessageDispatcher;
 import io.spine.server.command.CommandHandlingEntity;
 import io.spine.server.commandbus.CommandBus;
 import io.spine.server.commandbus.CommandDispatcherDelegate;
@@ -52,6 +54,7 @@ import javax.annotation.CheckReturnValue;
 import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.util.Exceptions.newIllegalStateException;
 
 /**
  * The abstract base for Process Managers repositories.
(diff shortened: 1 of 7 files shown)


armiol:
@alexander-yevsyukov PTAL.

armiol:
@dmdashenkov PTAL.

server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
```diff
@@ -100,36 +103,70 @@ protected ProcessManagerRepository() {
...
-                          .register(commandDispatcher);
+        final boolean handlesCommands = register(boundedContext.getCommandBus(),
+                                                 DelegatingCommandDispatcher.of(this));
+        final boolean handlesDomesticRejections = register(boundedContext.getRejectionBus(),
+                                                           rejDispatcher);
+        final boolean handlesExternalRejections = register(boundedContext.getIntegrationBus(),
+                                                           rejDispatcher.getExternalDispatcher());
+        final boolean handlesDomesticEvents = !getMessageClasses().isEmpty();
+        final boolean handlesExternalEvents = !getExternalEventDispatcher().getMessageClasses()
+                                                                          .isEmpty();
```
dmdashenkov: Please fix the alignment.

server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
```diff
@@ -100,36 +103,70 @@ protected ProcessManagerRepository() {
...
+
+        final boolean subscribesToEvents = handlesDomesticEvents || handlesExternalEvents;
+        final boolean reactsUponRejections = handlesDomesticRejections || handlesExternalRejections;
+
+        if (!handlesCommands && !subscribesToEvents && !reactsUponRejections) {
+            throw newIllegalStateException(
```
dmdashenkov: This branch appears not covered.

server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
```diff
@@ -100,36 +103,70 @@ protected ProcessManagerRepository() {
...
-                          .register(extRejectionDispatcher);
+    /**
+     * Registers the given dispatcher in the bus if there is at least one message class declared
+     * for dispatching by the dispatcher.
+     *
+     * @param bus        the bus to register dispatchers in
+     * @param dispatcher the dispatcher to register
+     * @param <D>        the type of dispatcher
+     * @return
+     * {@code true} if there are message classes to dispatch by the given dispatchers,
```
dmdashenkov: This alignment appears strange.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -91,6 +92,28 @@ EventStore getEventStore() {
...
+    @Override
+    public void onRegistered() {
+        super.onRegistered();
+
+        final boolean noEventSubscriptions = getMessageClasses().isEmpty();
+        if (noEventSubscriptions) {
+            final boolean noExternalSubscriptions = getExternalEventDispatcher().getMessageClasses()
+                                                                                .isEmpty();
+            if (noExternalSubscriptions) {
+                throw newIllegalStateException(
```
dmdashenkov: This branch appears not covered.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -91,6 +92,28 @@ EventStore getEventStore() {
...
+                        "Projections of the repository %s have neither domestic nor external " +
```
dmdashenkov: Let's have `Projections of type ...` instead `of repository ...`. It seems more natural.

server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
dmdashenkov: This might be a bug on __codecov__, but please check for sure.

armiol: Double-checked. 

Yes, that looks like an outdated **codecov** report.

server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
armiol: Adjusted. My goal was to have `{@code true}` and `{@code false}` inter-aligned.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
armiol: As per discussion, it actually is.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
armiol: As discussed, this an emphasis of a **repository**, which validation mechanism failed to complete a process of _Repository-in-Bounded-Context Registration_. 

So to me it makes sense.

PR #602, opened 2017-10-13
Remove an enrichment information from events before storing to `AggregateStorage`

dmytro-grankin:
Storage of `Aggregate` event records stores event enrichments, that aren't needed and take a lot of space.

To fix this behavior, `AggregateStorage.writeEvent(...)` method was changed and now it clears enrichments from the event context and the enrichment from the first-level origin of the specified event. But the event will still contain enrichments from second-level and deeper origins, because their removal is a heavy performance operation.

Other changes:
* `EEntity` was fixed. Now, it removes only enrichments from an event.
* `GivenEnrichment` utility was introduced.
* The framework version was increased to `0.9.78-SNAPSHOT`.

### DIFF
--- a/core/src/main/java/io/spine/core/Events.java
+++ b/core/src/main/java/io/spine/core/Events.java
@@ -262,9 +262,10 @@ public final class Events {
                 throw newIllegalStateException("Unsupported origin case is encountered: %s",
                                                originCase);
         }
-        return event.toBuilder()
-                    .setContext(resultContext)
-                    .build();
+        final Event result = event.toBuilder()
+                                  .setContext(resultContext)
+                                  .build();
+        return result;
     }
 
     /**
--- a/core/src/main/java/io/spine/core/Events.java
+++ b/core/src/main/java/io/spine/core/Events.java
@@ -221,8 +221,7 @@ public final class Events {
     /**
      * Clears enrichments of the specified event.
      *
-     * <p>This method may be helpful to decrease a size of an event,
-     * when enrichments aren't important.
+     * <p>Use this method to decrease a size of an event, if enrichments aren't important.
      *
      * <p>A result won't contain:
      * <ul>
--- a/core/src/main/java/io/spine/core/Events.java
+++ b/core/src/main/java/io/spine/core/Events.java
@@ -36,6 +36,7 @@ import java.util.UUID;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static io.spine.protobuf.AnyPacker.unpack;
+import static io.spine.util.Exceptions.newIllegalStateException;
 import static io.spine.validate.Validate.checkNotEmptyOrBlank;
 
 /**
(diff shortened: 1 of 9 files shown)


dmytro-grankin:
@armiol PTAL.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -218,6 +219,52 @@ public static boolean isExternal(EventContext context) {
...
+                                                         .clearEnrichment());
+                break;
+            case COMMAND_CONTEXT:
+                // Nothing to remove.
+                break;
+            case ORIGIN_NOT_SET:
+                // Does nothing because there is no origin for this event.
+                break;
+            default:
+                throw newIllegalStateException("Unsupported origin case is encountered: %s",
```
armiol: As long as this method is `public` this branch can be tested. Please do.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -218,6 +219,52 @@ public static boolean isExternal(EventContext context) {
...
+    public static Event compact(Event event) {
+        final EventContext context = event.getContext();
+        final EventContext.Builder resultContext = context.toBuilder()
+                                                          .clearEnrichment();
+        final EventContext.OriginCase originCase = resultContext.getOriginCase();
+        switch (originCase) {
+            case EVENT_CONTEXT:
+                resultContext.setEventContext(context.getEventContext()
+                                                     .toBuilder()
+                                                     .clearOrigin()
```
armiol: As discussed, do not clear this. And please mention that in docs.

testutil-core/src/main/java/io/spine/core/given/GivenEnrichment.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * <p>An enrichment instruction is invalid and has random values.
+     *
+     * @return an enabled enrichment
+     * @see Enrichment.ModeCase#getDoNotEnrich()
+     */
+    public static Enrichment enabledEnrichment() {
+        final String key = newUuid();
+        final Any value = pack(toMessage(newUuid()));
+        return Enrichment.newBuilder()
```
armiol: Please have a `result` for such a complex expression before returning it.

core/src/main/java/io/spine/core/Events.java
dmytro-grankin: Done.

testutil-core/src/main/java/io/spine/core/given/GivenEnrichment.java
dmytro-grankin: Done.

core/src/main/java/io/spine/core/Events.java
dmytro-grankin: It cannot be tested even with using of reflection.

dmytro-grankin:
@armiol PTAL.

testutil-core/src/main/java/io/spine/core/given/GivenEnrichment.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class GivenEnrichment {
+
+    private GivenEnrichment() {
+        // Prevent instantiation of this utility class.
+    }
+
+    /**
+     * Creates a non-{@linkplain io.spine.validate.Validate#isDefault(com.google.protobuf.Message)
+     * default} {@link Enrichment}.
```
armiol: Please describe what's in there. And then you'll be able to remove this "non-default" phrase.

testutil-core/src/main/java/io/spine/core/given/GivenEnrichment.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.google.protobuf.Any;
+import io.spine.core.Enrichment;
+
+import static com.google.protobuf.Any.pack;
+import static io.spine.Identifier.newUuid;
+import static io.spine.protobuf.TypeConverter.toMessage;
+
+/**
+ * Factory methods to create {@code Enrichment} instances for test purposes.
```
armiol: `in test purposes`.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -218,6 +219,53 @@ public static boolean isExternal(EventContext context) {
     }
 
     /**
+     * Obtains the compacted version of the event, which should be used for storing.
```
armiol: Let's discuss this description. Right now the name of the method does not correspond with it.

In particular, the description says that this is a procedure one would do before storing an event. And the rest of the content is about enrichments. So to me it looks more like a `clearEnrichments()` method. And even so I would adjust a few things.

However, we can stay with `compact()`. I like the name and the idea behind this name. However, there should be a more generic description, that this one.

Let's discuss this change.

dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -218,6 +219,55 @@ public static boolean isExternal(EventContext context) {
...
+                // Nothing to remove.
+                break;
+            case ORIGIN_NOT_SET:
+                // Does nothing because there is no origin for this event.
+                break;
+            default:
+                throw newIllegalStateException("Unsupported origin case is encountered: %s",
+                                               originCase);
+        }
+        return event.toBuilder()
```
armiol: Please extract a `result` variable.

testutil-core/src/main/java/io/spine/core/given/GivenEnrichment.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Factory methods to create {@code Enrichment} instances in test purposes.
+ *
+ * @author Dmytro Grankin
+ */
+public class GivenEnrichment {
+
+    private GivenEnrichment() {
+        // Prevent instantiation of this utility class.
```
armiol: Please reformat in a way `Events` utility does that.

PR #603, opened 2017-10-17
`README` updates

armiol:
This PR fixes markdown variable values to point to the latest release and a new site.

Also, the version is set to `0.10.1-SNAPSHOT` to avoid accidental re-publishing of previously released version.


armiol:
@alexander-yevsyukov PTAL.

PR #614, opened 2017-11-01
Make Command handling repositories post Rejections

dmdashenkov:
Currently, `CommandBus` is the only place where we post rejection into the `RejectionBus`. It catches `Throwable`s from the `CommandDispatcher.dispatch()` invocation and posts the rejections parsed from the `Throwable`.
This approach may not work if the command handling is postponed.

The suggested approach is following:
 - `AggregateRepository` and `ProcessManagerRepository` (which are command dispatchers) post rejections that are thrown upon their dispatching. 
 - `CommandBus` still posts the rejections which arrive back to it in the call stack.

### DIFF
--- a/core/src/main/java/io/spine/core/Commands.java
+++ b/core/src/main/java/io/spine/core/Commands.java
@@ -22,7 +22,6 @@ package io.spine.core;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicate;
-import com.google.common.base.Throwables;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Message;
 import com.google.protobuf.Timestamp;
@@ -226,36 +225,24 @@ public final class Commands {
     }
 
     /**
-     * Verifies if the exception has command rejection as its
-     * {@linkplain Throwables#getRootCause(Throwable) root cause} .
+     * Produces a {@link Rejection} for the given {@link Command} based on the given
+     * {@linkplain ThrowableMessage cause}.
      *
-     * @param exception the exception to analyze
-     * @return {@code true} if the exception was created because of a command rejection thrown,
-     *         {@code false} otherwise
+     * <p>The given {@link Throwable} should be
+     * {@linkplain Rejections#causedByRejection caused by a Rejection} or
+     * an {@link IllegalArgumentException} is thrown.
+     *
+     * @param command the command to reject
+     * @param cause the rejection cause (may be wrapped into other kinds of {@code Throwable})
+     * @return a {@link Rejection} for the given command
+     * @throws IllegalArgumentException upon an invalid rejection cause
      */
-    public static boolean causedByRejection(Throwable exception) {
-        //TODO:2017-07-26:alexander.yevsyukov: Check against CommandRejection
-        // instead of ThrowableMessage when code generation allows customizing a custom
-        // rejection types instead of `ThrowableMessage`.
-        // See: https://github.com/SpineEventEngine/base/issues/20
-        final Throwable rootCause = Throwables.getRootCause(exception);
-        final boolean result = rootCause instanceof ThrowableMessage;
-        return result;
-    }
-
-    public static ThrowableMessage getCauseRejection(Throwable throwable) {
-        checkNotNull(throwable);
-        checkArgument(causedByRejection(throwable));
-        final Throwable cause = Throwables.getRootCause(throwable);
-        return (ThrowableMessage) cause;
-    }
-
     @Internal
-    public static Rejection reject(Command command, Throwable cause) {
+    public static Rejection reject(Command command, Throwable cause) throws IllegalStateException {
         checkNotNull(command);
         checkNotNull(cause);
 
-        final ThrowableMessage rejectionThrowable = getCauseRejection(cause);
+        final ThrowableMessage rejectionThrowable = Rejections.getCauseRejection(cause);
         final Rejection rejection = Rejections.toRejection(rejectionThrowable, command);
         return rejection;
     }
--- a/core/src/main/java/io/spine/core/Commands.java
+++ b/core/src/main/java/io/spine/core/Commands.java
@@ -238,11 +238,12 @@ public final class Commands {
      * @throws IllegalArgumentException upon an invalid rejection cause
      */
     @Internal
-    public static Rejection reject(Command command, Throwable cause) throws IllegalStateException {
+    public static Rejection rejectWithCause(Command command, Throwable cause)
+            throws IllegalArgumentException {
         checkNotNull(command);
         checkNotNull(cause);
 
-        final ThrowableMessage rejectionThrowable = Rejections.getCauseRejection(cause);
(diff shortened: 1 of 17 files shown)


dmdashenkov:
@armiol, PTAL.

core/src/main/java/io/spine/core/Commands.java
```diff
@@ -225,21 +225,26 @@ public static CommandId checkValid(CommandId id) {
...
-    public static boolean causedByRejection(RuntimeException exception) {
-        //TODO:2017-07-26:alexander.yevsyukov: Check against CommandRejection
-        // instead of ThrowableMessage when code generation allows customizing a custom
-        // rejection types instead of `ThrowableMessage`.
-        // See: https://github.com/SpineEventEngine/base/issues/20
-        final Throwable rootCause = Throwables.getRootCause(exception);
-        final boolean result = rootCause instanceof ThrowableMessage;
-        return result;
+    @Internal
+    public static Rejection reject(Command command, Throwable cause) throws IllegalStateException {
```
armiol: `rejectWithCause`
Also, `IllegalStateException` makes not much sense. In case it's thrown, the _argument_ is wrong. Not the state.

core/src/main/java/io/spine/core/Rejections.java
```diff
@@ -192,4 +193,42 @@ public static boolean isExternal(RejectionContext context) {
...
+     *
+     * <p>Throws an {@link IllegalArgumentException} if the root cause is not
+     * a {@code ThrowableMessage}.
+     *
+     * @param throwable the {@link Throwable} wrapping a {@link ThrowableMessage}
+     * @return the wrapped {@link ThrowableMessage}
+     * @throws IllegalArgumentException upon an invalid {@link Throwable}
+     *                                  {@linkplain Throwables#getRootCause root cause}
+     */
+    static ThrowableMessage getCauseRejection(Throwable throwable)
```
armiol: It's either `getRejectionCause`, where "cause" is a noun. Or just`getCause` — the context is defined by the parent class anyway.

server/src/main/java/io/spine/server/commandbus/CommandBus.java
```diff
@@ -206,6 +207,9 @@ protected Ack doPost(CommandEnvelope envelope) {
             if (causedByRejection(e)) {
                 final ThrowableMessage throwableMessage = (ThrowableMessage) cause;
                 final Rejection rejection = toRejection(throwableMessage, envelope.getCommand());
+                final Class<?> rejectionClass = AnyPacker.unpack(rejection.getMessage())
+                                                         .getClass();
+                Log.log().info("Posting rejection {} to RejectionBus.", rejectionClass.getName());
```
armiol: `Log.log()` is weird.

Also, that's definitely not `info`. It's either "trace" or "debug". I'd think the former.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -287,8 +289,14 @@ public I dispatch(final CommandEnvelope envelope) {
     public void onError(CommandEnvelope envelope, RuntimeException exception) {
         checkNotNull(envelope);
         checkNotNull(exception);
-        if (!causedByRejection(exception)) {
+        if (causedByRejection(exception)) {
```
armiol: That looks like a code duplicate. Let's discuss how to do that better.

dmdashenkov:
@armiol, PTAL again.

armiol:
@dmdashenkov `...if the dispatcher is concurrent` is slang. Please add more details. Also it's not obvious to an average developer, so make sure you describe the whole issue.

server/src/main/java/io/spine/server/commandbus/CommandErrorHandler.java
```diff
@@ -0,0 +1,103 @@
...
+import static java.lang.String.format;
+
+/**
+ * A {@link CommandDispatcher} which handles the dispatching errors.
+ *
+ * @author Dmytro Dashenkov
+ * @see #onError(CommandEnvelope, RuntimeException)
+ */
+@Internal
+public final class CommandErrorHandler<I> implements CommandDispatcher<I> {
```
armiol: Let's discuss this in person. I don't see it like named like that. Given the responsibility it should have had — as we discussed it — it should have had a different name.

server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
```diff
@@ -408,4 +423,13 @@ public void onError(ExternalMessageEnvelope envelope, RuntimeException exception
                      envelope, exception);
         }
     }
+
+    private class CommandHandler implements Function<CommandEnvelope, I> {
```
armiol: We already have "command handler" in this domain.

server/src/test/java/io/spine/core/RejectionsShould.java
```diff
@@ -44,6 +45,12 @@
 import static org.junit.Assert.assertTrue;
 
 /**
+ * Tests for {@link Rejections} utility class.
+ *
+ * <p>The test suite is located under the "server" module since actor request generation
```
armiol: Let's discuss this as well. I don't really understand, why the issue exists in the first place.

dmdashenkov: As discussed, leaving this as is.

dmdashenkov:
@armiol, PTAL again.

server/src/test/java/io/spine/core/RejectionsShould.java
```diff
@@ -124,6 +131,23 @@ public void return_empty_optional_if_producer_not_set() {
         assertFalse(getProducer(freshRejection.getContext()).isPresent());
     }
 
+    @SuppressWarnings({
+            "NewExceptionWithoutArguments" /* No need to have a message for this test. */,
+            "SerializableInnerClassWithNonSerializableOuterClass" /* Does not refer anything. */
+    })
+    @Test
+    public void say_if_RuntimeException_was_called_by_command_rejection() {
```
armiol: `tell_if..`

server/src/main/java/io/spine/server/commandbus/CommandErrorHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    public void handleError(CommandEnvelope envelope, RuntimeException exception) {
+        checkNotNull(envelope);
+        checkNotNull(exception);
+        if (causedByRejection(exception)) {
+            final Rejection rejection = rejectWithCause(envelope.getCommand(), exception);
+            rejectionBus.post(rejection);
+        } else {
+            log().error(format("Error dispatching command (class: %s id: %s).",
+                               envelope.getMessage().getClass(),
+                               envelope.getId().getUuid()),
```
armiol: Can we please use `Stringifiers` on it? It'd be cool to know a type as well, not just a UUID value.

PR #618, opened 2017-11-13
Update `base` dependency

dmdashenkov:
In this PR we update dependency on `base` project to the `0.10.0` version.

The framework version is set to `0.10.4-SNAPSHOT`.


dmdashenkov:
This PR awaits the `base` PR fix.

PR #619, opened 2017-11-14
Storage test suits improvement

dmytro-grankin:
Previously, some tests for storages created instances of storages, which were not closed after execution of a test.

It could lead to issues, when the resources used by the storage were not released. This PR fixes this and improves the documentation.

For now, `AbstractStorageShould` manages creation and closing of the single storage for the test entity class, which is specified by `getTestEntityClass()` method. This storage can be obtained using `getStorage()` method.

In the case, if the storage for different entity class should be tested, it can be created using `newStorage(Class)` method, but closing of this storage is a responsibility of a caller.

All storages should be closed after a test to avoid the issues, which may occur due to unreleased resources. `AbstractStorageShould.close(AbstractStorage) ` simplifies this task.

The version of the framework was increased to `0.10.5-SNAPSHOT`.

### DIFF
--- a/server/src/test/java/io/spine/server/stand/StandStorageShould.java
+++ b/server/src/test/java/io/spine/server/stand/StandStorageShould.java
@@ -30,17 +30,14 @@ import com.google.protobuf.Message;
 import io.spine.Identifier;
 import io.spine.core.given.GivenVersion;
 import io.spine.protobuf.AnyPacker;
-import io.spine.server.entity.Entity;
 import io.spine.server.entity.EntityRecord;
 import io.spine.server.entity.FieldMasks;
-import io.spine.server.storage.RecordStorage;
 import io.spine.server.storage.RecordStorageShould;
 import io.spine.test.storage.Project;
 import io.spine.test.storage.ProjectId;
 import io.spine.test.storage.Task;
 import io.spine.test.storage.TaskId;
 import io.spine.type.TypeUrl;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import javax.annotation.Nullable;
@@ -60,7 +57,7 @@ import static java.lang.String.format;
  * @author Dmytro Dashenkov
  */
 public abstract class StandStorageShould extends RecordStorageShould<AggregateStateId,
-                                                                     RecordStorage<AggregateStateId>> {
+                                                                     StandStorage> {
 
     protected static final Supplier<AggregateStateId<ProjectId>> DEFAULT_ID_SUPPLIER
             = new Supplier<AggregateStateId<ProjectId>>() {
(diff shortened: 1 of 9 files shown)


dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

server/src/test/java/io/spine/server/storage/AbstractStorageShould.java
```diff
@@ -100,6 +99,9 @@ protected final S getStorage() {
     /** Creates a new read request with the specified ID. */
     protected abstract R newReadRequest(I id);
 
+    /** Returns the class of the default test {@link Entity}. */
+    protected abstract Class<? extends Entity> getDefaultEntityClass();
```
armiol: Let's try to avoid `default` at all.

server/src/test/java/io/spine/server/stand/StandStorageShould.java
```diff
@@ -60,7 +57,7 @@
  * @author Dmytro Dashenkov
  */
 public abstract class StandStorageShould extends RecordStorageShould<AggregateStateId,
-                                                                     RecordStorage<AggregateStateId>> {
+                                                                     StandStorage> {
```
armiol: I think this is a change, which will may not fit well into the dependent modules (such as `jdbc-storage`).

dmytro-grankin: I tested this for `jdbc-storage` and it's okay. For `gae-java` it should be also fine.

server/src/test/java/io/spine/server/storage/AbstractStorageShould.java
dmytro-grankin: Done.

dmytro-grankin:
@armiol PTAL.

server/src/test/java/io/spine/server/storage/AbstractStorageShould.java
```diff
@@ -42,7 +42,17 @@
 import static org.junit.Assert.fail;
 
 /**
- * Abstract storage tests.
+ * An abstract test suites testing storages.
+ *
+ * <p>The suit manages creation and closing of the {@linkplain #getStorage() storage}
```
armiol: Please fix the typo.

server/src/test/java/io/spine/server/storage/AbstractStorageShould.java
```diff
@@ -42,7 +42,17 @@
...
+ * for the {@linkplain #getTestEntityClass() test entity class}.
+ *
+ * <p>In case, if the storage for different entity class should be tested,
```
armiol: Please remove comma after `case`.

PR #623, opened 2017-11-22
Fix `Target.type` content

### LINKED ISSUE #622: Fix `Target.type`
There is a message `Target` defined in `spine/client/entities.proto`. It is used for bearing the information about the `Query`/`Subscription` target.

The message declares the field:
```protobuf
// Represents TypeUrl for the entity of interest.
string type = 1;
```

Though, in practice, the `TypeName`, not `TypeUrl` is used.
Now, either the declaration or the usage should be fixed.

Also, we should consider having Protobuf types for both `TypeName` and `TypeUrl` to avoid such errors in future.

dmdashenkov:
This PR fixes the usage of `type` field in `spine.client.Target`. By now we stored Protobuf type name in it instead of type URL.

Updates to the dependant repositories (storage implementations, etc.) may be required.

Fixes #622.

### DIFF
--- a/client/src/main/java/io/spine/client/Queries.java
+++ b/client/src/main/java/io/spine/client/Queries.java
@@ -23,14 +23,15 @@ import com.google.protobuf.FieldMask;
 import com.google.protobuf.Message;
 import io.spine.Identifier;
 import io.spine.annotation.Internal;
-import io.spine.type.TypeName;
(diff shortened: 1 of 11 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #626, opened 2017-11-27
Tenant ID callback

dmdashenkov:
This PR introduces the API for performing an operation for each tenant in a bounded context.

The `TenantIndex.forEachTenant` method applies the given `TenantIdConsumer` to all the tenants in the index. Also, all the subsequent `TenantIndex.keep` calls will apply the operation on the new `TenantId`s.

`TenantIdConsumer` itself is a functional interface, which is duck-type compatible to Java 8 `Consumer<TenantId>`, therefore can be substituted with the standard interface when migrating to Java 8.

Note that in most cases the consumer is applied only once to a single instance of `TenantId`. Though, in some cases, it may be applied several times. At-least-once invocation is guaranteed.

### DIFF
--- a/server/src/main/java/io/spine/server/tenant/TenantIndex.java
+++ b/server/src/main/java/io/spine/server/tenant/TenantIndex.java
@@ -47,12 +47,15 @@ public interface TenantIndex extends AutoCloseable {
     Set<TenantId> getAll();
 
     /**
-     * Applies the given {@code operation} onto all the tenant IDs, currently present and added
-     * in future.
+     * Applies the given {@code operation} to all the tenant IDs currently present in the index
+     * and those added in future.
      *
      * <p>The operation is applied to {@linkplain #getAll() all present tenant IDs} at once. When
      * new IDs are {@linkplain #keep(TenantId) added} to the index, the operation is applied
      * to them as well.
+     *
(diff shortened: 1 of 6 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

server/src/main/java/io/spine/server/tenant/TenantIndex.java
```diff
@@ -45,6 +46,21 @@
...
+     * <p>The operation is applied to {@linkplain #getAll() all present tenant IDs} at once. When
+     * new IDs are {@linkplain #keep(TenantId) added} to the index, the operation is applied
+     * to them as well.
+     *
+     * <p>Note that the operation may be applied to a single {@code TenantId} more then once. In
+     * general, the caller may rely on at-least-once behavior.
+     *
+     * @param operation the operation to apply to all the tenant IDs
+     */
+    void forEachTenant(TenantIdConsumer operation);
```
alexander-yevsyukov: Assuming we have `getAll()` which returns a set of all tenant IDs, why do we need this method? Why isn't `for` operator enough?

dmdashenkov:
As another approach has been chosen.

PR #627, opened 2017-11-29
Clarify EventPlayingEntity.getBuilder Javadoc

vlad-lubenskyi:
Prior to this PR, Javadoc for `EventPlayingEntity.getBuilder()` stated that this method must be called `only from within an event applier` and throws an exception if happens otherwise.

While it is true for Aggregates, it doesn't suite Process Managers. 

It would be more accurate to say that this method must be called inside of an active transaction. Then clarify that `Aggregate` fulfills this restriction only in event appliers, and `ProcessManager` does so in event/rejection reactors, and command handlers.

This pull request:

1. Generalizes Javadoc of `EventPlayingEntity.getBuilder()`. Javadoc claims that this method must be called inside of an active `Transaction`.
2. Indicates in Javadoc for overridden methods when these methods must be called in every particular case.


vlad-lubenskyi:
@armiol PTAL

server/src/main/java/io/spine/server/procman/ProcessManager.java
```diff
@@ -106,8 +106,17 @@ protected CommandBus getCommandBus() {
         return commandBus;
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>In {@code ProcessManager}, this method must be called from either an event applier or
```
armiol: I think, it's in fact not an `event applier`, but `event and rejection reactors`.

That's because `ProcessManager`s do not have `@Apply` methods. They only have `@React` serving to program a reaction upon an event or a rejection.

Please update the description here and below. Probably, PR description should be updated as well.

vlad-lubenskyi:
@armiol PTAL

server/src/main/java/io/spine/server/procman/ProcessManager.java
```diff
@@ -106,8 +106,17 @@ protected CommandBus getCommandBus() {
         return commandBus;
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>In {@code ProcessManager}, this method must be called from an event/rejection reactor or
```
armiol: Let's avoid `event/rejection` notation, as they are really different things :)

PR #629, opened 2017-11-30
Tenant events

dmdashenkov:
This PR introduces the `TenantAdded` event. It may be used by the framework users to handle a new tenant onboarding.

The users may choose to fire events as a reaction to a domain-specific event or produce them in response to a domain-specific command.

Currently, no entity fires `TenantAdded` events by default.

To subscribe to the events, a user may define an `EventSubscriber`:
```java
public class TenantOnboardingListener extends EventSubscriber {
    @Subscribe(external = true)
    public void on(TenantAdded event) {
         // Handle new tenant here.
    }
}
```
Then, an instance of `TenantOnboardingListener` should be registered in `IntegrationBus`:
```java
final EventSubscriber subscriber = new TenantOnboardingListener();
boundedContext.getIntegrationBus().register(subscriber);
```

Note that the defined event subscriber is [external](https://spine.io/core-java/javadoc/core/io/spine/core/Subscribe.html#external--). This example implies an existence of *`Tenants`* `BoundedContext`, which publishes these events.

The `TenantAdded` event should be posted in the scope of the added tenant. It is not recommended to handle tenant events in an `Entity` (`Aggregate`, `Projection` or `ProcessManager`) since an event will apparently create a new instance of the entity (for the newly created tenant) and that is probably not intended.


server/src/main/proto/spine/server/tenant/events.proto
```diff
@@ -0,0 +1,44 @@
...
+option java_package = "io.spine.server.tenant";
+option java_generate_equals_and_hash = true;
+option java_multiple_files = true;
+option java_outer_classname = "TenantEventsProto";
+
+import "google/protobuf/any.proto";
+
+import "spine/core/tenant_id.proto";
+
+// Notifies the subscribers of adding a new tenant to the multitenant system.
```
armiol: `An event telling that a new tenant has been added to a multi-tenant application.`

dmdashenkov:
@armiol, PTAL.

PR #634, opened 2017-12-14
Fix javadoc warnings

alexander-yevsyukov:
This PR fixes all but one Javadoc warnings.

The remaining warning

```
core-java/server/src/main/java/io/spine/server/entity/storage/ColumnType.java:96: warning - Tag @link: reference not found: java.sql.PreparedStatement PreparedStatement
```
seems to be a [bug in Javadoc](https://bugs.openjdk.java.net/browse/JDK-8006642), which is fixed in Java 8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ build/
 out/
 
 # Spine temporary artifact storage
+.spine/
 */.spine/
 model/*/.spine/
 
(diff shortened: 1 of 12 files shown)


alexander-yevsyukov:
@armiol, PTAL.

PR #637, opened 2017-12-14
Improve code layout

alexander-yevsyukov:
This PR:
 - fixes right margins;
 - addresses minor code issues discovered during cursory walkthrough;
 - advances the Spine version to `0.10.13-SHAPSHOT` as this PR changes Javadocs.

PR #640, opened 2017-12-14
Extend test model

alexander-yevsyukov:
This PR introduces minor improvements to the test model under `spine/test/event`.


alexander-yevsyukov:
@armiol, it's branched from #637. So it would make sense to finish with #637 first.

Issue #639, opened 2017-12-14
Test inner delegate of `DelegatingEventDispatcher`


alexander-yevsyukov:
See [this coverage report](https://codecov.io/gh/SpineEventEngine/core-java/compare/db0334169b898a904615e3fd29e124883692c2af...654b01efef1bdff6e7261db053bd5bd852dc0c0b/src/server/src/main/java/io/spine/server/event/DelegatingEventDispatcher.java) for details.

armiol:
Resolved in #641.

Issue #638, opened 2017-12-14
Document `ExternalMessageEnvelope.of()`


alexander-yevsyukov:
The method raises some questions like:
 *  Why do we need the second parameter `originalMessage` while `ExternalMessage` has `originalMessage`?
 * Can I send default values?

armiol:
Resolved in #641.

PR #641, opened 2017-12-15
`ExternalMessageEnvelope`: documentation and improved coverage

armiol:
This PR addresses #638 and #639.

In particular it clarifies the API of `ExternalMessageEnvelope`. 

Also it improves the test coverage of `DelegatingEventDispatcher` behavior, that dispatches `ExternalMessageEnvelope` instances.

As a bonus, `RejectionHandlerMethod` documentation was improved with explicit mentioning of `@React` and `@Subscribe` in different cases.

### DIFF
--- a/server/src/main/java/io/spine/server/rejection/RejectionHandlerMethod.java
+++ b/server/src/main/java/io/spine/server/rejection/RejectionHandlerMethod.java
@@ -152,11 +152,18 @@ class RejectionHandlerMethod extends HandlerMethod<RejectionContext> {
          * A rejection handler method which receives a rejection message
          * as a single parameter.
          *
-         * <p>The signature of such a method is following:
+         * <p>The signature of such a method is as follows, if used in {@link RejectionSubscriber}:
          * <pre>
-         * [{@literal @}HandlerAnnotation]
+         * [{@literal @}Subscribe]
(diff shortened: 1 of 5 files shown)


armiol:
@alexander-yevsyukov PTAL.

server/src/main/java/io/spine/server/rejection/RejectionHandlerMethod.java
```diff
@@ -152,11 +152,18 @@ Object doInvoke(Object target, Message rejectionMsg, RejectionContext context) {
          * A rejection handler method which receives a rejection message
          * as a single parameter.
          *
-         * <p>The signature of such a method is following:
+         * <p>The signature of such a method is as follows, if used in {@link RejectionSubscriber}:
          * <pre>
-         * [{@literal @}HandlerAnnotation]
+         * [{@literal @}Subscribe]
```
alexander-yevsyukov: Why do we need square brackets around `@Subscribe` here and in the other Javadocs below?

armiol: @alexander-yevsyukov they were previously used to mark `[HandlerAnnotation]`.

Removed.

armiol:
@alexander-yevsyukov PTAL.

PR #642, opened 2017-12-19
Dispatch events to projections with any visibility

dmytro-grankin:
Previously, there was no possibility to dispatch events to archived or deleted projections.

If a projection is marked as archived or deleted, it's still a valid case to dispatch events to such a projection (e.g. to cancel archiving/deletion or anything else).

Even though all projections should be visible during dispatching, only projections with default `LifecycleFlags` should be exposed to read side, so this behavior was not changed.

To obtain an entity for dispatching, endpoints use `RecordBasedRepository.findOrCreate(ID)` method. This method is used only for this purpose. Previously the method returned only visible entities, so it was decided to change this behavior. Now `findOrCreate(ID)` returns an entity despite its visibility.

This change resolves the problem with dispatching for all record based repositories (`ProjectionRepository` and `ProcessManagerRepository`).

The Spine version was bumped to `0.10.14-SNAPSHOT`.

### DIFF
--- a/server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
+++ b/server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
@@ -150,10 +150,26 @@ public abstract class RecordBasedRepository<I, E extends Entity<I, S>, S extends
             return Optional.absent();
         }
         final EntityRecord record = optional.get();
-        final boolean recordVisible = !isEntityVisible().apply(record.getLifecycleFlags());
-        if (recordVisible) {
+        final boolean recordVisible = isEntityVisible().apply(record.getLifecycleFlags());
+        if (!recordVisible) {
             return Optional.absent();
         }
+        final E entity = toEntity(record);
+        return Optional.of(entity);
+    }
+
+    /**
+     * Finds an entity with the passed ID despite its {@linkplain LifecycleFlags visibility}.
+     *
+     * @param id the ID of the entity to load
+     * @return the entity or {@link Optional#absent()} if there is no entity with the ID
+     */
+    protected Optional<E> findWithAnyVisibility(I id) {
+        Optional<EntityRecord> optional = findRecord(id);
+        if (!optional.isPresent()) {
+            return Optional.absent();
+        }
+        final EntityRecord record = optional.get();
         final E entity = toEntity(record);
         return Optional.of(entity);
     }
--- a/server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
+++ b/server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
@@ -141,7 +141,14 @@ public abstract class RecordBasedRepository<I, E extends Entity<I, S>, S extends
         store(entity);
     }
 
-    /** {@inheritDoc} */
+    /**
+     * Finds an entity with the passed ID if this entity is
+     * {@linkplain EntityWithLifecycle.Predicates#isEntityVisible() visible}.
+     *
+     * @param id the ID of the entity to load
+     * @return the entity or {@link Optional#absent()} if there's no entity with such ID
+     *         or this entity is not visible
+     */
     @Override
     @CheckReturnValue
     public Optional<E> find(I id) {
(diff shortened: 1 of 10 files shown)


dmytro-grankin:
@armiol PTAL.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -141,7 +141,14 @@ public void storeRecord(EntityRecord record) {
         store(entity);
     }
 
-    /** {@inheritDoc} */
+    /**
+     * Finds an entity with the passed ID if this entity is
+     * {@linkplain EntityWithLifecycle.Predicates#isEntityVisible() visible}.
+     *
+     * @param id the ID of the entity to load
```
armiol: `...to find`.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -141,7 +141,14 @@ public void storeRecord(EntityRecord record) {
...
+     * @return the entity or {@link Optional#absent()} if there's no entity with such ID
```
armiol: `..there is no..`

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -190,18 +197,21 @@ public EntityRecord findOrCreateRecord(I id) {
 
     /**
      * Loads an entity by the passed ID or creates a new one, if the entity was not found.
+     *
+     * <p>An entity will be loaded despite its {@linkplain LifecycleFlags visibility}.
```
armiol: This is not enough. Please add ```I.e. even if the entity is either {@linkplain ... hidden} or {@linkplain ... deleted}, it is loaded and returned. The new entity is created if and only if there is no record with the corresponding ID.```

server/src/test/proto/spine/test/procman/commands.proto
```diff
@@ -52,3 +53,8 @@ message PmDoNothing {
 message PmThrowEntityAlreadyArchived {
     ProjectId project_id = 1;
 }
+
+message PmChangeLifecycle {
+    ProjectId project_id = 1;
+    spine.server.entity.LifecycleFlags lifecycleFlags = 2;
```
armiol: The declaration style is wrong. It should NOT be camel-cased in the Protobuf definitions.

Also, can we please avoid having lifecycle flags as a parameter at all? Please have a command to archive and another to delete.

server/src/test/proto/spine/test/projection/events.proto
```diff
@@ -42,3 +43,8 @@ message PrjTaskAdded {
 message PrjProjectStarted {
     ProjectId project_id = 1;
 }
+
+message PrjProjectLifecycleChanged {
+    ProjectId project_id = 1;
+    spine.server.entity.LifecycleFlags lifecycleFlags = 2;
```
armiol: Same as above.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
dmytro-grankin: Done.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
dmytro-grankin: Done.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
dmytro-grankin: Done.

server/src/test/proto/spine/test/procman/commands.proto
dmytro-grankin: Done.

server/src/test/proto/spine/test/projection/events.proto
dmytro-grankin: Done.

dmytro-grankin:
@armiol PTAL.

server/src/test/proto/spine/test/procman/commands.proto
```diff
@@ -52,3 +53,11 @@ message PmDoNothing {
 message PmThrowEntityAlreadyArchived {
     ProjectId project_id = 1;
 }
+
+message ArchivePm {
```
armiol: `PmArchiveProcess`. `PmDeleteProcess` below.

Let's not violate our conventions.

dmytro-grankin:
@armiol PTAL.

PR #643, opened 2017-12-22
Fix in-memory read of an aggregate history

dmytro-grankin:
Previously, it was not possible to read a history of an archived or deleted aggregate using `InMemoryAggregateStorage`. Also, receiving of aggregate IDs was not correct and only IDs of aggregates with default `LifecycleFlags` were returned.

This PR fixes the issues and now behavior for `InMemoryAggregateStorage` and for `DsAggregateStorage` is consistent.

Also, the behavior of `AggregateRepository.find(ID)` was changed and now it returns archived and deleted aggregates either. Such a behavior is correct because the repository should provide an ability to load archived/deleted aggregates during enrichment and in fact, this method is used only for this purpose.

The Spine version was increased to `0.10.15-SNAPSHOT`.

### DIFF
--- a/server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
+++ b/server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
@@ -58,7 +58,6 @@ import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
-import static io.spine.server.entity.EntityWithLifecycle.Predicates.isEntityVisible;
 import static io.spine.util.Exceptions.newIllegalStateException;
 
 /**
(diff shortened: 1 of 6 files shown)


dmytro-grankin:
@armiol PTAL.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -501,29 +500,21 @@ void onModifiedAggregate(TenantId tenantId, A aggregate) {
     /**
      * Loads an aggregate by the passed ID.
      *
-     * <p>The method returns {@link Optional#absent()} if:
-     * <ul>
-     *     <li>there are no events stored for the aggregate with the passed ID, or
-     *     <li>the aggregate has at least one {@linkplain LifecycleFlags lifecycle flag} set.
-     * </ul>
+     * <p>An aggregate will be loaded despite its {@linkplain LifecycleFlags visibility}.
+     * I.e. even if the aggregate is either
```
armiol: Please remove `either`.

PR #648, opened 2018-01-10
Fix  `Queries.typeOf(Query)` doc

dmdashenkov:
In this PR we make the `Queries.typeOf(Querry)` doc consistent with the implementation.

Before, the doc stated that the method returns `null` if the type is unknown.
Now, the doc states the truth — the method throws `ISE` upon such event (since an unknown type is an illegal state for a system).
  
The framework version is _not_ changed in this PR.


dmdashenkov:
@armiol, PTAL.

PR #650, opened 2018-01-22
Remove the `io.spine.server.users` package as redundant.

armiol:
Previously `core-java` contained a number of messages, grouped around user- and tenant-related functionality. `io.spine.server.users` package was used for those.

However, as of now those messages have been migrated to other packages, leaving `io.spine.server.users` empty.

This PR removes the package and its `package-info` description.


armiol:
@alexander-yevsyukov PTAL.

PR #651, opened 2018-01-23
Tuples

alexander-yevsyukov:
This PR adds the first tuple class called `Pair` and the basis for other tuple implementations.

Other modifications:
  * Gradle version was bumped to `4.4.1`.
  * Spine version advanced to `0.10.16-SNAPSHOT`.


alexander-yevsyukov:
@armiol, PTAL.

PR #658, opened 2018-01-29
Configure IDE to align `{}` to different lines

dmdashenkov:
In this PR we fallback to the alignment rule which makes us align empty code blocks as follows:
```java
private ConstructorExample() {
}
```
Note that the curly braces are placed on __different__ lines.


dmdashenkov:
@armiol, PTAL.

PR #659, opened 2018-01-29
Allow `Either` as `Tuple` element

alexander-yevsyukov:
This PR allows `Either` to be an element of a tuple.

Spine version advanced to `0.10.20-SNAPSHOT`.


alexander-yevsyukov:
@armiol, PTAL.

Issue #607, opened 2017-10-20
Record origin ID in the `EventContext`


armiol:
In scope of this issue a traceability of actions needs to be improved.

Currently the chain of `Command` propagation (e.g. `Command` --> `@Assign` handler --> events --> `@React` handler --> events --> ...) is nearly impossible to trace back to the original command instance. Especially it is impossible, being deep down the execution path. 

However, such an opportunity is sometimes a crucial thing. In particular, this is a way to provide idempotence for command-handing entities in relation to commands they handle.

To address that we need to store the origin ID in the `EventContext` for each `Event`. E.g. 

`EventContext.getOriginId() : Any`

In this way the origin ID  will automatically get stored to `EventStore` and will always be available for back-traces.

mdrachuk:
This was implemented in https://github.com/SpineEventEngine/core-java/pull/660.

PR #660, opened 2018-01-29
Add Root Command ID to Event Context

mdrachuk:
This PR tackles the issue https://github.com/SpineEventEngine/core-java/issues/607 fixing some critical bugs along the way.

## Root Command ID
This PR introduces a new `EventContext` attribute — `root_command_id`. 

This property is aimed to track the root command which caused events in application to be spawned. This information is required to solve an issue where a command can be delivered more than once to the application. Previously to track down the initial command one had to go over all of the events in the event chain which is inefficient.

In this context, a **root command** is an initial command handled by the application which spawned some events. All of this events will have their contexts `root_command_id` property set to the root commands identifier. 

The identifier of the command is propagated through the event chain. This means that when an event occurs as a reaction to another event, its `root_command_id` is duplicated from the previous event.

## Dead Event Tap

Additionally, this PR changes the behavior of the `DeadEventTap`. 

The `DeadEventTap` is used by the `DeadMessageFilter`. The `DeadMessageFilter` acts in the same way as any `AbstractBusFilter` implementation filtering messages posted to a `Bus`. This means that _events_ filtered by the `DeadMessageFilter` did not get saved in the `EventStore`, which was a bug. 

The `DeadEventTap` now stores received events to the event store on its own.

## Storing filtered messages

Another critical bug was found as a part of `Bus#post()`. 

The messages received by a bus were filtered, but if at least one message wasn’t filtered out, all of the received messages were stored. This was changed to store only the messages which passed all the filters.

## React issues in multitenant environments

Previously the Tenant ID was obtained from the events origin Command Context only. This resulted in a bug where the events Event or Rejection Context were not accessed to retrieve the Tenant ID. Now the event context is traversed to obtain the Tenant ID value and returns an empty ID only when all of the origin values are empty.


mdrachuk:
@armiol PTAL

core/src/main/java/io/spine/core/Events.java
```diff
@@ -162,6 +162,20 @@ public static UserId getActor(EventContext context) {
         return id;
     }
 
+    /**
+     * Obtains origin ID of the Event — the id of the original message that spawned this event.
```
armiol: This is both wrong (as you used "origin", meaning something else in this context) and wrong (it should be `ID`, not `id`; we also don't use `—` symbols as much).

I have rewritten it like this. 

```
Obtains the ID of the root command, which lead to this event.

<p> In case the passed {@code Event} instance is a reaction to another {@code Event}, the identifier of the very first command in this chain is returned.
```

core/src/main/java/io/spine/core/CommandEnvelope.java
```diff
@@ -113,13 +114,18 @@ public ActorContext getActorContext() {
     }
 
     /**
-     * Sets the context of the enclosed command as the origin for the event being built.
+     * Passes data from the enclosed command to the event context being built.
      *
-     * @param builder event context builder into which set the event origin context
+     * <p>In particular it sets the root command identifier and the event context origin.
```
armiol: Please have an unordered list in HTML. See other `<li>` examples in Spine for reference.

core/src/main/java/io/spine/core/CommandEnvelope.java
```diff
@@ -113,13 +114,18 @@ public ActorContext getActorContext() {
     }
 
     /**
-     * Sets the context of the enclosed command as the origin for the event being built.
+     * Passes data from the enclosed command to the event context being built.
      *
-     * @param builder event context builder into which set the event origin context
+     * <p>In particular it sets the root command identifier and the event context origin. 
+     * The commands identifier is set as the root command identifier.
```
armiol: `The commands identifier` contains a typo.

core/src/main/java/io/spine/core/EventEnvelope.java
```diff
@@ -96,15 +96,21 @@ public ActorContext getActorContext() {
                                 .getActorContext();
     }
 
+    
     /**
-     * Sets the context of the enclosed event as origin for the event which context
-     * is being built.
+     * Passes data from the enclosed event to the event context being built.
+     * 
+     * <p>In particular it sets a root command identifier and an event context origin.
```
armiol: `the event context`.

core/src/main/proto/spine/core/event.proto
```diff
@@ -99,6 +99,9 @@ message EventContext {
         RejectionContext rejection_context = 7 [(required) = true];
     }
 
+    // An id of the root command which caused this event being spawned.
```
armiol: Please use the idea from one of my comments, where I explain what this ID is.

server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -148,7 +154,28 @@ boolean hasDispatchers(EventClass eventClass) {
...
+
+    /**
+     * Creates a {@link DeadMessageTap} instance to handle a dead event by producing an 
+     * {@link UnsupportedEventException}, and saving the event to the Event Store.
+     *
+     * <p>An event gets into this tap if it does not pass the {@link io.spine.server.bus.DeadMessageFilter}.
+     * Although, this stops the Event Bus from handling the event, it still must be added to the Event Store.
+     */
+    private DeadMessageTap<EventEnvelope> newDeadEventTap() {
+        return new DeadMessageTap<EventEnvelope>() {
```
armiol: Let's extract this as a inner (non-`static`) class. As I look at it right now, I don't like how complex this expression seems to be.

server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -148,7 +154,28 @@ boolean hasDispatchers(EventClass eventClass) {
...
     protected DeadMessageTap<EventEnvelope> getDeadMessageHandler() {
-        return DeadEventTap.INSTANCE;
+        return deadMessageHandler;
+    }
+
+    /**
+     * Creates a {@link DeadMessageTap} instance to handle a dead event by producing an 
+     * {@link UnsupportedEventException}, and saving the event to the Event Store.
+     *
+     * <p>An event gets into this tap if it does not pass the {@link io.spine.server.bus.DeadMessageFilter}.
```
armiol: Let's simplify this paragraph and follow DRY principle (because now you duplicate your point from the previous paragraph).

```
We must store dead events, as they are still emitted by some entity and therefore are a part of the history for the current bounded context.
```

server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -148,7 +154,28 @@ boolean hasDispatchers(EventClass eventClass) {
 
     @Override
     protected DeadMessageTap<EventEnvelope> getDeadMessageHandler() {
-        return DeadEventTap.INSTANCE;
+        return deadMessageHandler;
+    }
+
+    /**
+     * Creates a {@link DeadMessageTap} instance to handle a dead event by producing an
```
armiol: First we save, then we throw. Please switch places.

server/src/test/java/io/spine/server/event/EventRootCommandIdShould.java
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * @author Mykhailo Drachuk
+ */
+public class EventRootCommandIdShould {
+
+    private static final TestActorRequestFactory requestFactory =
+            TestActorRequestFactory.newInstance(EventRootCommandIdShould.class);
+
+    private static final ProjectId PROJECT_ID = projectId();
```
armiol: We don't have that. We use static methods from `TestEnv` directly.

server/src/test/java/io/spine/server/event/given/EventRootCommandIdTestEnv.java
```diff
@@ -0,0 +1,410 @@
...
+        return new ResponseObserver();
+    }
+
+    /**
+     * A {@link StreamObserver} implemented to query {@link io.spine.server.event.EventStore}.
+     *
+     * <p>It keeps all of the observed events, allowing access to the using
+     * {@link ResponseObserver#getResults()}.
+     */
+    public static class ResponseObserver implements StreamObserver<Event> {
```
armiol: Check out `MemoizingObserver` instead.

server/src/test/java/io/spine/server/event/given/EventRootCommandIdTestEnv.java
```diff
@@ -0,0 +1,410 @@
...
+        return EventStreamQuery.newBuilder()
+                               .build();
+    }
+
+    /**
+     * Creates a new {@link StreamObserver} instance, which keeps all observed events.
+     *
+     * <p>The observed events can be accessed using {@link ResponseObserver#getResults()}.
+     */
+    public static ResponseObserver newStreamObserver() {
```
armiol: Check out the `StreamObservers` API.

server/src/test/java/io/spine/server/event/given/EventRootCommandIdTestEnv.java
```diff
@@ -0,0 +1,410 @@
...
+    private static MemberInvitation memberInvitation(TeamId teamId) {
+        return ((MemberInvitation.Builder) Sample.builderForType(MemberInvitation.class))
+                .setTeamId(teamId)
+                .build();
+    }
+
+    /**
+     * Creates a new {@link EventStreamQuery} without any filters. 
+     */
+    public static EventStreamQuery newStreamQuery() {
```
armiol: This is `allEventsQuery`, isn't it?

server/src/test/java/io/spine/server/event/given/EventRootCommandIdTestEnv.java
```diff
@@ -0,0 +1,410 @@
...
+        public List<Event> getResults() {
+            return ImmutableList.copyOf(resultStorage);
+        }
+    }
+
+    public static class ProjectAggregateRepository
+            extends AggregateRepository<ProjectId, ProjectAggregate> { }
+
+    @SuppressWarnings("SerializableInnerClassWithNonSerializableOuterClass")
+    public static class TeamAggregateRepository
```
armiol: Some comments on routing would be very appropriate :)

armiol:
```
Additionally, this PR changes the behaviour of the DeadEventTap. The DeadEventTap is storing received events to the event store.
```
Please explain this in more details. In particular, there should be some history (previous behavior etc) and why we have changed it.

armiol:
Also, please update the PR title, so that it reads as a sentence. It's way to complex at the moment.

server/src/test/proto/spine/test/event/commands.proto
```diff
@@ -24,21 +24,59 @@ package spine.test.event.command;
...
+    repeated spine.net.EmailAddress email = 2;
+}
+
+message AcceptInvitation {
+    spine.core.UserId user_id = 1;
+
+    MemberInvitation invitation = 2;
+}
+
+message FinishSignUp {
```
armiol: `CompleteSignUp` is more common. Please rename.

server/src/test/java/io/spine/server/event/EventBusShould.java
```diff
@@ -79,7 +86,7 @@ public void setUp() {
 
     private void setUp(@Nullable EventEnricher enricher) {
         final BoundedContext bc = BoundedContext.newBuilder()
-                                                .setMultitenant(true)
+                                                .setMultitenant(false)
```
mdrachuk: This change was introduced because multitenant environments do not function properly at the moment. 

No tests were affected by this change.

mdrachuk:
@armiol PTAL

server/src/test/java/io/spine/server/event/given/EventRootCommandIdTestEnv.java
```diff
@@ -0,0 +1,375 @@
...
+                                   return singleton(msg.getTeamId());
+                               }
+                           });
+        }
+    }
+
+    /**
+     * Routes the {@link InvitationAccepted} event to the {@link TeamCreationProcessManager} which 
+     * created the invitation. This is done for the purposes of the 
+     * {@link EventRootCommandIdShould#match_the_id_of_an_external_event_handled_by_a_process_manager()}
```
armiol: ` {@link EventRootCommandIdShould#match_the_id_of_an_external_event_handled_by_a_process_manager()} 
 test.`
would better look like this:
```java
 {@linkplain EventRootCommandIdShould#match_the_id_of_an_external_event_handled_by_a_process_manager() test}.
```

server/src/test/proto/spine/test/bus/bus.proto
```diff
@@ -0,0 +1,58 @@
...
+}
+
+message BusMessage {
+    
+    BusMessageId id = 1 [(required) = true];
+    
+    google.protobuf.Any contents = 2 [(required) = true];
+    
+    BusMessageContext context = 3;
+
```
armiol: This line is redundant.

server/src/test/java/io/spine/server/bus/BusesShould.java
```diff
@@ -51,4 +73,140 @@ public void not_accept_nulls() {
...
+        assertBusPostErrs(bus, ErrorType.FAILING_FILTER);
+
+        assertTrue(passingFilter.passed());
+        assertTrue(passingFilter2.passed());
+    }
+
+    /**
+     * Asserts that bus acknowledges the error when posting a single message.
+     */
+    private static void assertBusPostErrs(TestMessageBus bus, ErrorType type) {
```
armiol: `...Errors`?

server/src/test/java/io/spine/server/bus/BusesShould.java
```diff
@@ -51,4 +73,140 @@ public void not_accept_nulls() {
...
+        final List<BusMessage> messages = asList(message, message2, message3);
+        final MemoizingObserver<Ack> observer = memoizingObserver();
+
+        bus.post(messages, observer);
+
+        final List<Ack> responses = observer.responses();
+        assertSize(3, responses);
+        assertSize(0, bus.storedMessages());
+    }
+
```
armiol: This line is redundant.

server/src/test/java/io/spine/server/bus/given/BusesTestEnv.java
```diff
@@ -0,0 +1,481 @@
...
+            }
+        }
+
+        private static class TestDeadMessageTap implements DeadMessageTap<TestEnvelope> {
+            @Override
+            public MessageUnhandled capture(TestEnvelope message) {
+                return new Exceptions.DeadMessageException();
+            }
+        }
+
```
armiol: This line is redundant.

server/src/test/java/io/spine/server/bus/given/BusesTestEnv.java
```diff
@@ -0,0 +1,481 @@
...
+        @SafeVarargs
+        public static Set<TestMessageClass> setOf(Class<? extends Message>... classes) {
+            return setOf(Arrays.asList(classes));
+        }
+    }
+
+    interface TestMessageDispatcher
+            extends MessageDispatcher<TestMessageClass, TestEnvelope, BusMessageId> {
+    }
+
```
armiol: This line is redundant.

server/src/test/java/io/spine/server/event/EventRootCommandIdShould.java
```diff
@@ -0,0 +1,192 @@
...
+        final MemoizingObserver<Event> observer = StreamObservers.memoizingObserver();
+
+        boundedContext.getEventBus()
+                      .getEventStore()
+                      .read(query, observer);
+
+        final List<Event> results = observer.responses();
+        return results;
+    }
+
```
armiol: Please fix all the similar cases. If you look through the codebase, you'll notice we don't allow extra empty lines.

server/src/test/proto/spine/test/bus/bus.proto
```diff
@@ -0,0 +1,58 @@
...
+    
+    BusMessageContext context = 3;
+    
+}
+
+message BusMessageContext {
+    
+}
+
+message TestMessageContents {
```
armiol: Can we please think of some real example? Such as `ProjectAdded` etc? Also please note that we name the test messages, which are aimed for testing a particular routine, with a prefix. E.g. `AggCreateProject` is a command for `AggreagateShould` tests, etc.

server/src/test/proto/spine/test/bus/bus.proto
```diff
@@ -0,0 +1,58 @@
...
+    
+    BusMessageId id = 1 [(required) = true];
+    
+    google.protobuf.Any contents = 2 [(required) = true];
+    
+    BusMessageContext context = 3;
+    
+}
+
+message BusMessageContext {
```
armiol: Please fill with the real attributes. Consider those defined in `CommandContext`, `EventContext` or `RejectionContext`.

server/src/test/proto/spine/test/event/commands.proto
```diff
@@ -24,21 +24,55 @@ package spine.test.event.command;
...
     Task task = 2;
 }
 
+message AddTasks {
+    ProjectId project_id = 1;
+
+    repeated Task task = 2;
+}
+
 message StartProject {
```
armiol: Please think of a prefix for the test messages. See the rest of examples and my comment on the same matter above.

That helps to distinguish the messages by their purpose. And avoid confusion in future, when someone creates another `StartProject` message with its specific purpose.

server/src/test/proto/spine/test/event/team.proto
```diff
@@ -0,0 +1,72 @@
+//
+// Copyright 2017, TeamDev Ltd. All rights reserved.
```
armiol: Copyright has changed. It's 2018 now. Please review the changeset on this matter.

server/src/test/java/io/spine/server/bus/BusesShould.java
mdrachuk: I can change it, but what I did mean to say [_errs_](https://www.merriam-webster.com/dictionary/err).

server/src/test/proto/spine/test/event/commands.proto
mdrachuk: I’ve added the prefix to the messages created by me, but I didn’t get to changing all existing messages in this test suite. If you want to I can update them after the PR is accepted, so that refactorings wouldn’t get in the way of the real changes.

mdrachuk:
@armiol PTAL

core/src/main/java/io/spine/core/Events.java
```diff
@@ -183,16 +201,78 @@ public static EventId checkValid(EventId id) {
...
+     * stored in the initial {@link CommandContext} and can be retrieved from the events origin 
+     * command or rejection context. 
      */
     @Internal
     public static TenantId getTenantId(Event event) {
-        final TenantId result = event.getContext()
-                                     .getCommandContext()
-                                     .getActorContext()
-                                     .getTenantId();
+        final Optional<CommandContext> commandContext = getCommandContext(event);
```
armiol: Let's append 
```java
checkNotNull(event);
```
as the first line.

`EventsShould` should have a "nullability check" test case (see other Spine tests for a sample).

server/src/test/java/io/spine/core/EventsShould.java
```diff
@@ -198,10 +201,119 @@ public void obtain_type_name_of_event() {
         assertEquals(Timestamp.class.getSimpleName(), typeName.getSimpleName());
     }
 
+    @Test
```
armiol: Please append your name as an `@author` value to the header of this class.

server/src/test/java/io/spine/core/EventsShould.java
```diff
@@ -48,6 +49,7 @@
 import static io.spine.core.Events.getTimestamp;
 import static io.spine.core.Events.nothing;
 import static io.spine.core.Events.sort;
+import static io.spine.core.given.EventsTestEnv.TEST_TENANT_NAME;
```
armiol: Please have a static method instead of referencing a constant.

I also suspect that it should be `EventsTestEnv.tenantId()` and it should return `TenantId`. Note the absence of "test" and a non-`String` return value.

server/src/test/java/io/spine/server/event/EventBusShould.java
```diff
@@ -355,6 +392,99 @@ public void create_validator_once() {
         assertSame(validator, eventBus.getValidator());
     }
 
+    @Test
+    public void store_an_event() {
+        final Command command = command(createProject());
+        final EBProjectCreatedSubscriber subscriber = new EBProjectCreatedSubscriber();
+        // Register an event subscribe for the event to pass the `DeadEventFilter`
```
armiol: Please append yourselves as an `@author`. 

Also, please move all of your comments to the Javadoc section of the method (remember to use `{@code ...}` instead of back-ticks).  Several versions ago we decided to move away from the inline comments, as they are hardly readable in comparison to the Javadocs. So let's not increase our technical debt.

server/src/test/java/io/spine/server/event/given/EventBusTestEnv.java
```diff
@@ -20,25 +20,240 @@
...
+                    final Ack result = reject(packedId, error);
+                    return Optional.of(result);
+                }
+            }
+            return absent();
+        }
+
+        private static Error error() {
+            return Error.newBuilder()
+                        .setMessage("Task can not be created done.")
```
armiol: `"The task cannot be created in a 'done' state."`

server/src/test/proto/spine/test/event/project.proto
```diff
@@ -61,4 +61,5 @@ message Task {
     TaskId task_id = 1;
     string title = 2;
     string description = 3;
+    bool done = 4 [(required) = true];
```
armiol: Let's call that `completed` or `released`. A project is not a task; it can hardly be "done".

mdrachuk: In fact, this is a task.

mdrachuk:
@armiol PTAL

mdrachuk:
@alexander-yevsyukov PTAL.

core/src/main/java/io/spine/core/CommandEnvelope.java
```diff
@@ -113,13 +114,20 @@ public ActorContext getActorContext() {
...
+     * <ul>
+     *     <li>the command identifier is set as the root command identifier;</li>
+     *     <li>the command origin is set to the context of the enclosed event.</li>
+     * </ul>
+     * 
+     * @param builder event context builder which is filled with the enclosed message data
      */
     @Override
-    public void setOriginContext(EventContext.Builder builder) {
+    public void passToEventContext(EventContext.Builder builder) {
```
alexander-yevsyukov: I am not sure about the new name. It doesn't sound right to be, but I cannot suggest better alternatives. What options did you consider?

core/src/main/java/io/spine/core/EventEnvelope.java
```diff
@@ -96,15 +96,23 @@ public ActorContext getActorContext() {
                                 .getActorContext();
     }
 
+    
     /**
-     * Sets the context of the enclosed event as origin for the event which context
-     * is being built.
+     * Passes data from the enclosed event to the event context being built.
```
alexander-yevsyukov: The previous method name highlighted the fact that the message within the envelope is the origin for the event being built. The new name lost this meaning.

The new description of the method makes me ask: ”Why does it do it?”

core/src/main/java/io/spine/core/MessageEnvelope.java
```diff
@@ -60,10 +60,9 @@
     C getMessageContext();
 
     /**
-     * Sets the context of the enclosed message as the origin into the context of an event
-     * to be build.
+     * Passes the data of the enclosed message to the event context being built.
```
alexander-yevsyukov: Again, why do we do this?

core/src/main/proto/spine/core/event.proto
```diff
@@ -99,6 +99,13 @@ message EventContext {
         RejectionContext rejection_context = 7 [(required) = true];
     }
 
+    // An ID of the root command, which lead to this event.
+    //
+    // If the `Event` is a reaction to another `Event`, then this attribute holds the identifier 
+    // of the very first command in this chain.
+    //
+    google.protobuf.Any root_command_id = 9 [(required) = true];
```
alexander-yevsyukov: `CommandId` is the built-in type. It's not a user-defined type, like, for example, an aggregate ID.

Why do we need `Any` here?

server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -107,6 +108,9 @@
     /** The validator for messages of posted events. */
     private final MessageValidator eventMessageValidator;
 
+    /** The handler for dead events. */
+    private final DeadMessageTap<EventEnvelope> deadMessageHandler;
```
alexander-yevsyukov: Why we have `handler` in the field name, and `Tap` in the type?

Can we hide the implementation details, and have more generic type for the field?

server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -148,7 +154,7 @@ boolean hasDispatchers(EventClass eventClass) {
 
     @Override
     protected DeadMessageTap<EventEnvelope> getDeadMessageHandler() {
```
alexander-yevsyukov: Can we have a more general type returned? Please also note my question about the `handler` vs `tap` names regarding the field.

mdrachuk:
@alexander-yevsyukov PTAL.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -183,16 +200,80 @@ public static EventId checkValid(EventId id) {
...
     }
 
+    /**
+     * Obtains the {@link TenantId} from the {@link CommandContext}.
+     * 
+     * <p>The {@link CommandContext} is accessible from the {@link Event} if the {@code Event} was 
+     * created as a result of some command or its rejection. This makes the {@code CommandContext} 
+     * a valid {@code TenantId} source inside of the {@code Event}.
+     */
+    private static TenantId getTenantId(CommandContext context) {
```
alexander-yevsyukov: I would put this into the `Commands` class. We have already `getTenantId(Command)` there. 

This way we localize the knowledge about whereabouts `TenantId` is in a command “guts” inside one utility class `Commands`.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -183,16 +200,80 @@ public static EventId checkValid(EventId id) {
...
+    }
+
+    /**
+     * Obtains the context of the command, which lead to this event.
+     * 
+     * <p> The context is obtained by traversing the events origin for a valid context source. 
+     * There are can be two sources for the command context:
+     * <ol>
+     *     <li>The command context set as the event origin.</li>
+     *     <li>The command set as a rejection property.</li>
```
alexander-yevsyukov: This isn't quite clear. It should say explain that the event was generated in response to a rejection, and the context of the command which was rejected we obtain.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -183,16 +200,80 @@ public static EventId checkValid(EventId id) {
...
+     * </ol>
+     * 
+     * <p>If at some point the event origin is not set the {@link Optional#absent()} is returned.
+     */
+    private static Optional<CommandContext> getCommandContext(Event event) {
```
alexander-yevsyukov: Assuming the description, the name `getOriginCommandContext()` would highlight the fact that we go deeper.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -183,16 +200,80 @@ public static EventId checkValid(EventId id) {
...
+                    commandContext = eventContext.getCommandContext();
+                    break;
+
+                case REJECTION_CONTEXT:
+                    commandContext = eventContext.getRejectionContext()
+                                                 .getCommand()
+                                                 .getContext();
+                    break;
+
+                case ORIGIN_NOT_SET:
```
alexander-yevsyukov: When is this possible?

mdrachuk: This is the case that occurs in existing tests.

mdrachuk: Other than that it doesn’t seem to be possible in real-world.

mdrachuk:
@alexander-yevsyukov I’ve implemented the requested changes. PTAL.

core/src/main/java/io/spine/core/Commands.java
```diff
@@ -106,12 +107,88 @@ public static Message ensureMessage(Message commandOrMessage) {
...
 
+    /**
+     * Obtains a {@link TenantId} from the given {@link Event}.
+     *
+     * <p>The {@code TenantId} is retrieved by traversing the passed {@code Event}s context. It is
+     * stored in the initial {@link CommandContext} and can be retrieved from the events origin 
+     * command or rejection context. 
+     */
+    @Internal
+    public static TenantId getTenantId(Event event) {
```
alexander-yevsyukov: What I was commenting about “guts” of `TenantId` and the `CommandContext` was related to the method with `CommandContext` parameter. 

This method, which accepts `Event`, looks like a part of `Events` utility class.

core/src/main/java/io/spine/core/Commands.java
```diff
@@ -106,12 +107,88 @@ public static Message ensureMessage(Message commandOrMessage) {
...
+     * There can be two sources for the command context:
+     * <ol>
+     *     <li>The command context set as the event origin.</li>
+     *     <li>The command set as a field of a rejection context if an event was generated in a 
+     *     response to a rejection.</li>
+     * </ol>
+     *
+     * <p>If at some point the event origin is not set the {@link Optional#absent()} is returned.
+     */
+    private static Optional<CommandContext> getOriginCommandContext(Event event) {
```
alexander-yevsyukov: Again, this method is part of `Events` “business”.

mdrachuk:
@alexander-yevsyukov, PTAL.

PR #666, opened 2018-02-07
Ensure exactly once delivery of commands to aggregates

mdrachuk:
This PR tackles an issue when a command can be delivered to an aggregate more than once in a short period of time. This is done by ensuring that since the last snapshot there are no events available in the aggregate event-store which have the `root_command_id` matching the ID of the command being dispatched. 

The root cause of this issue is message brokers giving a guarantee of at-least-once delivery. This means that whenever the message broker has an issue delivering a command it retries doing so until it succeeds.


server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -406,8 +406,13 @@ protected void setSnapshotTrigger(int snapshotTrigger) {
...
 
-    @VisibleForTesting
-    public AggregateStorage<I> aggregateStorage() {
+    /**
+     * Returns the storage assigned to this aggregate.
+     *
+     * @return storage instance
+     * @throws IllegalStateException if the storage is null
+     */
+    protected AggregateStorage<I> aggregateStorage() {
```
mdrachuk: This method is used outside of tests already. Although it doesn’t need to be publicly visible in the same manner as `ProjectionRepository#projectionStorage()` and `RecordBasedRepository#recordStorage()` are protected.

mdrachuk:
@alexander-yevsyukov PTAL.

server/src/main/java/io/spine/server/aggregate/AggregateCommandEndpoint.java
```diff
@@ -46,13 +55,51 @@ I handle(AggregateRepository<I, A> repository, CommandEnvelope command) {
...
-
     static <I, A extends Aggregate<I, ?, ?>>
     AggregateCommandEndpoint<I, A>
     of(AggregateRepository<I, A> repository, CommandEnvelope command) {
         return new AggregateCommandEndpoint<>(repository, command);
     }
 
+    @Override
+    protected void deliverNowTo(I aggregateId) {
+        ensureNotDuplicate(aggregateId);
```
alexander-yevsyukov: This means that we read events twice: once when checking, and another time when actually loading the aggregate. It's an expensive operation. We need to check it differently. Maybe query recent history of an aggregate after it's already loaded?

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
alexander-yevsyukov: Good to see.

server/src/main/java/io/spine/server/commandbus/DuplicateCommandException.java
```diff
@@ -0,0 +1,90 @@
...
+import io.spine.core.CommandEnvelope;
+import io.spine.core.CommandId;
+import io.spine.core.CommandValidationError;
+import io.spine.core.MessageInvalid;
+import io.spine.type.TypeName;
+
+import static java.lang.String.format;
+
+/**
+ * The exception for reporting duplicate commands.
```
alexander-yevsyukov: How about a shorter text like:

`Reports a command which was delivered more than once.`

server/src/main/java/io/spine/server/commandbus/DuplicateCommandException.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * <p> A command is considered a duplicate when it’s ID matches the ID of a command already
```
alexander-yevsyukov: `it's` -> `its`

server/src/main/java/io/spine/server/commandbus/DuplicateCommandException.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Generates a formatted duplicate command error message. 
+     */
+    private static String errorMessage(CommandId id, Message commandMessage) {
+        return format(
+                "The command (class: `%s`, type: `%s`, id: `%s`) was already handled by " +
+                "a Command Bus.",
+                CommandClass.of(commandMessage)
```
alexander-yevsyukov: You can pass a `CommandEnvelope` instance (which you should already have) and invoke `getMessageClass()`.

mdrachuk:
@alexander-yevsyukov, PTAL.

mdrachuk:
- [ ] Aggregate should have a `historyBackwards()` similar to `Storage`
- [ ] Command should be checked for duplicates in an endpoint
- [ ] The exception should extend `CommandException` stating that command cannot be dispatched twice
- [ ] There should be a client-server test with an acknowledgement observer check.

mdrachuk:
@armiol PTAL.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -392,19 +423,125 @@ protected String getMissingTxMessage() {
...
+        final List<AggregateEventRecord> items = queryBatch(batchSize);
+        return items.iterator();
+    }
+
+    /**
+     * Obtains a batch of aggregate records specified by the batch size.
+     * 
+     * @param batchSize the amount of events that must be retrieved
+     */
+    private List<AggregateEventRecord> queryBatch(int batchSize) {
```
armiol: What if the batch size is greater than the recent history depth?

That's another point to simplify the `historyBackward` public API. Let's not allow to set the size. At least for now, while we aren't able to load additional (older, than those "recent") events from the `EventStore`.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -392,19 +423,125 @@ protected String getMissingTxMessage() {
...
+     * <p> The check is performed by searching for an event caused by this command that was
+     * committed since last snapshot.
+     *
+     * <p> This functionality support the ability to stop duplicate commands from being dispatched
+     * to the aggregate.
+     *
+     * @param command the command to check
+     * @return {@code true} if the command was handled since last snapshot, {@code false} otherwise
+     */
+    public boolean didHandleSinceLastSnapshot(CommandEnvelope command) {
```
armiol: Let's have a separate (at least, inner) class called `IdempotencyGuard`. Its responsibility would be to provide idempotency for a certain type of input (now `Command`s, than later `Event`s and `Rejection`s to react on).

server/src/main/java/io/spine/server/aggregate/AggregateCommandEndpoint.java
```diff
@@ -46,13 +47,33 @@ I handle(AggregateRepository<I, A> repository, CommandEnvelope command) {
...
-
     static <I, A extends Aggregate<I, ?, ?>>
     AggregateCommandEndpoint<I, A>
     of(AggregateRepository<I, A> repository, CommandEnvelope command) {
         return new AggregateCommandEndpoint<>(repository, command);
     }
 
+    @Override
+    protected void deliverNowTo(A aggregate) {
+        ensureNotDuplicate(aggregate);
```
armiol: This isn't a responsibility of an endpoint. It's a responsibility of an `Entity`. It must not be possible to "forget" to call `ensureNotDuplicate`; it should be an implementation feature of `Aggregate`. 

With the current approach it's quite possible to feed a duplicate command.

server/src/main/java/io/spine/server/commandbus/DuplicateCommandException.java
```diff
@@ -0,0 +1,97 @@
...
+    private static String errorMessage(CommandEnvelope envelope) {
+        return format(
+                "The command (class: `%s`, type: `%s`, id: `%s`) cannot be dispatched twice.",
+                envelope.getMessageClass()
+                        .value()
+                        .getName(),
+                TypeName.of(envelope.getMessage()),
+                Identifier.toString(envelope.getId()));
+    }
+
```
armiol: This line is redundant.

server/src/main/java/io/spine/server/commandbus/DuplicateCommandException.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Generates a formatted duplicate command error message.
+     *
+     * @param envelope the envelope with a command which is considered a duplicate
+     * @return a string with details on what happened
+     */
+    private static String errorMessage(CommandEnvelope envelope) {
+        return format(
+                "The command (class: `%s`, type: `%s`, id: `%s`) cannot be dispatched twice.",
```
armiol: I would also try to mention the target, i.e. an `Aggregate` or `ProcessManager`, which history prevented the repeated dispatch.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -378,9 +396,22 @@ protected int uncommittedEventsCount() {
     List<Event> commitEvents() {
         final List<Event> result = ImmutableList.copyOf(uncommittedEvents);
         uncommittedEvents.clear();
+        addToHistory(result);
```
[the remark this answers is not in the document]
mdrachuk: `play` is called when the aggregate is loaded from storage. 
We do not play the events we’ve just applied to the aggregate. For this case, updating history of the aggregate when events are committed seems to me like the most appropriate place.

server/src/main/java/io/spine/server/aggregate/AggregateCommandEndpoint.java
mdrachuk: We‘ve discussed this in a personal meeting and decided the check must be performed at the endpoint. This is one of the points in the list of tasks higher in this conversation.
Aggregate doesn’t know anything about its repository, placing the check inside of aggregate would make it harder to check for duplicates if we need more then the events played when loading the aggregate.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -203,6 +214,7 @@ protected B getBuilder() {
      */
     @Override
     protected List<? extends Message> dispatchCommand(CommandEnvelope command) {
+        idempotencyGuard.ensureIdempotence(command);
```
alexander-yevsyukov: This is hard to read. We have to read `idempotent` twice. Does the guard have other public methods for checking?

Can it be something like?

```java
idempotencyGuard.check(command);
```

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -406,9 +418,8 @@ protected int uncommittedEventsCount() {
      *  @param events the list of the events
      */
     private void addToHistory(Collection<Event> events) {
-        final List<AggregateEventRecord> records = toAggregateEventRecords(events);
-        for (AggregateEventRecord record : records) {
-            historySinceLastSnapshot.addFirst(record);
+        for (Event event : events) {
+            historySinceLastSnapshot.addFirst(event);
```
alexander-yevsyukov: Can we have `recentHistory` instead of `historySinceLastSnapshot`? This way we would be able to extend the notion of the recency without renaming. For example, for some aggregates we may want to have the history until the grandfather of the last snapshot.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -448,98 +459,11 @@ Snapshot makeSnapshot() {
...
-                case SNAPSHOT:
-                    return false;
-                case KIND_NOT_SET:
-                default:
-                    throw newIllegalStateException("Aggregate record with kind not set.");
-            }
-        }
-        return false;
+    protected Iterator<Event> historyBackward() {
+        final ImmutableList<Event> events = ImmutableList.copyOf(historySinceLastSnapshot);
```
alexander-yevsyukov: Please encapsulate the impl. details into `RecentHistory` class, as we discussed vocally.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -519,4 +502,40 @@ public boolean apply(Message message) {
             return !message.equals(EMPTY);
         }
     }
+
+    private static class RecentHistory {
```
alexander-yevsyukov: Can we benefit from having this class used in `Projection`, for example?

server/src/main/java/io/spine/server/aggregate/IdempotencyGuard.java
```diff
@@ -45,13 +45,13 @@
...
     /**
-     * Ensures that the command was not dispatched to the aggregate.
+     * Checks that the command was not dispatched to the aggregate.
      * If it was a {@link DuplicateCommandException} is thrown.
      *
      * @param envelope an envelope with a command to check
      * @throws DuplicateCommandException if the command was dispatched to the aggregate
      */
-    void ensureIdempotence(CommandEnvelope envelope) {
+    void check(CommandEnvelope envelope) {
```
alexander-yevsyukov: Can we delegate the check to the `RecentHistory` class instead? I mean the history can answer on the question: “Did we have events that were originated with this command?” If yes, then the guard would fire the exception.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
mdrachuk: I think we can. But I’d like to stay within the bounds of this PR. 

Our goal was to implement this functionality for the aggregate commands.

Placing the Recent History outside of Aggregate to lay the groundwork for adding its support to Process Managers or Projections would require a lot of additional work. We would be exposing Recent History as part public API raising many new questions:
- Is this really how the RecentHistory should look like in this case?
- Maybe we would like to have some `PageableIterator` be returned by `iterator()`?  
- We shouldn’t give an iterator over `Event` instances.
- etc.

We will need to work all of this out, but this doesn’t help to solve the issue at hand.

server/src/main/java/io/spine/server/aggregate/IdempotencyGuard.java
mdrachuk: It sure can. 

But Recent History is just an aggregates `historyBackwards()` implementation detail for now.
Adding the `containsEventsCausedBy(Command)` to the aggregate would be bloating its API. There’s no other case for using this as an aggregate method.

alexander-yevsyukov:
I've moved the `RecentHistory` and updated dependencies. I think we need to wrap this PR, unless you see something critical for now.

@armiol, PTAL.

core/src/main/proto/spine/core/command.proto
```diff
@@ -147,6 +147,12 @@ enum CommandValidationError {
 
     // The scheduled command was not delivered to the target in time.
     EXPIRED = 5;
+
+    // The command is a duplicate of the command that was already dispatched.
+    //
+    // A command is considered a duplicate if its ID matches the id of another command.
```
armiol: `...matches the ID...`

server/src/main/java/io/spine/server/commandbus/DuplicateCommandException.java
```diff
@@ -0,0 +1,97 @@
...
+        super(messageText, command, error);
+    }
+
+    /**
+     * Creates an exception for a duplicate command dispatched to aggregate.
+     *
+     * @param command the duplicate command
+     * @return a newly constructed {@link DuplicateCommandException} instance
+     */
+    public static DuplicateCommandException forAggregate(Command command) {
```
armiol: I don't like the naming, as you are passing an instance of `Command`, not an instance of `Aggregate`. That's why let's rename it to `of`.

server/src/main/java/io/spine/server/entity/RecentHistory.java
```diff
@@ -0,0 +1,118 @@
...
+import io.spine.core.Event;
+
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.Objects;
+
+import static com.google.common.collect.Queues.newArrayDeque;
+
+/**
+ * Recent history of an {@linkplain io.spine.server.entity.EventPlayingEntity event-sourced entity}.
```
armiol: I would mention that this is just a copy of events, not a direct reference to the history of an entity. Thus, any modifications made to this object do not affect the real event stream.

server/src/main/java/io/spine/server/entity/RecentHistory.java
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @return an events iterator
+     */
+    public Iterator<Event> iterator() {
+        final ImmutableList<Event> events = ImmutableList.copyOf(history);
+        return events.iterator();
+    }
+
+    /**
+     * Adds events to the aggregate history.
```
armiol: We cannot mention an aggregate here, as we speak in terms of `event-sourced` entities.

server/src/main/java/io/spine/server/entity/RecentHistory.java
```diff
@@ -0,0 +1,118 @@
...
+            return false;
+        }
+        final RecentHistory other = (RecentHistory) obj;
+        return Objects.equals(this.history, other.history);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("history", history)
```
armiol: This implementation does not seem useful, as there may be dozens of those. I would print the number of events and that's it.

server/src/main/java/io/spine/server/entity/RecentHistory.java
```diff
@@ -0,0 +1,118 @@
...
+
+import static com.google.common.collect.Queues.newArrayDeque;
+
+/**
+ * Recent history of an {@linkplain io.spine.server.entity.EventPlayingEntity event-sourced entity}.
+ *
+ * @author Mykhailo Drachuk
+ * @author Alexander Yevsyukov
+ */
+public final class RecentHistory {
```
armiol: Judging on the API, this class seems to be a descendant of some `Iterable`, such as `List`.

server/src/test/java/io/spine/server/aggregate/IdempotencyGuardShould.java
```diff
@@ -0,0 +1,149 @@
...
+        final StreamObserver<Ack> noOpObserver = noOpObserver();
+        commandBus.post(createCommand, noOpObserver);
+
+        final TestAggregate aggregate = getAggregate(projectId, tenantId);
+
+        final IdempotencyGuard guard = new IdempotencyGuard(aggregate);
+        guard.check(of(startCommand));
+    }
+
+    private TestAggregate getAggregate(ProjectId id, TenantId tenantId) {
```
armiol: This seems to be `loadAggregate`.

server/src/main/java/io/spine/server/entity/RecentHistory.java
alexander-yevsyukov: I think we're going to have more to this class in the near future.

server/src/test/java/io/spine/server/aggregate/IdempotencyGuardShould.java
alexander-yevsyukov: This was actually a piece duplicated from another test environment.

@mdrachuk, please see `AbstractAggregateTestRepository`, which eliminates the duplication.

server/src/main/java/io/spine/server/entity/RecentHistory.java
alexander-yevsyukov: Yes, you're right. 

I've submitted an issue for possible extension of this output in the future (#679).

alexander-yevsyukov:
@armiol, I think I've addressed all your comments. PTAL again.

PR #667, opened 2018-02-17
Filter empty messages from results of aggregate dispatch

mdrachuk:
This PR implements empty filtering for the results of command handling, event reactions, and rejection reactions for aggregates. 

Empty messages cannot be returned directly but they can be created indirectly and should, therefore, be handled properly. A real-world case of empty messages falling into dispatching results are the returned tuples with empty `Optional` values.

### DIFF
--- a/server/src/main/java/io/spine/server/aggregate/Aggregate.java
+++ b/server/src/main/java/io/spine/server/aggregate/Aggregate.java
@@ -20,6 +20,7 @@
 package io.spine.server.aggregate;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.protobuf.Any;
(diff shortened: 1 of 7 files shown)


mdrachuk:
@armiol PTAL.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -122,6 +124,8 @@
                                 B extends ValidatingBuilder<S, ? extends Message.Builder>>
                 extends CommandHandlingEntity<I, S, B> {
 
+    private static final Empty EMPTY = Empty.getDefaultInstance();
```
armiol: Why this is here? It does not belong to an `Aggregate`. Can we please have a nested class for filtering? It should be a singleton. In this way this constant will go there.

server/src/test/proto/spine/test/aggregate/task.proto
```diff
@@ -0,0 +1,42 @@
...
+    string id = 1;
+}
+
+message AggTask {
+
+    AggTaskId id = 1;
+    
+    spine.core.UserId assignee = 2;
+}
+
```
armiol: This line is redundant.

mdrachuk:
@armiol PTAL.

PR #668, opened 2018-02-19
Make applier methods package-private

dmdashenkov:
Before, we recommended the applier methods to be `private`.
The issue with that is that the static code analyzers complain of that methods as "unused".

Now, we recommend the event applier methods to be package-private. This should resolve the problem with the static code analysis.


dmdashenkov:
@armiol, PTAL.

PR #671, opened 2018-02-27
Inherited annotations

dmdashenkov:
In this PR we change the Java code style configs in order to make the IDE automatically add some annotations to overridden method.

For example, let test class `BaseTest` define following method:
```java
@Before
public void setUp() {
    // Perform some test case setup.
}
```
If a class `DerivedTest extends BaseTest` overrides that method, it should look as follows:
```java
@Before
@Override
public void setUp() {
    super.setUp();
    // Perform some extra test case setup.
}
```

This behavior is enforced for following annotation types:
- `javax.annotation.OverridingMethodsMustInvokeSuper`
- `org.junit.After`
- `org.junit.Before`
- `org.junit.Test`
- `org.junit.jupiter.api.AfterEach`
- `org.junit.jupiter.api.BeforeEach`
- `org.junit.jupiter.api.Test`


dmdashenkov:
@armiol, PTAL.

PR #673, opened 2018-02-27
Fix `CommandFactory.create(...)` doc

### LINKED ISSUE #672: The `targetVersion` of `CommandFactory.create(Message, int)` is not documented properly
Currently it is documented as follows.

```java
 @param targetVersion the ID of the entity for applying commands if {@code null}
                      the commands can be applied to any entity
```

When clearly it isn't what we say about it.

Also, for some reason, we check this `int`-typed parameter for not being `null`.

In scope of this issue we need to fix the documentation and address the nullability check.

dmdashenkov:
This PR fixes an issue found in the `CommandFactory.create(command, targetVersion)` documentation. 

The description of `targetVersion` parameter was simply wrong and irrelevant.

Fixes #672.

### DIFF
--- a/client/src/main/java/io/spine/client/ActorRequestFactory.java
+++ b/client/src/main/java/io/spine/client/ActorRequestFactory.java
@@ -93,14 +93,29 @@ public class ActorRequestFactory {
         return result;
     }
 
+    /**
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@armiol, PTAL.

PR #676, opened 2018-03-02
Update dependent repositories own version along with the `core-java` dependency

dmitrykuzmin:
This PR is an addition to the https://github.com/SpineEventEngine/core-java/pull/670.

It introduces changes which will allow dependent repositories to auto-update their own version along with the `core-java` dependency.

On the successful build in the `master` branch of `core-java`, pull requests are opened in the `jdbc-storage` and `gae-java` repositories. These pull requests contain version update for the `core-java` dependency in the corresponding repository, and now, with the changes in this PR, the own version of the repository will be updated to the same value too.

PR #670, opened 2018-02-26
Auto-update core library dependency version in the `gae-java` and `jdbc-storage` repositories

dmitrykuzmin:
This PR introduces changes which allow to keep track of the `core-java` library version and automatically update it in the `gae-java` and `jdbc-storage` repositories.

Every time the code is merged into the `master` branch of the `core-java`, the script runs and obtains the current version of the core libraries from the `core-java` `ext.gradle` file. The obtained version is then compared to the core library dependencies version in the `gae-java` and `jdbc-storage` repositories. If the dependency version doesn't match the actual version, the script creates a PR updating it in the corresponding repository.

Currently, all PRs and commits are created under `dmitrykuzmin` GitHub account. The instruction on how to switch the GitHub user for the script can be found in the `scripts/dependant-repositories/update-dependant-repositories.sh` file.

PR #677, opened 2018-03-05
Add library version to the name of the dependency-updating branch

dmitrykuzmin:
This pull request is an addition to the https://github.com/SpineEventEngine/core-java/pull/670 and https://github.com/SpineEventEngine/core-java/pull/676.

Previously, the script updating `core-java` dependency in the `jdbc-storage` and `gae-java` repositories was always creating the branch with the same generic name, independent of the new version of the library. Now, the script will insert the new library version into the created branch name, for example: `update-to-core-java-0.10.26-SHAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

PR #678, opened 2018-03-06
Add lifecycle flags to test records

dmdashenkov:
In order to make tests pass upon non-in-memory storage implementations, the records used for tests should have the lifecycle entity columns, `archived` and `deleted`.

In this PR we add these columns where required.


dmdashenkov:
@armiol, PTAL.

server/src/test/java/io/spine/server/storage/RecordStorageShould.java
```diff
@@ -109,6 +118,22 @@ public EntityRecord apply(
                 }
             };
 
+    private static final EntityColumn ARCHIVED;
```
armiol: Please make it an enumeration.

Issue #635, opened 2017-12-14
Transform `Columns` into per-BoundedContext class


alexander-yevsyukov:
Now the `Columns` class mixes utilities and cache of known entity columns.  It should be transformed into a class which is instantiated per `BoundedContext` instance.

Think of having a class that represents columns per entity, to avoid static utility methods here that pass data to each other.

The Javadoc for this class says:
```
Inherited columns are taken into account too, but building entity hierarchies is strongly discouraged.
```
We need to elaborate on why it is discouraged OR disable such a behaviour at all.

alexander-yevsyukov:
@armiol, FYI.

armiol:
Resolved in #681.

PR #681, opened 2018-03-09
Extract entity columns cache to per-Storage class

dmitrykuzmin:
This PR fixes the issue https://github.com/SpineEventEngine/core-java/issues/635.

Previously, the class `Columns` mixed utility methods for working with `EntityColumns` and the global cache of known `EntityColumns`.

Now, the `Columns` class retains only utility methods, while the cache was moved to the separate class `EntityColumnCache`. The entity column cache is initialized per `Storage` instance and stores the column metadata for the `Entity` class with which the `Storage` works. 

As only `RecordStorage` and `RecordBasedRepository` support functionality related to `EntityColumns`, the cache is also supported only by these sub-classes of `Repository` and `Storage`.


server/src/main/java/io/spine/server/entity/storage/EntityColumnCache.java
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * <p>The data is stored this way for convenient querying of the specific columns.
+     *
+     * <p>This container is mutable and thread safe.
+     */
+    private final Map<String, EntityColumn> entityColumnData = synchronizedMap(
+            new LinkedHashMap<String, EntityColumn>());
+
+    @VisibleForTesting
+    private EntityColumnCache() {
```
armiol: I don't understand how private ctor can be `@VisibleForTesting`. It's worth commenting here.

server/src/main/java/io/spine/server/entity/storage/Columns.java
```diff
@@ -147,94 +107,158 @@ public static void checkColumnDefinitions(EntityClass<?> entityClass) {
...
+     * <p>Records the result to {@link Map}.
+     *
+     * @param columns {@link Collection collection} of columns to extract values from
+     * @param entity  {@link Entity} from which to extract the values
+     * @return a {@link Map} of the column {@linkplain EntityColumn#getStoredName()
+     *         names for storing} to their {@linkplain MemoizedValue memoized values}
+     * @see MemoizedValue
+     */
+    private static <E extends Entity<?, ?>> Map<String, MemoizedValue> recordColumnValuesToMap(
+            Collection<EntityColumn> columns,
```
armiol: We tend to move the method name to the next line instead of separating the list of arguments, as you did here.

server/src/main/java/io/spine/server/entity/storage/Columns.java
```diff
@@ -147,94 +107,158 @@ public static void checkColumnDefinitions(EntityClass<?> entityClass) {
...
+     *
+     * <p>Records the result to {@link Map}.
+     *
+     * @param columns {@link Collection collection} of columns to extract values from
+     * @param entity  {@link Entity} from which to extract the values
+     * @return a {@link Map} of the column {@linkplain EntityColumn#getStoredName()
+     *         names for storing} to their {@linkplain MemoizedValue memoized values}
+     * @see MemoizedValue
+     */
+    private static <E extends Entity<?, ?>> Map<String, MemoizedValue> recordColumnValuesToMap(
```
armiol: I wouldn't mention the `Map` in the method name, as it's obvious from the declared returning value. Besides, the name is too long already.

I would even rename it to `readValues`.

server/src/main/java/io/spine/server/entity/storage/Columns.java
```diff
@@ -147,94 +107,158 @@ public static void checkColumnDefinitions(EntityClass<?> entityClass) {
...
+        checkNotNull(entity);
+
+        final Class<? extends Entity> entityClass = entity.getClass();
+        if (!isPublic(entityClass)) {
+            return Collections.emptyMap();
         }
-        addToCache(entityType);
+
+        final Map<String, MemoizedValue> values = recordColumnValuesToMap(entityColumns, entity);
+
```
armiol: Please kill this empty line.

server/src/main/java/io/spine/server/entity/storage/Columns.java
```diff
@@ -147,94 +107,158 @@ public static void checkColumnDefinitions(EntityClass<?> entityClass) {
...
-                entityColumns.add(column);
-            }
+        return true;
+    }
+
+    /**
+     * Generates {@linkplain MemoizedValue memoized values} for the given {@linkplain EntityColumn columns}
+     * of the given {@link Entity}.
+     *
+     * <p>Records the result to {@link Map}.
```
armiol: This is neither here nor there. It's either redundant, since the method signature tells the same. Or, if you wanted to say more about some specifics in the returned `Map` instance (such as describing the keys and values, perhaps, with an example — which I think is redundant for now), then you should've done so.

server/src/main/java/io/spine/server/entity/storage/Columns.java
```diff
@@ -147,94 +107,158 @@ public static void checkColumnDefinitions(EntityClass<?> entityClass) {
...
+     * the given {@link Entity} class can be skipped.
+     *
+     * @param entity        an {@link Entity} to get the {@linkplain EntityColumn column} values from
+     * @param entityColumns {@linkplain EntityColumn entity columns} which values should be extracted
+     * @param <E>           the type of the {@link Entity}
+     * @return a {@link Map} of the column {@linkplain EntityColumn#getStoredName()
+     *         names for storing} to their {@linkplain MemoizedValue memoized values}
+     * @see MemoizedValue
+     */
+    static <E extends Entity<?, ?>> Map<String, MemoizedValue> extractColumnValues(
```
armiol: I think the documentation of this method lacks the fact that we don't process non-`public` classes (as I can tell from the implementation, of course; but still it's an important thing for those who judges by the docs).

server/src/main/java/io/spine/server/entity/storage/Columns.java
```diff
@@ -147,94 +107,158 @@ public static void checkColumnDefinitions(EntityClass<?> entityClass) {
...
+     * <p>Performs checks for entity column definitions correctness along the way.
+     *
+     * <p>If check for correctness fails, throws {@link IllegalStateException}.
+     *
      * @param entityClass the class containing the {@link EntityColumn} definition
      * @return a {@link Collection} of {@link EntityColumn} corresponded to entity class
+     * @throws IllegalStateException if entity column definitions are incorrect
      */
-    static Collection<EntityColumn> getColumns(Class<? extends Entity> entityClass) {
+    static Collection<EntityColumn> obtainColumns(Class<? extends Entity> entityClass) {
```
armiol: There is a lot of similar methods, which aim is to read/extract/obtain the columns. I would create an inner (?, or top-level?) class (such as `ColumnReader` or something) and use its instance (i.e. its instance methods!) instead of chaining the static method calls.

Also, there is another point. Some of these methods are package-private by design, others look `private`, but they are still package-private (most likely, for testing). I think this issue will fade away once the methods are extracted to one or more of such helper classes.

server/src/main/java/io/spine/server/entity/storage/EntityColumnCache.java
```diff
@@ -0,0 +1,201 @@
...
+ * @see Columns
+ */
+@Internal
+public class EntityColumnCache {
+
+    private final Class<? extends Entity> entityClass;
+    private boolean columnsCached = false;
+
+    /**
+     * A one to one container of the {@link EntityColumn} name and its data.
```
armiol: `one-to-one` isn't a clear statement. Please rewrite to make it more readable. It deserves a sentence.

E.g.
```
A container of {@link EntityColumn entity column} name and its data.

<p>Each {@code EntityColumn} has its own cache. <Or whatever you mean by `one-to-one`.>
...
```

server/src/main/java/io/spine/server/entity/storage/EntityColumnCache.java
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * <p>The data is stored this way for convenient querying of the specific columns.
+     *
+     * <p>This container is mutable and thread safe.
```
armiol: `thread-safe`.

server/src/main/java/io/spine/server/entity/storage/EntityColumnCache.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    private final Map<String, EntityColumn> entityColumnData = synchronizedMap(
```
armiol: I would move `synchronizedMap(` to the next line.

server/src/main/java/io/spine/server/entity/storage/EntityColumnCache.java
```diff
@@ -0,0 +1,201 @@
...
+import static io.spine.validate.Validate.checkNotEmptyOrBlank;
+import static java.lang.String.format;
+import static java.util.Collections.synchronizedMap;
+
+/**
+ * A class designated to store cached {@linkplain EntityColumn entity column metadata} for a single
+ * {@link Entity} class.
+ *
+ * <p>The cache remains empty on creation. The {@linkplain EntityColumn column metadata} is retrieved
+ * on the first access to the {@linkplain EntityColumnCache cache} instance and then is stored
```
armiol: `and stored in it`. There is no need to mention the implementation details here.

I would even remove the references to `Columns` usage in the paragraph below (or the whole paragraph).

dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/entity/Repository.java
```diff
@@ -371,6 +375,23 @@ protected void logError(String msgFormat,
...
+     * <p>This method relies on the {@linkplain Columns#checkColumnDefinitions(Class) Columns}
+     * utility to perform the check.
+     *
+     * <p>If {@link Column} definitions are incorrect, the {@link IllegalStateException}
+     * is thrown.
+     *
+     * @throws IllegalStateException in case entity column definitions are incorrect
+     */
+    @Internal
+    void checkEntityColumnDefinitions() {
```
armiol: It seems to me, that a generic `Repository` cannot have the entity columns in all cases. It's `RecordBasedRepository`, that has. Therefore I think this method is too high in the `Repository` hierarchy.

Let's discuss.

server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * <p>The reader can be further used to {@linkplain ColumnReader#readColumns() obtain}
+     * {@linkplain EntityColumn entity columns} for the given class.
+     *
+     * <p>{@code Null} entity class as an argument is not accepted by this method.
+     *
+     * @param entityClass {@link Entity} class for which to create the instance
+     * @return new instance of {@code ColumnReader} for the specified class
+     */
+    static ColumnReader createForClass(Class<? extends Entity> entityClass) {
```
armiol: `forClass` is good enough and shorter.

server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
```diff
@@ -0,0 +1,133 @@
...
+        this.entityClass = entityClass;
+    }
+
+    /**
+     * Creates an instance of {@link ColumnReader} for the given {@link Entity} class.
+     *
+     * <p>The reader can be further used to {@linkplain ColumnReader#readColumns() obtain}
+     * {@linkplain EntityColumn entity columns} for the given class.
+     *
+     * <p>{@code Null} entity class as an argument is not accepted by this method.
```
armiol: This is a general contract that we have for all the methods by default. See `package-info.java`. 

Please remove this sentence.

server/src/test/java/io/spine/server/entity/storage/ColumnValueExtractorShould.java
```diff
@@ -0,0 +1,117 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+/**
+ * @author Dmytro Kuzmin
+ */
+public class ColumnValueExtractorShould {
+
+    private static final String STRING_ID = "some-string-id-never-used";
```
armiol: Please rename it to reflect its meaning. I don't understand what it is for.

dmitrykuzmin:
@armiol PTAL.

dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -399,6 +415,31 @@ private E toEntity(EntityRecord record) {
...
     }
 
+    /**
+     * Retrieves the {@link EntityColumnCache} used by this repository's {@linkplain RecordStorage storage}.
+     *
+     * @return the entity column cache from the storage
+     * @throws IllegalStateException if the {@link EntityColumnCache} is not supported by this
+     *                               repository's storage
+     */
+    private EntityColumnCache retrieveEntityColumnCache() {
```
armiol: Can we just rename it to `columnCache()`?

armiol:
@dmitrykuzmin please make sure that before the merge the library version is increased, compared to one in `master`.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -313,7 +329,7 @@ protected E findOrCreate(I id) {
         checkNotNull(filters);
         checkNotNull(fieldMask);
 
-        final EntityQuery<I> entityQuery = EntityQueries.from(filters, getEntityClass());
+        final EntityQuery<I> entityQuery = EntityQueries.from(filters, columnCache());
```
armiol: Similar to the JDBC library, let's hide the entity column cache from the public API. `EntityQueries.from(...)` is a good example of external API.

dmitrykuzmin:
@armiol PTAL.

Issue #680, opened 2018-03-07
Remove `grpc-all` dependency


dmdashenkov:
Currently, the `server` module has the `io.grpc:grpc-all` dependency. This causes issues with the colliding third parties (such as Mockito), not to mention that `io.grpc:grpc-testing` is a part of `grpc-all`, making it a part of Spine-dependent projects production compile- and runtime classpath.

For now, the `io.grpc:grpc-testing` is excluded.

PR #685, opened 2018-03-13
Use resource file names from `Resources`

alexander-yevsyukov:
This PR...
* migrates to newer version of `base` for using resource file names declared in the `Resources` class.
* narrows down the gRPC dependency (see issue #680 for details), and bumps it to `1.10.0`.

The version of Spine advances to `0.10.32-SNAPSHOT`.


alexander-yevsyukov:
@dmdashenkov, PTAL.

PR #686, opened 2018-03-21
Allow to declare rejection handlers for the same rejection message, but for different command messages

### LINKED ISSUE #675: Unable to declare rejection subscribers for the same rejection message, but for different command messages
Now, it is impossible to declare rejection subscribers for the same rejection message, but for different command messages in a single class. For the following code, `MessageHandlerMap.scan(...)` will throw `DuplicateHandlerMethodError`:

```java
public class RejectionNotifier extends RejectionSubscriber {

    @Subscriber
    public void on(CannotModifyArchivedEntity f, StartTask c) { ... }

    @Subscriber
    public void on(CannotModifyArchivedEntity f, CancelTask c) { ... }
}
```

Such a code looks totally valid. It is useful when a rejection message is broad enough like `CannotModifyArchivedEntity` or `CannotModifyDeletedEntity`, but some specific data from the command message are required.

Also, consider a case, when there is a subscriber with and without the command message parameter (for the same rejection message):

```java
public class RejectionNotifier extends RejectionSubscriber {

    @Subscriber
    public void on(CannotModifyArchivedEntity f, StartTask c) { ... }

    @Subscriber
    public void on(CannotModifyArchivedEntity f) { ... }
}
```

Probably, for this case, if `CannotModifyArchivedEntity` was not caused by `StartTask` command, then it should be dispatched to the subscriber without the command message parameter.

dmytro-grankin:
This PR fixes #675.

Previously, a message handler was identified by a class of a handled message. `HandlerKey` class was introduced to be used instead of `MessageClass`. `HandlerKey` contains a message class of a handled message and a message class of its origin. For example, if `CannotStartProject` rejection was caused by `StartProject` command, in this case the origin for `HandlerKey` is `MessageClass` of `StartProject`.

Now, `MessageHandlerMap` allows obtaining of a handler by a key, a message class, a message and origin class.

The handler selection logic for `MessageHandlerMap.getMethod(messageClass, originClass)`:
1) Check whether there is a handler, which matches the message and origin class.
2) If there is no handler matching the origin class, select by the message class only.

`RejectionHandlerMethod`s construct `HandlerKey` with `CommandClass` as an origin, so a command message parameter in a handler will affect the selection of the handler. Other handler methods construct `HandlerKey` without an origin, hence the handlers are selected by a message class only.

The framework version was increased to `0.10.34-SNAPSHOT`.


dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -243,7 +244,10 @@ protected B getBuilder() {
      *         response to this rejection
      */
     List<? extends Message> reactOn(RejectionEnvelope rejection) {
-        final RejectionReactorMethod method = thisClass().getReactor(rejection.getMessageClass());
+        final CommandClass commandClass = CommandClass.of(rejection.getCommandMessage()
+                                                                   .getClass());
```
armiol: Let's remove `getClass()`call.

PR #693, opened 2018-04-25
Modify Transport and Sharding API for SPI users

armiol:
This pull request modifies the core concepts of `Transport`, so that it is more convenient to create custom `Subscriber` and `Publisher` implementations. 

Also, the `Sharding`-related tests are refactored, so that they are extensible and can be reused with custom implementations of sharding service and its underlying transport.

All the changes are aimed primarily for extension in Spine libraries (such as `gcloud-java`) and also for SPI users.

The framework version has been increased to `0.10.36-SNAPSHOT`.


armiol:
@alexander-yevsyukov PTAL.

PR #694, opened 2018-04-26
Update Encrypted Credentials

dmdashenkov:
One of the credential files (in particular, the GCS service account key) has become obsolete.

In this PR we update the whole encrypted `credentials.tar` archive, which now contains a new key.

Also, the private RSA key, used for the GitHub auth on Travis is now found in that archive as well.


dmdashenkov:
@armiol, PTAL.
This change is to be propagated to other repositories as well.

Issue #692, opened 2018-04-24
Empty event is not filtered out from event reaction results in Process Manager in case Either is used


yuri-sergiichuk:
Currently, in case `EitherOf` is used as a return type for `@React` method in Process Manager and is populated with `Empty` it will not be filtered out inside `ProcessManager#dispatchEvent` and a list with `empty` event is returned:
```java
List<Event> dispatchEvent(EventEnvelope event)  {
  final EventReactorMethod method = thisClass().getReactor(event.getMessageClass());
  final List<? extends Message> eventMessages =
        method.invoke(this, event.getMessage(), event.getEventContext());
  final List<Event> events = toEvents(eventMessages, event); // <--- we should filter results here
  return events;
}
```

In aggregates, it works, cause there is an additional filtration with a `NonEmpty` predicate:
```java
protected List<? extends Message> dispatchCommand(CommandEnvelope command) {
  idempotencyGuard.check(command);
  final CommandHandlerMethod method = thisClass().getHandler(command.getMessageClass());
  final List<? extends Message> messages =
        method.invoke(this, command.getMessage(), command.getCommandContext());
  return from(messages).filter(nonEmpty()).toList();
}
```

mdrachuk:
Implemented in https://github.com/SpineEventEngine/core-java/pull/697.

mdrachuk:
Implemented in #697.

PR #697, opened 2018-04-30
Filter Empty messages from events emitted by PM

mdrachuk:
This PR resolves an issue https://github.com/SpineEventEngine/core-java/issues/692 described in the subject.

The issue is fixed by implementing new `Dispatch` layer common for Aggregate and Process Manager. It takes care of dispatch of an envelope to a target, filtering and processing the dispatch results.

## Black Box Bounded Context
Additionally, in this PR new test utils are introduced. Black Box Bounded Context allows to register Bounded Context repositories, send Commands and Events to it, and verify the Bounded Context behaviour and state.

Example:
```java
BlackBoxBoundedContext
    .with(new TaskAggregateRepository())
    .receivesCommand(assignTask())
    .verifiesThat(acked(once()).withoutErrorsOrRejections())
    .verifiesThat(emitted(AggTaskAssigned.class, twice()))
    .verifiesThat(emitted(AggUserNotified.class))
    .close();
```


server/src/main/java/io/spine/server/utils/MessagePredicates.java
```diff
@@ -0,0 +1,62 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.utils;
```
mdrachuk: I know we‘re trying to get rid of the utils, but I couldn‘t figure out a better place for this. 
Any suggestions?

server/src/test/java/io/spine/server/procman/given/ProcessManagerTestEnv.java
```diff
@@ -60,7 +78,129 @@
...
+    }
+
+    private static TestActorRequestFactory requestFactory(TenantId tenantId) {
+        return TestActorRequestFactory.newInstance(AggregateTestEnv.class, tenantId);
+    }
+
+    /**
+     * Reads all events from the bounded context for the provided tenant.
+     */
+    public static List<Event> readAllEvents(final BoundedContext boundedContext,
```
mdrachuk: Should this methods be moved to `testutils`? 
We‘re using them in `examples-java` already and right now they‘re populating each test environment.

server/src/main/java/io/spine/server/utils/MessagePredicates.java
armiol: Please put that to `base` repository. You'll be looking for `io.spine.protobuf.Messages`.

server/src/test/java/io/spine/server/procman/given/ProcessManagerTestEnv.java
armiol: @mdrachuk I've browsed `testutil-server` for a good location, but this case is a bit orphaned at the moment and it's difficult to generalize it nicely. 

Let's create `io.spine.server.event.TestEventReader` utility class in `testutil-server`. To my understanding, that's the best and the most unobtrusive way to do it for now.

server/src/main/java/io/spine/server/utils/MessagePredicates.java
alexander-yevsyukov: I would suggest keeping `MessagePredicates` and putting them next to `Messages`. Two specialized utility classes is better than a `ParentGodOfAllUtils`.

server/src/test/java/io/spine/server/procman/given/ProcessManagerTestEnv.java
mdrachuk: @armiol It‘s not just reading events. It‘s also `newBoundedContext`, `newTenantId`, `closeContext` and `requestFactory`. All of those are identical throughout framework and its example tests.

mdrachuk:
@armiol PTAL.

server/src/main/java/io/spine/server/command/CommandHandlingEntity.java
```diff
@@ -172,4 +176,70 @@ protected ValueMismatch expectedNotEmpty(String expected) {
...
+             *
+             * @param message the message being checked
+             * @return {@code true} if the message is not empty, {@code false} otherwise
+             */
+            @Override
+            public boolean apply(Message message) {
+                return !message.equals(EMPTY);
+            }
+        }
+
```
armiol: This line is redundant.

server/src/main/java/io/spine/server/command/CommandHandlingEntity.java
```diff
@@ -172,4 +176,70 @@ protected ValueMismatch expectedNotEmpty(String expected) {
     protected ValueMismatch unexpectedValue(String expected, String actual, String newValue) {
         return StringMismatch.unexpectedValue(expected, actual, newValue, versionNumber());
     }
+
+    /**
+     * A object which performs filtering of the events emitted by a
+     * {@link CommandHandlingEntity command handling entity}.
+     */
+    protected static class EventFilter {
```
armiol: As long as it contains another class (i.e. an `enum` declaration) I'd move it to the upper level for it to become a standalone class.

Then I'd review the method visibility. Right now (while it's still a `protected` nested) its `public static` methods look confusing. Let's see how they will look after the class is extracted.

server/src/main/java/io/spine/server/command/CommandHandlingEntity.java
```diff
@@ -172,4 +176,70 @@ protected ValueMismatch expectedNotEmpty(String expected) {
     protected ValueMismatch unexpectedValue(String expected, String actual, String newValue) {
         return StringMismatch.unexpectedValue(expected, actual, newValue, versionNumber());
     }
+
+    /**
+     * A object which performs filtering of the events emitted by a
```
armiol: That isn't an object yet. 
```A filter for events emitted by ... ```

Also there should be an explanation of why we have it.

Another idea I have (looking at its current usage) is that the ideal way of incorporating it into the flow would be the following. Right now it's possible to "forget" to invoke the filter. What if we force SPI users to do that (those who may want to extend the framework).

Let's discuss in person.

File: server/src/test/java/io/spine/server/procman/ProcessManagerShould.java [hunk 6]
```diff
@@ -288,6 +307,107 @@ public void create_router() {
...
+     *     <li>{@link io.spine.test.procman.exam.command.PmStartExam Start Exam} — to initialize the PM;
+     *     <li>{@link io.spine.test.procman.exam.command.PmAnswerProblem Answer Problem } — a target
+     *         command that produces either of 3 events.
+     * </ol>
+     *
+     * <p>First command emits an {@link io.spine.test.procman.exam.event.PmExamStarted Exam Started}
+     * event.
+     *
+     * <p>Because the exam is started without any problems to solve, 
+     * {@link io.spine.test.procman.exam.command.PmAnswerProblem answer problem command} can not
```
armiol: `cannot`.

server/src/test/java/io/spine/server/procman/ProcessManagerShould.java
```diff
@@ -288,6 +307,107 @@ public void create_router() {
...
+     *
+     * <p>Because the exam is started without any problems to solve, 
+     * {@link io.spine.test.procman.exam.command.PmAnswerProblem answer problem command} can not 
+     * match any problems. This results in emitting {@link io.spine.server.tuple.EitherOfThree Either Of Three} 
+     * containing {@link com.google.protobuf.Empty Empty}.
+     *
+     * @see io.spine.server.procman.given.DirectExamProcman
+     */
+    @Test
+    public void not_create_event_if_command_handle_results_is_either_of_three_with_Empty() {
```
armiol: We tend to use `emit event` instead of `create`.

server/src/test/proto/spine/test/procman/exam/commands.proto
```diff
@@ -0,0 +1,42 @@
...
+option java_multiple_files = true;
+
+import "spine/test/procman/exam/exam.proto";
+
+message PmStartExam {
+    PmExamId exam_id = 1;
+    repeated PmProblemId problem = 2;
+}
+
+message PmAnswerProblem {
```
armiol: I'd say `PmAddressProblem`.

server/src/test/proto/spine/test/procman/exam/commands.proto
```diff
@@ -0,0 +1,42 @@
...
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.test.procman.exam.command";
+option java_outer_classname = "ProcmanExamCommandsProto";
+option java_multiple_files = true;
+
+import "spine/test/procman/exam/exam.proto";
+
+message PmStartExam {
+    PmExamId exam_id = 1;
+    repeated PmProblemId problem = 2;
```
armiol: It looks weird. Starting an exam assumes that it already exists.

I'd say `CreateTest`, `StartTest`, `ProposeSolution` etc. To me that's more common.

testutil-server/src/main/java/io/spine/server/TestBoundedContexts.java
```diff
@@ -0,0 +1,142 @@
...
+public class TestBoundedContexts {
+
+    /**
+     * Wraps the provided domain command message in a test command with a provided tenant ID.
+     *
+     * @param commandMessage a domain command
+     * @param tenantId       an ID of a tenant in multitenant bounded context
+     * @return a newly created command instance
+     */
+    public static Command command(Message commandMessage, TenantId tenantId) {
```
armiol: I don't think it belongs here. We already have `TestActorRequestFactory`. Let's discuss in person.

testutil-server/src/main/java/io/spine/server/TestBoundedContexts.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private static TestActorRequestFactory requestFactory() {
+        return TestActorRequestFactory.newInstance(TestBoundedContexts.class);
+    }
+
+    /**
+     * @return a new {@link TenantId Tenant ID} with a random UUID value convenient
+     * for test purposes.
+     */
+    public static TenantId newTenantId() {
```
armiol: This method (and those above) aren't about bounded contexts.

testutil-server/src/main/java/io/spine/server/TestBoundedContexts.java
```diff
@@ -0,0 +1,142 @@
...
+import static io.spine.Identifier.newUuid;
+import static io.spine.grpc.StreamObservers.memoizingObserver;
+
+/**
+ * Utilities for tests that deal with {@link BoundedContext Bounded Context}.
+ * 
+ * @author Mykhailo Drachuk
+ */
+@VisibleForTesting
+public class TestBoundedContexts {
```
armiol: Right now it's a bit of a mess. Those are just the methods joined together. 

Instead we need a helper for writing the integration tests (or something). Like an API for white-box / black-box testing approaches. We are going to invent it :) Let's discuss in person.

alexander-yevsyukov:
@armiol, can someone pick it up while Misha is not available?

mdrachuk:
@armiol PTAL.

server/src/test/java/io/spine/server/aggregate/AggregateTest.java
```diff
@@ -835,33 +830,14 @@ void acknowledgeExceptionForDuplicateCommand() {
...
-            Rejection emptyRejection = Rejection.getDefaultInstance();
-            assertEquals(emptyRejection, status.getRejection());
-
-            List<Event> events = readAllEvents(boundedContext, tenantId);
-            assertSize(1, events);
-            closeContext(boundedContext);
+            BlackBoxBoundedContext
+                    .with(new TaskAggregateRepository())
+                    .receivesCommand(createTask())
+                    .verifiesThat(acked(1).withoutErrorsOrRejections())
```
armiol: I think that `acked` and `emitted` API methods need to be more clear to users and code-readers. Right now I am not even sure that I can 100% tell what they do. I can only try and guess.

Let's discuss in person.

testutil-server/src/main/java/io/spine/server/integration/Acknowledgements.java
```diff
@@ -0,0 +1,222 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.integration;
```
armiol: It looks that this class should be in `testutil-client`, as it's a client who listens for Acks and other stuff. Therefore the package should differ from the current one.

testutil-server/src/main/java/io/spine/server/integration/AcknowledgementsVerifier.java
```diff
@@ -0,0 +1,612 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.integration;
```
armiol: Same as above.

testutil-server/src/main/java/io/spine/server/integration/AcknowledgementsVerifier.java
```diff
@@ -0,0 +1,612 @@
...
+import java.util.Collection;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.Lists.newArrayList;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
+ * An abstract verifier of acknowledgements.
```
armiol: Please tell a bit about what these verifiers serve for.

testutil-server/src/main/java/io/spine/server/integration/AcknowledgementsVerifier.java
```diff
@@ -0,0 +1,612 @@
...
+     * @param predicate     a predicate filtering domain rejections
+     * @param <T>           a domain rejection type
+     * @return a new {@link AcknowledgementsVerifier} instance
+     */
+    public <T extends Message> AcknowledgementsVerifier
+    withRejections(int expectedCount, Class<T> type, RejectionPredicate<T> predicate) {
+        AcknowledgementsVerifier oneRejection = ackedWithRejections(expectedCount, type, predicate);
+        return this.and(oneRejection);
+    }
+}
```
armiol: This class seems to be a bit bigger than seems manageable. Can we please extract some of its bits as package-private classes under the same package?

testutil-server/src/main/java/io/spine/server/integration/ErrorQualifier.java
```diff
@@ -0,0 +1,150 @@
...
+public abstract class ErrorQualifier {
+
+    /**
+     * A message describing the qualifier.
+     *
+     * <p>Used to report an assertion error.
+     */
+    public abstract String description();
+
+    public abstract boolean test(Error error);
```
armiol: This is its main API element, but it is undocumented.

I also think that qualifiers (as a term) tell to which group of object a particular one belongs. That is, qualify it.

What you imply by its API is that the instances of this class act as predicates. The method `test(T):boolean` is the same as in `java.util.function.Predicate` (we are in Java 8 now, you are welcome to use it). 

Therefore I am confused by the naming.

testutil-client/src/test/java/io/spine/client/blackbox/AcknowledgementsTest.java
```diff
@@ -48,9 +48,9 @@
 import static io.spine.client.blackbox.given.CommandAcksTestEnv.projectAlreadyStarted;
 import static io.spine.client.blackbox.given.CommandAcksTestEnv.taskCreatedInCompletedProject;
 import static io.spine.client.blackbox.given.CommandAcksTestEnv.taskLimitReached;
-import static io.spine.testing.client.blackbox.Rejections.BbProjectAlreadyStarted;
-import static io.spine.testing.client.blackbox.Rejections.BbTaskCreatedInCompletedProject;
-import static io.spine.testing.client.blackbox.Rejections.BbTaskLimitReached;
+import static io.spine.test.testutil.blackbox.Rejections.BbProjectAlreadyStarted;
```
alexander-yevsyukov: @mdrachuk , @armiol , what is `testutil` in this context? 

I understand, classes under `test` are test data, but what kinds of utilities we have under `test`?
If they are really utilities, why aren't they parts of `testing.server` or `testing.client` according to our new naming conventions?

armiol: @mdrachuk @alexander-yevsyukov as [discussed previously](https://github.com/SpineEventEngine/core-java/pull/740#discussion_r201697877), the repackaging of `testutil-*` modules is currently in progress. This package naming will be changed in scope of the repackaging in separate PRs.

mdrachuk: @vlad-lubenskyi is currently working on repackaging existing test-utils, so I’ve set them to match an existing convention to avoid any confusion.

Issue #491, opened 2017-05-29
Enum Entity Columns


dmdashenkov:
For now, Entity Columns that end up to be of `enum` type are not supported for the `Any` packing reasons.
It's worth notifying the user about this.
Also, we should come up with the policy of handling the `enum` types. The options are:
 - making them invalid and throwing an exception;
 - making them work through the `int ordinal` or `String name` representation (the way it's done in Hibernate etc.);
 - handling them invisibly from the user using the strategy defined above.

armiol:
Let's mimic [`@Enumerated`](http://docs.oracle.com/javaee/6/api/javax/persistence/Enumerated.html) behavior.

dmitrykuzmin:
Currently, the [Java Enum](https://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html) type of Entity Columns will be supported. The [Protobuf Enum](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Enum) type support can be added later after the functionality supporting Java enums is ready.

dmitrykuzmin:
Done.

PR #698, opened 2018-05-08
Add support for Enum Entity Columns

dmitrykuzmin:
This PR resolves the issue https://github.com/SpineEventEngine/core-java/issues/491.

Now the `@Column` annotation can be used on `getters` which return the [Java Enum](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) type.

Enum values are persisted in the storage through their `ordinal` or their `name`. A new annotation `@Enumerated` allows user to specify the storage type; by default all Enum Entity Columns are `ordinal` (even those not annotated with `@Enumerated`). This behaviour imitates the [JPA Enumerated](https://docs.oracle.com/javaee/6/api/javax/persistence/Enumerated.html).

`ColumnFilter` instances are created with Enum values directly, so they do not depend on the Column storage type.

The Spine version was bumped to `0.10.38-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

client/src/main/java/io/spine/client/FilterValueConverter.java
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * <p>This class performs all necessary actions preceding the value conversion as well as the type
+ * conversion itself.
+ *
+ * <p>{@link ColumnFilter} values can either be {@linkplain com.google.protobuf.Message Protobuf
+ * Messages}, Java primitives, or {@link Enum enumerated values}.
+ *
+ * @author Dmytro Kuzmin
+ */
+public final class FilterValueConverter {
```
armiol: See `Messages` and `TypeConverter` in `io.spine.protobuf` package (from `base` repository).

I think that most of the code in this class will go away. The rest should probably be non-public (i.e. package-private). Or some pieces should be moved to `base`. 

Let's discuss if needed.

client/src/test/java/io/spine/client/FilterValueConverterShould.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Test
+    public void retrieve_enum_value_from_any() {
+        final TestEnum value = TestEnum.ONE;
+        final Any convertedValue = TypeConverter.toAny(value.name());
+        final Object unpackedValue = toValue(convertedValue, TestEnum.class);
+        assertEquals(value, unpackedValue);
+    }
+
+    private enum TestEnum {
```
armiol: We don't use `Test...` naming for test classes anymore. Please think of some example in the projects-tasks bounded contexts, such as `TaskStatus`.

Also, we use `TestEnv`-based approach. In particular, this enum should be an inner member of `FilterValueConverterTestEnv`, which in turn should reside in the `... .given` package.

client/src/main/java/io/spine/client/FilterValueConverter.java
```diff
@@ -0,0 +1,130 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.client;
```
armiol: See my comment below first, but still I am sure that this package isn't the right place, as the task is way too specific.

Let's see how you address the next comment. Maybe this class will go away completely, and we'll have no problem with the package.

server/src/main/java/io/spine/server/entity/storage/ColumnValueConverter.java
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * <p>The input value is assumed to be the one {@linkplain EntityColumn#memoizeFor(Entity)
+     * retrieved} via the {@link EntityColumn} getter.
+     *
+     * <p>The output value is the corresponding value to be used in the data storage.
+     *
+     * @param value the value to convert
+     * @return the value used for the persistence in the data storage
+     */
+    Serializable convert(Object value);
```
armiol: I don't understand how `Serializable` is a "persisted type" (according to the documentation, it is).

I also think that this interface should not be `public`, as its usages should be limited to the package.

I am confused, honestly. Let's discuss in person.

server/src/main/java/io/spine/server/entity/storage/ColumnValueConverter.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.io.Serializable;
+
+/**
+ * An interface for converting the {@link EntityColumn} values to their persisted type.
+ *
+ * @author Dmytro Kuzmin
+ */
+@Internal
+public interface ColumnValueConverter extends Serializable {
```
armiol: I don't understand why this converter should be serializable.

server/src/main/java/io/spine/server/entity/storage/PersistenceInfo.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * An info about how the {@link EntityColumn} is persisted in the data storage.
+ *
+ * <p>This class provides user with the info about {@link EntityColumn} persistence type as well
+ * as the conversion method between the entity column values and the persistence values.
+ *
+ * @author Dmytro Kuzmin
+ * @see Column
+ */
+@Internal
```
armiol: We don't use `@Internal` for package-private classes. We assume they are internal by their definition.

server/src/main/java/io/spine/server/entity/storage/PersistenceInfo.java
```diff
@@ -0,0 +1,137 @@
...
+final class PersistenceInfo implements Serializable {
```
armiol: The name is way too generic. 

Please think of this class as of something performing an action, not just as of storage for some ephemeral information (as it is now). Then you'll be able to give it some responsibility and name it properly.

dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/entity/storage/EntityColumn.java
```diff
@@ -198,13 +199,29 @@
...
+     * <p>The only place where this field is updated, except the constructor, is
+     * {@link #readObject(ObjectInputStream)} method.
+     */
+    private transient ColumnValuePersistor valuePersistor;
+
+    private EntityColumn(Method getter,
+                         String name,
+                         String storedName,
+                         boolean nullable,
+                         ColumnValuePersistor valuePersistor) {
```
armiol: Let's avoid the 5th argument as per vocal discussion.

server/src/main/java/io/spine/server/entity/storage/EntityColumn.java
```diff
@@ -340,6 +363,51 @@ public Class getType() {
...
     }
 
+    /**
+     * Returns the type under which the column values are persisted in the data storage.
+     *
+     * <p>For the non-{@link Enumerated} entity columns
+     * this type will be equal to the one retrieved via the {@link #getType()}.
+     *
+     * <p>For {@link Enumerated} columns, see {@link
+     * EnumPersistenceTypes}.
```
armiol: I think the content of this line can be moved back to the previous line.

server/src/main/java/io/spine/server/entity/storage/EntityColumn.java
```diff
@@ -340,6 +363,51 @@ public Class getType() {
...
+     *
+     * <p>The value type should be the same as the one obtained through the {@link #getType()}. The
+     * output value type will be the same as {@link #getPersistedType()}.
+     *
+     * <p>For the {@code null} input argument the method will always return {@code null},
+     * independently of the declared types.
+     *
+     * <p>The method is accessible outside of the {@link EntityColumn} class to enable the proper
+     * {@link io.spine.client.ColumnFilter} conversion for the {@link
+     * Enumerated} column values.
```
armiol: Please move `Enumerated}` back to the previous line.

server/src/main/java/io/spine/server/entity/storage/EntityColumn.java
```diff
@@ -340,6 +363,51 @@ public Class getType() {
...
+     *
+     * @param columnValue the column value to convert
+     * @return the column value converted to the form used for persistence in the data storage
+     * @throws IllegalArgumentException if the value type is not equal to the {@linkplain #getType()
+     *                                  entity column type}
+     */
+    @Nullable
+    public Serializable toPersistedValue(@Nullable Object columnValue) {
```
armiol: We need to additionally explain why the argument may be `null` and why it makes sense to return `null` in this case back as well.

server/src/main/java/io/spine/server/entity/storage/EntityColumn.java
```diff
@@ -387,6 +475,12 @@ private Method restoreGetter() {
         }
     }
 
+    private ColumnValuePersistor restoreValuePersistor() {
+        checkState(valuePersistor == null, "Column value persistor is already restored.");
```
armiol: I thought we should have just ignored it. But now we throw an exception...

dmitrykuzmin:
@armiol PTAL. I've added the test to the record storage, so all the storage implementations such as `jdbc` and `gcloud` inherit the contract of matching enum-typed filters against the entity columns. All the new functionality is still tested.

dmitrykuzmin:
@armiol PTAL.

server/src/test/java/io/spine/server/entity/storage/ColumnValueConvertersShould.java
```diff
@@ -0,0 +1,82 @@
...
+        assertEquals(OrdinalEnumConverter.class, converter.getClass());
+    }
+
+    private static ColumnValueConverter ofGetter(String name) {
+        try {
+            final Method getter = TestEntity.class.getDeclaredMethod(name);
+            final ColumnValueConverter converter = of(getter);
+            return converter;
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException(e);
```
armiol: I think, that's `Exceptions.illegalArgumentWithCauseOf(...)` (and we usually import this method statically).

dmitrykuzmin:
@armiol PTAL again. A very few minor changes since approval.

Issue #589, opened 2017-09-27
Fix `ProcessManager` versioning


dmdashenkov:
Currently, `ProcessManager`s never update their versions as both events and commands are dispatched bypassing the `Transaction`.

armiol:
@illia-shepilov Seems that event reaction now updates the version. We need to ensure that command handling also leads to a version increment. And have tests for both of these cases. 

In tests please make use of `testutil-...` modules. There are some helpers that test entities against commands (passed to `@Assign` method) and events (fed to `@React` method). Maybe you'll find it helpful to append the test utilities with some API to check (or "expect") a particular version.

armiol:
@dmitrykuzmin please feel free to close this one once #893 is merged.

dmitrykuzmin:
Resolved via #893.

PR #707, opened 2018-05-23
Refactor Entity Hierarchy

dmdashenkov:
## Summary

In this PR we split the concepts of `EventPlayer` and `TransactionalEntity`.

## Motivation

Historically, all the entities which had transactions (such as `Aggregate`s, `Projection`s, and `ProcessManager`s) were able to play to events (handle events, change own state, and produce no response). 

However, now the `ProcessManager`s are not able to declare any kind of such subscriptions. This is why the concept of `ProcessManager.play(events)` has lost its meaning.

## Resolution

The `EventPlayingEntity` (which was previously the supertype of all the transactional entities, see above) is split into two parts: 
 - `TransactionalEntity` is the superclass for all the entities capable of transactions; 
 - `EventPlayer` is the common interface for components which play events.

`Aggregate`, `Projection`, and `ProcessManager` extend `TransactionalEntity` class but only `Aggregate` and `Projection` implement `EventPlayer` contract.

## Implementation Details

The common functionality of playing event streams is extracted into a separate class - `TransactionalEvenPlayer`. It has a common contract with the event playing entities (declared in `EventPlayer` interface). See the diagram below for more details.

diagram1

## Next Changes

As a follow-up change, we should make `ProcessManager` receive messages through a transaction. Currently, they are sent to the handlers directly (see #589).

## Other Changes

 - Commonly used `List<Event>` type is wrapped into the `EvantStream` value type.
 - `@VisibleForTesting` annotation is removed from `getBuilder()` method in `Aggregate`, `Projection`, and `ProcessManager`, as it causes warnings each time it's called in the client code.
 - From now on, `Projection` dispatches events though transaction instead of invoking the handler directly.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/entity/TransactionalEventPlayer.java
```diff
@@ -0,0 +1,63 @@
...
+    public void play(EventStream events) {
+        events.events()
+              .stream()
+              .map(EventEnvelope::of)
+              .forEach(transaction::apply);
+    }
+
+    @Override
+    public String toString() {
+        return "EventPlayer.forTransactionOf(...)";
```
armiol: This is weird. 

Let's print something that:
* points to the fact that this is an `TransactionalEventPlayer`;
* mentions the `transaction` — I'd override the `Transaction.toString()` as it's its responsibility to print itself.

server/src/test/java/io/spine/server/entity/EventPlayingEntityShould.java
```diff
@@ -17,225 +17,57 @@
...
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 /**
- * @author Alex Tymchenko
+ * @author Dmytro Dashenkov
  */
 public class EventPlayingEntityShould {
```
armiol: I think that's a `TransactionalEventPlayerShould` now, as it's the one that fails to play events with no transaction and it's the one to delegate the execution to `tx` when playing.

server/src/test/java/io/spine/server/entity/EventPlayingEntityShould.java
```diff
@@ -245,18 +77,21 @@ private EventPlayingEntity entityWithActiveTx(boolean txChanged) {
         return entity;
     }
 
+    private static void verifyEventApplied(Transaction txMock, Event event) {
+        verify(txMock).apply(eq(EventEnvelope.of(event)));
+    }
+
     private static class EpeEntity
-            extends EventPlayingEntity<Long, StringValue, StringValueVBuilder> {
+            extends TransactionalEntity<Long, StringValue, StringValueVBuilder>
```
armiol: And this class is `TxPlayingEntity` or something.

server/src/main/java/io/spine/server/entity/EventPlayer.java
```diff
@@ -0,0 +1,57 @@
...
+public interface EventPlayer {
+
+    /**
+     * Plays the given {@link EventStream} upon the underlying entity.
+     *
+     * <p>Typically, the entity state is changed upon this operation.
+     *
+     * @param events the event stream to play
+     */
+    void play(EventStream events);
```
armiol: As discussed, let's try to deal with IDEA warning.

dmdashenkov:
@armiol, PTAL again.

server/src/main/java/io/spine/server/entity/EventPlayers.java
```diff
@@ -0,0 +1,54 @@
...
+     * Prevents the utility class instantiation.
+     */
+    private EventPlayers() {
+    }
+
+    /**
+     * Creates a transactional {@link EventPlayer} for the given
+     * {@linkplain TransactionalEntity entity}.
+     *
+     * <p>It is expected that the given entity is currently in a transaction.
```
armiol: Let's mention what happens if this method is invoked with no active transaction. Maybe provide a `{@link}` to a corresponding API element.

PR #714, opened 2018-05-29
Migrate `core` module to new test structure based on JUnit 5

dmitrykuzmin:
This PR continues changes required to migrate to JUnit 5 test structure as described in the https://github.com/SpineEventEngine/core-java/issues/701.

All `core` module tests are migrated to the new structure which takes advantage of JUnit 5 capabilities such as [`@DisplayName`](https://junit.org/junit5/docs/5.0.3/api/org/junit/jupiter/api/DisplayName.html) and [`@Nested`](https://junit.org/junit5/docs/5.0.3/api/org/junit/jupiter/api/Nested.html) classes.

Changes in this PR rely on `base` version `0.10.40-SNAPSHOT` to obtain common test display names.

PR #715, opened 2018-05-30
Migrate `client` to new test naming pattern

dmitrykuzmin:
This PR migrates `client` module tests to the new naming pattern proposed in https://github.com/SpineEventEngine/core-java/pull/714.

From now all `@Nested` classes of the test are named after some action, and their test cases are supposed to represent some condition or clarification (previously it was vice-versa).

A small number of `@Nested` tests that could not fit into this pattern, were removed and their test cases were moved to the outer class.


dmitrykuzmin:
@armiol PTAL.

client/src/test/java/io/spine/client/ColumnFiltersTest.java
```diff
@@ -87,42 +85,42 @@ void passNullToleranceCheck() {
     }
 
     @Nested
-    @DisplayName("when creating column filter")
-    class CreateFilterTest {
+    @DisplayName("create column filter of type")
+    class Create {
```
armiol: `CreateFilterOfType`

client/src/test/java/io/spine/client/ColumnFiltersTest.java
```diff
@@ -138,27 +136,27 @@ private void checkCreatesInstance(ColumnFilter filter,
     }
 
     @Nested
-    @DisplayName("when creating composite column filter")
-    class CreateCompositeFilterTest {
+    @DisplayName("create composite column filter of type")
+    class CreateComposite {
```
armiol: `CreateCompositeFilterOfType`

PR #719, opened 2018-06-01
Remove GCS Plugin

dmdashenkov:
In this PR we remove usage of the Spine GCS plugin.

See [this `base` PR](https://github.com/SpineEventEngine/base/pull/122) for more details.


dmdashenkov:
@armiol, PTAL.

Issue #725, opened 2018-06-21
Command routing in PM blocks command handler execution in local environment


vlad-lubenskyi:
In a single threaded application with `Direct` delivery (a typical local configuration), 

When a command handler in PM routes another command, the command handler execution is blocked until the routed command and all subsequent events are routed and handled.

Let's say we have 2 entities:

`UserAggregate`:
- Has a handler for `CreateUser` which produces `UserCreated`.

`SignInPm`:

- Has a handler for `SignInNewUser` which routes `CreateUser` and changes PM's status to `IN_PROGRESS`.
- Reacts on `UserCreated` and produces `NewUserSignedIn` if the status is `IN_PROGRESS`, `SignInFailed` otherwise.

### Current execution flow:

![copy of untitled diagram 1](https://user-images.githubusercontent.com/11441732/41720314-564b4fb2-756b-11e8-8e59-616c746cfc7d.png)

### Expected execution flow:

![untitled diagram 2](https://user-images.githubusercontent.com/11441732/41720311-546189be-756b-11e8-8d22-b10145473290.png)

Issue #696, opened 2018-04-27
Command routing causes unpredictable results


dmytro-grankin:
If an application works in a single thread and uses `Direct` delivery (test configuration), command routing causes unpredictable results.

Let's say, there are 3 entities: `TaskAggregate`, `TaskProjection` and `TaskCreationProcessManager`.

`TaskAggregate`:
- Has a handler for `CreateTask`, which produces `TaskCreated`.
- Has a handler for `AssignTask`, which produces `TaskAssigned`.

`TaskProjection`:
- Has a subscriber on `TaskCreated`.
- Has a subscriber on `TaskAssigned`.

`TaskCreationProcessManager`:
- Reacts on `TaskCreated` and routes `AssignTask`.

With this configuration, there are two possible scenarios (all in the order of the execution):

1) `TaskCreated` is dispatched to `TaskCreationProcessManager`, the PM routes `AssignTask` and `TaskAssigned` is emitted by the aggregate. Then `TaskAssigned` is dispatched to `TaskProjection` and only then source `TaskCreated` is dispatched to the projection. In this case, `TaskCreated` was emitted before `TaskAssigned`, but the projection handled `TaskAssigned` and only then `TaskCreated`.

![bus queue wrong](https://user-images.githubusercontent.com/25301769/39423808-5a44318c-4c7c-11e8-9527-8f9e6a4f7a92.PNG)

2) `TaskCreated` is dispatched to `TaskProjection`, then it dispatched to `TaskCreationProcessManager`. The PM routes `AssignTask` and `TaskAssigned` is emitted by the aggregate. Then `TaskAssigned` is dispatched to `TaskProjection`. This case is valid unlike the first one, the projection handled `TaskCreated` and then `TaskAssigned`.

![bus queue ok](https://user-images.githubusercontent.com/25301769/39423826-774264e8-4c7c-11e8-8a50-b2b3a843e5e2.PNG)

These two scenarios are possible because `DispatcherRegistry` returns dispatchers in an unpredictable order. So, depending on the order of dispatchers, `TaskCreated` will be dispatched first either to the process manager or to the projection.

The second questionable thing is that a command router posts commands to the bus, before the execution of the handler (which routes commands), is finished.

dmytro-grankin:
The proposed solution is to add a FIFO queue for messages at the bus level. In this case, commands, events and rejections will be dispatched in the order, in which they were posted into a bus.

[...]

PR #731, opened 2018-06-29
Dispatch messages out of the `Bus` in FIFO order

vlad-lubenskyi:
Addresses https://github.com/SpineEventEngine/core-java/issues/696 and https://github.com/SpineEventEngine/core-java/issues/725.

The mentioned issues have the same root cause: in a single-threaded environment the `CommandBus` doesn't acknowledge the command until dispatching is finished; thus, routing a nested command from the command handler will block command handler until all nested commands and events are handled.

This PR addresses the problem by ensuring that `Bus` will always dispatch messages in FIFO order. Every message is positively acknowledged as soon as it was posted (that is enqueued for dispatching).


vlad-lubenskyi:
@armiol PTAL

server/src/main/java/io/spine/server/bus/Bus.java
```diff
@@ -207,6 +208,16 @@ public void close() throws Exception {
         return filterChain;
     }
 
+    /**
+     * Obtains the queue of the envelopes.
+     */
+    private DispatchingQueue<E> envelopeQueue() {
```
armiol: I would easily skip the `envelope` part from its name, as it's the only queue at the moment.

Also, please add some description of what's going on here. And give the `@see` reference to the `DispatchingQueue` class.

server/src/main/java/io/spine/server/bus/DispatchingQueue.java
```diff
@@ -0,0 +1,111 @@
...
+    private boolean dispatchingInProgress;
+
+    DispatchingQueue(DispatchAction<E> dispatchAction) {
+        this.dispatchAction = dispatchAction;
+    }
+
+    /**
+     * Enqueues the envelope for dispatching and dispatches it in its turn.
+     *
+     * <p>The given envelope is dispatched right away if and only if:
```
armiol:
```
<p>Once an envelope is being added to the queue, a number of actions may be performed.
<ul>
     <li>Queue is empty and nothing is being dispatched at the moment -> the envelope is dispatched right away;</li>
     <li>Queue is non-empty and nothing is being dispatched at the moment -> the envelope is enqueued as the first item + the last item of the queue starts to be dispatched.</li>
     <li>Another envelope is being dispatched at the moment -> the given envelope is enqueued as the first item and its dispatching is postponed.</li>
</ul>

<p>Once the envelope dispatching action is finished, the next available item is picked from the queue and its dispatching is started.

...     
```

server/src/test/proto/spine/test/command/commands.proto
```diff
@@ -43,8 +43,28 @@ message CmdStartProject {
...
+    ProjectId id = 1;
+
+    string name = 2;
+}
+
+message FCmdStartProject {
+
+    ProjectId id = 1;
+}
+
```
armiol: This line is redundant.

server/src/test/proto/spine/test/command/commands.proto
```diff
@@ -43,8 +43,28 @@ message CmdStartProject {
...
+// A command to create a project to test the command posting order.
+//
+message FCmdCreateProject {
+
+    ProjectId id = 1;
+
+    string name = 2;
+}
+
+message FCmdStartProject {
```
armiol: If `F` stands for `First` I would say so. And its `TestEnv` factory methods starting with `f` are also unreadable. Let's make their naming more user-friendly.

And the Message description is needed.

vlad-lubenskyi:
@armiol PTAL

server/src/main/java/io/spine/server/bus/DispatchingQueue.java
```diff
@@ -49,14 +49,18 @@
...
+     *     <li>if nothing is being dispatched and the queue is empty the envelope is dispatched
+     *         right away;
+     *     <li>if nothing is being dispatched and the queue is non-empty the envelope is enqueued
+     *         as the first item, and the queue dispatches the last item.
      * </ul>
      *
-     * <p>Otherwise, the envelope is only enqueued and will be dispatched according to its position
-     * in the queue.
+     * <p>Once the envelope dispatching action is finished, the queue picks the next available item
+     * and starts to dispatch it, until the queue is empty.
```
armiol: I think the comma is redundant here.

PR #735, opened 2018-07-03
Migrate `model` and `testutil-...` modules to JUnit 5

dmitrykuzmin:
This PR finishes the migration of the `core-java` to the new JUnit 5 test structure.

The following modules were migrated:
1. `model-assembler`
2. `model-verifier`
3. `testutil-core`
4. `testutil-server`

The only class that could not be migrated is [`ModelVerifierPluginShould`](https://github.com/SpineEventEngine/core-java/blob/master/model/model-verifier/src/test/java/io/spine/model/verify/ModelVerifierPluginShould.java), as it relies on the [`GradleProject`](https://github.com/SpineEventEngine/base/blob/master/tools/plugin-testlib/src/main/java/io/spine/tools/gradle/GradleProject.java) class from `base` which uses JUnit 4 API. After this problem is solved, the JUnit 4 API can be eliminated from this test.

In all other places of the `core-java`, the JUnit 5 is now used instead of JUnit 4, and the `junit-vintage` engine is now used only in the `model-verifier` module for the reasons stated above.

### DIFF
--- a/model/model-verifier/src/test/java/io/spine/model/verify/ModelVerifierTest.java
+++ b/model/model-verifier/src/test/java/io/spine/model/verify/ModelVerifierTest.java
@@ -23,7 +23,11 @@ package io.spine.model.verify;
 import com.google.common.base.Function;
 import com.google.common.io.Files;
 import io.spine.model.CommandHandlers;
-import io.spine.model.verify.given.ModelVerifierTestEnv;
+import io.spine.model.verify.ModelVerifier.GetDestinationDir;
+import io.spine.model.verify.given.ModelVerifierTestEnv.AnyCommandHandler;
+import io.spine.model.verify.given.ModelVerifierTestEnv.DuplicateAnyCommandHandler;
+import io.spine.model.verify.given.ModelVerifierTestEnv.Int32HandlerAggregate;
+import io.spine.model.verify.given.ModelVerifierTestEnv.Int64HandlerProcMan;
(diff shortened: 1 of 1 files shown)


dmitrykuzmin:
@armiol PTAL.

model/model-verifier/src/test/java/io/spine/model/verify/ModelVerifierTest.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Test
+    @DisplayName("verify model from classpath")
+    void verifyModel() {
+        ModelVerifier verifier = new ModelVerifier(project);
+
+        verify(project).getSubprojects();
+
+        String commandHandlerTypeName = ModelVerifierTestEnv.AnyCommandHandler.class.getName();
```
armiol: Please use static import for the nested classes.

testutil-server/src/test/java/io/spine/server/command/given/CommandTestTestEnv.java
```diff
@@ -0,0 +1,55 @@
...
+import io.spine.client.ActorRequestFactory;
+import io.spine.server.command.CommandTest;
+
+/**
+ * @author Alexander Yevsyukov
+ * @author Dmytro Kuzmin
+ */
+public class CommandTestTestEnv {
+
+    /** Prevents instantiation on this utility class. */
```
armiol: `of this`

PR #736, opened 2018-07-03
Update Base

dmdashenkov:
In this PR we update the version to `0.10.44-SNAPSHOT` along with the `base` module.


dmdashenkov:
@armiol, PTAL.

Issue #739, opened 2018-07-05
Model verifier can't load classes apart of its own dependencies and project's output


vlad-lubenskyi:
### Prerequisites

- The `spine-users` module declares an abstract process manager:

    ```java
    abstract public class RemoteIdentityProviderProcMan extends ProcessManager<...> {
        ...
    }
    ```
* The `my-application` application depends on `spine-users` and implements the PM:

    ```java
    class GoogleIdentityProviderProcMan extends RemoteIdentityProviderProcMan {
        ...
    }
    ```

### Issue

The `my-application:verifyModel` task fails with the following error:
```
13:09:38.983 [DEBUG] [io.spine.model.verify.ModelVerifier] Trying to load class 'com.teamdev.users.signin.GoogleIdentityProviderProcMan'
13:09:38.991 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] FAILURE: Build failed with an exception.
13:09:38.991 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] 
13:09:38.991 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] * What went wrong:
13:09:38.991 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] Execution failed for task ':users:verifyModel'.
13:09:38.991 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] > io/spine/users/signin/RemoteIdentityProviderProcMan
```

The possible reason for this could be that `ModelVerifier` simply [doesn't expect](https://github.com/SpineEventEngine/core-java/blob/master/model/model-verifier/src/main/java/io/spine/model/verify/ModelVerifier.java#L132) any classes apart of its own `implementation` dependencies and project's output directories.

PR #742, opened 2018-07-10
Use Gradle project's class loader in model verifier

vlad-lubenskyi:
Addresses https://github.com/SpineEventEngine/core-java/issues/739.

### Before

Right now, the `ModelVerifier` uses it's own class loader appended with projects' output directories.

### After
In this PR, I make it use the class loader of the current project's `buildscript`.


vlad-lubenskyi:
@armiol PTAL

Issue #747, opened 2018-07-11
Add `ZoneId` to `ActorContext`


alexander-yevsyukov:
An offset does not give enough information for time computation, because of daylight saving time transitions.

In addition to the offset, we need a zone ID for the case of possible time-related computations.

armiol:
Closing as resolved.

PR #749, opened 2018-07-12
Add `ZoneId` to `ActorRequest`

alexander-yevsyukov:
This adds support of `ZoneId` in `ActorRequest` (#747).

The zone id is necessary for time-related calculations, since only offset does not provide enough information when it comes to rules of daylight saving time switching.

This PR also switches to newer version of Spine Time.

Other notable changes:
  * `Change` types related to date/time were moved under Spine Time library.
  * `TestActorRequestFactory` moved under `testing.client` package.

The version advances to `0.10.47-SNAPSHOT`.


alexander-yevsyukov:
@mdrachuk, PTAL.

client/src/main/java/io/spine/client/ActorRequestFactory.java
```diff
@@ -57,35 +70,61 @@
...
+     * which corresponds to new location of the actor.
      *
-     * @param zoneOffset the offset of the time zone
-     * @return new factory at new time zone
+     * @param zoneOffset the offset of the new time zone
+     * @param zoneId     the ID of the new time zone
+     * @return new factory at the new time zone
      */
-    public ActorRequestFactory switchTimezone(ZoneOffset zoneOffset) {
+    public ActorRequestFactory switchTimezone(ZoneOffset zoneOffset, ZoneId zoneId) {
```
armiol: As discussed, we'll need to have an overloaded "shortcut" switching method, that takes `zoneId` only. The reason behind this decision is that just a `zoneId` is enough for the most cases. And those who need custom offsets, will use this one.

Issue #737, opened 2018-07-04
Test `rejectDuplicateHandlingMethods` of `ModelVerifierPluginTest` is failing other tests in the suite


dmitrykuzmin:
The test uses `GradleProject` class to build the project with duplicate handlers, which is supposed to fail on the model verification step.

The build fails as expected, and the test passes, but the process in which the Gradle build was executed remains open for some reason even after the test has been finished.

This leads to Gradle builds from other tests "seeing" classes of the failing project and thus failing too.

Prior to the JUnit 5 migration the bug did not occur because the bugged test was always run last in the suite.

Prior to the JUnit 5 migration the test had name `ModelVerifierPluginShould.halt_build_on_duplicate_command_handling_methods()`.

dmitrykuzmin:
Closed via https://github.com/SpineEventEngine/core-java/pull/751.

Issue #726, opened 2018-06-25
One `Model` per bounded context


dmdashenkov:
Currently, `Model` is a system-wide singleton. The problem with this is revealed when trying to add more complex validation, e.g. find the correspondence between commands and bounded contexts.

In order to make `Model` agiler, and also prohibit inadequate access to the `Model` instances, we should turn it into a per-BC concept.

dmdashenkov:
The suggested approach is to produce strict rules on what a bounded context is at compile time. 

This can be done by introducing [module annotation](https://docs.oracle.com/javase/9/docs/api/java/lang/annotation/ElementType.html#MODULE) `@BoundedContext` and consider this module to contain the definition of the bounded context.

PR #751, opened 2018-07-14
Clear global `Model` before applying `ModelVerifier`

dmitrykuzmin:
This PR fixes issue https://github.com/SpineEventEngine/core-java/issues/737.

Currently, we have `Model` as a system-wide singleton. Thus, if multiple Spine projects are built under the same JVM (for example via the [`GradleRunner`](https://docs.gradle.org/current/javadoc/org/gradle/testkit/runner/GradleRunner.html) in tests), the `Model` gets polluted by the classes from all the projects and searches for duplicates among them. This leads to unexpected `verifyModel` task failures.

With this PR, the `Model` will be cleared before every `ModelVerifier` application, so each project is  verified independently of others.

This PR should be counted as a temporary fix, because with the introduction of https://github.com/SpineEventEngine/core-java/issues/726 and one `Model` per `BoundedContext` the problem will go away automatically.

### DIFF
--- a/model/model-verifier/src/main/java/io/spine/model/verify/ModelVerifier.java
+++ b/model/model-verifier/src/main/java/io/spine/model/verify/ModelVerifier.java
@@ -76,6 +76,7 @@ final class ModelVerifier {
      */
     void verify(CommandHandlers spineModel) {
         Logger log = log();
+        model.clear();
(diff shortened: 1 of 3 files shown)


dmitrykuzmin:
@armiol PTAL.

PR #754, opened 2018-07-16
Disable `ClassCanBeStatic` in build

alexander-yevsyukov:
This PR:
 * disables the `ClassCanBeStatic` warning until `ErrorProne` can [support JUnit nested classes](https://github.com/google/error-prone/issues/956).
* Adds an ability to create `ActorRequestFactory` by an `ActorContext`.


alexander-yevsyukov:
@armiol, PTAL.

build.gradle
```diff
@@ -216,7 +216,12 @@ subprojects {
     tasks.withType(JavaCompile) {
         // Exclude generated sources from being analyzed by Error Prone.
         options.compilerArgs += [
-                '-XepExcludedPaths:.*/generated/.*'
+                // Turn Error Prone off for generated code.
+                '-XepExcludedPaths:.*/generated/.*',
```
armiol: We should also think of putting this configuration into `config` somehow.

client/src/main/java/io/spine/client/ActorRequestFactory.java
```diff
@@ -81,6 +81,18 @@ public static Builder newBuilder() {
         return new Builder();
     }
 
+    /**
+     * Creates an instance by the passed {@code ActorContext}.
+     */
+    public static ActorRequestFactory fromContext(ActorContext actorContext) {
+        Builder builder = newBuilder()
```
armiol: Let's check for `actorContext` not being `null` first.

build.gradle
alexander-yevsyukov: Yes, I'm thinking of this as I'm doing the same under the `base` right now. Don't have good ideas at the moment. Need to dive into Gradle docs once more.

Issue #683, opened 2018-03-12
Include entity version data to the `QueryResponse`


armiol:
**Preface**

`Aggregate`s and `ProcessManager`s are the versioned entities, which can handle commands. Each modification of their state leads to the version increment. 

At the same time the target version number can be used as a parameter for the `Command`, being a way to escape from updating the stale entities (see `CommandContext.target_version` attribute). I.e. the client may specify the expected version of the target entity, in which as he believes this entity resides. In case the actual entity version differs from the expected version set for the command, the command is rejected — as it clearly has been created with the stale entity state in mind.

**Problem**

Right now it is possible to specify the target version for the command via `CommandFactory.create(Message message, int targetVersion) : Command` API call.

However, at the moment it's not possible to tell the version of entities, which states are returned as a result of `Query` execution. So effectively client-side users aren't able to use the version targeting at all.

dmitrykuzmin:
Resolved via #758.

PR #758, opened 2018-07-18
Include entity version data to the `QueryResponse`

dmitrykuzmin:
This PR addresses #683.

`QueryResponse` now stores `EntityStateWithVersion` instances instead of just entity state:
```proto
message QueryResponse {
    ...
    repeated EntityStateWithVersion messages = 5;
}

message EntityStateWithVersion {
    google.protobuf.Any state = 1;
    core.Version version = 2;
}
```

To enable this functionality, the `MirrorProjection` and related events now also store the corresponding aggregate version.


dmitrykuzmin:
@armiol PTAL.

client/src/main/proto/spine/client/query.proto
```diff
@@ -87,9 +88,21 @@ message QueryResponse {
     // Reserved for more query response attributes, e.g. to describe paginated response etc.
     reserved 2 to 4;
 
-    //TODO:2017-02-20:alexander.yevsyukov: Use EntityRecord instead of Any.
```
armiol: Please move the `todo` item into the `EntityStateWithVersion`.

dmitrykuzmin:
@armiol PTAL again. The `web` will need a couple of minor fixes, but is compatible with this change overall.

armiol:
@dmitrykuzmin I have recalled why we had left this change out.

dmitrykuzmin:
@armiol PTAL again.

PR #764, opened 2018-07-18
Fix script updating `core` dependency in `jdbc-storage`

dmitrykuzmin:
After the `jdbc-storage` [migrated](https://github.com/SpineEventEngine/jdbc-storage/pull/64) to the new Gradle project structure, the format of its `ext.gradle` changed.

Thus we need to update the script updating `core` dependency for the `jdbc-storage` repository.


dmitrykuzmin:
@armiol PTAL.

Issue #711, opened 2018-05-27
Move tests of `EntityBuilder` to proper test suite(s)


alexander-yevsyukov:
There are methods of `AggregateShould`  like: 
 * `throw_InvalidEntityStateException_if_state_is_invalid()`
 * `update_valid_entity_state()`

They test the `EntityBuilder`, and not the `Aggregate`. 
Similar methods are in the `AggregatePartShould` test suite.

Because of this they need to be moved into corresponding test suite.

dmitrykuzmin:
Resolved via https://github.com/SpineEventEngine/core-java/pull/767.

PR #767, opened 2018-07-19
Move tests of `EntityBuilder` to proper test suite

dmitrykuzmin:
This PR fixes issue https://github.com/SpineEventEngine/core-java/issues/711.

The `AggregateTest` and `AggregatePartTest` classes had methods that in fact checked `EntityBuilder` functionality.

Now these tests are moved to the `EntityBuilderTest` and merged (because there is no substantial difference whether to use `Aggregate` or `AggregatePart` for test).


dmitrykuzmin:
@armiol PTAL.

PR #768, opened 2018-07-19
 Replacement of optionals for `testutil` modules

illia-shepilov:
This PR provides the following changes for the `testutil-core`, `testutil-server` and `testutil-client` modules:

1. The `final` keyword on local variables was removed;
2. Guava's `Optional` was replaced with JDK 8 `Optional`;
3. The version was advanced to `0.10.54-SNAPSHOT`.


illia-shepilov:
@armiol PTAL.

PR #770, opened 2018-07-20
Replace anonymous classes with lambdas

illia-shepilov:
This PR provides the following changes:

1. Anonymous classes were replaced with lambdas;
2. Modifiers were resorted.


illia-shepilov:
@armiol PTAL.

PR #772, opened 2018-07-20
Commander methods

alexander-yevsyukov:
This PR introduces the `@Command` annotation for marking methods as those that produce command messages in response to incoming messages. It also introduces `CommandSubstMethod` and `Commander` class, but these classes will be finished in the following PRs devoted to command generation.

The main focus of this PR is to improve Model-related classes in terms of:
  * improving naming;
  * avoiding unnecessary redirections in `Method` classes;
  * extracting common abstract bases, that will be used by future classes related to command-generation;
  * migrating to Java `Function` and `Predicate`.


server/src/main/java/io/spine/server/model/ClassMap.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Returns a unique identifying name of the given class.
+     *
+     * <p>The returned value is guaranteed to be unique per class and non-null.
+     *
+     * @param cls the {@link Class} to identity
+     * @return a non-null class name
+     */
+    private static String nameOf(Class<?> cls) {
+        return cls.getName();
```
alexander-yevsyukov: @dmdashenkov, did you mean to do something more eventually in this method? The Javadoc hits on a bigger future for this method.

alexander-yevsyukov:
@dmdashenkov, @armiol, PTAL.

The topic to discuss: move Model-related classes under `model` sub-packages (in the following PR).

server/src/main/java/io/spine/server/model/ClassMap.java
dmdashenkov: The method is used to extract the `Class` key in a unified way for all the occurrences. We need this to avoid programming errors when using `Class.getName()` VS `Class.getSimpleName()` VS `Class.getCanonicalName()`.

server/src/main/java/io/spine/server/command/Command.java
```diff
@@ -0,0 +1,88 @@
...
+ *     <li>be annotated with {@link Command @Command};
+ *     <li>return a command message derived from {@link com.google.protobuf.Message Message},
+ *     if there is only one command to be generated;
+ *     <strong>or</strong> an {@link java.lang.Iterable Iterable} of command messages for two or
+ *     more commands;
+ *     <li>accept one of the following as the first parameter:
+ *     <ul>
+ *        <li>a {@linkplain com.google.protobuf.Message command message}
+ *        <li>an {@linkplain com.google.protobuf.Message event message}
+ *        <li>a {@link com.google.protobuf.Message rejection message}
```
dmdashenkov: `linkplain` as well.

server/src/main/java/io/spine/server/command/Command.java
```diff
@@ -0,0 +1,88 @@
...
+ * <p>If a commanding method accepts a command type which is also handled by the application
+ * by another method, a run-time error will occur. This also means that an application cannot have
+ * two commanding methods that accept the same command type.
+ *
+ * @author Alexander Yevsyukov
+ * @see io.spine.server.command.Assign Handling Commands
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Command {
```
dmdashenkov: This name is confusing. I bet there will be a file in the `server` module which uses both `io.spine.server.command.Command` and `io.spine.core.Command`. I do understand that one of these is a verb and another is a noun. Though, my understanding does not make the situation better.

Can we rename this to something like `@Transform`?

server/src/main/java/io/spine/server/command/CommandAcceptingMethod.java
```diff
@@ -0,0 +1,104 @@
...
+import java.lang.reflect.Method;
+import java.util.Optional;
+
+import static com.google.common.base.Throwables.getRootCause;
+
+/**
+ * @author Alexander Yevsyukov
+ */
+@Internal
+public abstract class CommandAcceptingMethod
```
dmdashenkov: Please document.

server/src/main/java/io/spine/server/command/CommandSubstMethod.java
```diff
@@ -0,0 +1,107 @@
...
+import java.lang.reflect.Method;
+import java.util.function.Predicate;
+
+/**
+ * A method that produces one or more command messages in response to an incoming command.
+ *
+ * @author Alexander Yevsyukov
+ */
+@Internal
+public final class CommandSubstMethod
```
dmdashenkov: Please use the whole word. `Command Substance Method` is what comes into my head first.

server/src/main/java/io/spine/server/command/dispatch/DispatchResult.java
```diff
@@ -40,30 +44,37 @@
...
      */
-    public List<Event> asEvents(Any producerId, Version version) {
-        return HandlerMethod.toEvents(producerId, version, messages, origin);
+    public List<Event> asEvents(Any producerId, @Nullable Version version) {
+        checkNotNull(producerId);
+        EventFactory eventFactory = EventFactory.on(origin, producerId);
+        List<Event> result =
+                messages.stream()
+                        .map(eventMessage -> eventFactory.createEvent(eventMessage, version))
+                        .collect(Collectors.toList());
```
dmdashenkov: Please add static import for `toList()`.

server/src/main/java/io/spine/server/event/EventFactory.java
```diff
@@ -119,45 +117,33 @@ private static void validate(Message messageOrAny) throws ValidationException {
...
-    }
-
-    private static EventContext toEventContext(IntegrationEventContext value) {
-        Timestamp timestamp = value.getTimestamp();
-        Any producerId = toAny(value.getBoundedContextName());
-        return EventContext.newBuilder()
-                           .setTimestamp(timestamp)
-                           .setProducerId(producerId)
-                           .build();
+    public static Event toEvent(IntegrationEvent ie) {
```
dmdashenkov: Please use the whole words.

server/src/main/java/io/spine/server/event/IntegrationEventConverter.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.io.Serializable;
+
+import static io.spine.protobuf.TypeConverter.toAny;
+import static io.spine.util.Exceptions.unsupported;
+
+/**
+ * Converts {@code IntegrationEvent} into {@code Event}.
+ *
+ * <p>Does not support backward conversion.
```
dmdashenkov: How come it's a `Converter` then?

server/src/main/java/io/spine/server/event/IntegrationEventConverter.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @author Alexander Yevsyukov
+ */
+class IntegrationEventConverter extends Converter<IntegrationEvent, Event> implements Serializable {
```
dmdashenkov: Please explain why this is `Serializable` in the doc.

server/src/main/java/io/spine/server/model/MethodPredicate.java
```diff
@@ -35,11 +35,11 @@
...
 
     @Override
-    public boolean apply(@Nullable Method method) {
+    public boolean test(@Nullable Method method) {
         checkNotNull(method);
         boolean result = verifyAnnotation(method)
-                         && verifyParams(method)
-                         && verifyReturnType(method);
+                && verifyParams(method)
+                && verifyReturnType(method);
```
dmdashenkov: Seems to be an accidental change. Please undo.

server/src/main/java/io/spine/server/model/ClassMap.java
```diff
@@ -0,0 +1,110 @@
...
+     * there are already handlers registered in the model.
+     *
+     * @param rawClass raw class for which to obtain a model class
+     * @param supplier a supplier of a model class
+     * @param commandHandler if {@code true} the passed raw class handles commands
+     * @return the instance of the model class corresponding to the raw class
+     * @throws DuplicateCommandHandlerError
+     *         if the passed class handles commands that are already handled
+     */
+    ModelClass<?> get(Class<?> rawClass, Supplier<ModelClass<?>> supplier, boolean commandHandler) {
```
dmdashenkov: Please have two separate methods instead of a `boolean` flag parameter.

server/src/test/java/io/spine/server/aggregate/AggregateTest.java
```diff
@@ -487,9 +491,23 @@ void beingCommitedAfterDispatch() {
...
             List<Event> events = aggregate().commitEvents();
 
-            assertContains(getEventClasses(events),
-                           AggProjectCreated.class, AggTaskAdded.class, AggProjectStarted.class);
+            assertEventClasses(getEventClasses(events),
+                               AggProjectCreated.class, AggTaskAdded.class, AggProjectStarted.class);
         }
+
+        private Collection<EventClass> getEventClasses(Collection<Event> events) {
+            return transform(events, new Function<Event, EventClass>() {
```
dmdashenkov: Please use `Stream` API for such transformations.

server/src/main/java/io/spine/server/command/Command.java
alexander-yevsyukov: Yes, it's a verb. There will be other methods that issue command messages in response to events or rejections. It is better to have one annotation for all the cases.

The client code is not going to see the `core.Command` at all.

server/src/main/java/io/spine/server/event/IntegrationEventConverter.java
alexander-yevsyukov: Guava's `Converter` does allow to throw in `doBackward()`.

server/src/main/java/io/spine/server/event/IntegrationEventConverter.java
alexander-yevsyukov: We make converters `Serializable` for the possible use in Apache Beam transformations.

server/src/main/java/io/spine/server/event/IntegrationEventConverter.java
alexander-yevsyukov: See the issue for making it a read `Converter` (#777).

server/src/main/java/io/spine/server/model/MethodPredicate.java
alexander-yevsyukov: It's the current IDEA layout.

server/src/main/java/io/spine/server/model/ClassMap.java
alexander-yevsyukov: It turns out we can do without the parameter at all.

server/src/main/java/io/spine/server/command/CommandSubstMethod.java
alexander-yevsyukov: `subst` is an accepted shortening. This API isn't end user, and I don't expect much of a confusion anyway. I'm going to visit this issue once again when adding more command generation methods.

server/src/test/java/io/spine/server/aggregate/AggregateTest.java
```diff
@@ -496,16 +494,11 @@ void beingCommitedAfterDispatch() {
...
-                    if (record == null) {
-                        return null;
-                    }
-                    Message eventMessage = unpack(record.getMessage());
-                    return EventClass.of(eventMessage);
-                }
-            });
+            List<EventClass> result =
+                    events.stream()
+                          .map(EventClass::of)
```
dmdashenkov: This seems to be `EventClass.of(Event)` for any invocation.

server/src/test/java/io/spine/server/aggregate/AggregateTest.java
```diff
@@ -496,16 +494,11 @@ void beingCommitedAfterDispatch() {
...
+                          .collect(Collectors.toList());
```
dmdashenkov: Please add a static import here as well.

PR #781, opened 2018-07-23
Obtain default entity state via entity model class

alexander-yevsyukov:
This PR eliminates `DefaultStateRegistry` in favor of storing a default entity state in `EntityClass`.

Other notable changes:
 * `EntityQueryMatcher` now uses packed identifiers for filtering.
 * Partial migration to Java `Predicate` and `Function`
 * Fixed return values of model classes returned by `AggregatePart`.
 * `RecordStorage` now allows querying without passing default `FieldMask`.


alexander-yevsyukov:
@armiol, PTAL.

server/src/test/java/io/spine/server/procman/ProcessManagerTest.java
```diff
@@ -191,12 +191,6 @@ void setUp() {
         return events;
     }
 
-    @Test
```
alexander-yevsyukov: This common `Entity` behavior test moved under `EntityTest`.

server/src/test/java/io/spine/server/stand/StandStorageTest.java
```diff
@@ -164,116 +168,63 @@ private void checkByTypeRead(FieldMask fieldMask) {
...
+            EntityRecord record = newRecord(task);
+            storage.write(id, record);
         }
 
-        return ids;
-    }
+        private List<AggregateStateId> fill(StandStorage storage,
+                                            int count,
+                                            Supplier<AggregateStateId<ProjectId>> idSupplier) {
+            List<AggregateStateId> ids = new LinkedList<>();
```
armiol: Strictly saying, this should be `new ArrayList<>(count)`;

Issue #782, opened 2018-07-23
Hide `updateState()` from end user API


alexander-yevsyukov:
`AbstractVersionableEntity` exposes `updateState()` as `public`. Even though, it has the method marked `@Internal`, there is a great risk of misusing the API via end users.

The reason for exposure is documented in `AbstractVersionableEntityTest.haveUpdateStatePackagePrivate()` as

```java
@Disabled // The `updateState` method was made public to be accessible from `testutil-server`
```

Current calls of `updateState()` from the `testutil` base are limited to a single usage in `EntityBuilder`:

```java
    protected void setState(E result, S state, Version version) {
        result.updateState(state, version);
    }
```
At the same time this particular method implementation seems to be never used, but overridden by each of its descendants, such as `AggregateBuilder`. In _their_ implementation of `setState()` no calls to `updateState()` are made.

To do:

1. Make `EntityBuilder.setState()` no-operational with the comment. Or try to make it `abstract` along with `EntityBuilder` itself — this will take some effort, but seems doable.
2. Rollback `AbstractVersionableEntity.updateState()` back to package-private.
3. Re-enable `AbstractVersionableEntityTest.haveUpdateStatePackagePrivate()` .

alexander-yevsyukov:
@armiol, FYI.

alexander-yevsyukov:
@vlad-lubenskyi, can this issue be closed, assuming #784 is merged?

PR #784, opened 2018-07-24
Hide 'updateState' method of AbstractVersionableEntity

vlad-lubenskyi:
Addresses https://github.com/SpineEventEngine/core-java/issues/782

This PR makes `updateState()` package-private and deals with consequences:

- `EntityBuilder` becomes `abstract`;
- `Given.entityOfClass(...)` method is removed;
- `GivenTest.createEntityBuilder()` test case is removed;
- `RecordBasedRepositoryTest` delegates entity creation to the subclasses (which use `EntityBuilder` implementation);
- `ColumnTest.invokeGetter` now works with `Aggregate` instead of an abstract entity;
- `EntityTest.ProvidesSuchEqualsThat.equalsToSame` now works with `Aggregate` instead of an abstract entity.

Spine version is advanced to 0.10.60-SNAPSHOT


vlad-lubenskyi:
@armiol PTAL

server/src/test/java/io/spine/server/entity/RecordBasedRepositoryTest.java
```diff
@@ -301,6 +295,8 @@ void noEntitiesIfEmpty() {
         }
     }
 
+    protected abstract E entity(I id);
```
armiol: Please move to the `protected` methods section (see above). Also, I don't understand why it is not an overload of `createEntity() : E` method, which already exists above. `createEntity(id) : E` would look much more consistent.

server/src/test/java/io/spine/server/entity/given/EntityTestEnv.java
```diff
@@ -100,4 +104,35 @@ public EntityWithMessageId() {
...
     }
+
+
+    public static class TestAggregate extends Aggregate<String, StringValue, StringValueVBuilder> {
+
+        protected TestAggregate(String id) {
+            super(id);
+        }
+
+        public static TestAggregate withStateOf(TestAggregate entity) {
```
armiol: It's not only a "state", it's a copy. Therefore I'd call it `copyOf(...)`.

server/src/test/java/io/spine/server/entity/storage/given/EntityRecordWithColumnsTestEnv.java
```diff
@@ -47,4 +49,12 @@ public EntityWithoutColumns(String id) {
             super(id);
         }
     }
+
+    public static class TestEntityBuilder extends EntityBuilder<TestEntity, String, Project> {
```
armiol: There are two similar classes in this changeset. Another one is in `TestEntity`. Not sure we need both.

server/src/test/java/io/spine/server/entity/TestEntity.java
```diff
@@ -34,32 +35,40 @@
...
     static TestEntity withStateOf(TestEntity entity) {
-        TestEntity result = Given.entityOfClass(TestEntity.class)
-                                 .withId(entity.getId())
-                                 .withState(entity.getState())
-                                 .withVersion(entity.getVersion()
+        TestEntity result = new TestEntityBuilder().setResultClass(TestEntity.class)
+                                                   .withId(entity.getId())
+                                                   .withState(entity.getState())
+                                                   .withVersion(entity.getVersion()
                                                     .getNumber())
```
armiol: I think the formatting is broken for this line.

server/src/test/java/io/spine/server/entity/given/EntityTestEnv.java
```diff
@@ -100,4 +104,35 @@ public EntityWithMessageId() {
             super(Sample.messageOfType(ProjectId.class));
         }
     }
+
```
armiol: This line is redundant.

armiol:
@vlad-lubenskyi also before merging, please be careful and make sure that the bumped version in this PR is indeed an increment in relation to `master`. There is a concurrent PR which also has `60`.

vlad-lubenskyi:
@armiol PTAL

server/src/test/java/io/spine/server/entity/given/EntityTestEnv.java
```diff
@@ -100,4 +104,34 @@ public EntityWithMessageId() {
...
+    public static class TestAggregate extends Aggregate<String, StringValue, StringValueVBuilder> {
```
armiol: Please create an issue saying that currently we use aggregates in tests for `Entity`, but it seems odd. There should be a cleaner way of doing it. And put the reference to the issue as a todo item here.

vlad-lubenskyi:
@armiol PTAL

Issue #773, opened 2018-07-21
A `ProcessManager` can ignore a command


alexander-yevsyukov:
The body of `CommandHandlerMethod.invoke()` currently can accept an empty list of event messages generated as the result of the method execution.

Attempt to prevent this via
```java
checkState(!events.isEmpty();
```
Results in several broken tests in `ProcessManagerRepositoryTest`. It turns out that test environment classes if this test suite do violate the contract of command handling methods. See `ProcessManagerRepositoryTestEnv.TestProcessManager`.

The problem is that even if handling method of a `ProcessManager` returns `Empty`, it would result in no events generated. Chances are that the same would be possible for `Aggregate`s.

The following needs to be done:
  1. Test environment classes for `ProcessManagerRepositoryTest` need to be adjusted to stop violating the contract of command handling methods.
  2. The `CommandHandlerMethod.invoke()` should have the check for non-empty list of events. 
      (I'm adding TODO with the reference to this issue in this method.)
  3. Inform current users of the framework on the change of the behaviour of `ProcessManager`s.

dmitrykuzmin:
Resolved via https://github.com/SpineEventEngine/core-java/pull/785.

PR #785, opened 2018-07-24
Disallow returning empty event list from command handlers

dmitrykuzmin:
This PR fixes issue https://github.com/SpineEventEngine/core-java/issues/773.

Previously, it was possible to violate the `@Assign` contract by returning empty event list from the command handler method. Thus, the command was basically ignored.

This PR adds the check which prevents contract violation and fixes all `...TestEnv` classes which did that and thus became invalid.

Spine version advances to `0.10.58-SNAPSHOT`.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -25,7 +25,7 @@
  *  as we want to manage the versions in a single source.
  */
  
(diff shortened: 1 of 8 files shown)


dmitrykuzmin:
@armiol PTAL.

Issue #783, opened 2018-07-24
Column functionality demands Entity class to be public


dmitrykuzmin:
Currently, the `ColumnValueExtractor` class does not extract column values for entity classes which are not public.

The following method

```
private boolean canExtractColumns(Class<? extends Entity> entityClass) {
    checkNotNull(entityClass);

    int modifiers = entityClass.getModifiers();
    if (!Modifier.isPublic(modifiers)) {
        logNonPublicClass(entityClass);
        return false;
    }
    return true;
}
```

denies the extraction of columns for all non-public entities and logs a warning.

We should remove this check as entities should be able to have any access modifier, and also write tests for the implemented change.

dmitrykuzmin:
Resolved via https://github.com/SpineEventEngine/core-java/pull/787.

PR #787, opened 2018-07-25
Process columns of non-`public` entities

dmitrykuzmin:
This PR fixes issue https://github.com/SpineEventEngine/core-java/issues/783.

Previously, if the entity class was non-`public`, its columns would not be extracted and the warning would be printed to the log.

This PR removes the check and ensures that all entities are processed by the `ColumnValueExtractor` regardless of their access level.

The Spine version advances to `0.10.61-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

PR #789, opened 2018-07-25
Allow `ProcessManager` classes return `Empty` in command handlers

dmitrykuzmin:
In https://github.com/SpineEventEngine/core-java/pull/785 the command handler methods producing empty event list were forbidden. An attempt to return `Empty` or empty `List<Event>` from command handlers now results in the `IllegalStateException`.

This PR adds the exception to this rule: `ProcessManager` instances are now allowed to return `Empty` from the command handler methods, as their state may or may not be modified after handling a command. 

Returning empty event `List` from command handler methods is still forbidden though, as the `Empty` "event" should be used for this functionality (it's then internally transformed into the empty event `List`).

Spine version advances to `0.10.62-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

alexander-yevsyukov:
@armiol, let's discuss the principle of handling commands _before_ merging this in.

PR #791, opened 2018-07-25
Hide `Model` from entity classes

alexander-yevsyukov:
This PR hides `Model` from entity classes. Model class do not interact with the model indirectly via their abstract base (`ModelClass`).

The `Model` got a new instance lookup method `getInstance(Class)` which still returns the singleton.
Elimination of `Model.getInstance()` will be possible after implementing support of bounded context name lookup in `@BoundedContext` annotations in `package-info.java`.

Also this PR migrates the code of repositories to Java `Function` and `Predicate`.


alexander-yevsyukov:
@armiol, PTAL.

model/model-verifier/src/main/java/io/spine/model/verify/ModelVerifier.java
```diff
@@ -98,19 +101,19 @@ void verify(CommandHandlers spineModel) {
             "CheckReturnValue" /* Returned values for asXxxClass() are ignored because we use
                                   these methods only for verification of the classes. */
     })
-    private void verifyClass(Class<?> cls) {
+    private static void verifyClass(Class<?> cls) {
         Logger log = log();
         if (Aggregate.class.isAssignableFrom(cls)) {
             Class<? extends Aggregate> aggregateClass = (Class<? extends Aggregate>) cls;
-            model.asAggregateClass(aggregateClass);
+            asAggregateClass(aggregateClass);
```
armiol: Instead of `static` import with this method name I'd rather have `AggregateClass.of(aggregateClass)` and use no static import.

armiol:
@alexander-yevsyukov I also wonder about `CommandSubstMethod` naming, which seems weird to me. `CommandSubstituteMethod` isn't much longer, but more comprehensible.

alexander-yevsyukov:
@armiol, 
> ... an issue found by Codacy to address 
That's going to be solved when commanding methods will be implemented in full.

PR #796, opened 2018-07-26
Rework dispatch results

alexander-yevsyukov:
This PR improves the way method dispatch results are handled.

Notable changes are:
 * New class `MethodResult` is introduced. This class and its descendants are based on the ideas from the `command.dispatch` package (which is no longer with us).
 * The `HandlerMethod` interface got first generic parameter describing a type of target objects.
 * New interfaces `EventProducer`, `EventReactor`, `RejectionReactor` are introduced and used on existing message handling classes.


server/src/main/java/io/spine/server/aggregate/model/EventApplier.java
```diff
@@ -151,4 +159,5 @@ protected boolean verifyReturnType(Method method) {
             return isVoid;
         }
     }
+
```
mdrachuk: Get rid of this empty line.

server/src/main/java/io/spine/server/command/model/CommandHandlerMethod.java
```diff
@@ -161,4 +135,42 @@ protected boolean verifyReturnType(Method method) {
...
+     * The result of a command handler method execution.
+     */
+    public static final class Result extends EventsResult {
+
+        private Result(EventProducer producer, Object rawMethodResult) {
+            super(producer, rawMethodResult);
+            List<Message> eventMessages = toMessages(rawMethodResult);
+            List<Message> withoutEmpty = filterEmpty(eventMessages);
+            checkResultNonEmpty(withoutEmpty, rawMethodResult, producer);
+            setMessages(withoutEmpty);
```
mdrachuk: `withoutEmtpy` is confusing in 2 lines above. They read like "check result non-empty in without empty" and "set messages to without empty".
Can we have it `filteredMessages`, `filtered` or something else? 
It would read like
```java
checkResultNonEmpty(filteredMessages, rawMethodResult, producer);
setMessages(filteredMessages);
```

server/src/main/java/io/spine/server/rejection/RejectionReactor.java
```diff
@@ -0,0 +1,31 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.rejection;
+
+import io.spine.server.model.EventProducer;
+
+/**
+ * Objects that react on rejections may produce events.
```
mdrachuk: This line sounds more like a general statement instead of describing the interface.

server/src/main/java/io/spine/server/model/MethodResult.java
```diff
@@ -0,0 +1,141 @@
...
+public abstract class MethodResult<V extends Message> {
+
+    private final @Nullable Object rawMethodOutput;
+    private @MonotonicNonNull ImmutableList<V> messages;
+
+    protected MethodResult(@Nullable Object output) {
+        rawMethodOutput = output;
+    }
+
+    protected final void setMessages(List<V> messages) {
```
mdrachuk: `setMessages` seems to be intended to be used only during object construction. Can we mention that in JavaDoc or even throw an exception in case of re-assign?

server/src/main/java/io/spine/server/rejection/model/RejectionSubscriberMethod.java
```diff
@@ -57,20 +59,29 @@
...
+    MethodResult<Empty> invoke(Object target, Message rejectionMessage, RejectionContext context) {
+        ensureExternalMatch(context.getExternal());
+        doInvoke(target, rejectionMessage, context);
+        return MethodResult.empty();
+    }
 
-        Object result = doInvoke(target, rejectionMessage, context);
-        return result;
+    /**
+     * @implNote This method is never called because
```
mdrachuk: Seems like this tag fails our CI builds.

server/src/main/java/io/spine/server/rejection/RejectionReactor.java
alexander-yevsyukov: What would you need as a description of this interface? Since this interface is just an alias for `EventProducer` it serves mostly as a marker interface.

This interfaces say: “There are event reactors...”

alexander-yevsyukov:
@mdrachuk, PTAL.

server/src/main/java/io/spine/server/rejection/model/RejectionSubscriberMethod.java
alexander-yevsyukov: I initially removed the tags in the source, but now enabled the tag options for `javadoc` in `build.gradle`. So, it should work now. It's better with new tags.

PR #801, opened 2018-08-01
Model per BoundedContext

alexander-yevsyukov:
This PR makes `Model` being created per a name of a `BoundedContext` specified in package annotations of corresponding model classes.

System entities belong to the `_System` Model, which is shared for all system entities.

This PR also introduces `server.security` package with initial ability to limit classes that can call a method of interest. See `InvocationGuard` for details.


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/io/spine/server/annotation/BoundedContext.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author Alexander Yevsyukov
+ */
+@Target(ElementType.PACKAGE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface BoundedContext {
+
+    /**
+     * The name of the Bounded Context to which the package belongs.
```
armiol: From this sentence I get the impression that several different  packages may belong to a single Bounded Context (e.g. `com.foo` and `bar.bla` — that have no common "root" part).

If so, we should test that use case.

If not, we should rewrite this sentence.

server/src/main/java/io/spine/server/model/DuplicateCommandHandlerError.java
```diff
@@ -51,22 +55,67 @@ private static String fmt(CommandHandlingClass duplicatingClass,
...
+                                  .sum();
+        checkState(totalDuplicatedCommands >= 1);
+        builder.append(
+                totalDuplicatedCommands == 1
+                ? "method for the command which is "
+                : "methods for commands that are "
+        );
+        builder.append("already handled by ");
+
+        // How many already handling classes do we have?
```
armiol: It seems it better reads as `How many handling classes do we have already?` Not sure I get your original idea though.

server/src/main/java/io/spine/server/model/DuplicateCommandHandlerError.java
```diff
@@ -51,22 +55,67 @@ private static String fmt(CommandHandlingClass duplicatingClass,
...
             }
-            builder.append(registeredHandlers.get(commandClasses));
+            CommandHandlingClass handlingClass = registeredHandlers.get(commandClasses);
+            builder.append(backTick(handlingClass));
+            builder.append('.');
         }
         return builder.toString();
     }
+
+    private static String backTick(Object obj) {
```
armiol: I have looked through a [couple](https://en.wiktionary.org/wiki/backtick) of [different](https://www.techopedia.com/definition/17836/backtick) [sources](https://en.wikipedia.org/wiki/Grave_accent), and it seems that there is a `backtick`, but there isn't a `back tick`. 

So this method should probably be `private static String backtick(Object obj)`.

server/src/main/java/io/spine/system/server/package-info.java
```diff
@@ -0,0 +1,42 @@
...
+ * <p>{@link io.spine.server.BoundedContext BoundedContext}s serving entities
+ * of this package will be created with
+ * {@linkplain io.spine.core.BoundedContextNames#system(io.spine.core.BoundedContextName)
+ * synthetic names} created after names of parent {@code BoundedContext}s.
+ */
+@Internal
+@BoundedContext("_System")
+@CheckReturnValue
+@ParametersAreNonnullByDefault
+package io.spine.system.server;
```
armiol: I can feel Spring in the air :)

server/src/main/java/io/spine/server/annotation/BoundedContext.java
alexander-yevsyukov: As discussed, I've added the test for this case, and placed additional comments to Javadoc of `@BoundedContext`.

PR #803, opened 2018-08-02
Interfaces for command handler and event subscriber

alexander-yevsyukov:
This PR:
 1.  introduces interfaces `CommandHandler` and `EventSubscriber`;
 2. renames classes that had these names before to have `Abstract` prefix;
 3. moves the `EventProducer` interface from `server.model` to `server` package.


alexander-yevsyukov:
@dmdashenkov, PTAL. 
@armiol, FYI.

server/src/main/java/io/spine/server/event/AbstractEventSubscriber.java
```diff
@@ -0,0 +1,118 @@
...
+ * The abstract base for objects that can be subscribed to receive events from {@link EventBus}.
+ *
+ * <p>Objects may also receive events via {@link EventDispatcher}s that can be
+ * registered with {@code EventBus}.
+ *
+ * @author Alexander Yevsyukov
+ * @author Alex Tymchenko
+ * @see EventBus#register(MessageDispatcher)
+ */
+public abstract class AbstractEventSubscriber implements EventDispatcher<String>, EventSubscriber {
```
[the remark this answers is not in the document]
alexander-yevsyukov: We cannot enforce making descendants stateless. We're not composing a will here :)

On a more serious note, this naming pattern is a set practice. People will understand what's this thing is about. Have a look at classes that start from `Abstract` in libraries and tools we use.

Another reason for such naming is to encourage people using `Projection` for subscribing, and `Aggregate` for command handling.

alexander-yevsyukov:
@dmdashenkov, PTAL.

server/src/main/java/io/spine/server/model/HandlerMethodFailedException.java
```diff
@@ -59,6 +69,40 @@ public HandlerMethodFailedException(Object  target,
...
+        Throwable rootCause = getRootCause(exception);
+        if (rootCause instanceof ThrowableMessage) {
+            ThrowableMessage thrownMessage = (ThrowableMessage) rootCause;
+            Any producerId = idOf(target);
+            thrownMessage.initProducer(producerId);
+        }
+    }
+
+    /**
+     * Obtains an identity of an object which thrown {@link ThrowableMessage}.
```
dmdashenkov: Please check the grammar around `thrown`.

server/src/main/java/io/spine/server/model/HandlerMethodFailedException.java
```diff
@@ -59,6 +69,40 @@ public HandlerMethodFailedException(Object  target,
...
+        if (target instanceof Entity) {
+            Object entityId = ((Entity) target).getId();
+            return Identifier.pack(entityId);
+        }
+
+        if (target instanceof EventProducer) {
+            return ((EventProducer) target).getProducerId();
+        }
+
+        return TypeConverter.toAny(target.toString());
```
dmdashenkov: Let's use `Identifiers.pack(...)`. There are no differences between the results of those two, but it feels better to use `Identifiers` for IDs.

Issue #792, opened 2018-07-25
`RecordBasedRepository.findRecords()` is never used


alexander-yevsyukov:
This method is a part of public API. However, it is only used in Datastore- and JDBC-specific implementations.

We need to test it in`core-java` as well.

alexander-yevsyukov:
@armiol, FYI.

armiol:
Closing as resolved.

PR #814, opened 2018-08-10
Remove the unused `RecordBasedRepository.findRecords()` method

dmitrykuzmin:
This PR resolves issue https://github.com/SpineEventEngine/core-java/issues/792.

The method `RecordBasedRespository.findRecords()`, as it turns out, is used only in outdated Spine repositories (more specifically, [deployment](https://github.com/SpineEventEngine/deployment)). 

So it was deleted from the class.

Also there is a minor test refactoring as a side effect of initially writing test for the deleted method.


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -317,6 +317,7 @@ protected E findOrCreate(I id) {
      * @param fieldMask the mask to apply for returned records
      * @return an iterator over the matching records
      */
+    @SuppressWarnings("WeakerAccess") // Used in the code outside of the repository.
```
armiol: Please clarify the usages.

dmitrykuzmin:
@armiol PTAL again. I deleted the method.

armiol:
@dmitrykuzmin @alexander-yevsyukov as discussed previously, this method was only used in an outdated version of `deployment` module. For now we delete this method from the public API. And we'll get back to `development` implementation in its good time.

@dmitrykuzmin LGTM.

Issue #813, opened 2018-08-10
Impossible to reach `ActorContext` from an enrichment function


vlad-lubenskyi:
To enrich the events in `BoundedContextOne`, I would like to query `BoundedContextTwo` via `QueryService` from my enrichment function.

However, building queries requires an `ActorContext` which is not accessible from the enrichment function. And I'm not quite sure why it is so (snippet from [`FieldEnrichment`](https://github.com/SpineEventEngine/core-java/blob/master/server/src/main/java/io/spine/server/outbus/enrich/FieldEnrichment.java#L87)):
```java
@Override
public T apply(S message, C context) {
    ensureActive();
    T result = function.apply(message); // `context` parameter is simply ignored.
    return result;
}
```

My enrichment is defined as:

```proto
message UserDetailsEnrichment {
    option (enrichment_for) = "my.company.order.OrderPlaced"  
    my.company.users.UserDetails = 1 [(by) = "user_id"];
}
```

And I would like my Java code to look like this:
```java
EventEnricher createEnricher() {
    EventEnricher enricher =
        EventEnricher.newBuilder()
                     .add(UseId.class, UserDetails.class, userIdToDetails())
                     .build();
    return enricher;
}
private BiFunction<UseId, ActorContext, UserDetails> userIdToDetails() {
    Function<UseId, UserDetails> result = (productId, context) -> {
        QueryFactory queryFactory = createFromContext(context);
        Query query = queryFactory.byIds(...);
        ...
    };
    return result;
}
```

PR #815, opened 2018-08-10
Add `context` parameter to the enrichment functions

vlad-lubenskyi:
Adresses #813 

In this PR, we pass `EventContext` as a second parameter to enrichments functions.

Thus:
 - the enrichment functions are not longer `Function<Source, Target>`, but `BiFunction<Source, Context, Target>`;
 - `FieldEnrichment` updated respectively;
 - unit tests updated respectively.


vlad-lubenskyi:
@armiol PTAL

server/src/main/java/io/spine/server/outbus/enrich/FieldEnrichment.java
```diff
@@ -37,9 +37,10 @@
 final class FieldEnrichment<S, T, C extends Message> extends EnrichmentFunction<S, T, C> {
```
armiol: While you are here, please document the `C` generic parameter.

server/src/test/java/io/spine/server/outbus/enrich/EnricherBuilderTest.java
```diff
@@ -51,15 +52,15 @@
...
     private Enricher.AbstractBuilder builder;
-    private Function<Timestamp, StringValue> function;
+    private BiFunction<Timestamp, Int32Value, StringValue> function;
     private FieldEnrichment<Timestamp, StringValue, ?> fieldEnrichment;
 
     @BeforeEach
     void setUp() {
         this.builder = EventEnricher.newBuilder();
-        this.function = new Function<Timestamp, StringValue>() {
+        this.function = new BiFunction<Timestamp, Int32Value, StringValue>() {
```
armiol: Please use the value second parameter type (e.g. `EventContext`) here and below.

Once we get rid of rejections (and they become rejection events), the second parameter will always be `EventContext`; that includes `C` parameter in the `FieldEnrichment` class, which will be removed from the generic params at all.

vlad-lubenskyi:
@armiol PTAL, the last build failed because of dependency fetching timeout

ext.gradle
```diff
@@ -25,7 +25,7 @@
  *  as we want to manage the versions in a single source.
  */
  
-def final SPINE_VERSION = '0.10.75-SNAPSHOT'
+def final SPINE_VERSION = '0.10.76-SNAPSHOT'
```
armiol: Please make sure that your version is an increment — if you have to merge from `master`.

PR #822, opened 2018-08-20
Fix posting of multiple commands in response to an event

alexander-yevsyukov:
This PR fixes the bug which did not allow to post more than one command in response to incoming event.


alexander-yevsyukov:
@vlad-lubenskyi, PTAL.

server/src/test/java/io/spine/server/procman/ProcessManagerTest.java
```diff
@@ -244,22 +251,54 @@ void setUp() {
...
+             * @see TestProcessManager#split(PmCencelIteration)
+             */
+            @Test
+            @DisplayName("when splitting incoming command")
+            void splitCommand() {
+                boundedContext.receivesCommand(cancelIteration())
+                              .assertThat(emittedCommands(PmScheduleRetrospective.class,
+                                                          PmPlanIteration.class));
+            }
+
```
vlad-lubenskyi: Redundant empty line.

PR #826, opened 2018-08-22
Importing events to Aggregates

alexander-yevsyukov:
This PR implements importing events into aggregates.

Importing aggregates should have `allowTrue` flag set in `@Apply` annotations. Events to be imported are dispatched by `ImportBus`, which accepts registrations of `AggregateRepositories`.

By default, dispatching is based on `producerId` of `Event`s to be imported, but can be easily changed to using first event message field.

Other notable changes:
  * `UnicastBus` was introduced as an abstract base for unicast dispatching.
  * `Bus.AbstractBuilder` was pushed to the top level and renamed to `BusBuilder`.
  * `idAsString()` was added to `Entity` to simplify diagnostic output.
  * Consolidated code in classes and interfaces responsible for event routing. 
  * Introduced `SystemRepository` abstract base for system aggregate repository classes.
  * Removed `Bounded` infix in classes extending `BoundedContext`.
  * `ImportEvent`, `CommandRounted` types are removed.
  * added factory methods for creating `TestEventFactory` from `BlackBoxBoundedContext`.


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/io/spine/server/bus/DispatcherRegistry.java
```diff
@@ -111,6 +113,32 @@ protected void unregisterAll() {
...
+        int size = dispatchers.size();
+        checkState(size == 1,
+                   "More than one (%s) dispatchers found for the message class `%s`.",
+                   size, messageClass);
+
+        // Since there can be only one dispatcher per message class, and the set is not empty,
+        // we get the single element.
+        D result = dispatchers.stream()
+                              .findFirst()
+                              .get();
```
armiol: As long as you return an optional result in the next line, there is no need to `.get` the result of `findFirst()`. You can simply return the `findFirst()` outcome, as you assured there is a single element in the collection.

PR #827, opened 2018-08-23
Entity Idempotency

dmdashenkov:


dmdashenkov:
[Another approach](#829) was taken.

PR #829, opened 2018-08-27
System-Backed Entity Idempotency

dmdashenkov:
# Entity Idempotency
From now, `Projection`s and `ProcessManager`s are idempotent to commands and events dispatched to them (as well as the `Aggregates`).

## Message dispatching flow
 1. A message is dispatched to a record-based repository.
 2. The repository sends a system command to the associated `EntityHistory`
 3. The `EntityHistory` checks if the message has been recently handled.
 4. If the `EntityHistory` discovers that the message is a duplicate, it produces a rejection. Otherwise, it allows the message dispatching by producing a corresponding event.
 5. The repository listens to those events and rejections and either completes the message dispatching or reports the error via `Repository.onError(...)`.

For more details, see the schema below:
Dispatch Flow

## System Event Types

When dispatching a command to an `Aggregate` or a `ProcessManager`, the system bounded context produces either `CommandDispatchedToHandler` event or `CannotDispatchCommandTwice` rejection.

When dispatching an event to an `Aggregate` or a `ProcessManeger`, the system bounded context produces either `EventDispatchedToReactor` event or `CannotDispatchEventTwice` rejection. 

When dispatching an event to a `Projection`, the system bounded context produces either `EventDispatchedToSubscriber` event or `CannotDispatchEventTwice` rejection. 

A framework user may subscribe to these events and rejections.
_Warning:_ never use entities to listen to the system events, since that would produce an endless stream of commands and events traveling between the domain and system contexts. The only safe way to listen to a system event is to use an `AbstractEventSubscriber` descendant.

## Implementation Details

The depth of the duplicate lookup is the size of the `RecentHistory` of the `EntityHistory` aggregate. It is currently impossible to configure this value.

## Aggregate Idempotency

The mechanism of `Aggregate` idempotency has not changed. However, `Aggregate`s are now idempotent to events on which they react, as well as to the commands.

Though the system events and rejections are produced for the aggregate as well as for the record-based entities, the aggregate repository does not rely on the system context to deliver its messages.

## Concurrent Dispatching

Since a record-based repository relies on the system context messages, it does not block the execution in order to dispatch a given message to its targets. For instance, an event is dispatched to a `Projection` as follows:
 1. The repository receives the event and determines its target(s).
 2. The repository sends a system command. After this stage, the thread is released.
 3. The repository receives the system event or rejection and performs (or does not, in case of the rejection) the event delivery.

Between steps 2 and 3, the thread typically can perform some other work.
_Note:_ the dispatching is truly concurrent if the `TransportFactory` used by the bounded context is non-blocking. In case of the default in-memory transport (or any other _blocking_) transport, the dispatching is performed as a single operation.

# Other Changes
## Routing

From now on, the `EntityMessageEndpoint`s are not responsible for the message routing. Instead, they are told whom to deliver the message to. The routing is performed by the `Repository` itself.

## `EventContext`

In the `EventContext`, we introduce a new `origin_id` `oneof` group. The group consists of `command_id` and `event_id` fields, which represent the ID of the origin message of the associated event.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL.

armiol:
@dmdashenkov 
1. PR description: `A message is dispatched to a record-based entity.` — I think it's to "repository".
2. PR description: `...reports about the error.` -> `reports the error <to where?>.`
3. On a diagram it should be `non-routed`.
4. Please tell about red-green labels in the legend.

And I still don't understand, why a repository is the first in this chain and not the endpoint.

armiol:
@dmdashenkov perhaps, we should rename endpoints into something else, if the proposed flow stays after the discussion.

armiol:
@dmdashenkov and you could reuse `...Watcher` naming that you had previously.

Let's discuss all that in person.

server/src/main/java/io/spine/server/delivery/SystemEventWatcher.java
```diff
@@ -0,0 +1,92 @@
...
+ * {@linkplain io.spine.core.Subscribe#external() external} events.
+ *
+ * @author Dmytro Dashenkov
+ */
+@Internal
+public abstract class SystemEventWatcher<I> extends AbstractEventSubscriber {
+
+    private final TypeUrl targetType;
+
+    protected SystemEventWatcher(TypeUrl targetType) {
```
armiol: Please document constructor. I am not sure what `targetType` is.

server/src/main/java/io/spine/server/delivery/SystemEventWatcher.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Tells whether or not the given {@link io.spine.system.server.EntityHistoryId EntityHistoryId}
+     * belongs to an entity of the {@link #targetType}.
+     *
+     * @param historyId
+     *         the entity history ID to check
+     * @return {@code true} if the type of the entity with given ID is the target type of this
+     *         subscriber, {@code false} otherwise
+     */
+    protected boolean isCorrectType(EntityHistoryId historyId) {
```
armiol: As discussed, let's try to avoid calling this method is every subscriber method (for instance, override the dispatching).

And let's rename it, as it seems to be a sort of matching, while now the method looks like a validation of the passed message :)

dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/delivery/SystemEventWatcher.java
```diff
@@ -47,24 +53,48 @@
...
+     *         of this watcher
      */
-    protected boolean isCorrectType(EntityHistoryId historyId) {
-        String typeUrlRaw = historyId.getTypeUrl();
-        TypeUrl typeUrl = TypeUrl.parse(typeUrlRaw);
-        return typeUrl.equals(targetType);
+    private boolean producerOfCorrectType(EventEnvelope event) {
+        EventContext context = event.getEventContext();
+        Any anyId = context.getProducerId();
+        boolean idTypeMatches = anyId.is(EntityHistoryId.class);
```
armiol: The whole method and this expression in particular seem an expensive operation.

Let's discuss.

PR #831, opened 2018-08-27
Fix optional return value in a commanding method result

### LINKED ISSUE #830: Cannot return empty `Optional` in command reaction method
`CommandingMethod.Result` does not take into account optional nature of a returned value.

Even though, there's a test for optional return value (see `ReOptionalResult` class) from a commanding method, the full cycle is not tested.

alexander-yevsyukov:
This PR fixes #830.


alexander-yevsyukov:
@vlad-lubenskyi, PTAL.

PR #833, opened 2018-08-30
Hide aggregate command endpoint

alexander-yevsyukov:
This PR hides `AggregateEndpoint` classes from public API (which were used by `testutil-server`) and introduces the internal class `AggregateTestSupport` instead.


alexander-yevsyukov:
@armiol, PTAL.

PR #834, opened 2018-08-31
Validate Events

### LINKED ISSUE #824: `EventValidator` should check event message, not outer object
Please see `EventValidator.validate()`. Event though the class Javadoc says:

```java
/**
 ...
 * Checks if the message of the passed event is {@linkplain MessageValidator#validate(Message)
 * valid}.
 */
```
... the code validates the output of `getOuterObject()`. This way, we can pack something non-valid into `Any` and get it through.

dmdashenkov:
In this PR we mark the `Event.message` field as `valid`. This ensures that the event message is validated whenever the `Event` is validated.

Also, in this PR we migrate to `base` of version `0.10.68-SNAPSHOT`.

Fixes #824.


dmdashenkov:
@armiol, PTAL.

PR #835, opened 2018-09-03
Rename `EntityMessageEndpoint` to `EntityProxy`

dmdashenkov:
Since the `EntityMessageEndpoint`s are no longer endpoints, strictly speaking, we:
 1. rename `EntityMessageEndpoint` to `EntityProxy` (and all the descendants accordingly);
 2. make proxies per-entity, not per-message;
 3. rename methods, such as `AggregateRepository.getCommandEndpointDelivery()` so that they don't contain "endpoint" word.


dmdashenkov:
@armiol, PTAL.

Note, that this PR is targeted into the `system-backed-idempotency` (#829), not `master`.

dmdashenkov:
This change is denied.

PR #841, opened 2018-09-10
Event Filter

dmdashenkov:
Introducing `EventFilter`. The filter exists one per repository. The goal of an event filter is to sift or/and modify events posted by entities managed by the given repository.

The same event filter is applied both to the domain events produced by an entity and to the system events posted by an instance of `EntityLifecycle`.

When an event is posted to the event bus or to the system context, the filter is applied to it first. The result of filtering (if any) is then posted.

## Domain Events Application

A typical use case of filtering domain events is securing sensitive data from being accessed by other entities or bounded contexts.

For example:
```protobuf
// --- billing_events.proto ---

message CreditCardUpdated {

    BillingAccountId account_id = 1;

    PersonName new_holder = 2;
    CreditCardNumber new_card_number = 3;
    Cvv2 new_cvv_code = 4;
    MonthYear new_expiration = 5;
}
```
The `CreditCardUpdated` event contains data which should not be accessed from other entities. Yet still, we want to store the whole event in the aggregate history __and__ notify other entities about the fact that the card was updated, but without providing them with the credit card data.

In order to do that, we specify an `EventFilter` which would trim the sensitive information off the event message before it's posted to the bus.

```java
class BillingRepository extends AggregateRepository<BillingAccountId, BillingAccountAggregate> {

    private static final EventFilter filter = EventFieldFilter
        .newBuilder()
        .putMask(CreditCardUpdated.class, 
                 FieldMasks.maskOf(CreditCardUpdated.getDescriptor(),
                                   CreditCardUpdated.ACCOUNT_ID_FIELD_NUMBER))
        .build();

    @Override
    protected EventFilter eventFilter() {
        return filter;
    }
}

```

Note that an event filter __DOES NOT__ prevent events from being exposed to the transport of the current bounded context.

## System Events Application

Some system events carry redundant information about an entity. Framework users who do not need all of the features provided by the system context may disable redundant events from being posted in favor of a better performance.

For example, the `EntityStateChanged` event is emitted each time an entity state is updated (effectively, on each domain event). The event contains the whole entity state in it. This may bring the performance down, as well as fill the storage up with unnecessary redundant data.

In order to avoid this, blacklist the `EntityStateChanged` event for those events, where storing the entity state is not required:

```java

// --- repository class ---

@Override
protected EventFilter eventFilter() {
    return EventBlackList.forbiddenEvents(EntityStateChanged.class);
}
```

Note that the above is the default behavior of a `ProcessManagerRepository`.

A more complex case may involve implementing own `EventFilter`. The only abstract method of this interface is `Optional<Event> filter(Event event)`. It gives the SPI user the freedom of changing the event contents or restricting a certain event from being posted:
```java

// --- repository class ---

@Override
protected EventFilter eventFilter() {
    return event -> {
        if (isSensitive(event)) {
            Event encrypted = encrypt(event);
            return Optional.of(encrypted);
        }
        return Optional.of(event);
    };
}
```

## Other Changes
 - some redundant `{@inheritDoc}` expressions are removed;
 - the recommended way of including the model-assembler annotation processor into a project is now via `annotationProcessor` Gradle configuration, not `compileOnly`; see the [Gradle doc](https://discuss.gradle.org/t/regarding-the-annotation-processors-on-compile-classpath-warning-in-gradle-4-6/26144) for reasoning.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/entity/EventFilter.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * A filter accepting {@link Event}s posted by a {@link Repository}.
+ *
+ * <p>A filter may {@linkplain #allowAll() allow any event}, reject certain types of events, or
+ * change the event content.
+ *
+ * @author Dmytro Dashenkov
+ * @apiNote This type is a {@link FunctionalInterface}, so that an event filter may be defined
+ *          with a lambda expression.
+ * @see EventWhiteList
```
armiol: I'd remove the links to implementations at all. This should go to `package-info.java` instead.

dmdashenkov:
@armiol, PTAL.

armiol:
@dmdashenkov as far as I understand from your recent changes, repositories are now filtering the outgoing events themselves. It just difficult to trace from based on the changeset, honestly speaking.

If so, then LGTM.

If not, let's discuss.

PR #844, opened 2018-09-16
Query limit and ordering

mdrachuk:
This PR adds an ability to sort the query results and limit them to a specified amount of entries.

The ordering is specified using a new `Order` message, and limit is specified using `page_size` parameter in the `Pagination` message. Both are attributes of a `Query`.

The `QueryBuilder` was extended to support newly introduced parameters.

Their support was also implemented for an existing in-memory storage implementation.


mdrachuk:
@armiol, PTAL.

client/src/main/java/io/spine/client/QueryBuilder.java
```diff
@@ -54,22 +64,109 @@
...
+        return self();
+    }
+
+    /**
+     * Limits the number of results returned by the query.
+     *
+     * @param count an amount of the results to be returned
+     * @return this builder instance
+     */
+    public QueryBuilder limit(Integer count) {
```
armiol: `int`.

client/src/main/java/io/spine/client/QueryBuilder.java
```diff
@@ -54,22 +64,109 @@
...
+        return self();
+    }
+
+    /**
+     * Limits the number of results returned by the query.
+     *
+     * @param count an amount of the results to be returned
+     * @return this builder instance
+     */
+    public QueryBuilder limit(Long count) {
```
armiol: `long`

client/src/main/proto/spine/client/query.proto
```diff
@@ -46,30 +47,75 @@ message QueryId {
...
+    //
+    Direction direction = 1 [(required) = true];
+
+    // A column name which value is used for ordering the Query results. 
+    //
+    string column = 2;
+    
+    enum Direction {
+        OD_UNKNOWN = 0;
+        ASCENDING = 1;
```
armiol: Please comment on the enum values.

client/src/main/proto/spine/client/query.proto
```diff
@@ -46,30 +47,75 @@ message QueryId {
...
+
+// Pagination parameters for Query results.
+//
+message Pagination {
+
+    // An amount of entities to be returned by executing a Query.
+    //
+    // A minimum of 1 items per page can be requested
+    //
+    uint64 page_size = 1 [(required) = true];
```
armiol: Please use `(min).value = 1` option.

client/src/main/proto/spine/client/query.proto
```diff
@@ -46,30 +47,75 @@ message QueryId {
...
     // Service information about the environment in which the query was created.
+    //
     core.ActorContext context = 4 [(required) = true,
                                    (valid) = true,
                                    (if_invalid).msg_format = "Invalid actor context"];
 
+    // An order for entities returned by the Query.
+    //
+    Order order = 5 [(valid) = true,
+                     (if_invalid).msg_format = "Invalid pagination order"];
```
armiol: That's just an order. It's no "pagination order". Users can set just the order, without pagination.

client/src/main/proto/spine/client/query.proto
```diff
@@ -46,30 +47,75 @@ message QueryId {
     string value = 1;
 }
 
+// An order for Query results.
+//
+message Order {
```
armiol: Please rename to `OrderBy`.

server/src/test/java/io/spine/server/entity/storage/EntityQueriesTest.java
```diff
@@ -73,8 +79,20 @@
 
     private static EntityQuery<?> createEntityQuery(EntityFilters filters,
                                                     Class<? extends Entity> entityClass) {
+        return createEntityQuery(filters, Order.getDefaultInstance(),
+                                 Pagination.getDefaultInstance(), entityClass);
+    }
+
+    /**
+     * Is not placed test environment because it uses package-private
```
armiol: Something with the grammar here. Please rephrase.

client/src/main/java/io/spine/client/QueryBuilder.java
```diff
@@ -41,6 +49,8 @@
  *                                  .where(eq("type", "permanent"),
  *                                         eq("discountPercent", 10),
  *                                         eq("companySize", Company.Size.SMALL))
+ *                                  .orderedBy("name", DESCENDING)
```
armiol: How come this is `orderedBy`, when we use `SELECT`, `WHERE` and `LIMIT` wording? This should be `orderBy` as well.

PR #850, opened 2018-09-24
Reduce warnings

alexander-yevsyukov:
This PR addresses Error Prone and Javadoc warnings. 

It also improves work with `FieldMask`s in the way suggested by the standard `FieldMaskUtil` class. Previously, we constructed `FieldMask` manually via its `Builder`, assuming that field names are fully-qualified names. It turned out that they should be non-qualified names, and `FieldMaskUtility` class addresses our needs.


alexander-yevsyukov:
@armiol, @dmdashenkov, PTAL.

alexander-yevsyukov:
There are still three Error Prone warnings left. I'm planning to address them in one of the future PRs.

server/src/test/java/io/spine/server/QueryServiceTest.java
```diff
@@ -77,9 +70,10 @@ void setUp() {
         boundedContexts.add(projectsContext);
 
         // Create Customers Bounded Context with one repository.
-        customersContext = BoundedContext.newBuilder()
-                                         .setName("Customers")
-                                         .build();
+        BoundedContext customersContext = BoundedContext
+                .newBuilder()
```
armiol: Let's format this bounded context initialization expression in the same manner as the one a few lines above.

PR #858, opened 2018-10-04
Subscribe to entity state updates

dmdashenkov:
This pull request brings major enhancement to the `@Subscribe` API.

# Conditional subscriptions

It is now possible to subscribe to a fraction of events of a certain type.
When declaring a subscriber method in a `Projection` or in an `AbstractEventSubscriber`, declare a message field filter for received events as follows:
```proto
message BillingStrategy {
    // ...
    bool free_trial = 27;
}

// -- events.proto --

message CustomerCreated {
    // ...
    BillingStrategy billing = 42;
}
```
```java
@Subscribe(
        filter = @ByField(path = "billing.free_trial", value = "false")
)
public void onPaidCustomer(CustomerCreated event) {
    // ...
}
```
The `onPaidCustomer` method will only receive the events `CustomerCreated`, which have the `billing` field such that its `free_trial` field value is `false`.

The `ByField.path` property receives the dot-separated path to the field to filter by. The path may be arbitrarily long.
The `ByField.value` property receives the stringified expected field value.

Any combination of `Subscribe.external` and `Subscribe.filter` is valid, i.e. it is possible to filter external event subscriptions.

A single subscribing class may define a number of subscriber methods with different field filters. Though, all the field filters must target the same field.

# Entity state subscriptions

It is now possible to subscribe to changes in entity states.
In order to do that, just specify the entity type as the first parameter in a subscriber method.

```proto
message Order {
    option (entity).kind = AGGREGATE;
    // ...
}
```
Then in a projection:
```java
@Subscribe
public void onUpdate(Order aggregateState) {
    // ...
}
```
An `AbstractEventSubscriber` is capable of subscribing to entity states as well.

It is also possible to subscribe to updates of entities of other bounded contexts by specifying `@Subscribe(external = true)`.

It is not possible to filter entity states by a field.

### Routing

In order to route entity state updates to projections, a user should route the `EntityStateChanged` events:

```java
class MyProjectionRepository extends ProjectionRepository<...> {
    @Override
    public void onRegistered() {
        getEventRouting().route(EntityStateChanged.class, (event, context) -> {
            Message newState = AnyPacker.unpack(event.getNewState());
            if (newState instanceof Project) {
                return ImmutableSet.of(((Project) newState).getId());
            } else {
                return ImmutableSet.of();
            }
        });
    }
}
```

### Next steps:
 - #862;
 - #863.


dmdashenkov:
@armiol, PTAL.

Note that the coverage flaws (which are not connected to the matter of this PR) are to be fixed shortly.

armiol:
@dmdashenkov please append the PR description with the code snippet and the description on how to prevent _every_ `Order` state update to abuse the subscriber.

server/src/main/java/io/spine/server/event/model/EntitySubscriberMethod.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private Class<? extends Message> entityType() {
+        return getFirstParamType(getRawMethod());
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>An entity state subscriber is always external since {@link EntityStateChanged} is
```
armiol: Let's discuss this. I don't think it's a valid point.

server/src/main/java/io/spine/server/model/HandlerFieldFilterClashError.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package io.spine.server.model;
+
+import java.lang.reflect.Method;
+
+import static java.lang.String.format;
+
+/**
+ * An error indicating that two message handlers have clashing {@link io.spine.core.ByField ByField}
```
armiol: Can you please have a code example of clashing here? I don't think I am following this OK.

server/src/main/java/io/spine/server/route/EntityStateRoute.java
```diff
@@ -0,0 +1,30 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.route;
+
+import com.google.protobuf.Message;
+import io.spine.core.EventContext;
+
+/**
+ * @author Dmytro Dashenkov
```
armiol: While we are at it.

Please document. And also feel free to eliminate all `@author` tags in the files you have changed. We are going to do that anyway.

server/src/main/proto/spine/server/model/message_handler.proto
```diff
@@ -0,0 +1,40 @@
...
+
+import "google/protobuf/any.proto";
+
+import "spine/base/field_path.proto";
+
+// A message handler token.
+//
+// All the message handlers of an entity must have distinct tokens.
+//
+message HandlerToken {
```
armiol: Let's discuss this as well.

armiol:
@dmdashenkov after looking through code I wonder whether two domain contexts (Projects and Tasks) are able to receive the entity state updates of each other. Let's include this question into our discussion tomorrow.

server/src/main/java/io/spine/server/model/HandlerFieldFilterClashError.java
dmdashenkov: Done.

dmdashenkov:
@armiol, PTAL again.

PR #864, opened 2018-10-10
Convenient API for entity subscriptions

### LINKED ISSUE #862: Provide API for entity state updates routing
Currently, in order to route an entity state update to an entity, we have to route the `EntityStateChanged` event directly.

It would be better to give the user the ability to route entity state updates seamlessly without knowing the under-the-hood details.

### LINKED ISSUE #863: Enable entity state subscribers to receive the origin message
Currently, the only valid entity state subscription signature is the one with a single parameter - the entity state itself.

It would be useful to enable the framework users to see the origin message and its context when handling an entity update.

dmdashenkov:
In this PR we improve the API of the entity state subscriptions.

## Signatures

Now it is possible to obtain the context of the `EntityStateChanged` system event:
```java
@Subscribe
public void on(Project state, EventContext systemEventContext) {
    // ...
}
```

Obtaining the origin message is not possible right now.
Fixes #863 (in a way).

## Routing

Now it's easier to route the entity state updates:

```java

// -- ProjectViewRepository.java --

@Override
public void onRegistered() {
    getEventRouting().routeEntityStateUpdates(
        EntityStateRouting
            .<I>newInstance()
            .route(Project.class, (state, context) -> state.getId())
    );
}
```

Fixes #862.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/route/EntityStateRoute.java
```diff
@@ -0,0 +1,34 @@
...
+import io.spine.core.EventContext;
+
+/**
+ * Obtains a set of entity IDs for which to deliver an entity state update.
+ *
+ * @param <I> the type of entity IDs
+ * @param <M> the type of entity states to get IDs from
+ */
+@FunctionalInterface
+public interface EntityStateRoute<I, M extends Message> extends Multicast<I, M, EventContext> {
```
armiol: Can we rename it to `StateUpdateRoute`? I think `Update` is a key part.

server/src/main/java/io/spine/server/route/EntityStateRouting.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * <p>A routing schema consists of a default route and custom routes per event class.
+ * When calculating a set of update targets, {@code EntityStateRouting} would see if there is
+ * a custom route set for the type of the event. If not found, the default route will be
+ * {@linkplain EventRoute#apply(Message, Message) applied}.
+ *
+ * @param <I>
+ *         the type of the entity IDs to which the updates are routed
+ */
+public class EntityStateRouting<I>
```
armiol: `StateUpdateRouting` similar to one above.

PR #867, opened 2018-10-12
Delete empty `Rejection`-related Java files

armiol:
Some `.java` files that were previously declaring `Rejection`-related classes are now empty in the VCS. Most likely, they weren't deleted, but left empty after some merge procedure.

This PR deletes them from the VCS.

The framework version is the same.


armiol:
@alexander-yevsyukov PTAL.

PR #868, opened 2018-10-12
Swap "visible" for "active" in respect to Lifecycle Flags

mdrachuk:
This PR replaces whatever "visible" mentions of lifecycle flags were left (after LifecycleFlags replaced Visibility) for "active" and updates the documentation of Record Storage specifying how the lifecycle flags should be treated when querying.

Also, the creation of a complete EntityQuery, which includes lifecycle flags, was moved to in-memory record storage implementation to match the behaviour specified by documentation.


mdrachuk:
@armiol, PTAL.

server/src/main/java/io/spine/server/entity/EntityWithLifecycle.java
```diff
@@ -29,12 +29,17 @@
 /**
  * An entity which has {@linkplain LifecycleFlags lifecycle flags}.
  *
+ * <p>Lifecycle flags determine if an entity is active.
+ * An entity is considered to be active if the lifecycle flags are missing.
+ * If an entity is {@linkplain #isArchived() archived} or {@linkplain #isDeleted() deleted}, 
+ * then its regarded to be inactive.
```
armiol: `it's`

server/src/main/java/io/spine/server/entity/storage/QueryParameters.java
```diff
@@ -75,6 +75,13 @@ public static Builder newBuilder() {
         return new Builder();
     }
 
+    public static Builder newBuilder(QueryParameters parameters) {
+        return new Builder()
+                .addAll(parameters)
```
armiol: It seems to be misaligned. Maybe I am wrong. Please check.

mdrachuk: This alignment is performed by our IDEA config.

PR #870, opened 2018-10-16
Add missing nullable annotations to storage `readMultiple(...)`

mdrachuk:
This PR adds missing Nullable annotations and documents the situations in which nulls can be returned.


mdrachuk:
@armiol, PTAL.

PR #871, opened 2018-10-16
System Read/Write Sides

dmdashenkov:
Introducing `SystemReadSide`—the way for a domain context to receive events and read domain aggregate states from its system counterpart.

## Usage

Before this PR, a domain context would listen to system events through the `IntegrationBus`. This approach is not quite efficient, since all the system events were sent through transport.

From now on, the domain context may subscribe to the system events directly through `SystemReadSide`, which acts in memory.

## API changes

`SystemGateway` is renamed to `SystemWriteSide`.  The method `readDomainAggregate` is moved from the ex-gateway to read side.

`BoundedContext.getSystemGateway` is deleted. Instead, the `BoundedContext` has a composite entity—`SystemClient`, which, effectively, is an umbrella term for the system read and write sides.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/BoundedContextBuilder.java
```diff
@@ -350,7 +351,7 @@ private void initEventBus(StorageFactory storageFactory) {
         }
     }
 
-    private void initStand(SystemGateway systemGateway) {
+    private void initStand(SystemReadSide systemWriteSide) {
```
armiol: The parameter name is wrong.

server/src/main/java/io/spine/server/bus/BusBuilder.java
```diff
@@ -88,15 +88,15 @@ public final B appendFilter(BusFilter<E> filter) {
     }
 
     /**
-     * Inject the {@link SystemGateway} of the bounded context to which the built bus belongs.
+     * Inject the {@link io.spine.system.server.SystemWriteSide} of the bounded context to which the built bus belongs.
```
armiol: Please have an alt text.

server/src/main/java/io/spine/server/delivery/SystemEventWatcher.java
```diff
@@ -102,6 +93,16 @@ public boolean canDispatch(EventEnvelope envelope) {
             );
         }
         return classes;
+
```
armiol: Please kill this line.

server/src/main/java/io/spine/system/server/SystemMonitor.java
```diff
@@ -0,0 +1,62 @@
...
+    void closeSystemContext() throws Exception;
+
+    /**
+     * Creates a new instance of {@code SystemMonitor} for the given system context.
+     *
+     * @param context
+     *         the system context to monitor
+     * @return new system monitor
+     */
+    static SystemMonitor newInstance(SystemContext context) {
```
armiol: Let's make `SystemContext` to return the value instead of this factory method.

server/src/main/java/io/spine/system/server/SystemMonitor.java
```diff
@@ -0,0 +1,62 @@
...
+
+import io.spine.annotation.Internal;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * The entry point of a system context API exposed to its domain counterpart.
+ */
+@Internal
+public interface SystemMonitor {
```
armiol: I think it's more of `SystemClient`, as long as monitor just monitors. And this one closes, exposes modification operations etc.

server/src/test/java/io/spine/system/server/TenantAwareSystemReadSideTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @DisplayName("in multitenant env")
+    class Multitenant {
+
+        private TenantId tenantId;
+
+        @BeforeEach
+        void setUp() {
+            delegate = MemoizingReadSide.multitenant();
+            tenantId = TenantId
+                    .newBuilder()
```
armiol: I think the code isn't aligned here and there. Maybe it would be cool to reformat the test code you have changed automatically in IDEA by multiselecting the changed files.

dmdashenkov:
@armiol, PTAL.

PR #873, opened 2018-10-19
Public QueryParameters factory with active lifecycle flags

mdrachuk:
This PR adds a public QueryParameters factory with active lifecycle flags that to use in other storage implementations.

The reason behind this change is that we don’t have any means to find out what entity state is considered active in public API.

Also:
- a missing `@Nullable` annotation is added to the RecordStorage
- Grpc, Spine, and test generated directories are registered as generated by the IDEA


mdrachuk:
@armiol, PTAL.

server/src/test/java/io/spine/server/entity/storage/QueryParametersTest.java
```diff
@@ -190,9 +202,44 @@ void keepManyFiltersForColumn() throws ParseException {
         assertTimeFilters.containsExactly(startTimeFilter, deadlineFilter);
     }
 
-    private static EntityColumn mockColumn() {
-        EntityColumn column = mock(EntityColumn.class);
-        return column;
+    @Test
+    @DisplayName("create an parameters with active lifecycle flags")
```
armiol: Please fix the grammar.

server/src/main/java/io/spine/server/storage/RecordStorage.java
```diff
@@ -377,7 +377,7 @@ public EntityColumnCache entityColumnCache() {
      *
      * @see BulkStorageOperationsMixin#readMultiple(java.lang.Iterable)
      */
-    protected abstract Iterator<EntityRecord> readMultipleRecords(Iterable<I> ids);
+    protected abstract Iterator<@Nullable EntityRecord> readMultipleRecords(Iterable<I> ids);
```
armiol: Please explain why it is `@Nullable`. It's not obvious from here.

PR #874, opened 2018-10-19
Fix a projection e2e test

dmdashenkov:
In this PR we fix a poorly implemented test case.


dmdashenkov:
@armiol, please check upon your changes, since that how we discovered the issue in the first place.

PR #875, opened 2018-10-23
Release `1.0.0-pre1`

dmytro-grankin:
This PR updates `base` dependency to `1.0.0-pre1`.

Changes:
- `Identifier.generate(Class) is used where possible.
- Tests updated since now entity IDs are required by default.
- The usages of `Identifier` and `TypeConverter` updates after the [API](https://github.com/SpineEventEngine/base/pull/232) changes.

The framework version was bumped to `1.0.0-pre1`.


dmytro-grankin:
@armiol PTAL.

server/src/main/proto/spine/system/server/command_lifecycle.proto
```diff
@@ -46,7 +46,9 @@ message CommandLifecycle {
     option (entity).kind = AGGREGATE;
 
     // The ID of the command.
-    spine.core.CommandId id = 1;
+    //
+    // The ID is not required since it is set only after the command was received.
```
armiol: This is an issue, I think. Let's discuss.

server/src/test/proto/spine/test/system/server/q/projections.proto
```diff
@@ -1,41 +0,0 @@
...
-
-option (type_url_prefix) = "type.spine.io";
-option java_package = "io.spine.system.server";
-option java_outer_classname = "ProjectionsProto";
-option java_multiple_files = true;
-
-import "spine/people/person_name.proto";
-import "spine/test/system/server/entity_history_test.proto";
-
-message PersonDetails {
```
armiol: I don't understand how this file isn't needed anymore.

dmytro-grankin: It was just moved to a more appropriate file — `entity_history_test.proto`. I think it is more appropriate to have all definitions for a test suite in a single file while the number of definitions isn't big.

server/src/main/proto/spine/system/server/command_lifecycle.proto
dmytro-grankin: I have created the issue #877 since it isn't easy to make it required now.

dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

PR #878, opened 2018-11-09
Remove duplicated javac config

alexander-yevsyukov:
This PR removes duplicated ErrorProne javac configuration.

`config` dependency was also updated.


alexander-yevsyukov:
@dmdashenkov, PTAL.

PR #884, opened 2018-11-15
Add `Server` and `TestClient`

alexander-yevsyukov:
This PR:
  * Teaches `StorageFactory` to create its copies with a name of the target Bounded Context and its multi-tenancy status. This is needed for addressing mutual dependency with `InMemoryStorageFactory` needed in test environments.
  * Adds `Server` class which exposes one or more `BoundedContext`s via Command- and Query- services running over gRPC.
  * `TestClient` testing utility class, which is a simple client for local testing of gRPC communications
  * Adds ability to add a `Repository` to a `BoundedContextBuilder` so that the repository is registered with the `BoundedContext` produced by the builder.
  * Removes unnecessary static factory methods for in-memory storages.


alexander-yevsyukov:
@armiol, PTAL.

Issue #889, opened 2018-11-19
Kill `GrpcService` potentially exposing `EventStore`


armiol:
Currently when building `EventStore` we also build a service definition for gRPC service, exposing the `EventStore`. The class is `io.spine.server.event.GrpcService`.

Looking like this, the feature is harmful, as we don't want to expose the whole event stream to the outer world. In future maybe a `SubscriptionService` will allow to subscribe to events, but such a functionality is a matter of design to do.

For now let's remove this feature by deleting the `GrpcService` class and getting rid of its usages.

armiol:
@alexander-yevsyukov FYI.

armiol:
Closing as resolved.

PR #890, opened 2018-11-19
Kill ES gRpc service

alexander-yevsyukov:
This PR removes support of exposure of `EventStore` as a gRPC service (see #889). Queries must be performed only via `QueryService`.

Other notable changes:
  * `EventStore` and its builder were made `final`.


alexander-yevsyukov:
@armiol, PTAL.

PR #892, opened 2018-11-26
Optimize aggregate read operation

dmdashenkov:
In this PR we make the `AggregateRepository` read the required amount of history events, even if this number is greater than the snapshot trigger.

Previously, the number of read events was always equal to the snapshot trigger.

Also, now we cache the `eventCountAfterLastSnapshot` in the `Aggregate` instance. This helps us not to re-fetch the value when appending the aggregate history.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -120,6 +120,16 @@
...
 
+    /**
+     * The count of events stored to the {@linkplain AggregateStorage storage} since the last
+     * snapshot.
+     *
+     * <p>This field is set in {@link #play(AggregateStateRecord)} and is effectively final.
+     *
+     * @see AggregateStorage#readEventCountAfterLastSnapshot(Object)
+     */
+    private int persistedEventCount = 0;
```
armiol: To me, the name of the variable does not correspond to its description and usage.

dmdashenkov:
@armiol, PTAL.

Issue #880, opened 2018-11-12
Why all Process Managers always have version 0?


yuri-sergiichuk:
Currently, any Process Manager, in any case, has its version equal to `0`. Is it a desired behavior or bug?

P.S. Aggregates and Projections versions are advanced.

armiol:
Process Managers now act by advancing their versions upon any dispatched message. Each advance is an auto-increment.

Closing the issue as addressed. For more details see #893.

PR #893, opened 2018-11-28
Add version update for `ProcessManager` instances

dmitrykuzmin:
This PR handles issue #880.

Previously, the entity versioning didn't work for the `ProcessManager` implementations.

Now, the version of the `ProcessManager` is incremented by one upon any command or event dispatched to it (as process managers can update their state in command handlers, unlike aggregates).

The other side of the changes is that now `ProcessManager` state must be consistent with the model requirements upon handling any command. 

For instance, if the `QuizProcman` receives a command `PmStartQuiz`, and the `id` field of procman is required, the handler method should set the ID from the command right away and not wait until `PmQuizStarted` event.

This is done because the version update requires the entity state to be valid.


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/entity/Transaction.java
```diff
@@ -494,6 +495,32 @@ public void setDeleted(boolean deleted) {
                                        .build();
     }
 
+    /**
+     * Advances the transaction version.
```
armiol: I think we should be more precise here, as transactions do not have versions themselves.

server/src/main/java/io/spine/server/entity/EntityVersioningContext.java
```diff
@@ -0,0 +1,69 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.entity;
+
+import io.spine.core.EventEnvelope;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * The context which {@link EntityVersioning} uses for version increment.
```
armiol: Can we turn this value object into something that acts? E.g. can this be a `VersionIncrement`, which is responsible for setting a new version to an entity in a transaction?

dmitrykuzmin:
@armiol PTAL.

dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/entity/EventPlayingTransaction.java
```diff
@@ -0,0 +1,101 @@
...
+import io.spine.core.EventEnvelope;
+import io.spine.core.Version;
+import io.spine.server.event.EventDispatch;
+import io.spine.server.model.Nothing;
+import io.spine.validate.ValidatingBuilder;
+
+/**
+ * A transaction that supports event {@linkplain EventPlayer playing}.
+ *
+ * <p>Event playing is a process of applying the events to the certain entity with no expected
```
armiol: Let's remove this sentence. It's clear from the referenced `EventPlayer`.

server/src/main/java/io/spine/server/entity/CommandDispatchingPhase.java
```diff
@@ -0,0 +1,70 @@
...
+ * <p>The result of such dispatch is always a {@link List} of {@linkplain Event events} as
+ * described in the {@code CommandHandlingEntity}
+ * {@linkplain io.spine.server.command.CommandHandlingEntity#dispatchCommand(CommandEnvelope)
+ * contract}.
+ *
+ * @param <I>
+ *         the type of entity ID
+ */
+@Internal
+public class CommandDispatchingPhase<I> extends Phase<I, List<Event>> {
```
armiol: I think the problem here would be `@Command`-er methods in Process Manager, which return commands (split into several or transform the given one into another).

The same issue is in `EventDispatchingPhase`. As Process Managers may `@Command` upon events, turning events into commands.

armiol:
@dmitrykuzmin let's create some tests for `@Command`-ing scenarios and then see, what needs updates in the code.

PR #897, opened 2018-11-30
Add shortcut methods for asserting one event or rejection

alexander-yevsyukov:
This PR:
1. clarifies multi-tenancy status in API of creating `BlackBoxBoundedContext`.
2. addresses generics usage warning from ErrorProne.
3. adds shortcut methods for testing emitted events or rejections. So that instead of:

```java
     context.assertThat(emittedEvent(TaskCreated.class, once()));
```
... it would be possible to write:
```java
   context.assertEmitted(TaskCreated.class);
```
It is also possible to write:
```java
   context.assertRejectedWith(SomeRejection.class);
```

### DIFF
--- a/testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
+++ b/testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
@@ -90,28 +90,28 @@ public abstract class BlackBoxBoundedContext<T extends BlackBoxBoundedContext> {
     /**
      * Creates a single-tenant instance with the default configuration.
      */
-    public static SingleTenantBlackBoxContext newInstance() {
+    public static SingleTenantBlackBoxContext singleTenant() {
         return new SingleTenantBlackBoxContext(emptyEnricher());
     }
 
     /**
      * Creates a single-tenant instance with the specified enricher.
      */
-    public static SingleTenantBlackBoxContext newInstance(Enricher enricher) {
+    public static SingleTenantBlackBoxContext singleTenant(Enricher enricher) {
         return new SingleTenantBlackBoxContext(enricher);
     }
 
     /**
      * Creates a multitenant instance the default configuration.
      */
-    public static MultitenantBlackBoxContext multitenant() {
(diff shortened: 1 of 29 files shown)


alexander-yevsyukov:
@armiol, there is some strange anomaly of collecting rejections in `BlackBoxBoundedContext`. Please see my TODO note in the `BlackBoxBoundedContext.asserRejectedWith()`. We're getting two messages instead of one. If you can have somebody to look into this, I'd appreciate that.

PTAL to my changes in this PR. I decided to introduce them in relation to my current works on improving Kanban Board example.

armiol:
@alexander-yevsyukov I have addressed your `todo` item. Please see [this commit](https://github.com/SpineEventEngine/core-java/pull/897/commits/ad31ee0c04fd579958a9deced009f849c7ea3395) for more details.

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
```diff
@@ -84,35 +88,35 @@ protected BlackBoxBoundedContext(boolean multitenant, Enricher enricher) {
     }
 
     /**
-     * Creates a single tenant bounded context with the default configuration.
+     * Creates a single-tenant instance with the default configuration.
      */
-    public static SingleTenantBlackBoxContext newInstance() {
+    public static SingleTenantBlackBoxContext singleTenant() {
```
armiol: Well, I agree. But I have specifically asked to leave `newInstance` instead of `singleTenant` to keep our API compatible with the existing testing code and ease the migration. The reason is that we are `1.0.0-SNAPSHOT` and each published build breaks projects which reference it immediately — if an API element like this one is changed. This is the case for at least `users` library and one more proprietary project.

What if we introduce `singleTenant()`, keep `newInstance()` as an alias and mark the latter as `@Deprecated` to remove in `pre2`? In this way we'll still be able to compile our code (i.e. `users` library) and schedule a migration once we are ready to do it.

alexander-yevsyukov: Did so.

PR #904, opened 2018-12-05
Stop treating received events as emitted in BlackBoxBoundedContext

mdrachuk:
This PR filters out the events sent by `BlackBoxBoundedContext` from emitted events verifiers.

Such a change is required to match the expected behaviour when calling methods treating `emitted` events.


mdrachuk:
@armiol, PTAL.

Issue #848, opened 2018-09-21
Unsupported `is` prefix for `Column` getters


mdrachuk:
Column documentation states that methods with `get` and `is` prefixes are supported as columns, but in fact the `is` prefix is not supported.

dmitrykuzmin:
Resolved via #910.

PR #910, opened 2018-12-07
Clarify `is` prefix usage for entity columns

dmitrykuzmin:
This PR handles the issue #848.

Our entity columns lookup is based on the [Java Bean](https://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/) getter spec which allows prefix `is` for `boolean` properties only.

The `Column` doc and the tests were updated accordingly.

### DIFF
--- a/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
+++ b/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
@@ -44,8 +44,7 @@ import static io.spine.util.Exceptions.newIllegalStateException;
  * checks verifying that {@link EntityColumn} definitions in the processed {@link Entity} class are
  * correct. If column definitions are incorrect, the exception is thrown upon
  * {@linkplain EntityColumn entity columns} reading.
- *
- * @author Dmytro Kuzmin
+ * 
  * @see Columns
  * @see EntityColumn
  */
--- a/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
+++ b/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
@@ -44,7 +44,7 @@ import static io.spine.util.Exceptions.newIllegalStateException;
  * checks verifying that {@link EntityColumn} definitions in the processed {@link Entity} class are
  * correct. If column definitions are incorrect, the exception is thrown upon
  * {@linkplain EntityColumn entity columns} reading.
- * 
+ *
  * @see Columns
  * @see EntityColumn
  */
@@ -57,7 +57,7 @@ class ColumnReader {
     }
 
     /**
-     * Creates an instance of {@link ColumnReader} for the given {@link Entity} class.
+     * Creates an instance of {@code ColumnReader} for the given {@link Entity} class.
      *
      * <p>The reader can be further used to {@linkplain ColumnReader#readColumns() obtain}
      * {@linkplain EntityColumn entity columns} for the given class.
--- a/server/src/main/java/io/spine/server/entity/storage/ColumnRecords.java
+++ b/server/src/main/java/io/spine/server/entity/storage/ColumnRecords.java
@@ -36,7 +36,7 @@ import static java.lang.String.format;
  */
 public final class ColumnRecords {
 
-    /** Prevent initialization of the utility class */
+    /** Prevents initialization of the utility class. */
     private ColumnRecords() {
     }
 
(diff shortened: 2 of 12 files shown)


server/src/main/java/io/spine/server/entity/storage/ColumnRecords.java
```diff
@@ -36,7 +36,7 @@
  */
 public final class ColumnRecords {
 
-    /** Prevent initialization of the utility class */
+    /** Prevents initialization of the utility class. */
```
alexander-yevsyukov: It's “instantiation” (i.e. we cannot create instances of this class), not “initialization”. The class _is_ initialized.

Please also look for such errors around.

dmitrykuzmin:
@alexander-yevsyukov, @armiol PTAL.

server/src/main/java/io/spine/server/entity/storage/Methods.java
```diff
@@ -107,6 +106,9 @@ static void checkGetter(Method getter) {
                                            .find()
                               && getter.getParameterTypes().length == 0,
                       "Method `%s` is not a getter.", getter);
+        checkArgument(!getter.getName().startsWith(IS) ||
+                              boolean.class.isAssignableFrom(getter.getReturnType()),
```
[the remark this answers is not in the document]
dmitrykuzmin: No, as the column lookup doesn't allow `Boolean` columns with `is` prefix. In the JavaBean spec they mention only `boolean` too.

We could probably extend their column lookup with our own functionality to support `Boolean` columns starting with `is`, if it's important.

dmitrykuzmin:
@alexander-yevsyukov, @armiol PTAL.

dmitrykuzmin:
@alexander-yevsyukov , @armiol PTAL.

server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
```diff
@@ -85,32 +112,52 @@ static ColumnReader forClass(Class<? extends Entity> entityClass) {
...
-        ImmutableSet<EntityColumn> columns = Arrays
-                .stream(propertyDescriptors)
+    /**
+     * Gathers entity properties and creates the entity columns where appropriate.
+     *
+     * <p>The entity properties search is based on the Java Bean
+     * <a href="https://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/">
+     * specification</a>.
+     */
+    private Set<EntityColumn> gatherColumnsFromProperties() {
```
armiol: Can't we join both of `gather...` methods into one?

And also, I would rename the method to something like `scanColumns`.

server/src/main/java/io/spine/server/entity/storage/Methods.java
```diff
@@ -44,13 +44,25 @@
...
+     * One of the possible getter prefixes - {@code get-}.
+     */
+    private static final String GET_PREFIX = "get";
+
+    /**
+     * The other possible getter prefix - {@code is-}.
+     *
+     * <p>Allowed for {@code boolean} and {@link Boolean} entity columns.
+     *
+     * <p>Package-private access because we manually search for some of such methods.
```
armiol: At first glance this explanation doesn't really explain the reason. Can you please rephrase?

server/src/main/java/io/spine/server/event/MessageEnrichment.java
```diff
@@ -146,7 +146,8 @@ private void setFields(Message.Builder builder, S sourceMessage, C context) {
                                                              Field.getFieldClass(targetField)));
                 EnrichmentFunction fieldEnrichment = function.get();
                 @SuppressWarnings("unchecked"
```
armiol: Please see how a long comment to such a suppression is formatted in `RecordBasedRepository`.

dmitrykuzmin:
@armiol PTAL.

server/src/test/java/io/spine/server/entity/storage/ColumnReaderTest.java
```diff
@@ -74,7 +79,9 @@ void fromClassWithColumns() {
             ColumnReader columnReader = forClass(EntityWithManyGetters.class);
             Collection<EntityColumn> entityColumns = columnReader.readColumns();
 
-            assertThat(entityColumns).hasSize(3);
+            assertThat(entityColumns).hasSize(5);
```
armiol: Not sure why my comment on this disappeared...

I asked to rewrite this piece and avoid counting objects. Instead we can compare two `Set`s.

dmitrykuzmin:
@armiol PTAL.

PR #913, opened 2018-12-11
Remove redundant `VisibleForTesting` annotation

mdrachuk:
This PR removes `@VisibleForTesting` annotation from `ProcessManager#getBuilder()`.


armiol:
@mdrachuk 
See https://youtrack.jetbrains.com/issue/IDEA-204081. We don't need to remove the annotation, but let's add an `@apiNote`.

PR #914, opened 2018-12-11
Clean up `WithLifecycle`

### LINKED ISSUE #818: Mark generated messages with corresponding marker interfaces
Those from `commands.proto` should be `CommandMessage`, `events.proto` -> `EventMessage`, 
`rejections.proto` -> `RejectionMessage`.

As per #733, `RejectionMessage` should extend `EventMessage`.

`EventContext` and `CommandContext` should implement `MessageContext` marker interface.

### LINKED ISSUE #909: Remove build hack for `WithLifecycle` after Spine Proto Compiler supports optional generation of interfaces
The code of `server/build.gradle` has the task for removing generated `WithLifecycle.java`. The task is needed because there is handcrafted `WithLifecycle` class, which is used in code generation, so that `EntityRecord` implements it.

The hack can be removed after Spine Proto Compiler supports optional generation of interfaces via extended syntax of `is` option. Once it's available the following needs to be done:

- [ ]  New option should be used in `entity.proto`.
   Presumably, it would look like:
  ```proto
  option (is).java_type = "WithLifecycle" ;
  ```
  This syntax assumes that `generate` attribute equals `false`.

- [ ]  The task `deleteGeneratedWithLifecycleJava` and dependence on it is removed from `server/build.gradle`.

dmdashenkov:
### Summary

This PR removes a workaround which allowed interface `WithLifecycle` to be implemented in Protobuf message classes.

Also, versions of `base` and `time` are updated.

Fixes #909.
Fixes #818.

### `WithLifecycle`

Previously, `WithLifecycle` marker interface was both handcrafted __and__ generated by the Protobuf compiler. The generated version was then deleted in the Gradle plugin.

Since `(is)` option now allows using handcrafted interfaces instead of blindly generating them, this workaround is removed.

### `MessageContext`

`EventContext` and `CommandContext` message classes now both implement a common interface—`MessageContext`. The interface is, for now, generated by Spine Protobuf compiler plugin. Though, the contract of `MessageContext` may be changed in future if needed.

### Migration to the fresh `base`

In this PR we update the version of `base` and `time` libs to `1.0.0-SNAPSHOT`. This implies changing the API of `MessageValidator`. Previously, the validator could accept multiple messages in `validate(Message)` method. Now, the validator is created for a particular message instance and only validates that instance.

#### Removal of event validator in `EventBus`

It is now not possible to use a single `MessageValidator` for many messages. Thus, the validator instance stored in an `EventBus` does not make sense now.

It was decided not to replace the validator with some kind of new API because:
 - we have `Bus` filters to apply additional constraints to the posted events;
 - the user may declare more soft validation rules for events in order to skip some of the validation.

### DIFF
--- a/core/src/main/proto/spine/core/command.proto
+++ b/core/src/main/proto/spine/core/command.proto
@@ -91,6 +91,7 @@ message Command {
 
 // Meta-information about the command and the environment, which generated the command.
 message CommandContext {
+    option (is) = {java_type: "io.spine.core.MessageContext" generate: true};
 
     // Information about the environment of the user who created the command.
     ActorContext actor_context = 1 [(required) = true];
--- a/core/src/main/proto/spine/core/event.proto
+++ b/core/src/main/proto/spine/core/event.proto
@@ -81,6 +81,7 @@ message Event {
 
 // Meta-information for an event.
 message EventContext {
+    option (is) = {java_type: "io.spine.core.MessageContext" generate: true};
 
     // When the event occurred.
     google.protobuf.Timestamp timestamp = 1 [(required) = true];
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -44,17 +44,3 @@ javadoc {
         }
     }
 }
-
-//TODO:2018-12-07:alexander.yevsyukov: Remove this task and dependency on it after Spine Proto Compiler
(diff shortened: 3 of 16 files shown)


dmdashenkov:
@armiol, PTAL.

Issue #885, opened 2018-11-15
Calculate `zoneOffset` by `zoneId` in `TestActorRequestFactory`


alexander-yevsyukov:
Now we pass two parameters: `zoneOffset` and `zoneId`. Before `ZoneId` type was introduced, there used to be only one parameter, `zoneOffset`. It is possible to calculate the current offset by the zone ID. So, there's no sense in passing two parameters.

serhii-lekariev:
https://github.com/SpineEventEngine/core-java/pull/915 has addressed the described issue.
` TestActorRequestFactory`'s ctor now accepts 2 parameters, and the zone offset is calculated based on the specified zone ID.

PR #915, opened 2018-12-13
Calculate `zoneOffset` based on `zoneId` in `TestActorRequestFactory`

serhii-lekariev:
This PR addresses https://github.com/SpineEventEngine/core-java/issues/885.

Before, to create an instance of `TestActorRequestFactory`, two timezone-related parameters were needed:
zone ID and zone offset. 

Since it is possible to calculate offset based on ID, two parameters have been substituted with a single one - zone ID. Zone offset is now calculated based on the ID.


serhii-lekariev:
@dmdashenkov, PTAL.

Issue #912, opened 2018-12-11
Unexpected NPE when scanning for entity columns


dmdashenkov:
If an entity class declares a `public` setter of getter method, `ColumnReader` assumes the method is a property of a bean, i.e. an entity column. When trying to find an annotated version of the method in the class hierarchy (and failing to do so), `ColumnReader` throws an NPE with no message.

The expected behaviour is either
 - ignoring the method at all, possibly, with a warning log;
 - failing with a more descriptive exception and message.

dmdashenkov:
@armiol, let's discuss this issue, please.

armiol:
@dmdashenkov In case there is a getter with no `@Column` annotation, we just ignore it.

serhii-lekariev:
This issue has been addressed in https://github.com/SpineEventEngine/core-java/pull/917.

PR #917, opened 2018-12-14
`ColumnReader` `NullPointerException` when finding a `setSomething()` method that is not a property mutator

serhii-lekariev:
This PR addresses https://github.com/SpineEventEngine/core-java/issues/912.

Before, when `ColumnReader` was scanning the class in an attempt to parse `Columns`, it has tried to get all property descriptors of that class, and once obtained, call a `getReadMethod()` on every descriptor, to check whether it's annotated with `@Column`.

If a class declares a public `setSomething(Something something)` method, `something` becomes a property, and the declared method becomes a write method for this property.

A read method, however, remains undeclared, and `getReadMethod()` returns a `null`, hence the NPE.

Now, `ColumnReader` checks whether a property has an accessor, before attempting to get that accessors annotations and check whether it's actually a `@Column`.

### DIFF
--- a/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
+++ b/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
@@ -130,7 +130,6 @@ class ColumnReader {
      *         if columns contain repeated names
      */
     private void checkRepeatedColumnNames(Iterable<EntityColumn> columns) {
-
         Collection<String> checkedNames = newLinkedList();
         for (EntityColumn column : columns) {
             String columnName = column.getStoredName();
--- a/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
+++ b/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
@@ -98,15 +98,15 @@ class ColumnReader {
         }
 
         PropertyDescriptor[] propertyDescriptors = entityDescriptor.getPropertyDescriptors();
-        ImmutableSet<EntityColumn> annotatedColumns = Arrays
+        ImmutableSet<EntityColumn> columns = Arrays
                 .stream(propertyDescriptors)
                 .map(PropertyDescriptor::getReadMethod)
                 .filter(Objects::nonNull)
                 .filter(ColumnReader::isAnnotated)
                 .map(EntityColumn::from)
                 .collect(toImmutableSet());
-        checkRepeatedColumnNames(annotatedColumns);
-        return annotatedColumns;
+        checkRepeatedColumnNames(columns);
+        return columns;
     }
 
     private static boolean isAnnotated(Method method) {
--- a/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
+++ b/server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
@@ -20,14 +20,18 @@
 
 package io.spine.server.entity.storage;
 
+import com.google.common.collect.ImmutableSet;
 import io.spine.server.entity.Entity;
 
 import java.beans.BeanInfo;
 import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
+import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Set;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


serhii-lekariev:
@dmdashenkov, PTAL.

serhii-lekariev:
@dmdashenkov, PTAL.

server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
```diff
@@ -91,27 +93,34 @@ static ColumnReader forClass(Class<? extends Entity> entityClass) {
...
-                EntityColumn column = EntityColumn.from(getter);
-                entityColumns.add(column);
-            }
-        }
+        PropertyDescriptor[] propertyDescriptors = entityDescriptor.getPropertyDescriptors();
+        Arrays.stream(propertyDescriptors)
+              .filter(ColumnReader::hasAccessor)
+              .filter(ColumnReader::accessorIsAnnotated)
+              .map(PropertyDescriptor::getReadMethod)
+              .forEach(readMethod -> entityColumns.add(EntityColumn.from(readMethod)));
```
dmdashenkov: Please use a combination of `map` and `collect` instead. `forEach` must be stateless. In practice, in order to be honest with the functional style coding, `forEach` is almost never applicable.

server/src/test/java/io/spine/server/entity/storage/ColumnReaderTest.java
```diff
@@ -114,6 +115,14 @@ void fromImplementedInterface() {
         }
     }
 
+    @Test
+    @DisplayName("not confuse a `setSomething()` method with a property mutator")
```
dmdashenkov:
```suggestion
    @DisplayName("not confuse a setter method with a property mutator")
```

server/src/test/java/io/spine/server/entity/storage/given/ColumnsTestEnv.java
```diff
@@ -53,6 +53,33 @@ public int getValue() {
...
+    public static class EntityWithASetterButNoGetter extends AbstractEntity<String, Any> {
+
+        private Integer secretNumber;
+
+        protected EntityWithASetterButNoGetter(String id) {
+            super(id);
+        }
+
+        @Column
+        public Integer getFortyThree(){
```
dmdashenkov: The presence of this `@Column` makes the test a bit confusing. Can we just drop it and assert that there are _no_ columns?

server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
dmdashenkov: Also, you may extract the method from the descriptor as the first step, filter out nulls and not annotated methods and collect them at once. This is also a performance improvement, since `PropertyDescriptor.getReadMethod()` looks quite complex and not necessarily fast.

serhii-lekariev:
@dmdashenkov, PTAL again.

server/src/test/java/io/spine/server/entity/storage/given/ColumnsTestEnv.java
```diff
@@ -53,6 +53,28 @@ public int getValue() {
         }
     }
 
+    /**
+     * This entity declares a {@linkplain #setSecretNumber(Integer) mutator method}, however
```
dmdashenkov: `An entity type which ...`

serhii-lekariev:
@dmdashenkov, PTAL again.

server/src/main/java/io/spine/server/entity/storage/ColumnReader.java
```diff
@@ -90,28 +97,31 @@ static ColumnReader forClass(Class<? extends Entity> entityClass) {
...
-            if (isEntityColumn) {
-                EntityColumn column = EntityColumn.from(getter);
-                entityColumns.add(column);
-            }
-        }
-
-        checkRepeatedColumnNames(entityColumns);
-        return entityColumns;
+        PropertyDescriptor[] propertyDescriptors = entityDescriptor.getPropertyDescriptors();
+        ImmutableSet<EntityColumn> annotatedColumns = Arrays
```
dmdashenkov:
```suggestion
        ImmutableSet<EntityColumn> columns = Arrays
```

Issue #921, opened 2018-12-17
Events.getActor may return empty actor


yuri-sergiichuk:
Current functionality assumes that `ActorContext` is filled in the nearest to the event `CommandContext` and for all events emitted as a result of the command processing it is true. But, if an event was emitted as a result of the `@React`ion to another event, the nearest `CommandContext` of the current `EventContext` will be empty, because the `EventContext` `OriginCase` is equal to `EVENT_CONTEXT` is such case.

Maybe it worth changing implementation in a way, that will allow obtaining actor from the nested command contexts if needed, e.g. smth like that may work:
```
public static UserId getActor(EventContext context) {
        EventContext nestedContext = getNestedContext(context);
        CommandContext commandContext = nestedContext.getCommandContext();
        return commandContext.getActorContext()
                             .getActor();
    }

private static EventContext getNestedContext(EventContext context) {
        EventContext currentContext = context;
        while (currentContext.getOriginCase() == EventContext.OriginCase.EVENT_CONTEXT) {
            currentContext = currentContext.getEventContext();
        }
        return currentContext;
    }
```

armiol:
We have to use `Events.findCommandContext()` in `Events.getActor()` implementation. And tweak `getActor` documentation accordingly — as currently it tells nothing about the `@React` use-case.

armiol:
Also, we should check if there is a case where an empty actor is returned. I would return `Optional.empty()` if such a scenario is possible.

dmitrykuzmin:
Resolved in #922.

PR #922, opened 2018-12-18
Process nested contexts in `Events.getActor`

dmitrykuzmin:
This PR fixes issue #921.

Previously, `Events.getActor(EventContext)` worked properly only for events generated from commands, as it did not traverse any nested contexts of the event.

Now, we analyze the origin of the event to retrieve any nested context and thus actor ID.

Also, the events without origin set are not allowed as an input to this method. Such events are not valid according to our [model](https://github.com/SpineEventEngine/core-java/blob/master/core/src/main/proto/spine/core/event.proto#L83) and the only way to create them is to use non-validating `Builder`. Some tests where this was overlooked had to be repaired.


dmitrykuzmin:
@armiol PTAL.

Issue #592, opened 2017-09-29
MultitenantStorage Java 8 improvement


dmitry-ganzha:
At the moment `MultitenantStorage.getStorage()` looks like this:
```java
S getStorage() {
        final TenantFunction<S> func = new TenantFunction<S>(isMultitenant()) {
            @Nullable
            @Override
            public S apply(@Nullable TenantId tenantId) {
                checkNotNull(tenantId);
                lock.lock();
                try {
                    S storage = tenantSlices.get(tenantId);
                    if (storage == null) {
                        storage = createSlice();
                        tenantSlices.put(tenantId, storage);
                    }
                    return storage;
                } finally {
                    lock.unlock();
                }
            }
        };
        final S result = func.execute();
        return result;
    }
```

When the Spine is migrated to Java 8, it will be necessary to refactor the lock mechanics to `Map.putIfAbsent()`.

PR #924, opened 2018-12-21
Use Java 8 feature to check whether a value is absent in the `MultitenantStorage`

serhii-lekariev:
This PR closes https://github.com/SpineEventEngine/core-java/issues/592.

As opposed to manually checking whether a value is present by calling:
```java
if(tenantSlices.get(tenantId) == null) {
...
}
```
A `Map.computeIfAbsent()` method introduced in Java 8 does it in 1 nice `tenantSlices.computeIfAbsent(tenantId, ()->...)` call.


serhii-lekariev:
@armiol, PTAL.

PR #926, opened 2018-12-21
Bump Base

alexander-yevsyukov:
This PR advances dependency on Base to newest snapshot version, which removed validating builders for standard Protobuf classes, and introduced a number of other changes. See #252 for details.

The main work does in this PR is migrating to specific entity types (instead of standard ones that can no longer be used because of absence of validating builders).

Other notable changes:
  * Affected test environment classes were pushed to package level.
  * Error Prone warnings `BadImport` and `EqualsGetClass` were addressed.
  * Naming for exception factories was simplified.


alexander-yevsyukov:
@armiol, it builds in local environment. Once you review this PR, let me know and I'll merge Base first to make it the build of this PR pass. PTAL.

PR #931, opened 2019-01-03
Fix model verifier tests

alexander-yevsyukov:
This PR fixes model verifier test project to use new approach for working with dependencies.


alexander-yevsyukov:
@armiol, PTAL. 

It appears that this PR does not need newer version of `base` (as I wrote in https://github.com/SpineEventEngine/base/pull/272).

PR #933, opened 2019-01-04
Clarify the API of `...Endpoint`s

armiol:
`EntityMessageEndpoint` had numerous `dispatchNow`, `dispatchTo`, `deliverNow` etc methods that serve different purposes, but are named in very similar manner.

This changeset is devoted to renaming and/or removing these methods in accordance with their purpose. 

Also, the internal structure of repositories, which communicate with endpoints is slightly modified.

This is the second PR in a chain of requests leading to `Inbox` introduction. In the following requests the hierarchy of `EntityMessageEndpoint` descendants will be modified even further to work with `Inbox`-provided routines.


armiol:
@dmdashenkov PTAL.

PR #935, opened 2019-01-07
Set once adjustments

serhii-lekariev:
This PR adjusts the current codebase to the changes brought by https://github.com/SpineEventEngine/base/pull/275.

## `MirrorProjection`

`MirrorProjection`'s state message - `Mirror`, now has an explicit `(set_once) = false` applied to its identifier.
This is the case because the creation of a `Mirror` can happen in one of the multiple ways, and thus, its identifier has to be updated in every case, sometimes overlapping.

## `EntityHistoryAggregate`

`EntityHistoryAggregate`'s state message -`EntityHistory`, is now also explicitly `(set_once) = false`.

### Test entities

Some entities used for testing underwent the same change - had their `(set_once)` options set to be `false`.

### DIFF
--- a/server/src/main/proto/spine/system/server/entity_history.proto
+++ b/server/src/main/proto/spine/system/server/entity_history.proto
@@ -44,7 +44,7 @@ import "spine/server/entity/entity.proto";
 message EntityHistoryId {
 
     // The ID of the entity.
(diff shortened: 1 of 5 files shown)


serhii-lekariev:
@dmdashenkov, PTAL.

PR #939, opened 2019-01-10
`PmSubject` and forceful entity search

alexander-yevsyukov:
This PR:
  1. extends API of the `RecordBasedRepository` with an ability to find deleted or archived entities.
  2. Removes the `RepositoryView` interface. So far we have not see any need for a read-only interface to a repository. The interface is redundant.
  3. Introduces `PmSubject` (Truth-based) which can be obtained from `BlackBoxBoundedContext` by the class of the Process Manager and its ID.

The abstract base for entity subjects is provided for providing `Subject`s for other types of entities.


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -162,16 +162,19 @@ public void store(Collection<E> entities) {
...
+        Optional<E> result = record.filter(WithLifecycle::isActive)
+                                   .map(this::toEntity);
+        return result;
+    }
+
+    /**
+     * Finds an entity with the passed ID even if the entity is not
+     * {@linkplain WithLifecycle#isActive() active}.
+     */
+    public Optional<E> doFind(I id) {
```
armiol: I would name this differently. An option coming to my mind is `findRaw`, meaning that no filtering is applied, as opposed to `find`, which filters by `archive == false && deleted == false`.

PR #942, opened 2019-01-11
Update gradle wrapper version

serhii-lekariev:
This PR updates version of the Gradle wrapper to `5.1.1`, along with a respective PR in [base](https://github.com/SpineEventEngine/base/pull/282) and [config](https://github.com/SpineEventEngine/config) repositories.


serhii-lekariev:
@dmdashenkov, PTAL.
The build fails because of the old jacoco script taken from `config`, once that is merged, I believe that the build passes.

PR #944, opened 2019-01-14
Update base

### LINKED ISSUE #551: Use `@OverridingMethodsMustInvokeSuper` instead of warning suppression
We have several warning suppressions like this

```java
@SuppressWarnings("MethodDoesntCallSuperMethod")
```
we should use `@OverridingMethodsMustInvokeSuper` and enforcing the check.

dmdashenkov:
This PR prepares `core-java` modules for `1.0.0-pre3` release.

In this PR we update `base`, `time`, and `config` dependencies.

Also, in this PR we declare Spine-specific Maven repositories so that only the Spine modules are looked up in those repositories.

Also, this PR removes some unnecessary warnings and fixes #551.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL.

PR #947, opened 2019-01-15
Fix error message in `QueryService`

### LINKED ISSUE #687: Improve Handling of Not-Registered Projection Repositories
Currently, if a projection repository wasn't registered within the `BoundedContext` and one tries to use `QueryService` in order to find the projection only a `NullPointerException` is thrown from the `QueryService`:
```java
java.lang.NullPointerException
    at io.spine.server.QueryService.read(QueryService.java:66)
```

Here is the code that throws the NPE:
```java
@SuppressWarnings("MethodDoesntCallSuperMethod")
// as we override default implementation with `unimplemented` status.
@Override
public void read(Query query, StreamObserver<QueryResponse> responseObserver) {
    log().debug("Incoming query: {}", query);

    final TypeUrl type = Queries.typeOf(query);
    final BoundedContext boundedContext = typeToContextMap.get(type);
    final Stand stand = boundedContext.getStand();
    try {
        stand.execute(query, responseObserver);
    } catch (@SuppressWarnings("OverlyBroadCatchBlock") Exception e) {
        log().error("Error processing query", e);
        responseObserver.onError(e);
    }
}
```

The problem here is that `typeToContextMap` has no `BoundedContext` for the given `TypeUrl` if corresponded `ProjectionRepository` wasn't registered.

I believe it would be better to throw some kind of `IllegalStateException` with appropriate exception details instead of having the NPE.

dmdashenkov:
In this PR we make the unknown entity type query error handling adequate according to the gRPC service contract expectations.

Before this PR, if the queried entity type was unknown, `QueryService.read` threw NPE.
Now, the method reports the error to the response observer.

Fixes #687.


dmdashenkov:
@armiol, PTAL.

PR #955, opened 2019-01-22
Subscriptions based on the `EntityStateChanged` event

dmitrykuzmin:
Until now subscriptions relied on the Aggregate and Projection repositories posting their changes to the Stand directly.

With this PR the Stand will subscribe to `EntityStateChanged` events and notify matching subscriptions when it receives a new one.

Posting directly to Stand has the same effect as before (via the `Stand.post(...)` method), but is now deprecated.

This PR is the first step required to make event subscriptions work. The API is unchanged for now except the entity filters which were transformed to the general-purpose filters (which can apply to both event message and entity state). See [filters.proto](https://github.com/SpineEventEngine/core-java/blob/351493c99584574276ffa2651220b1786223b59c/client/src/main/proto/spine/client/filters.proto) for details.


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/stand/SubscriptionMatcher.java
```diff
@@ -0,0 +1,170 @@
...
+
+import static io.spine.protobuf.FieldPaths.fieldAt;
+import static io.spine.server.storage.OperatorEvaluator.eval;
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+import static java.lang.String.join;
+
+/**
+ * Matches the incoming event against a subscription criteria.
+ *
+ * <p>The class is abstract because event and entity subscriptions match through the different
```
armiol: To me this paragraph should say that descendants decide on how to turn the incoming event envelope into a set of data pieces suitable for filtering.

And not that "this class is abstract because etc".

dmitrykuzmin:
@armiol PTAL.

PR #956, opened 2019-01-22
More enrichment code improvements

alexander-yevsyukov:
This PR introduces a number of refactorings in enrichment-related code.

Decision on the enrichment is isolated in the `Enricher`. `EventBus` no longer asks `Enricher` if it can enrich or not. `Enricher` may return the same instance if it cannot be enriched.

Working with `Multimap`s is encapsulated in corresponding classes. See `Schema` and `EnrichmentMap`.

The order of generic parameters in `EnrichmentFunction`s is sorted to have message context type after the message type. It used to be the 3rd after the target message type, which was confusing.

Method objects were extracted to minimize the code related to passing parameters.

Also inactive `EnrichmentFunctions` are eliminated when `Enricher` is created, to avoid repeated filtering.


alexander-yevsyukov:
@armiol, PTAL.

The build failure was probably because of somehow cached snapshot dependency.

alexander-yevsyukov:
@dmdashenkov, PTAL.

server/src/main/java/io/spine/server/event/enrich/MessageEnrichment.java
```diff
@@ -104,74 +81,111 @@ public T apply(S eventMsg, C context) {
...
-                    fieldFunctions.get(sourceFieldClass);
-            Collection<FieldDescriptor> targetFields = fieldMap.get(srcField);
+    /**
+     * Method object for {@link MessageEnrichment#apply(Message, Message)}.
+     *
+     * @param <S> the type of the source message
+     * @param <T> the type of the target message
+     * @param <C> the type of the source message context
+     */
+    private static final class Helper<S extends Message, C extends Message, T extends Message> {
```
dmdashenkov: A message context should always implement `MessageContext` interface. Please use it instead of plain `Message`.

server/src/test/java/io/spine/server/event/enrich/EnrichmentAssertion.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.common.collect.ImmutableSet;
+import com.google.protobuf.Message;
+import io.spine.type.TypeName;
+
+import static com.google.common.truth.Truth.assertThat;
+
+/**
+ * Asserts that the passed enrichment class enriches message classes.
+ */
+final class EnrichmentAssertion {
```
dmdashenkov: Wouldn't it be easier if we make this class extend `Subject`?

server/src/main/java/io/spine/server/event/enrich/EnrichmentFileSet.java
```diff
@@ -0,0 +1,201 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Sets.newHashSet;
+import static io.spine.io.PropertyFiles.loadAllProperties;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Loads enrichment information from resource files named
+ * {@link Resources#ENRICHMENTS enrichments.properties}.
+ */
+final class EnrichmentFileSet {
```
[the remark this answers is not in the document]
alexander-yevsyukov: As discussed vocally, let's keep the name since we don't have clearer better alternatives. And it is a set or _property_ files placed in resources (in a mathematical sense of the word).

server/src/test/java/io/spine/server/event/enrich/EnrichmentAssertion.java
alexander-yevsyukov: As discussed vocally, there is no much value in having `EnrichmentSubject` in `testutil-server` as end users are not going to test enrichments directly. Having it just inside the framework testing package looks like over-engineering.

alexander-yevsyukov:
@dmdashenkov, comments addressed. PTAL again.

server/src/main/java/io/spine/server/event/enrich/EnrichmentFileSet.java
```diff
@@ -53,9 +54,8 @@
     private static final Pattern pipeSeparatorPattern = Pattern.compile("\\|");
 
     /**
-     * Constant indicating a package qualifier.
-     *
-     * <p>Must be a postfix of the qualifier.
+     * Constant indicating a package qualifier. If this string is used as a suffix
```
dmdashenkov: Please add a paragraph after `.`.

PR #957, opened 2019-01-23
Enrichment improvements

alexander-yevsyukov:
This PR introduces incremental improvements in the code related to enrichments.

Notable changes:
  * Now it is possible to have a field enrichment function which has the same source and target types. For example, there could be a function which obtains a `person_name` (as `string`) from a `user_id` (which is also `string`).
  * `ValidationResult` class was renamed to `FieldTransitions`.
  * `RefernceValidator` class was renamed to `Linker`. It provides field transitions required for creating an enrichment type from a source message type.

This PR requires API introduced in [this PR from `base`](https://github.com/SpineEventEngine/base/pull/303).


server/src/main/java/io/spine/server/event/enrich/Linker.java
```diff
@@ -0,0 +1,247 @@
...
+    private Set<FieldDescriptor>
+    findSourceFields(ImmutableList<FieldRef> references, FieldDescriptor enrichmentField) {
+        int refCount = references.size();
+        checkArgument(refCount > 0, "References may not be empty");
+        Set<FieldDescriptor> result = new HashSet<>(refCount);
+
+        FieldDescriptor.Type basicType = null;
+        Descriptor messageType = null;
+        for (FieldRef ref : references) {
+            boolean strict = refCount == 1;
```
yuri-sergiichuk: I guess `strict` should be taken out of the `for` cycle. It is not modified anywhere and the `refCount` is also a constant.

server/src/main/java/io/spine/server/event/enrich/Linker.java
```diff
@@ -0,0 +1,247 @@
...
+                continue;
+            }
+
+            if (basicType == null) { // Get type of the first field
+                basicType = field.getType();
+                if (basicType == MESSAGE) {
+                    messageType = field.getMessageType();
+                }
+            } else { // Compare the type with each of the next
+                checkState(basicType == field.getType(),
```
yuri-sergiichuk: Is there a test for such a situation?

PR #961, opened 2019-02-04
Event subscriptions

dmitrykuzmin:
This PR enables the subscriptions to events.

Event subscriptions are created using the same `Topic` message as entity subscriptions, specifying event type URL and filters for the event message fields.

Once such subscription is activated, the client will start receiving all matching events via the `SubscriptionUpdate`. The `SubscriptionUpdate` now consists of 2 possible update values as follows:
```proto
message SubscriptionUpdate {
...
    oneof update {

       // The entity state updates if it's an entity subscription.
       EntityUpdates entity_updates = 12;

        // The new events occurred in the system if it's an event subscription.
        EventUpdates event_updates = 13;
    }
}

message EntityUpdates {

    // The entity state update list.
    repeated EntityStateUpdate updates = 1 [(valid) = true];
}

message EventUpdates {

    // The event list.
    repeated core.Event events = 1 [(valid) = true];
}
```

Manually subscribing to system events (like `EntityStateChanged`) is forbidden - only domain events are exposed by the `Stand`.


dmitrykuzmin:
@armiol PTAL. The build will pass once https://github.com/SpineEventEngine/base/pull/313 is merged.

server/src/main/java/io/spine/server/command/model/CommandHandlingClass.java
```diff
@@ -21,21 +21,31 @@
...
 
     /**
      * Obtains classes of commands handled by the class.
      */
     Set<CommandClass> getCommands();
 
+    /**
+     * Obtains classes of all messages produced as a result of command handle.
+     */
+    Set<P> getHandleProducts();
```
armiol: I'd like to avoid a `product` term. Let's rename this (and similar) occurrences to `getCommandOutput()` (similar case would be `getReactionOutput()` etc).

server/src/main/java/io/spine/server/command/model/CommandHandlingClass.java
```diff
@@ -21,21 +21,31 @@
...
+public interface CommandHandlingClass<P extends MessageClass<?>,
+                                      H extends CommandAcceptingMethod<?, P, ?>> {
 
     /**
      * Obtains classes of commands handled by the class.
      */
     Set<CommandClass> getCommands();
 
+    /**
+     * Obtains classes of all messages produced as a result of command handle.
```
armiol: `of command handling.`

PR #968, opened 2019-02-10
Enable PMD static code analysis during the build

armiol:
This PR enables and configures PMD code analysis at  build-time.

Some of the discovered issues were fixes, others were suppressed with an explanation.


armiol:
@alexander-yevsyukov PTAL.

Issue #860, opened 2018-10-06
Reduce console output during build


alexander-yevsyukov:
There are way too many stacktraces for exceptions thrown during testing edge cases. Similarly to recent build polishing in `base`, we should make cleaner console output in `core-java`.

alexander-yevsyukov:
@armiol, please see who can address this request soon.

dmitrykuzmin:
Resolved via #969.

PR #969, opened 2019-02-13
Reduce console output during build

dmitrykuzmin:
This PR addresses #860.

The build output was polished to contain no excessive output from tests that are designed to throw an error.

Also, multiple model verification warnings were either fixed or suppressed with `MuteLogging` where appropriate.


dmitrykuzmin:
@armiol PTAL.

model/model-verifier/src/test/java/io/spine/model/verify/ModelVerifierTest.java
```diff
@@ -102,6 +102,7 @@ void setUp() {
     }
 
     @Test
+    @MuteLogging
```
armiol: Not sure why logging is muted for this test.

server/src/test/java/io/spine/server/SubscriptionServiceTest.java
```diff
@@ -206,6 +206,7 @@ private void checkSubscribesTo(Class<? extends Message> aClass) {
     }
 
     @Test
+    @MuteLogging
     @DisplayName("not subscribe to a system event")
     void notSubscribeToSystemEvent() {
```
armiol: This test also does not seem to throw "expected" exceptions, so I don't know why mute.

server/src/test/java/io/spine/server/aggregate/model/EventApplierTest.java
```diff
@@ -121,6 +122,7 @@ void hasOneMessageParameter() {
         }
 
         @Test
+        @MuteLogging
```
armiol: Same as above.

server/src/test/java/io/spine/server/command/model/CommandHandlerMethodTest.java
```diff
@@ -106,6 +107,7 @@ void passNullToleranceCheck() {
     }
 
     @Nested
+    @MuteLogging
```
armiol: Same thing.

server/src/test/java/io/spine/server/command/model/CommandHandlerMethodTest.java
```diff
@@ -201,6 +203,7 @@ void allowEmptyInProcman() {
...
```
armiol: And once again.

testutil-server/src/test/java/io/spine/testing/client/grpc/TestClientTest.java
```diff
@@ -44,6 +45,7 @@
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+@MuteLogging
```
armiol: Not sure again. Is this due to a shutdown? Let's comment on muting in this case.

dmitrykuzmin:
@armiol PTAL. I fixed/clarified the usage of `MuteLogging` across the changeset.

server/src/main/java/io/spine/server/SubscriptionService.java
```diff
@@ -120,6 +121,9 @@ private BoundedContext selectBoundedContext(Subscription subscription) {
     private BoundedContext selectBoundedContext(Target target) {
         TypeUrl type = TypeUrl.parse(target.getType());
         BoundedContext result = typeToContextMap.get(type);
+        if (result == null) {
+            throw newIllegalArgumentException("Trying to subscribe to an unknown type: %s", type);
```
armiol: This method is going to be called from `cancel` via `selectBoundedContext(Subscription subscription)`. So the text is misleading.

Also, I don't think that's a good idea to throw an exception. I'd rather return `Optional.empty()`. And if called from `subscribe` I would throw an exception (and return a respective error via gRPC observer's `onError`). But if called from `cancel` I would emit a warning, as such a use case isn't a show-stopper.

server/src/test/java/io/spine/server/command/model/CommandHandlerMethodTest.java
```diff
@@ -216,6 +216,9 @@ void messageParam() {
         }
 
         @Test
+        @MuteLogging
```
armiol: We rarely use single-line comments.

I would rewrite this one (and the one above) like that:

```java
@MuteLogging  /** Signature mismatch warnings are expected. */
```

testutil-server/src/test/java/io/spine/testing/client/grpc/TestClientTest.java
```diff
@@ -45,7 +45,7 @@
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-@MuteLogging
+@MuteLogging // Mute technical logs about starting the server, etc.
```
armiol: Let's not mute any logs on starting the server. They are NOT exceptions.

dmitrykuzmin:
@armiol PTAL again.

PR #970, opened 2019-02-19
Eliminate usages of `Options` utility.

serhii-lekariev:
This PR is a follow-up to https://github.com/SpineEventEngine/base/pull/333.

Though `Options` utility is removed from `base`, `core-java` still uses it.

This PR eliminates said usages in favor of using `Option` subclasses, namely `EntityOption`.


serhii-lekariev:
@alexander-yevsyukov, after https://github.com/SpineEventEngine/base/pull/333 is merged, the build should pass. Meanwhile, let's discuss the naming of the `EntityOption`.

PR #973, opened 2019-02-20
Structure and naming improvements

alexander-yevsyukov:
This PR introduces the following changes.

### Naming changes
  * Accessor methods were renamed removing the `get` prefix.
    We still have `get` in the generated code, but let's migrate to more humanistic naming convention where and when we can. 

### Structure changes
  * Envelope types are moved under the `server.type` package. This change is in-line with this [type improvements in `base`](https://github.com/SpineEventEngine/base/pull/340).
  * Factory methods of `TestActorRequestFactory` are deprected in favor of constructors. We may need  classes that descend from `TestActorRequestFactory`. Static factories would hinder doing this.
  * `Event` and `Command` now implement `MessageWithContext` interface.

### Enrichment changes
It is expected that we are going to support command enrichments sometime after v1.0 is out. Correspondingly, the following changes were made in the enrichment-related code:
  * The package moved under `server` to be `io.spine.server.enrich`.
     Previously, it was under `server.event` (`io.spine.server.event.enrich`).
  * The API was updated to take into account recently introduced `EnrichableMessageContext`. Now only `EventContext` implements this interface. In order to support command enrichment, `CommandContext` code generation should be adjusted to implement the interface too.

This PR depends on [API changes in `base`](https://github.com/SpineEventEngine/base/pull/340).


alexander-yevsyukov:
@armiol, PTAL. Apologies on the number of changes. Most of them are due to renamings.

core/src/main/java/io/spine/core/MessageWithContext.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.protobuf.Message;
+
+/**
+ * Base interfaces for outer objects of messages with contexts.
+ */
+public interface MessageWithContext extends Message {
+
+    Message getId();
+
+    Any getMessage();
```
armiol: Can you please document the methods.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -84,19 +84,20 @@ static Timestamp nullToDefault(@Nullable Timestamp timestamp) {
     }
 
     /** Obtains {@link EventStore} from which to get events during catch-up. */
-    EventStore getEventStore() {
-        return getBoundedContext().getEventBus()
-                                  .getEventStore();
+    EventStore eventStore() {
+        return this.boundedContext()
```
armiol: Not sure why we need `this.`. We don't refer to the instance methods anywhere else like this.

Issue #920, opened 2018-12-17
Allow deleting/archiving process managers upon a rejection or an event


mdrachuk:
This issue opens up a topic of "terminal" events for processes.

The vast majority of the processes have a distinct end-of-life marker. In most cases that's an event (or several events) emitted by the process manager. Typically the framework users would have to write a code to mark the process manager deleted before emitting the terminal events. It is quite possible right now, it just some boilerplate to keep in mind.

However a process may also end emitting a rejection. In this case an engineer is not able to change neither the entity state nor its lifecycle, as they are automatically reverted after an exception — which is a rejection by its nature.

We may automate this.

An event or a rejection declared as terminal signalizes the framework this process manager can be safely deleted or archived. In this case the framework marks the process manager correspondingly in scope of the same transaction. 

I would imagine the process manager declaration as follows.
```
message SignInProcess {
    option (entity).kind = PROCESS_MANAGER;

    // The process is deleted in case `UserSignedIn` is emitted or `CannotAuthenticateUser` is thrown.
    option (lifecycle).delete_upon = "UserSignedIn, CannotAuthenticateUser";

    // Another option is to archive automatically:
    // option (lifecycle).archive_upon = "..."

    // Here goes the process state description.
}
```

dmitrykuzmin:
In the end, we took the other approach and abandoned the idea of having options that describe entity behavior. 

Automatic entity archivation/deletion via repository was implemented in #974.

PR #974, opened 2019-02-22
Archive/delete process manager upon certain events or rejections

dmitrykuzmin:
This PR addresses #920.

It adds the configurable lifecycle rules to the `ProcessManagerRepository`.

The rules can be configured as follows:
```java
repository.lifecycle()
          .archiveOn(Event1.class, Rejection1.class)
          .deleteOn(Rejection2.class); 
```
Now, all managed entities will be automatically archived upon emitting `Event1` or `Rejection1` and will be automatically deleted on producing `Rejection2`.

The archivation/deletion takes place in scope of the same transaction where events were emitted.


dmitrykuzmin:
@armiol PTAL. The build will pass once https://github.com/SpineEventEngine/base/pull/339 is merged.

dmitrykuzmin:
@armiol PTAL again.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -257,6 +258,16 @@ public static RejectionEventContext rejectionContext(CommandMessage commandMessa
                                     .build();
     }
 
+    /**
+     * Obtains the type URL of the event message enclosed by a given {@code Event}.
+     */
+    public static TypeUrl typeUrl(Event event) {
```
armiol: This is weird. I think it belongs to a message interface, such as `EventMessage`.

Also I suspect this isn't the first time we want to find out the `TypeUrl` of an `Event`. So please have a look around, maybe this task is already solved somehow.

server/src/main/java/io/spine/server/procman/Lifecycle.java
```diff
@@ -0,0 +1,140 @@
...
+     */
+    @SafeVarargs
+    @CanIgnoreReturnValue
+    public final Lifecycle archiveOn(Class<? extends EventMessage>... messageClasses) {
+        archiveOn.addAll(asList(messageClasses));
+        return this;
+    }
+
+    /**
+     * Checks if the process hould become archived when the given {@code events} are emitted.
```
armiol: `should`.

dmitrykuzmin:
@armiol PTAL.

PR #975, opened 2019-02-27
Update base

dmdashenkov:
In this PR we update `base`.

Changes:
 - `(events)` and `(rejections)` options are deleted;
 - generated checks of `(set_once)` in `VBuilder`s are changed, thus the builders must be regenerated.

Also, in this PR, some Javadoc warnings are fixed and the `config` module is updated.

Also, this PR releases version `1.0.0-pre6`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,10 +8,15 @@
 *.html text
 *.xml text
 *.css text
+*.scss text
 *.js text
 *.properties text
-*.bat text
 *.rtf text
+*.yaml text
+*.yml text
+*.md text
+
+LICENSE text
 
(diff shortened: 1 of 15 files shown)


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@dmitrykuzmin, FYI. `protoc` does not tolerate usage of non-existing options.

dmdashenkov:
@armiol, this change is to be done to many Spine modules. I suggest calling this `1.0.0-pre6` in order to preserve the change integrity and propagate it to the framework users more easily.

PR #976, opened 2019-02-27
Abstract event reactor

serhii-lekariev:
This PR introduces `AbstractEventReactor`.

### `AbstractEventReactor`

An abstract base for all classes that react on events with other events.

We currently have two abstract bases for event-reacting classes: `ProcessManager` and `Aggregate`, both of which are inappropriate for simple cases becase:
- Aggregates often map to entities that exist in a domain, and using one for just reacting on events is not only overkill but also misleading.
- Process managers imply some state of a multi-state operation. Simple event reactors might not want to have state at all.

To use a simple event reactor, one would need to:
1) Extend from `AbstractEventReactor`. Its constructor takes an `EventBus` instance, which is the event bus to which emitted events are posted.
2) Register in a desired `BoundedContext`, by using `BoundedContext#registerEventDispatcher`, which takes care of both domestic and external subscriptions.

Also in this PR:
- Change access level of some `BlackBoxBoundedContext` methods to `public` to ease testing;
- Renaming some accessor methods from `getAttribute()` to just `attribute()`,


serhii-lekariev:
@armiol, PTAL.

serhii-lekariev:
@armiol, PTAL.

server/src/test/proto/spine/test/event/events.proto
```diff
@@ -97,3 +99,105 @@ message EBTaskAdded {
 message ProjectFinished {
     ProjectId project_id = 1;
 }
+
```
armiol: Let's extract into a separate file. And let's have a file-level comment on the purpose of those events.

serhii-lekariev:
@armiol, PTAL.

server/src/test/java/io/spine/server/event/given/AbstractReactorTestEnv.java
```diff
@@ -0,0 +1,286 @@
...
+
+        public FaultyNotifier(EventBus eventBus) {
+            super(eventBus);
+        }
+
+        @SuppressWarnings("NewExceptionWithoutArguments") /* Does not matter for this test case. */
+        @React
+        Pair<CustomerNotified, DeliveryServiceNotified>
+        notifyAboutOrder(OrderReadyToBeServed orderReady) {
+            throw new RuntimeException();
```
armiol: Can't you just add a message to the exception and remove the suppression?

Just in case, I don't understand why you throw the exception. Let's discuss.

server/src/test/java/io/spine/server/event/given/AbstractReactorTestEnv.java
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        @SuppressWarnings("NewExceptionWithoutArguments") /* Does not matter for this test case. */
+        @React(external = true)
+        DonationMade makeDonation(OrderPaidFor orderPaidFor) {
+            throw new RuntimeException();
+        }
+    }
+}
+
```
armiol: Please kill the empty line.

server/src/test/java/io/spine/server/event/given/AbstractReactorTestEnv.java
```diff
@@ -0,0 +1,286 @@
...
+            DeliveryServiceNotified result = DeliveryServiceNotified
+                    .newBuilder()
+                    .setMessage(format(messageFormat, order.getOrderId()))
+                    .setOrder(order)
+                    .build();
+            return result;
+        }
+    }
+
+    /** Obtains an event that signgifes that some order is ready to be served. */
```
armiol: `signgifes` has problems.

server/src/test/java/io/spine/server/event/AbstractEventReactorTest.java
```diff
@@ -0,0 +1,246 @@
...
+import static io.spine.testing.client.blackbox.Count.twice;
+import static io.spine.testing.server.blackbox.VerifyEvents.emittedEvent;
+import static junit.framework.TestCase.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.slf4j.event.Level.ERROR;
+
+@DisplayName("Abstract event reactor should")
+class AbstractEventReactorTest {
+
+    private BlackBoxBoundedContext<SingleTenantBlackBoxContext> restaurantContext;
```
armiol: Please use `SingleTenantBlackBoxContext` instead of this monster.

server/src/test/java/io/spine/server/event/AbstractEventReactorTest.java
```diff
@@ -0,0 +1,246 @@
...
+        @DisplayName("receive several")
+        void receiveSeveral() {
+            ImmutableList<OrderServed> eventsToEmit = ImmutableList.of(
+                    someOrderServedInTime(), someOrderServedInTime(), someOrderServedLate()
+            );
+            eventsToEmit.forEach(restaurantContext::receivesEvent);
+            ImmutableList<OrderServed> ordersServed = performanceTracker.ordersServed();
+            ImmutableList<OrderServedLate> ordersServedLate = performanceTracker.ordersServedLate();
+            assertThat(ordersServed).hasSize(eventsToEmit.size());
+            @SuppressWarnings("Guava") /* provides a prettier API. */
```
armiol: Let's not do that and use JDK API.

server/src/test/java/io/spine/server/event/AbstractEventReactorTest.java
```diff
@@ -0,0 +1,246 @@
...
+
+        private boolean isServedLate(OrderServed orderServed) {
+            Order order = orderServed.getOrder();
+            Timestamp placedOn = order.getTimePlaced();
+            Timestamp servedOn = orderServed.getServedOn();
+
+            Duration difference = Timestamps.between(placedOn, servedOn);
+            long differenceInMins = Durations2.toMinutes(difference);
+            return differenceInMins > 50;
+
```
armiol: Please review this request for an excessive empty lines.

server/src/test/java/io/spine/server/event/AbstractEventReactorTest.java
```diff
@@ -0,0 +1,246 @@
...
+            @SuppressWarnings("Guava") /* provides a prettier API. */
+                    long ordersInTime = ordersServed
+                    .stream()
+                    .filter(not(this::isServedLate))
+                    .count();
+            assertThat(ordersInTime).isEqualTo(2);
+            assertThat(ordersServedLate).hasSize(1);
+        }
+
+        private boolean isServedLate(OrderServed orderServed) {
```
armiol: Let's define the logic in a single place.

server/src/test/java/io/spine/server/event/AbstractEventReactorTest.java
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Makes sure that the error has been correctly logged.
+     *
+     * <p>Checks that:
+     * <ul>
+     *     <li>only 1 message has been logged;
+     *     <li>logged message is of the {@code ERROR} level.
+     * </ul>
+     */
+    private static void assertLoggedCorrectly(Queue<SubstituteLoggingEvent> messages) {
```
armiol: `assertSingleError(...)`

serhii-lekariev:
@armiol, PTAL.

PR #977, opened 2019-03-02
Simpler enrichments

alexander-yevsyukov:
This PR finalizes changes towards easier working with enrichments. It depends on API changes from [this PR from `base`](https://github.com/SpineEventEngine/base/pull/349).

### Changes to `Enricher`

`Enricher` is now an abstract class, with `EventEnricher` descending from this class. The change is made to allow easier introduction of `CommandEnricher` should we see such a need in the near future.

`EventEnricher` has the `Builder` which accepts one or more functions producing an enrichment message per enrichable message class or one or more super-interfaces of such a class. When the `EventEnricher` is built all such functions are aggregated per event message class to minimize reflection calls when an actual enrichment is to be made.

### Other Notable changes

  * Now it is possible to get an enrichment message from `EnrichableMessageContext`, if the caller is sure that there is such a message.  Previously, it was needed to `find()` which returned `Optional`.
  * `CommandEnrichment` type is eliminated in favor of getting just `Command` as an enrichment of `CommandScheduled` event.
  * `AbstractEntity.state()` made `final`.


alexander-yevsyukov:
@armiol, @yuri-sergiichuk, PTAL.

server/src/main/java/io/spine/server/bus/UnicastBus.java
```diff
@@ -48,7 +48,7 @@ protected D getDispatcher(E envelope) {
 
     private static IllegalStateException noDispatcherFound(MessageEnvelope envelope) {
         String id = Identifier.toString(envelope.id());
-        String msg = format("No dispatcher found for the command (class: %s id: %s).",
+        String msg = format("No dispatcher found for the command (class: `%s` id: `%s`).",
```
yuri-sergiichuk: I'd personally rather extract `format` string as a separate variable and inline all 3 parameters of the `format()` call:
```
String msg = format(messageFormat, envelope.messageClass(), id);
```

server/src/main/java/io/spine/server/enrich/EnricherBuilder.java
```diff
@@ -20,89 +20,178 @@
...
+public abstract class EnricherBuilder<M extends Message,
+                                      C extends EnrichableMessageContext,
+                                      B extends EnricherBuilder<M, C, B>> {
 
-    /** Functions which perform the enrichment. */
-    private final Set<EnrichmentFunction<?, ?, ?>> functions = Sets.newHashSet();
+    private static final String SUGGEST_REMOVAL = " Please call `remove(Class, Class)` first.";
+
+    /**
+     * Maps a pair of [source class, enrichment class] to a function which produces this enrichment.
```
yuri-sergiichuk: Isn't `{key : value}` the default way to express pairs? AFAIK `[value, value]` is generally understood as a tuple.

server/src/main/java/io/spine/server/enrich/EnricherBuilder.java
```diff
@@ -20,89 +20,178 @@
...
+        functions.remove(new Key(eventClass, enrichmentClass));
+        return self();
     }
 
+    /** Creates a new {@code Enricher}. */
+    public abstract Enricher build();
+
     /**
-     * Ensures that the passed enrichment function is not yet registered in this builder.
+     * Obtains immutable functions of functions added to the builder by the time of the call.
```
yuri-sergiichuk: `immutable map of functions`.

server/src/main/java/io/spine/system/server/SystemEnricher.java
```diff
@@ -47,25 +47,26 @@ private SystemEnricher() {
...
      */
-    public static Enricher create(CommandLifecycleRepository commandRepository) {
-        checkNotNull(commandRepository);
-        Enricher enricher = Enricher
+    public static EventEnricher create(CommandLifecycleRepository repo) {
+        checkNotNull(repo);
+        EventEnricher enricher = EventEnricher
                 .newBuilder()
-                .add(CommandId.class, Command.class, commandLookup(commandRepository))
+                .add(CommandScheduled.class, Command.class,
```
yuri-sergiichuk: Do we really need to place the last parameter on the new line? It looks like it should fit.

PR #978, opened 2019-03-05
Package-private subscriber methods

### LINKED ISSUE #953: Sort out visibility for handler methods
`@Assign`, `@Command`, `@Subscribe`, `@React` -> package-private
`@Applier` -> private

dmdashenkov:
Currently, it is recommended to use `public` access modifier for `@Subscribe` methods.

From now on, it is decided that we use package-private access level for all the handler methods with an exception of `@Apply` methods which still should be `private`.
This way there is less confusion and hesitation for newcomer users on which access modifier to use.
Also, subscriber methods are not supposed to be called from outside the entity context anyway.

Fixes #953.


dmdashenkov:
@armiol, PTAL.

client/src/test/java/io/spine/client/given/TestEntities.java
```diff
@@ -0,0 +1,44 @@
...
+
+public final class TestEntities {
+
+    /**
+     * Prevents the utility class instantiation.
+     */
+    private TestEntities() {
+    }
+
+    public static TestEntityId randomId() {
```
armiol: Please document.

PR #980, opened 2019-03-07
Close System context

### LINKED ISSUE #872: System context is not closed
Currently, when a domain bounded context is `close`-d, its system counterpart stays active.

To fix this, modify `DefaultSystemMonitor.closeSystemContext` to call `close()` on the system context instance.

While sounding quite easy, this operation causes a number of tests to fail silently at runtime without actually failing the test suite.

dmdashenkov:
Currently, System context is not closed upon the domain context `close()`.

This debt was caused by some kind of an error in message delivery. Right now, this error is gone, thus, in this PR, we make System context close when its domain counterpart is closed.

Fixes #872.

### DIFF
--- a/scripts/assemble-proto.gradle
+++ b/scripts/assemble-proto.gradle
@@ -56,7 +56,7 @@ buildscript {
     }
 }
 
-println "Assembling Protobuf definitions in ${project.name}."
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@armiol, PTAL.

PR #981, opened 2019-03-07
Improve naming

alexander-yevsyukov:
This PR:
 * Improves method naming, eliminating `get` prefix in non-generated API.
 * Updates API dependency on `base`, which introduces similar API changes.


server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -123,7 +123,7 @@
         implements EventPlayer, EventReactor {
 
     /**
-     * The count of events stored to the {@linkplain AggregateStorage storage} since the last
+     * The count of events stored to the {@linkplain AggregateStorage storage} sincee last
```
yuri-sergiichuk: `since`

PR #983, opened 2019-03-08
Default entity constructor

alexander-yevsyukov:
This PR introduces default constructor for entities.

Other notable changes:
  * Improved method names in Model classes.
  * Implementation of `AggregatePart` now does not violate the LSP.
  * Simplified naming related to storage converters.


alexander-yevsyukov:
@armiol, PTAL.

This PR depends on [this one from `base`](https://github.com/SpineEventEngine/base/pull/356).

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -142,6 +142,27 @@
     /** A guard for ensuring idempotency of messages dispatched by this aggregate. */
     private IdempotencyGuard idempotencyGuard;
 
+    /**
+     * Creates a new instance.
+     *
+     * @apiNote Constructors of derived classes are likely to have package-private access
+     *         level because of the following reasons:
+     *         <ol>
+     *         <li>These constructors are not public API of an application.
```
armiol: Let's try to indent the list items. Not sure it would read better, but it may.

server/src/main/java/io/spine/server/aggregate/model/AggregateClass.java
```diff
@@ -69,30 +69,36 @@ protected AggregateClass(Class<A> cls) {
...
+     */
     @Override
-    public final Set<EventClass> eventClasses() {
-        return delegate.eventClasses();
+    public final Set<EventClass> incomingEvents() {
+        return delegate.incomingEvents();
     }
 
+    /**
+     * Obtains the set of <em>external</em> event classes on which this aggregate class reacts.
```
armiol: As long as we are using `<b>`, let's continue to do that here as well.

server/src/main/java/io/spine/server/entity/AbstractEntity.java
```diff
@@ -64,8 +68,13 @@
     @LazyInit
     private volatile @MonotonicNonNull EntityClass<?> thisClass;
 
-    /** The ID of the entity. */
-    private final I id;
+    /**
+     * The ID of the entity.
+     *
+     * <p>Assigner either through the {@linkplain #AbstractEntity(Object)} constructor which
```
armiol: `Assigned`

server/src/main/java/io/spine/server/aggregate/model/AggregateClass.java
alexander-yevsyukov: We use bold for simulating headers. Here is the semantic markup with the emphasis on the word. The way it's rendered is up to styles.

PR #985, opened 2019-03-11
Default repositories

alexander-yevsyukov:
This PR allows to create an instance of a repository by its entity class. Previously, we always require creation of a new repository class.

This feature allows to save of creating classes for test environments, and for the cases when no specific routing is needed for the repository.


alexander-yevsyukov:
@armiol, PTAL.

alexander-yevsyukov:
@dmdashenkov, PTAL.

server/src/main/java/io/spine/server/entity/DefaultRecordBasedRepository.java
```diff
@@ -43,19 +45,20 @@
...
     }
 
     @Override
     protected EntityFactory<E> entityFactory() {
-        return this.entityFactory;
+        return entityModelClass().factory();
     }
 
     @Override
     protected StorageConverter<I, E, S> entityConverter() {
```
dmdashenkov: Please consider renaming this method into `storageConverter()`.

server/src/main/java/io/spine/server/entity/DefaultRecordBasedRepository.java
```diff
@@ -43,19 +45,20 @@
...
-        return this.storageConverter;
+        if (storageConverter == null) {
+            EntityClass<E> entityClass = entityModelClass();
+            TypeUrl stateType = entityClass.stateType();
+            storageConverter = DefaultConverter.forAllFields(stateType, entityFactory());
```
dmdashenkov: Let's perform the initialization in `onRegistered()` method.

The only thing that the getter should do is to make sure the converter has been initialized when the method is called.

server/src/main/java/io/spine/server/entity/Repository.java
```diff
@@ -104,6 +104,7 @@ protected Repository() {
      *
      * @param obj an instance to store
      */
+    @SuppressWarnings("AbstractMethodWithMissingImplementations") // work-around IDEA bug.
```
dmdashenkov: Please suppress it with a `noinspection` comment then. It's strange that no static analyzer complains about the redundant suppression.

server/src/main/java/io/spine/server/procman/DefaultProcessManagerRepository.java
```diff
@@ -0,0 +1,30 @@
...
+                                                   P extends ProcessManager<I, S, ?>,
+                                                   S extends Message>
+    extends ProcessManagerRepository<I, P, S> {
+
+    private final ProcessManagerClass<P> modelClass;
+
+    /**
+     * Creates a new repository managing process managers of the passed class.
+     */
+    public DefaultProcessManagerRepository(Class<P> cls) {
```
dmdashenkov: If we delegate the default repository instantiation to the `EntityClass`es, this constructor should accept a `ProcessManagerClass`, not just a `Class`.

server/src/main/java/io/spine/server/aggregate/DefaultAggregatePartRepository.java
```diff
@@ -18,13 +18,35 @@
...
+import io.spine.server.aggregate.model.AggregatePartClass;
+
+import static io.spine.server.aggregate.model.AggregatePartClass.asAggregatePartClass;
 
 /**
- * A repository of {@link CompanyAggregate}.
+ * Default implementation of {@code AggregatePartRepository}.
  */
-public final class CompanyRepository extends AggregateRepository<CompanyId, CompanyAggregate> {
+public final class DefaultAggregatePartRepository<I,
```
dmdashenkov: Please consider making all the default repositories `@Internal`. I don't recognize this as a part of the public API.

server/src/main/java/io/spine/server/procman/DefaultProcessManagerRepository.java
alexander-yevsyukov: As discussed vocally, we deliberately violate OOP in this case in favor of concise syntax for creating default repositories. This way it would be convenient for the user of the framework, and the code reads much better this way.

server/src/main/java/io/spine/server/DefaultRepository.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param cls
+     *         the class of entities
+     * @param <I>
+     *         the type of entity identifiers
+     * @param <E>
+     *         the type of entity
+     * @return new repository instance
+     */
+    @SuppressWarnings("unchecked") // Casts are ensured by class assignability checks.
+    public static <I, E extends Entity<I, ?>> Repository<I, E> of(Class<E> cls) {
```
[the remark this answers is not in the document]
alexander-yevsyukov: As discussed vocally, we deliberately violate OOP in this case in favor of concise syntax for creating default repositories. This way it would be convenient for the user of the framework, and the code reads much better this way.

server/src/main/java/io/spine/server/aggregate/DefaultAggregatePartRepository.java
```diff
@@ -18,13 +18,35 @@
...
 
-import io.spine.server.aggregate.AggregateRepository;
-import io.spine.system.server.CompanyId;
+import io.spine.server.aggregate.model.AggregatePartClass;
+
+import static io.spine.server.aggregate.model.AggregatePartClass.asAggregatePartClass;
 
 /**
- * A repository of {@link CompanyAggregate}.
+ * Default implementation of {@code AggregatePartRepository}.
```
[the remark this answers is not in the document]
alexander-yevsyukov: As discussed, we cannot follow this path, since `Repository` is not an interface. Let's make documentation backlinks to the `DefaultRepository` factory class. Together with the suggested `@Internal` annotations, it should explain the intent well.

server/src/main/java/io/spine/server/entity/DefaultRecordBasedRepository.java
```diff
@@ -53,12 +55,25 @@ protected DefaultRecordBasedRepository() {
...
             TypeUrl stateType = entityClass.stateType();
             storageConverter = DefaultConverter.forAllFields(stateType, entityFactory());
         }
         return storageConverter;
     }
+
+    /**
+     * @inheritDoc
+     *
+     * <p>Performs validation of the entity class, and initializes the storage converter.
```
dmdashenkov: The comma is redundant.

server/src/main/java/io/spine/server/DefaultRepository.java
```diff
@@ -0,0 +1,82 @@
...
+            return (Repository<I, E>) new DefaultAggregateRepository(cls);
+        }
+        if (ProcessManager.class.isAssignableFrom(cls)) {
+            return (Repository<I, E>) new DefaultProcessManagerRepository(cls);
+        }
+        if (Projection.class.isAssignableFrom(cls)) {
+            return (Repository<I, E>) new DefaultProjectionRepository(cls);
+        }
+        throw newIllegalArgumentException(
+                "No default repository implementation available for the class `%s`.", cls
```
dmdashenkov: This would print `class` twice. Please call `cls.getCanonicalName()`.

PR #989, opened 2019-03-13
Drop Identifier#generate usage

yuri-sergiichuk:
In this PR I have dropped usage of the `Identifier#generate` method and replaced it with the new generated `UuidValue` methods introduced in [base#357](https://github.com/SpineEventEngine/base/pull/357).

Now we should be ready to remove it completely.


yuri-sergiichuk:
@alexander-aleksandrov PTAL.

PR #990, opened 2019-03-13
Make publishing configuration order-agnostic

dmdashenkov:
Gradle project configuration process is complex. The details of the process, such as the order of script execution, may differ from an environment to an environment and from a computer to a computer.

In some environments, the script `assemble-proto.gradle` is executed before the `publishing.gradle`. In others, vice versa.

This PR addresses the build crash which occurred when the `maven-publish` plugin was not yet applied to the project when the `assemble-proto.gradle` was run. In this case, the `publishing` configuration block could not be resolved.

### DIFF
--- a/.idea/dictionaries/common.xml
+++ b/.idea/dictionaries/common.xml
@@ -18,7 +18,7 @@
       <w>flushables</w>
       <w>googleapis</w>
       <w>gradle</w>
-      <w>grpc</w>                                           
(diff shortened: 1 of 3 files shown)


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@yuri-sergiichuk, FYI.

Issue #806, opened 2018-08-05
Changes to the `system` package


alexander-yevsyukov:
* Use sub-packages `command` and `event` for generated code. Now all the generated code is in the same package.

* Use nesting when referencing proto types
There's no need to type `spine.core.CommandId` under the `spine.server.system` package.
`core.CommandId` works as well. It's more compact and shows that the type and the current package are inside the framework.

alexander-yevsyukov:
@armiol, let's discuss.

alexander-yevsyukov:
Can we initialize `when` fields from `CommandContext` of corresponding `Mark` commands, instead of taking the current time?

alexander-yevsyukov:
> Can we initialize `when` fields from `CommandContext` of corresponding `Mark` commands...
This is no longer applicable.

alexander-yevsyukov:
Addressed by #995.

PR #995, opened 2019-03-18
System package commands and events

alexander-yevsyukov:
This PR moves System context commands and events into sub-packages (as requested in #806).

It also introduces `EventRoute.none()` method for ignoring messages when routing, also moving `withId()` method from `Repository` to `EventRoute`.


alexander-yevsyukov:
@armiol, PTAL.

PR #997, opened 2019-03-25
Fix mirroring aggregates for with same IDs

### LINKED ISSUE #996: Aggregate parts with common ID override each other for querying
Say, we have `TaskPart` and `TaskLabelsPart` in our project which share a common `TaskId`.

After we send a command to a `TaskPart` with some `ID`, it will be stored in the repository and we can query for it.

If, after that, we send a command to a `TaskLabelsPart` with same `ID`, it will be also stored and available for querying, but, from now on, querying for `TaskPart` will yield 0 results.

The corresponding test to reproduce the issue was added to the `aggregate-parts-query-test` branch. See [`AggregatePartTest.notOverrideOtherParts()`](https://github.com/SpineEventEngine/core-java/blob/aggregate-parts-query-test/server/src/test/java/io/spine/server/aggregate/AggregatePartTest.java#L107).

It's worth noting that both `TaskPart` and `TaskLabelsPart` actually exist in their repositories. The issue most probably is related to `MirrorProjection` overriding its counterpart with the same `TaskId` (see [`MirrorRepository.idFrom(...)`](https://github.com/SpineEventEngine/core-java/blob/master/server/src/main/java/io/spine/system/server/MirrorRepository.java#L111) and related).

dmdashenkov:
Previously, the ID of a `Mirror` projection consisted only of the ID of the mirrored Aggregate, wrapped into `Any`. This caused a bug: if two aggregate instances had the same ID (e.g. parts belonging to one aggregate root), the `Mirror`s of those aggregates would override each other.

This PR adds the entity type URL field to the `MirrorId`. This way, there is no valid scenario in which two aggregates would have the same `Mirror`.

Fixes #996.

Also, this PR performs several test improvements:
 - test model for Aggregate part is cleaned up;
 - test entity state definitions are now marked with the `(entity)` option;
 - most of the tests now ensure that entities have IDs set.


dmdashenkov:
@armiol, PTAL.

PR #998, opened 2019-03-26
Enable Model Compiler defaults

yuri-sergiichuk:
With this PR I have enabled the Model Compiler default configurations and changed `EventContext` validation rules:
1. `origin` oneof field now uses `required_field` syntax as per [base#378](https://github.com/SpineEventEngine/base/issues/378).
1. `version` field is not `required` anymore, but instead should be `valid` if set. Further analysis is required here, see #999 for details.

Also, I've switch non-test Protobuf messages to `vBuilder` and `toVBuilder` usage. As a result, some tests were also modified to create actually valid messages.

Further migration to `vBuilder` usage will be done in a separate PR.

### DIFF
--- a/core/src/main/java/io/spine/core/Events.java
+++ b/core/src/main/java/io/spine/core/Events.java
@@ -88,7 +88,8 @@ public final class Events {
     /**
      * Sorts the given event record list by the event timestamps.
      *
-     * @param events the event record list to sort
+     * @param events
+     *         the event record list to sort
      */
     public static void sort(List<Event> events) {
         checkNotNull(events);
@@ -116,7 +117,8 @@ public final class Events {
     /**
      * Extracts the event message from the passed event.
      *
-     * @param event an event to get message from
+     * @param event
+     *         an event to get message from
      */
     public static EventMessage getMessage(Event event) {
         checkNotNull(event);
@@ -129,7 +131,7 @@ public final class Events {
      * Extract event messages from the passed events.
      */
     public static List<? extends EventMessage> toMessages(List<Event> events) {
-       checkNotNull(events);
+        checkNotNull(events);
         List<EventMessage> result =
                 events.stream()
                       .map(Events::getMessage)
@@ -177,7 +179,8 @@ public final class Events {
     /**
      * Obtains event producer ID from the passed {@code EventContext}.
      *
-     * @param context the event context to to get the event producer ID
+     * @param context
+     *         the event context to to get the event producer ID
      * @return the producer ID
      */
     public static Object getProducer(EventContext context) {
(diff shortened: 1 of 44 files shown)


core/src/main/java/io/spine/core/Events.java
```diff
@@ -393,9 +407,9 @@ protected EventId fromString(String str) {
      * <p>The context is obtained by traversing the events origin for a valid context source.
      * There can be three sources for the actor context:
      * <ol>
-     *     <li>The command context set as the event origin.
-     *     <li>The command context of an event which is an origin of this event.
-     *     <li>The import context if the event is imported to an aggregate.
+     * <li>The command context set as the event origin.
```
dmdashenkov: Please revert this formatting.

server/src/main/java/io/spine/server/entity/Transaction.java
```diff
@@ -380,7 +380,7 @@ private InvalidEntityStateException of(ValidationException exception) {
 
     private S currentBuilderState() {
         @SuppressWarnings("unchecked")  // OK, as `AbstractValidatingBuilder` is the only subclass.
-        AbstractValidatingBuilder<S, ?> abstractBuilder = (AbstractValidatingBuilder<S, ?>) builder;
+                AbstractValidatingBuilder<S, ?> abstractBuilder = (AbstractValidatingBuilder<S, ?>) builder;
```
dmdashenkov: Please revert this formatting.

server/src/main/java/io/spine/server/type/EventEnvelope.java
```diff
@@ -112,11 +112,12 @@ public ActorContext actorContext() {
      *
      * <p>In particular:
      * <ul>
-     *     <li>the root command identifier replicates the one defined in the enclosed event;
-     *     <li>the context of the enclosed event is set as the origin.
+     * <li>the root command identifier replicates the one defined in the enclosed event;
+     * <li>the context of the enclosed event is set as the origin.
```
dmdashenkov: Please revert.

server/src/test/java/io/spine/server/aggregate/AggregateStorageTest.java
```diff
@@ -75,9 +78,9 @@
 
 public abstract class AggregateStorageTest
         extends AbstractStorageTest<ProjectId,
-                                    AggregateHistory,
-                                    AggregateReadRequest<ProjectId>,
-                                    AggregateStorage<ProjectId>> {
+        AggregateHistory,
+        AggregateReadRequest<ProjectId>,
+        AggregateStorage<ProjectId>> {
```
dmdashenkov: Please revert.

PR #1002, opened 2019-03-28
Bus message listeners

alexander-yevsyukov:
This PR adds an ability to add a `Consumer` of envelopes of messages posted to a `Bus`. This is needed to avoid having a `BusFilter` which isn't filtering but simply consuming a posted message.

Other notable changes:
  * Improved method naming avoiding `get` prefixes.
  * Improved obtaining of generated events in `BlackBoxBoundedContext`. Previously, they were loaded from `EventStore`, now they are collected using Bus message listeners.


alexander-yevsyukov:
@armiol, PTAL.

PR #1003, opened 2019-04-05
Entity subject tests

dmdashenkov:
This PR adds unit tests for the `EntitySubject`.

In order to make the class testable, the way it performs checks is changed from 
```
assertThat(actual)
```
to
```
check().about(factory).that(actual)
```

Also, the `doesNotExist()` check method is added to the `EntitySubject`.


dmdashenkov:
@alexander-yevsyukov, PTAL.

testutil-server/src/test/java/io/spine/testing/server/entity/EntitySubjectTestEnv.java
```diff
@@ -0,0 +1,49 @@
...
+
+final class EntitySubjectTestEnv {
+
+    /**
+     * Prevents the utility class instantiation.
+     */
+    private EntitySubjectTestEnv() {
+    }
+
+    static final class ProjectView
```
alexander-yevsyukov: Please put this class into a sub-package under `given` or something.

PR #1004, opened 2019-04-05
Truth-based assertions

alexander-yevsyukov:
This PR:
 1. Eliminates `PmSubject` in favor of `EntitySubject` which can be used for asserting all kinds of entities.
 2. Introduces `CommandSubject` and `EventSubject` for asserting iterables of corresponding type.
 3. Improves gathering messages in `BlackBoxBoundedContext`. Previously, we loaded events from `EventStore`. Now we use `CommandCollector` and `EventCollector` which are `Consumer`s of corresponding messages, sitting on a `Bus`.

Other notable changes:
  * The code migrated to use latest Spine Time API.
  * Utilities for obtaining time of commands and events were renamed.
  * A temporarily utility `Utils.toTemporal(Timestamp)` was introduced, expecting that Spine Time would introduce the conversion method in `TimestampTemporal` class. See [the issue on that](https://github.com/SpineEventEngine/time/issues/57).
 * `MessageWithContext.getContext()` returns now `MessageContext` instead of just `Message`.


testutil-server/src/test/java/io/spine/testing/server/entity/EntitySubjectTest.java
```diff
@@ -0,0 +1,145 @@
...
+    private Entity<?, ?> entity;
+
+    @Override
+    protected Subject.Factory<EntitySubject, Entity<?, ?>> subjectFactory() {
+        return entities();
+    }
+
+    @BeforeEach
+    void setUp() {
+        BbProjectId id = BbProjectId
```
alexander-yevsyukov: Can we migrate these test IDs so that they are UUID-based, and use `generate()` instead?

alexander-yevsyukov:
@armiol, PTAL.

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
```diff
@@ -105,26 +106,47 @@
...
+    /**
+     * Collects all commands, including posted to the context during its setup or
+     * generated by the entities in response to posted or generated events or commands.
+     */
     private final CommandCollector commands;
+
+    /**
+     * Commands received by this instance and posted to Command Bus during the test setup.
+     *
+     * <p>These commands are filtered out from those which are generated in the Bounded Context
```
armiol: Let's remove `which are` from this line to avoid doubling the "which".
```
These commands are filtered out from those generated in the Bounded Context...
```

PR #1006, opened 2019-04-09
Assertions for generated messages

alexander-yevsyukov:
This PR extends `BlackBoxBoundedContext` to provide:
  * access to generated command- and event messages;
  * `EntitySubject` via its state class and ID. 
      Previously, it was possible to obtain an entity subject only via its class. So entity classes had to be made `public` for tests of orchestrating Process Managers just for that. The state class is public anyway.


alexander-yevsyukov:
@armiol, PTAL.

PR #1009, opened 2019-04-11
Obtain sub-subject of generated messages

alexander-yevsyukov:
This PR:
  * Extends `MessageWithContext` interface with an ability to obtain unpacked message and test its type.
  * Extends `EmittedMessageSubject` with an ability to create a subject of the same type on messages filtered by the passed type.

### DIFF
--- a/core/src/main/java/io/spine/core/MessageWithContext.java
+++ b/core/src/main/java/io/spine/core/MessageWithContext.java
@@ -23,7 +23,10 @@ package io.spine.core;
 import com.google.protobuf.Any;
 import com.google.protobuf.Message;
 import io.spine.base.MessageContext;
+import io.spine.protobuf.AnyPacker;
 import io.spine.type.TypeUrl;
+
+import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Base interfaces for outer objects of messages with contexts, such as commands or events.
@@ -48,9 +51,30 @@ public interface MessageWithContext extends Message {
     MessageContext getContext();
 
     /**
+     * Obtains the unpacked form of the enclosed message.
+     *
+     * <p>The method {@link #getMessage()} returns the packed version.
+     */
+    default Message enclosedMessage() {
+        Message enclosed = AnyPacker.unpack(getMessage());
+        return enclosed;
+    }
+
+    /**
      * Obtains the type URL of the enclosed message.
      */
     default TypeUrl typeUrl() {
         return TypeUrl.ofEnclosed(getMessage());
+    }
+
+    /**
+     * Verifies if the enclosed message has the same type as the passed, or the passed type
+     * is the super-type of the message.
+     */
+    default boolean hasType(Class<? extends Message> enclosedMessageClass) {
+        checkNotNull(enclosedMessageClass);
+        Message enclosed = enclosedMessage();
+        boolean result = enclosedMessageClass.isAssignableFrom(enclosed.getClass());
+        return result;
     }
 }
--- a/core/src/main/java/io/spine/core/MessageWithContext.java
+++ b/core/src/main/java/io/spine/core/MessageWithContext.java
@@ -22,6 +22,7 @@ package io.spine.core;
 
 import com.google.protobuf.Any;
 import com.google.protobuf.Message;
+import io.spine.annotation.GeneratedMixin;
 import io.spine.base.MessageContext;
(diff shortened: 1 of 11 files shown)


alexander-yevsyukov:
@dmdashenkov, PTAL.

core/src/main/java/io/spine/core/MessageWithContext.java
```diff
@@ -47,10 +50,31 @@
...
      */
     default TypeUrl typeUrl() {
         return TypeUrl.ofEnclosed(getMessage());
     }
+
+    /**
+     * Verifies if the enclosed message has the same type as the passed, or the passed type
+     * is the super-type of the message.
+     */
+    default boolean hasType(Class<? extends Message> enclosedMessageClass) {
```
dmdashenkov: I'd name it `isOfType` as in `event.isOfType(UserCreated.class)`.

core/src/main/java/io/spine/core/MessageWithContext.java
```diff
@@ -47,10 +50,31 @@
      */
     MessageContext getContext();
 
+    /**
+     * Obtains the unpacked form of the enclosed message.
+     *
+     * <p>The method {@link #getMessage()} returns the packed version.
```
dmdashenkov: Please use the `@see` tag for this reference.

testutil-server/src/test/java/io/spine/testing/server/EmittedMessageSubjectTest.java
```diff
@@ -101,15 +112,59 @@ void retrieveMessage() {
...
+        int messageCount = 3;
+        int otherMessageCount = 2;
+        Iterable<W> outerObjects = ImmutableList
+                .<W>builder()
+                .addAll(messages(messageCount))
+                .addAll(messages(otherMessageCount, this::createAnotherMessage))
+                .build();
+
+        Class<M> cls1 = typeOf(createMessage());
+        Class<M> cls2 = typeOf(createAnotherMessage());
```
dmdashenkov: Please use `message`  and `other` instead of `1` and `2` for better readability.

testutil-server/src/test/java/io/spine/testing/server/EmittedMessageSubjectTest.java
```diff
@@ -101,15 +112,59 @@ void retrieveMessage() {
...
+        subSubject1.hasSize(messageCount);
+
+        S subSubject2 = subject.withType(cls2);
+        subSubject2.hasSize(otherMessageCount);
+    }
+
+    @Test
+    @DisplayName("fail when trying to obtain filtered sub-subject over null actual")
+    void failWithNull() {
+        Class<M> cls1 = typeOf(createMessage());
```
dmdashenkov: There is only one `cls` in this test case.

core/src/test/java/io/spine/core/MessageWithContextTest.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    private Event stubEvent() {
+        Event event = Event
+                .newBuilder()
+                .setMessage(AnyPacker.pack(
+                        ProjectCreated
+                                .newBuilder()
+                                .setProjectId(ProjectId.newBuilder()
+                                                       .setId(getClass().getName()))
+                                .build()))
```
dmdashenkov: Please extract some values to make this more readable.

core/src/main/java/io/spine/core/MessageWithContext.java
```diff
@@ -47,10 +54,31 @@
      */
     MessageContext getContext();
 
+    /**
+     * Obtains the unpacked form of the enclosed message.
+     *
+     * @see #getMessage()
+     */
+    default Message enclosedMessage() {
```
dmdashenkov: Please take a look at the `Events` and `Commands` utilities. Those define methods `getMessage()` for the same thing. I suggest we deprecate those methods (or maybe, the entire utility classes).

alexander-yevsyukov: We cannot get rid of these utilities just yet. The `enclosedMessage()` returns `Message`, and utilities return specific types like `CommandMessage` and `EventMessage`. The Model Compiler does not support passing generic parameters, so I don't know how we'd overload the `enclosedMessage()` method returning more narrow type from `Command` or `Event`.

PR #1012, opened 2019-04-18
Update `base`

dmdashenkov:
This PR updates the `base` dependency. Some classes in `base` have been repackaged and some—removed.

Also, this PR moves the Gradle script files from `scripts` subfolder into `gradle` subfolder. Some of the Gradle scripts are substituted with the ones from `config`. `scripts` subfolder now only contains Shell scripts.


dmdashenkov:
@armiol, PTAL.

PR #1013, opened 2019-04-19
Improve error diags text

alexander-yevsyukov:
This PR fixes diagnostic messages to mention state type. Previously it mentioned entity types referring to state types.


alexander-yevsyukov:
@dmdashenkov, PTAL.

PR #1014, opened 2019-04-22
Prohibit returning Nothing from Process Managers

### LINKED ISSUE #1007: Prohibit returning `Nothing` from command handling methods
This issue is similar to #790. Now it is possible to return `Nothing` from a command handling method of a Process Manager. We need to disallow it, requiring a valid event message in response to a command. This is one of the fundamental principles of the framework, which we cannot relax. 

We still want `Expect<> ` some day. But while we expecting it, we need to ensure the command handling contract.

yuri-sergiichuk:
This PR resolves #1007.

It is now not possible to return `Nothing` from process managers.

As a part of the PR, I have separated test Protobuf messages for `command` and `commandbus` tests to fix issues with `model-verifier` and ensure test entities are used only for test cases they were intended for.

Also, I've cleaned up old Codacy-related configurations and bumped Gradle to the latest 5.4 version.


yuri-sergiichuk:
@armiol PTAL.

PR #1018, opened 2019-04-23
API comments revised and fixed to match grammar and style rules in CommandFactory.java

meizabel:
In `CommandFactory`.java file:


client/src/main/java/io/spine/client/CommandFactory.java
```diff
@@ -126,7 +126,7 @@ public Command createBasedOnContext(CommandMessage message, CommandContext conte
      * <p>If an instance of {@link Any} is passed as the {@code message} parameter, the packed
      * message is used for the command construction.
      *
-     * <p>The ID of the new command instance is automatically generated.
+     * <p>The ID of a new command instance is automatically generated.
```
dmdashenkov: Please revert. We're talking about a specific, command, not __a__ command.

PR #1024, opened 2019-04-24
API comments revised and edited in TopicFactory.java

meizabel:
In `TopicFactory.java` file:

### DIFF
--- a/client/src/main/java/io/spine/client/TopicFactory.java
+++ b/client/src/main/java/io/spine/client/TopicFactory.java
@@ -34,7 +34,7 @@ import static java.lang.String.format;
 /**
  * A factory of {@link Topic} instances.
  *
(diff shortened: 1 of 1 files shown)


client/src/main/java/io/spine/client/TopicFactory.java
```diff
@@ -83,7 +83,7 @@ public Topic allOf(Class<? extends Message> targetType) {
     }
 
     /**
-     * Creates a {@link Topic} for the specified {@link Target}, updates for which will include
+     * Creates a {@link Topic} for the specified {@link Target}; updates for it will include
```
dmdashenkov: Please split into 2 sentences. I'm not sure if the `javadoc` tool will or will not cut the doc "summary" at a `;`.

PR #1025, opened 2019-04-24
Update dependencies, restrain visibility, and release `1.0.0-pre7`

dmdashenkov:
This PR updates the `base` and `time` modules to `1.0.0-pre7` version. Core modules are updated to the `1.0.0-pre7` version as well.

## Entity Visibility

Previously, the default visibility of an entity was `FULL`. It was decided to make `NONE` the new default for greater security and fool tolerance. Projections still have `FULL` as the default.

Also, the query-side services now check if the visibility is sufficient.

## Test Fixes

This PR cleans up the test models in which the `Nothing` event was still produced as a reaction on commands in process managers. We don't do that any more.

## gRPC 

Previously, invalid queries caused multiple different exceptions in `QueryService`. Now, there is only one exception, which is fed to the `StreamObserver`.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/entity/EntityVisibility.java
```diff
@@ -0,0 +1,208 @@
...
+    public boolean is(Visibility visibility) {
+        checkNotNull(visibility);
+        checkNotDefault(visibility);
+        return value == visibility;
+    }
+
+    /**
+     * Checks if the visibility is at least as allowing as the given one.
+     *
+     * <p>Call {@code isAsLeast(NONE)} always returns {@code true}. Call {@code isAtLeast(FULL)}
```
armiol: Let's kill both `Call` words.

server/src/test/proto/spine/test/entity/visibility_test.proto
```diff
@@ -0,0 +1,99 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+syntax = "proto3";
+
+package spine.test.serverentity;
```
armiol: That's odd.

dmdashenkov:
@armiol, PTAL again.

server/src/test/proto/spine/test/entity/visibility_test.proto
dmdashenkov: I've added a comment on this matter. We cannot use the `.entity` package _and_ the `entity` option at the same time in a single file.

server/src/test/proto/spine/test/entity/visibility_guard_test.proto
```diff
@@ -19,7 +19,7 @@
  */
 syntax = "proto3";
 
-//Cannot use `.entity` package due to a confict with `(entity)` annotation.
+// Cannot use `.entity` package due to a conflict with `(entity)` annotation.
 package spine.test.serverentity;
```
armiol: We can have `spine.test.guard` for this Proto file.

dmdashenkov:
@armiol, PTAL again.

armiol:
@dmdashenkov LGTM. Let's also include #1038 into the release as well. So please let it be merged first.

dmdashenkov:
@armiol, PTAL.

PR #1027, opened 2019-04-24
API comments revised and edited in StreamObservers.java

meizabel:
in `StreamObservers.java` file:


client/src/main/java/io/spine/grpc/StreamObservers.java
```diff
@@ -58,7 +58,7 @@ private StreamObservers() {
 
     /**
      * A utility method which sends {@linkplain Responses#ok() acknowledgement}
-     * to the client via the {@code responseObserver} provided and
+     * to the client using the {@code responseObserver} provided and
```
armiol: `via` has the meaning closer to `through`. `responseObserver` serves as something like a pipe, transmitting the responses from one side to another. So `using` is an over-simplification to me.

PR #1029, opened 2019-04-25
Remove version column for EEntity

### LINKED ISSUE #1017: All records in event store have default version value
All records in the event store have default `version` value.

`EEntity.Factory.create()` doesn't write entity version from the past event.

yuri-sergiichuk:
This PR fixes #1017.

The version column definition is now applied only to `Projection` and `ProcessManager` entities as a separate `HasVersionColumn` mixin.


yuri-sergiichuk:
@armiol PTAL.

yuri-sergiichuk:
@armiol PTAL.

server/src/test/java/io/spine/server/entity/storage/ColumnsTest.java
```diff
@@ -102,7 +101,7 @@ void getAllColumns() {
 
         assertContainsColumns(
                 entityColumns,
-                version.name(), archived.name(), deleted.name(),
+                 archived.name(), deleted.name(),
```
armiol: Let's fix the indentation.

PR #1033, opened 2019-04-26
API comments revised and edited in io.spine.server.command package.

meizabel:
Language-, grammar- and style-related fixes in `io.spine.server.command` package classes and interfaces.


server/src/main/java/io/spine/server/command/AbstractCommandDispatcher.java
```diff
@@ -94,7 +94,7 @@ public void onError(CommandEnvelope envelope, RuntimeException exception) {
     }
 
     /**
-     * Indicates whether some another command handler is "equal to" this one.
+     * Indicates whether some other command handler is "equal to" this one.
```
alexander-yevsyukov: Can it be just “another” instead of “some other”?

PR #1035, opened 2019-04-26
API comments revised and edited in io.spine.server.enrich package.

meizabel:
Some language-, grammar- and style-related fixes in API comments ofio.spine.server.enrich package classes and interfaces.


server/src/main/java/io/spine/server/enrich/Enricher.java
```diff
@@ -51,7 +51,7 @@ protected Enricher(EnricherBuilder<M, C, ?> builder) {
      * <ol>
      *     <li>There is one or more enrichments defined in Protobuf using
```
alexander-yevsyukov: We don't have this feature any more.

server/src/main/java/io/spine/server/enrich/Enricher.java
```diff
@@ -51,7 +51,7 @@ protected Enricher(EnricherBuilder<M, C, ?> builder) {
...
      *     {@code enrichment_for} and/or {@code by} options.
-     *     <li>There is one or more field enrichment schema registered for
+     *     <li>There is one or more field enrichment schemas registered for
```
alexander-yevsyukov: We don't have field enrichments either.

server/src/main/java/io/spine/server/enrich/Enricher.java
```diff
@@ -46,13 +46,9 @@ protected Enricher(EnricherBuilder<M, C, ?> builder) {
     /**
      * Enriches the passed message if it can be enriched. Otherwise, returns the passed instance.
      *
-     * <p>An message can be enriched if the following conditions are met:
+     * <p>An message can be enriched if the following condition is met:
      *
      * <ol>
```
alexander-yevsyukov: Why are we having the ordered list at all, assuming that there is only one condition left?

PR #1036, opened 2019-04-26
API comments revised and edited in io.spine.server.entity package

meizabel:
Several language-, grammar- and style-related fixes in io.spine.server.entity package classes and interfaces.


server/src/main/java/io/spine/server/entity/EventBlackList.java
```diff
@@ -43,7 +43,7 @@ private EventBlackList(ImmutableSet<EventClass> forbiddenEvents) {
      * Creates a new instance of {@code EventBlackList} discarding events of the given types.
      *
      * @param eventClasses
-     *         the not allowed event classes
+     *         the list of Event classes which cannot be used
```
alexander-yevsyukov: I don't think we need to have event capitalized here. We're not referring to `Event` as a class. We are speaking about “event” as a specific kind of messages.

server/src/main/java/io/spine/server/entity/EventFieldFilter.java
```diff
@@ -42,8 +42,8 @@
  * An {@link EventFilter} which allows all the events but trims some of their message fields.
  *
  * <p>By default, the input event is returned unchanged. Specify a {@link FieldMask} for an event
- * type in order to trim certain fields off the event messages (the fields in the mask are
- * the retained, all the others are cleared).
+ * type to trim certain fields off the event messages (the fields in the mask are
```
alexander-yevsyukov: As long as we're improving the text, can we avoid braces in the new version, please?

server/src/main/java/io/spine/server/entity/FieldMasks.java
```diff
@@ -64,13 +64,13 @@ private FieldMasks() {
...
      *
-     * <p>In case the {@code FieldMask} instance contains invalid field declarations,
+     * <p>If the {@code FieldMask} instance contains invalid field declarations,
      * they are ignored and do not affect the execution result.
      *
      * @param mask    the {@code FieldMask} to apply.
      * @param message the {@link Message} to apply given mask to.
-     * @return the message of the same type as the given one with only selected fields if
-     * the {@code mask} is valid, original message otherwise.
+     * @return the message of the same type as the given one with only selected fields. Otherwise if
```
alexander-yevsyukov: If think new text alters the original meaning. Also, the previous style follows the way it's done in standard [JDK Javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html).

server/src/main/java/io/spine/server/entity/EventFieldFilter.java
alexander-yevsyukov: It's not the braces which is the problem. This text is hard to understand because the sentence is long.

PR #1041, opened 2019-04-29
API comments revised and edited in io.spine.server.event package.

meizabel:
Some language-, grammar- and style-related fixes in API comments of io.spine.server.event package classes, interfaces, and exceptions.


server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -447,8 +447,8 @@ protected Builder self() {
      * Handles a dead event by saving it to the {@link EventStore} and producing an
      * {@link UnsupportedEventException}.
      *
-     * <p> We must store dead events, as they are still emitted by some entity and therefore are
-     * a part of the history for the current bounded context.
+     * <p> We must store dead events as they can still be emitted by some entities and therefore are
```
dmdashenkov: Please remove the space at the start of the paragraph. They use it in some projects. We usually do not.

PR #1045, opened 2019-04-30
API comments revised and edited in io.spine.server.procman package.

meizabel:
Some language-, grammar- and style-related fixes in API comments of io.spine.server.procman package classes.


server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -63,8 +63,8 @@ public LifecycleRules() {
     /**
      * Configures the repository to archive its entities when certain events/rejections occur.
      *
-     * <p>Subsequent calls to this method do not clear the previously added event classes (i.e. not
-     * override each other).
+     * <p>Subsequent calls to this method do not clear the previously added event classes,
+     * for example, in order not to override each other.
```
alexander-yevsyukov: Isn't it “that is” in this context? “For example” doesn't feel right here. The same comment applies to the similar case below.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -63,8 +63,8 @@ public LifecycleRules() {
     /**
      * Configures the repository to archive its entities when certain events/rejections occur.
      *
-     * <p>Subsequent calls to this method do not clear the previously added event classes (i.e. not
-     * override each other).
+     * <p>Subsequent calls to this method do not clear the previously added event classes,
+     * i.e, in order not to override each other.
```
alexander-yevsyukov: Why don't we have the period after `e` and before comma?

Also, it still does not read well. It was meant to say that we do not override events. Please ask @dmitrykuzmin for the meaning he had when writing this.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -97,8 +97,8 @@ boolean shouldArchiveOn(ThrowableMessage rejection) {
     /**
      * Configures the repository to delete its entities when certain events/rejections occur.
      *
-     * <p>Subsequent calls to this method do not clear the previously added event classes (i.e. not
-     * override each other).
+     * <p>Subsequent calls to this method do not clear the previously added event classes
+     * i.e, in order not to override each other.
```
alexander-yevsyukov: Same comment as above.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
dmitrykuzmin: I meant that in case we make two subsequent calls
```
lifecycle.archiveOn(Event1.class);
lifecycle.archiveOn(Event2.class);
```
the repository will archive when _either_ of `Event1` and `Event2` occurs and not only on `Event2`. I.e. subsequent calls to the method do not override the previous calls.

The same applies to the `deleteOn` method below.

meizabel: @dmitrykuzmin 
Thank you for clarifying! 
I will include more details in both entries then.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -63,8 +63,12 @@ public LifecycleRules() {
     /**
      * Configures the repository to archive its entities when certain events/rejections occur.
      *
-     * <p>Subsequent calls to this method do not clear the previously added event classes (i.e. not
-     * override each other).
+     * <p>Subsequent calls to this method do not clear the previously added event classes,
+     * i.e, in order not to override each other. For example, if we make two subsequent calls:
```
dmitrykuzmin: This `i.e,` looks like a typo to me.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -63,8 +63,12 @@ public LifecycleRules() {
...
+     * `lifecycle.archiveOn(Event1.class)`;
```
dmitrykuzmin: It's probably better to use the `{@code}` tag like this: `{@code lifecycle.archiveOn(Event1.class);}`, so it's properly formatted in the IDE tool-tip.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -63,8 +63,12 @@ public LifecycleRules() {
...
+     * `lifecycle.archiveOn(Event2.class)`;
```
dmitrykuzmin: Same as the above.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -63,8 +63,12 @@ public LifecycleRules() {
...
+     * The repository will archive when either `Event1` or `Event2` occurs and  and not only when `Event2` occurs.
```
dmitrykuzmin: We probably can use `{@code Event1}` and `{@code Event2}` here either.

Also, this line seems to be longer than allowed 100 symbols.

server/src/main/java/io/spine/server/procman/ProcessManagerRepository.java
```diff
@@ -173,10 +173,10 @@ private void checkNotDeaf() {
     }
 
     /**
-     * Obtains classes of external events on which process managers managed by this repository
+     * Obtains classes of external events to which the process managers managed by this repository
      * react.
      *
-     * @return a set of event classes or an empty set, if process managers do not react on
+     * @return a set of event classes or an empty set if process managers do not react on
```
dmitrykuzmin: Judging from the changes above, maybe it should be `...react to` here also.

server/src/main/java/io/spine/server/procman/LifecycleRules.java
```diff
@@ -97,8 +101,12 @@ boolean shouldArchiveOn(ThrowableMessage rejection) {
     /**
      * Configures the repository to delete its entities when certain events/rejections occur.
      *
-     * <p>Subsequent calls to this method do not clear the previously added event classes (i.e. not
-     * override each other).
+     * <p>Subsequent calls to this method do not clear the previously added event classes
+     * i.e, in order not to override each other. For example, if we make two subsequent calls:
```
dmitrykuzmin: Same as the above for this paragraph.

meizabel:
@alexander-yevsyukov 
The comments have been addressed and approved by a reviewer.

PR #1046, opened 2019-04-30
API comments revised and edited in io.spine.server.projection package

meizabel:
Several language-, grammar- and style-related fixes in API comments of io.spine.server.projection package classes.


server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -273,7 +273,7 @@ public Timestamp readLastHandledEventTime() {
     EventStreamQuery createStreamQuery() {
         Set<EventFilter> eventFilters = createEventFilters();
 
-        // Get the timestamp of the last event. This also ensures we have the storage.
+        // Gets the timestamp of the last event. This also ensures we have the storage.
```
alexander-yevsyukov: This is not Javadoc of a method. It deliberately starts with imperative to convey the meaning of the following line.

PR #1051, opened 2019-05-01
API comments revised and edited in io.spine.server.tenant package.

meizabel:
Some language-, grammar- and style-related fixes in API comments of `io.spine.server.tenant` package.


server/src/main/java/io/spine/server/tenant/TenantAwareOperation.java
```diff
@@ -46,7 +46,7 @@ protected TenantAwareOperation() throws IllegalStateException {
     /**
      * Creates an instance of an operation for the tenant specified by the passed ID.
      *
-     * <p>If a default instance of {@link TenantId} is passed (because the application works in
+     * <p>If a default instance of {@link TenantId} is passed because the application works in
```
alexander-yevsyukov: We need to keep braces in this context since the main idea is the substitution of the value, and not single-tenant mode.

PR #1052, opened 2019-05-01
API comments revised and edited in io.spine.server.transport package.

meizabel:
Some language-, grammar- and style-related fixes in API comments of `io.spine.server.transport` package.


server/src/main/java/io/spine/server/transport/GrpcContainer.java
```diff
@@ -38,7 +38,7 @@
 /**
  * Wrapping container for gRPC server.
  *
- * <p>Maintains and deploys several of gRPC services within a single server.
+ * <p>Maintains and deploys some of gRPC services within a single server.
```
armiol: It should be `several gRPC services`. We want to emphasize it's more than one.

PR #1056, opened 2019-05-01
Fix column lookup

dmdashenkov:
Since recent changes, the `version` entity column was missing from the descendants of `Aggregate`, `ProcessManager`, and `Projection`. This PR fixes the column lookup mechanism so that this column is present.

This PR patches the version `1.0.0-pre7`.


dmdashenkov:
@armiol, PTAL.

PR #1059, opened 2019-05-06
Improve `BoundedContext` API

alexander-yevsyukov:
This  with:
 * Extends `BoundedContext` with a shortcut method for creating and registering `DefaultRepository` for an entity class.
 * `InMemoryStorageFactory` with a factory method which accepts Bounded Context name as a `String`.

Published version returned back to `1.0.0-SNAPSHOT`.


server/src/test/java/io/spine/server/BoundedContextTest.java
```diff
@@ -176,12 +181,27 @@ void processManagerRepository() {
...
         @DisplayName("ProjectionRepository")
         void projectionRepository() {
             boundedContext.register(DefaultRepository.of(ProjectReport.class));
+            assertRegisteredRepositoryOf(ProjectReport.class);
+        }
+
+        @Test
+        @DisplayName("DefaultRepository via passed entity class")
+        void entityClass() {
+            boundedContext.register(ProjectAggregate.class);
```
armiol: Looks, the assertion call is missing.

PR #1060, opened 2019-05-07
Test presence of repositories

alexander-yevsyukov:
This PR adds an ability to test the presence of a repository, instead of finding it (which takes into account visibility of entity states).

### DIFF
--- a/server/src/main/java/io/spine/server/BoundedContext.java
+++ b/server/src/main/java/io/spine/server/BoundedContext.java
@@ -362,22 +362,30 @@ public abstract class BoundedContext implements AutoCloseable, Logging {
     /**
      * Verifies if there is a repository managing the passed entity class is registered
      * with this Bounded Context.
+     *
+     * <p>This method does not take into account visibility of entity states.
+     *
+     * @see #findRepository(Class)
      */
     @VisibleForTesting
     public boolean hasRepositoryOf(Class<? extends Entity<?, ?>> entityClass) {
         EntityClass<? extends Entity<?, ?>> cls = EntityClass.asEntityClass(entityClass);
-        Optional<Repository> repository = findRepository(cls.stateClass());
-        return repository.isPresent();
+        boolean result = guard.hasRepository(cls.stateClass());
+        return result;
     }
 
     /**
      * Verifies if there a repository managing entities with the state of the passed class
      * registered with this Bounded Context.
+     *
+     * <p>This method does not take into account visibility of entity states.
+     *
+     * @see #findRepository(Class)
      */
     @VisibleForTesting
     public boolean hasRepositoryOfEntityWithState(Class<? extends Message> stateClass) {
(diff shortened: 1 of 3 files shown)


alexander-yevsyukov:
@dmdashenkov, PTAL.

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -350,13 +352,42 @@ public CommandErrorHandler createCommandErrorHandler() {
...
+    /**
+     * Verifies if there a repository managing entities with the state of the passed class
+     * registered with this Bounded Context.
+     *
+     * <p>This method does not take into account visibility of entity states.
+     *
+     * @see #findRepository(Class)
+     */
+    @VisibleForTesting
+    public boolean hasRepositoryOfEntityWithState(Class<? extends Message> stateClass) {
```
dmdashenkov: This is too long. Also, the default behaviour for the `findRepository` method is to look up by the entity state type, thus the `has...` method looks up by the entity state type should have the shorter name.

server/src/test/java/io/spine/server/BoundedContextTest.java
```diff
@@ -188,21 +186,38 @@ void projectionRepository() {
 
         <I, E extends Entity<I, ?>> void registerAndAssertRepository(Class<E> cls) {
             boundedContext.register(DefaultRepository.of(cls));
-            assertRegisteredRepositoryOf(cls);
+            assertThat(boundedContext.hasRepositoryOf(cls))
+                    .isTrue();
```
dmdashenkov: `assertTrue` is even better, I think. It's shorter and easier to read.

server/src/test/java/io/spine/server/BoundedContextTest.java
```diff
@@ -324,63 +339,73 @@ void useEventStoreIfSet() {
...
-                                               .isMultitenant());
+            context = singleTenant();
 
-            bc = BoundedContext.newBuilder()
-                               .setMultitenant(false)
-                               .build();
+            assertMultitenancyEqual(context::isMultitenant, context.stand()::isMultitenant);
+        }
+
+        void assertMultitenancyEqual(Supplier<Boolean> s1, Supplier<Boolean> s2) {
```
dmdashenkov: Please use `BooleanSupplier` instead.

server/src/test/java/io/spine/server/BoundedContextTest.java
```diff
@@ -324,63 +339,73 @@ void useEventStoreIfSet() {
...
+                    .build();
         }
+
+        private BoundedContext singleTenant() {
+            return BoundedContext
+                    .newBuilder()
+                    .setMultitenant(false)
+                    .build();
+        }
+
```
dmdashenkov: This line is redundant.

PR #1062, opened 2019-05-08
Own repository cache for the BlackBox context

dmdashenkov:
Previously, `BlackBoxBoundedContext` used the underlying `BoundedContext` to find an entity repository to fetch entity states.

The issue with this is that the `BoundedContext.findRepository(..)` does not obtain repositories for non-visible repositories.

Now, the `BlackBoxBoundedContext` has its own cache of the registered repositories. This way, the test util is capable of accessing any repository on demand.


dmdashenkov:
@armiol, PTAL.

PR #1065, opened 2019-05-15
Encapsulate routing setup

alexander-yevsyukov:
This PR improves the way routing is customized. Previously, we relied on exposed `eventRouting()` or `commandRouting()` methods that gave access to customizing the routing schemas either in repository constructors or in `onRegistered()`. This worked, but unnecessarily exposed the routing schemas to user's repository classes.

New way of routing customization exposes `setupEventRouting()`, `setupCommandRouting()`, and `setupImportRouting()` methods that are called during the `Repository.init()`  (which is also introduced by this PR).

The `init()` is meant to be the place for general purpose customization which may be needed by a user's repository class.

The `onRegistered()` callback is left for compatibility reasons, and for the cases when additional configuration is needed after a repository is registered in the Bounded Context.

Other notable changes:
  * `BoundedContext` got `toString()` which returns the context name.
  * Simplified routing by first message field.
  * When closed `Repository` clears its reference to its `BoundedContext`.
  * Common interface was introduced for `StandardRejection`s.


alexander-yevsyukov:
@armiol, @dmdashenkov, PTAL.

server/src/main/java/io/spine/server/entity/Repository.java
```diff
@@ -64,9 +64,10 @@
      * The {@link BoundedContext} to which the repository belongs.
      *
      * <p>This field is null when a repository is not {@linkplain
-     * BoundedContext#register(Repository) registered} yet.
+     * BoundedContext#register(Repository) registered} yet and
+     * after the repository is {@linkplain #close() closed}.
      */
-    private @MonotonicNonNull BoundedContext boundedContext;
```
armiol: Not sure why this has changed to `@Nullable`.

server/src/main/java/io/spine/server/entity/rejection/StandardRejection.java
```diff
@@ -0,0 +1,45 @@
...
+import io.spine.protobuf.AnyPacker;
+
+/**
+ * Interface common for {@link StandardRejections}.
+ */
+@Immutable
+@GeneratedMixin
+public interface StandardRejection extends RejectionMessage {
+
+    Any getEntityId();
```
armiol: Please document.

server/src/main/java/io/spine/server/route/FirstField.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.protobuf.Descriptors.FieldDescriptor;
+import com.google.protobuf.Message;
+
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.util.Exceptions.newIllegalStateException;
+
+/**
+ * Obtains the first field from multiple messages expecting each field
```
armiol: While I understand what you meant, this is only because we have discussed this behaviour previously.

This should be something like this:
```
Routes messages to a single target, which ID is the same as the first field of the routed message.

<p>It is expected that the types of the first field and the identifier are the same.
```

server/src/main/java/io/spine/server/route/StateUpdateRouting.java
```diff
@@ -104,4 +106,21 @@ private StateUpdateRouting() {
...
+     * Validates routing schema for types of state messages.
+     *
+     * @param stateClasses
+     *         the set of classes that this routing is expected to serve
+     * @throws IllegalStateException
+     *         if one of the state type cannot be dispatched by the current schema configuration
+     */
+    public void validate(Set<StateClass> stateClasses) throws IllegalStateException {
+        checkNotNull(stateClasses);
+        //TODO:2019-05-15:alexander.yevsyukov: Implement
```
armiol: Let's have an issue in GitHub and reference it here.

server/src/main/java/io/spine/server/entity/rejection/StandardRejection.java
```diff
@@ -0,0 +1,45 @@
...
+package io.spine.server.entity.rejection;
+
+import com.google.errorprone.annotations.Immutable;
+import com.google.protobuf.Any;
+import io.spine.annotation.GeneratedMixin;
+import io.spine.base.RejectionMessage;
+import io.spine.protobuf.AnyPacker;
+
+/**
+ * Interface common for {@link StandardRejections}.
```
armiol: I am not sure why we reference a generated type, which even does not have any class-level documentation.

I would briefly document the purpose of the mixin here or link some other class.

server/src/main/java/io/spine/server/entity/Repository.java
alexander-yevsyukov: It's documented in the Javadoc of the field. We now clear the reference after the repository is closed.

alexander-yevsyukov:
@armiol, PTAL again.

PR #1071, opened 2019-05-23
Test for the Bus dead lock

dmdashenkov:
This PR adds a test for the deadlock found in the `DispatchingQueue.add(..)`.

The test is at `DispatchingQueueSynchronisationTest.deadlock()`.
The test is currently disabled, as the issue is not fixed. It is expected that the `Inbox` feature will address this dead lock.


dmdashenkov:
@armiol, PTAL.

PR #1072, opened 2019-05-23
Update base

alexander-yevsyukov:
This PR fixes test data to match new version of `base`.


alexander-yevsyukov:
@dmdashenkov, @armiol, PTAL.

server/src/test/java/io/spine/test/validation/FakeOption.java
```diff
@@ -42,20 +43,20 @@
  * thrown by the constraint produced by this option. Otherwise, the constraint never discovers any
  * violations.
  */
-public final class FakeOption extends FieldValidatingOption<Void, Object> {
+public final class FakeOption extends FieldValidatingOption<Void, Empty> {
```
dmdashenkov: Why `Empty`? A field option must be applicable to any kind of fields, including primitives.

alexander-yevsyukov: These type parameters are marked as `@ImmutableTypeParameter`. `Object` is mutable.

alexander-yevsyukov: It took the first matching `Message` type. It could probably be `Any`.

dmdashenkov: In this case, we can suppress the warning caused by `ImmutableTypeParameter` as we only use `Object`s which in fact are immutable, such as `Boolean`, `String`, `ByteString`, etc.

PR #1074, opened 2019-05-24
State ID field

alexander-yevsyukov:
This PR teaches the `Transaction` to set the first field of an entity state, if this field is required or assumed required. Such an ability eliminates the need to call methods like `setId()` in every message handler of an entity.

Other changes:
  * Improved method names (eliminating the `get` prefix).
  * Reworked test environment for the `TransactionTest` hierarchy.
  * Methods related to work with entities were made `final` where possible (assuming Mockito).


alexander-yevsyukov:
@armiol, @dmdashenkov, PTAL.

server/src/main/java/io/spine/server/entity/Transaction.java
```diff
@@ -285,21 +282,30 @@ private void commitChangedState(@NonValidated S newState) {
             commitAttributeChanges();
             EntityRecord newRecord = createRecord();
             afterCommit(entityBeforeTransaction, newRecord);
-        } catch (ValidationException exception) {  /* Could only happen if the state
+        } catch (ValidationException validationException) {  /* Could only happen if the state
                                                       has been injected not using
                                                       the builder setters. */
-            InvalidEntityStateException invalidStateException = of(exception);
```
dmdashenkov: Please fix the comment alignment.

server/src/main/java/io/spine/server/entity/Transaction.java
```diff
@@ -285,21 +282,30 @@ private void commitChangedState(@NonValidated S newState) {
...
-            throw invalidStateException;
-        } catch (RuntimeException genericException) {
+            InvalidEntityStateException se = toStateException(validationException);
+            rollback(se);
+            throw se;
+        } catch (InvalidEntityStateException stateException) {
+            rollback(stateException);
+            throw stateException;
+        }
+        catch (RuntimeException genericException) {
```
dmdashenkov: Eh?

server/src/main/java/io/spine/server/entity/TransactionalEntity.java
```diff
@@ -160,40 +167,58 @@ final boolean isTransactionInProgress() {
...
      * Updates own {@code stateChanged} flag from the underlying transaction.
      */
-    void updateStateChanged() {
+    final void updateStateChanged() {
         this.stateChanged = tx().isStateChanged();
     }
 
-    B builderFromState() {
+    /**
+     * Obtains the builder for being used by a transaction.
```
dmdashenkov: This seems too specific to me. Do we assume it is only used in transactions?

server/src/main/java/io/spine/server/entity/model/EntityClass.java
```diff
@@ -57,6 +57,10 @@
     /** Type of the entity state. */
     private final TypeUrl entityStateType;
 
+    /** The helper for working with ID field of the entity state. */
```
dmdashenkov: It's an `IdFieldManager` than?

server/src/main/java/io/spine/server/entity/model/IdField.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.Serializable;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Helps to initialize ID field of an entity state, if such a field is declared.
+ */
+@Immutable
+@SuppressWarnings("SynchronizeOnThis") // Double-check idiom for lazy init.
```
dmdashenkov: 1. It does not have to use `this` anyway.
2. Why do we need a lazy init at all?

server/src/main/java/io/spine/server/entity/model/IdField.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public <I, S extends Message, B extends ValidatingBuilder<S>>
+    void initBuilder(B builder, I id) {
+        checkNotNull(builder);
+        checkNotNull(id);
+        if (!declared()) {
+            return;
+        }
+        FieldDescriptor idField = declaration.descriptor();
+        @SuppressWarnings("Immutable") // all supported types of IDs are immutable.
```
dmdashenkov: It's strange that this warning is present here.
Maybe, annotate `I` instead?

server/src/test/java/io/spine/server/aggregate/AggregateTransactionTest.java
```diff
@@ -19,121 +19,86 @@
...
 
     @Override
-    protected Transaction<ProjectId,
-                          Aggregate<ProjectId, Project, Project.Builder>,
-                          Project,
-                          Project.Builder> createTxWithState(
-            Aggregate<ProjectId, Project, Project.Builder> entity, Project state,
-            Version version) {
+    protected Transaction<Id,
+            Aggregate<Id, AggregateState, AggregateState.Builder>,
```
dmdashenkov: Please fix the alignment.

server/src/test/java/io/spine/server/procman/PmTransactionTest.java
```diff
@@ -19,124 +19,86 @@
...
     @Override
-    protected Project createNewState() {
-        return Project.newBuilder()
-                      .setId(ID)
-                      .setName("The new project name for procman tx tests")
-                      .build();
+    protected PmState newState() {
+        return PmState.newBuilder()
+                             .setId(id())
+                             .setName("The new project name for procman tx tests")
```
dmdashenkov: Please fix the alignment.

server/src/test/java/io/spine/server/projection/ProjectionShould.java
```diff
@@ -285,7 +285,11 @@ void notDeliverIfNotFits() {
                 .setValue("BBB")
                 .build();
         dispatch(projection, eventFactory.createEvent(skipped));
-        ProtoTruth.assertThat(projection.state()).isEqualTo(SavedString.getDefaultInstance());
+        ProtoTruth.assertThat(projection.state())
+                  // Ignore the different in the ID field of the state which
```
dmdashenkov:
```suggestion
                  // Ignore the difference in the ID field of the state which
```

server/src/test/java/io/spine/server/projection/ProjectionTransactionTest.java
```diff
@@ -50,94 +45,72 @@
...
     @Override
-    protected Transaction<ProjectId,
-                          Projection<ProjectId, Project, Project.Builder>,
-                          Project,
-                          Project.Builder>
-    createTxWithState(Projection<ProjectId, Project, Project.Builder> entity,
-                      Project state,
-                      Version version) {
+    protected Transaction<Id,
+            Projection<Id, ProjectionState, ProjectionState.Builder>,
```
dmdashenkov: Same as above.

server/src/test/proto/spine/test/tx/tx_events.proto
```diff
@@ -0,0 +1,49 @@
...
+
+import "spine/test/tx/tx_entities.proto";
+
+message TxCreated {
+    Id id = 1;
+    string name = 2;
+}
+
+// The event which causes the handling entity to fail with a `RuntimeException`.
+message TxErrorRequested {
```
dmdashenkov: This and the one below are passive aggressive events ((c) - Martin Fowler). Please think of a better naming for them.

server/src/main/java/io/spine/server/entity/Transaction.java
alexander-yevsyukov: The catch block no longer has the branch related to `ValidationException`. We do not throw it anymore.

server/src/main/java/io/spine/server/entity/Transaction.java
alexander-yevsyukov: I assume you meant variable names. Since we have now only two branches, both of variables are now `e`.

server/src/main/java/io/spine/server/entity/TransactionalEntity.java
alexander-yevsyukov: It is used only in the `Transaction` and tests related to it.

server/src/main/java/io/spine/server/entity/model/EntityClass.java
alexander-yevsyukov: I'd rather avoid calling things `Manager` as much as possible. You allow it once and there's no way to stop it after...

server/src/test/java/io/spine/server/aggregate/AggregateTransactionTest.java
alexander-yevsyukov: Сорсы этого тест-класса выглядят с этими всеми generic parameters как помятая газета... :) 

Anyway, aligned.

server/src/test/proto/spine/test/tx/tx_events.proto
alexander-yevsyukov: They are meant to be that way. We generate errors because of them.

server/src/main/java/io/spine/server/entity/model/IdField.java
alexander-yevsyukov: I guess you're right. The intent was to extend the `EntityClass` with information on the state field. After some thought, I've simplified it all, removing the access to ID field information from the public API.

server/src/main/java/io/spine/server/entity/IdField.java
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Helps to initialize ID field of an entity state, if such a field is declared.
+ */
+@Immutable
+final class IdField {
+
+    private final @MonotonicNonNull FieldDeclaration declaration;
+
+    IdField(EntityClass<?> entityClass) {
+        Message defaultState = entityClass.defaultState();
```
dmdashenkov: Please move the complex initialization logic into a static factory method from the constructor.

server/src/main/java/io/spine/server/entity/TransactionalEntity.java
dmdashenkov: Can it be used elsewhere?
Some methods of `TransactionalEntity` may only be used while in a transaction and throw exceptions otherwise. Is it the case with this method?

server/src/main/java/io/spine/server/entity/model/EntityClass.java
```diff
@@ -67,6 +68,7 @@
     @SuppressWarnings("Immutable") // effectively
     private transient volatile @MonotonicNonNull EntityFactory<E> factory;
 
+
```
dmdashenkov: Please remove this empty line.

server/src/test/java/io/spine/server/aggregate/AggregateRepositoryTest.java
```diff
@@ -431,10 +435,11 @@ void onEvents() {
...
                                                                     getClass());
-            AggProjectArchived msg = AggProjectArchived.newBuilder()
-                                                       .setProjectId(parentId)
-                                                       .addChildProjectId(childId)
-                                                       .build();
+            AggProjectArchived msg = AggProjectArchived
+                    .newBuilder()
+                    .setProjectId(parentId)
+                    .addChildProjectId(childId)
+                    .build();
```
dmdashenkov: Please use `vBuild()`. The ErrorProne check is to be enabled soon.

server/src/main/java/io/spine/server/entity/DefaultRecordBasedRepository.java
```diff
@@ -73,8 +71,9 @@ protected DefaultRecordBasedRepository() {
      * @param context
      *         the Bounded Context of this repository
      */
-    @OverridingMethodsMustInvokeSuper
     @Override
+    @OverridingMethodsMustInvokeSuper
+    @SuppressWarnings("CanIgnoreReturnValue") // See Javadoc.
```
armiol: The Javadoc does not really address the confusion.

server/src/main/java/io/spine/server/entity/TransactionalEntity.java
alexander-yevsyukov: Moved to `Transaction`.

PR #1076, opened 2019-05-29
Pom with dependencies

serhii-lekariev:
This PR updates the [config](https://github.com/SpineEventEngine/config) version and applies the `generatePom` task. It also adds the generated `pom` file.


serhii-lekariev:
@armiol, PTAL.

PR #1083, opened 2019-06-06
Add the ability to drop excessive aggregate event records

dmitrykuzmin:
This PR adds the functionality to drop redundant `AggregateEventRecord` instances from the storage.

The "cleaning" occurs on per-`AggregateStorage` basis, dropping the records which occur before the Nth snapshot for each entity. It's also possible to clip all records older than some `date`.


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/aggregate/AggregateStorage.java
```diff
@@ -204,4 +208,52 @@ private static AggregateEventRecord toStorageRecord(Snapshot snapshot) {
...
+     * the latest snapshot.
+     *
+     * <p>The snapshot number higher than the overall snapshot count of the entity is allowed, the
+     * entity records remain intact in this case.
+     *
+     * @throws IllegalArgumentException
+     *         if the {@code snapshotNumber} is {@code 0} or less
+     */
+    @Internal
+    public void clipRecordsBeforeSnapshot(int snapshotNumber) {
```
armiol: `truncateOlderThan(int snapshotIndex)`
and the one below will be
`truncateOlderThan(int snapshotIndex, Timestamp date)`

Please update the naming of the `protected` methods and the documentation accordingly.

dmitrykuzmin:
@armiol PTAL again.

dmitrykuzmin:
@armiol PTAL again.

PR #1084, opened 2019-06-12
System purge

dmdashenkov:
This PR starts a big System context rebuilt in anticipation of `Inbox`.

In this PR we:
 1. Transform System aggregates into projections and post events instead of commands.
 2. Strip system entities of their power to dispatch (or deny dispatching of) events and commands to non-event-sourced domain entities.

Currently, idempotency tests are disabled, since there are no more idempotency checks except for `IdempotencyGuard` for aggregates.

The System context receives non-handled events (e.g. `EventCausedCommand`). Those will be required later for integration with tracing services.


dmdashenkov:
@armiol, PTAL.

pom.xml
```diff
@@ -0,0 +1,250 @@
...
+
+<!-- 
+ 
+This file was generated using the Gradle `generatePom` task. 
+This file is not suitable for `maven` build tasks. It only describes the first-level dependencies of 
+all modules and does not describe the project structure per-subproject.
+ 
+ -->
+
+<groupId>io.spine</groupId>
```
armiol: I wonder how come this file hasn't been added to VCS previously.

server/src/main/java/io/spine/server/DefaultRepositoryString.java
```diff
@@ -0,0 +1,60 @@
...
+package io.spine.server;
+
+import io.spine.annotation.Internal;
+import io.spine.server.entity.model.EntityClass;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.String.format;
+
+/**
+ * Handles the string representation of the {@linkplain DefaultRepository default repositories}.
```
armiol: Let's discuss this concept.

server/src/main/java/io/spine/system/server/EntityLogRepository.java
```diff
@@ -20,29 +20,38 @@
...
 
 import static io.spine.server.route.EventRoute.withId;
 
 /**
- * The repository for {@link EntityHistoryAggregate}s.
+ * The repository for {@link EntityLogProjection}s.
  */
-final class EntityHistoryRepository
-        extends SystemRepository<EntityHistoryId, EntityHistoryAggregate> {
+final class EntityLogRepository
```
armiol: I think this one should die at all.

dmdashenkov:
@armiol, PTAL again.

PR #1087, opened 2019-06-13
Singular names for repeated and map fields

alexander-yevsyukov:
This PR:
  * Fixes naming for repeated and map fields to be singular nouns.
  * Adds useful methods for accessing properties of `QueryResponse` and `SubscriptionUpdate` using mixins.


armiol:
@alexander-yevsyukov please revert the changes made to `pom.xml`, as you have accidentally deleted it.

PR #1089, opened 2019-06-14
Remove notion of a separately stored event count

dmitrykuzmin:
This PR removes a separately stored `eventCountAfterLastSnapshot` for each aggregate.

We used this count for batch size optimization during the `Aggregate` history reads. The problem is, while the optimization itself rarely comes into play, it costs us one constant extra request to the storage during each read.

This PR removes the optimization and the notion of a separately stored event count. 

The method altering the repository snapshot trigger, which was the primary cause for optimization, now has the extra doc about performance concernings, and is left up to user judgement.

### DIFF
--- a/server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
+++ b/server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
@@ -457,7 +457,7 @@ public abstract class AggregateRepository<I, A extends Aggregate<I, ?, ?>>
      *
      * <p>The default value is defined in {@link #DEFAULT_SNAPSHOT_TRIGGER}.
      *
-     * <p>Please <b>NOTE</b>: repository read operations are optimized around the current snapshot
+     * <p><b>NOTE</b>: repository read operations are optimized around the current snapshot
      * trigger. Setting the snapshot trigger to a new value may cause read operations to perform
      * sub-optimally, until a new snapshot is created. This doesn't apply to newly created
      * repositories.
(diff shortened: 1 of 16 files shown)


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -458,7 +457,13 @@ protected int snapshotTrigger() {
      *
      * <p>The default value is defined in {@link #DEFAULT_SNAPSHOT_TRIGGER}.
      *
-     * @param snapshotTrigger a positive number of the snapshot trigger
+     * <p>Please <b>NOTE</b>: repository read operations are optimized around the current snapshot
```
armiol: Let's remove `Please`.

server/src/main/java/io/spine/server/aggregate/AggregateStorage.java
```diff
@@ -84,9 +85,11 @@ protected void checkNotClosed() throws IllegalStateException {
     /**
      * Writes events into the storage.
      *
-     * <p>NOTE: does not rewrite any events. Several events can be associated with one aggregate ID.
+     * <p>NOTE: does not rewrite any events. Several events can be associated with one
```
armiol: Let's settle down on whether we use `<b>` around "NOTE" or the capitalization is good enough even with no boldening. See the similar case above.

PR #1090, opened 2019-06-18
Trace API

dmdashenkov:
### Introducing Spine Trace API

When building a complex Spine-based system, it is often easy to lose track of how are domain messages propagated. In order to provide developers with insights about the system flow and performance of separate message handlers, we introduce the Trace API.

Trace API is set up separately for each bounded context. It is an extension point which allows integration with third-party services, such as Stackdriver Trace, Zipkin, etc.

By default, no tracing is performed. In order to integrate a bounded context with a tracing mechanism, pass a `TracerFactory` supplier to the context builder.

Also, this PR:
 - adds some more tests on the recently added/changed concepts of System context, such as the `ConstraintViolated` event;
 - prepares the ground for migrating from `Event-`/`CommandEnvelope`s to `Event`s and `Command`s.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/commandbus/CommandAckMonitor.java
```diff
@@ -50,10 +57,14 @@
 final class CommandAckMonitor extends DelegatingObserver<Ack> {
 
     private final SystemWriteSide writeSide;
+    private final ImmutableMap<CommandId, Command> commands;
 
     private CommandAckMonitor(Builder builder) {
         super(builder.delegate);
         this.writeSide = delegatingTo(builder.systemWriteSide).get(builder.tenantId);
+        this.commands = builder.commands
+                .stream()
```
armiol: Please fix the alignment.

server/src/main/java/io/spine/server/commandbus/CommandAckMonitor.java
```diff
@@ -117,6 +132,7 @@ static Builder newBuilder() {
         private StreamObserver<Ack> delegate;
         private TenantId tenantId;
         private SystemWriteSide systemWriteSide;
+        private ImmutableSet<Command> commands;
```
armiol: Let's discuss. A `Command` may be posted to the instance A, but its outcome will be received on the instance B (the one responsible for processing the entities with the respective ID).

server/src/main/java/io/spine/server/trace/TracerFactory.java
```diff
@@ -0,0 +1,61 @@
...
+import io.spine.core.BoundedContextName;
+import io.spine.core.Signal;
+
+/**
+ * A factory of {@link Tracer}s of signal messages.
+ *
+ * <p>Each tracer is created for a single signal and should not be reused.
+ *
+ * <p>Implementations may choose to perform I/O operations (i.e. send traces via network) on
+ * {@code close()}. It is expected nin general that each call to {@link #trace} results in some
```
armiol: `in`

server/src/main/java/io/spine/server/trace/TracerFactory.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * <p>The method may return a new factory with the same behaviour as this one but suited for
+     * the given context, or simply this instance. The method may NOT change the state of this
+     * instance to suite the given context.
+     *
+     * @param context
+     *         the name of the context to trace
+     * @return new instance
+     */
+    TracerFactory inContext(BoundedContextName context);
```
armiol: This looks weird. Let's discuss why we would need that.

server/src/test/java/io/spine/system/server/given/AbstractEventAccumulator.java
```diff
@@ -102,24 +103,26 @@ public void assertEventCount(int expectedCount) {
...
 
     @CanIgnoreReturnValue
-    public <E extends Message> E assertNextEventIs(Class<E> eventType) {
-        if (eventIterator == null) {
-            eventIterator = ImmutableList.copyOf(events).iterator();
-        }
-        assertTrue(eventIterator.hasNext(), errorMessage());
-        Message next = eventIterator.next();
-        assertThat(next, instanceOf(eventType));
+    public <E extends Message> E assertExistEvent(Class<E> eventType) {
```
armiol: The naming is confusing from both grammar and contextual point of view.

As there are no docs for the method, I am not able to suggest a name.

server/src/main/java/io/spine/server/commandbus/CommandAckMonitor.java
dmdashenkov: BTW, this is how automatic formatting works (since `commands` is a field). I'll take a look if we could fix it via the IDEA config later.

dmdashenkov:
@armiol, PTAL at the changes and let's discuss the leftover comments.

dmdashenkov:
@armiol, PTAL again.

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -73,8 +72,10 @@
  * Ubiquitous Language. Any term within a single bounded context has a single meaning and
  * may or may not map to another term in the language of another Bounded Context.
  *
- * <p>The Ubiquitous Language of a Bounded Context is represented by such concepts as the entity state, event,
- * and command types, entity types, and others. An entity and its adjacent types belong to the Bounded
+ * <p>The Ubiquitous Language of a Bounded Context is represented by such concepts as the entity
+ * state, event,
+ * and command types, entity types, and others. An entity and its adjacent types belong to the
```
armiol: Please fix the alignment.

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -120,24 +113,26 @@
      *
      * @throws IllegalStateException
      *         if called from a derived class, which is not a part of the framework
-     * @apiNote This constructor is for internal use of the framework. Application developers
-     *          should not create classes derived from {@code BoundedContext}.
+     * @apiNote This constructor is for internal use of the framework. Application
+     *         developers
```
armiol: Same as above.

server/src/main/java/io/spine/server/ContextSpec.java
```diff
@@ -0,0 +1,101 @@
...
+package io.spine.server;
+
+import com.google.common.base.Objects;
+import io.spine.core.BoundedContextName;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.core.BoundedContextNames.newName;
+
+/**
+ * Properties of a bounded context.
```
armiol: It's not quite.

It's much more about the facts which end-users let us know to create the context. "Properties" are much wider. Let's discuss to wrap this description up and avoid confusion and further misuse.

dmdashenkov:
@armiol, PTAL again

PR #1093, opened 2019-06-21
Fix the `EnricherBuilder` duplication check

dmitrykuzmin:
This PR fixes a small issue with the `EnricherBuilder` duplication check.

Right now it is impossible to create multiple enrichments of different type for a single source class. This PR fixes it.


dmitrykuzmin:
@armiol PTAL.

PR #1097, opened 2019-06-24
Gradual state trasitions

alexander-yevsyukov:
This PR improves phased transactions so that each dispatching method sees the state of the entity as if it was already committed during the previous phase.

Previously, the entity (such as an `Aggregate`) in a multi-message transaction saw its `state()` before the transaction was started. This could have been confusing and did not reflect atomicity of message handling.


server/src/main/java/io/spine/server/entity/Transaction.java
```diff
@@ -257,16 +258,51 @@ final Version version() {
...
+            throw propagate(t);
         } finally {
             phases.add(phase);
             listener.onAfterPhase(phase);
         }
     }
 
+    /**
+     * Rethrows the passed {@code Throwable} wrapped into {@code IllegalStateException},
+     * if it's not an instacen of {@code InvalidEntityStateException}.
```
armiol: Please fix the typo.

server/src/main/java/io/spine/server/entity/VersionIncrement.java
```diff
@@ -22,41 +22,92 @@
...
+     * @return the incremented {@code Version} of the entity
+     */
+    protected abstract Version nextVersion();
 
-    VersionIncrement(Transaction transaction) {
-        this.transaction = transaction;
+    /**
+     * Creates a version increment which sets the new version from the given event.
+     *
+     * <p>Such increment strategy is applied to the {@link Entity} types which represent
```
armiol: `Such an increment strategy`

server/src/main/java/io/spine/server/entity/VersionIncrement.java
```diff
@@ -22,41 +22,92 @@
...
+     */
+    public static VersionIncrement fromEvent(EventEnvelope e) {
+        return new IncrementFromEvent(e);
     }
 
     /**
-     * Advances the version of the entity in transaction.
+     * Creates a version increment which always advances the given entity version by 1.
+     *
+     * <p>Such increment strategy is applied to the {@link Entity} types which cannot use the event
```
armiol: Same as above.

server/src/main/java/io/spine/server/entity/VersionIncrement.java
```diff
@@ -22,41 +22,92 @@
...
-        Version nextVersion = nextVersion();
-        checkIsIncrement(transaction.version(), nextVersion);
-        transaction.setVersion(nextVersion);
+    public static VersionIncrement sequentially(Transaction tx) {
+        return new AutoIncrement(tx);
     }
 
     /**
-     * Returns the transaction on which the {@code VersionIncrement} operates.
+     * Increments version from a passed event.
```
armiol: Basically, it takes the version from the passed event. It does not _increment_ it.

server/src/main/java/io/spine/server/entity/VersionIncrement.java
```diff
@@ -22,41 +22,92 @@
...
+
+        private final EventEnvelope event;
+
+        private IncrementFromEvent(EventEnvelope event) {
+            super();
+            this.event = event;
+        }
+
+        @Override
+        public Version nextVersion() {
```
armiol: Please check if this method really should be `public` in a `private static final` class. The parent method is `protected`.

server/src/main/java/io/spine/server/entity/VersionIncrement.java
```diff
@@ -22,41 +22,92 @@
...
+
+        private final Transaction transaction;
+
+        private AutoIncrement(Transaction transaction) {
+            super();
+            this.transaction = checkNotNull(transaction);
+        }
+
+        @Override
+        public Version nextVersion() {
```
armiol: Same as above.

Issue #1099, opened 2019-06-25
Allow register abstract dispatchers in `BoundedContextBuilder`


dmdashenkov:
Currently, `BoundedContext` allows registering entity repositories, `EventDispatcher`s, `CommandDispatcher`s, and entity classes. `BoundedContextBuilder`, on the other hand, allows only to register entity repositories.

We need to add `add(..)` method overloads for the other registration options.

Once this is done, we need to make sure that `BlackBoxBoundedContext.from(BoundedContextBuilder)` also will take the registered abstract dispatchers.

dmitrykuzmin:
Done with #1100.

PR #1100, opened 2019-06-26
Expand `BoundedContextBuilder` API

dmitrykuzmin:
This PR addresses #1099.

The `BoundedContextBuilder` has been expanded with the API allowing to:
1. Register/unregister repositories by the entity class.
2. Register/unregister `CommandDispatcher`s.
3. Register/unregister `EventDispatcher`s.

The dispatcher registration works the same as repository registration, i.e. dispatchers are added to the `BoundedContextBuilder` but registered within `BoundedContext` only after the `build()` method is called.

If a `Repository` is passed in place of command or event dispatcher, it is still registered as a repository. The same now applies to the `BlackBoxBoundedContext` methods.


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -144,6 +144,15 @@ boolean hasDispatchers(EventClass eventClass) {
         return !dispatchers.isEmpty();
     }
 
+    /**
+     * Obtains a view {@code Set} of events that are known to this {@code EventBus}.
+     *
+     * <p>This set is changed when event dispatchers or handlers are registered or un-registered.
+     */
+    public Set<EventClass> registeredEventClasses() {
```
armiol: Is this a part of an internal or public API? Looks very internal to me, similar to `hasDispatchers(..)` above.

server/src/main/java/io/spine/server/event/EventDispatcherRegistry.java
```diff
@@ -66,6 +66,17 @@ public void unregister(EventDispatcher<?> dispatcher) {
         return super.getDispatchersForType(messageClass);
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Overrides to expose the method to
+     * {@link EventBus#registeredEventClasses() EventBus}.
```
armiol: The contents of this line fit the previous one.

server/src/test/java/io/spine/server/bc/given/Given.java
```diff
@@ -53,4 +66,56 @@ public static BcProjectStarted projectStarted(ProjectId id) {
                                    .build();
         }
     }
+
+    public static class NoopCommandDispatcher implements CommandDispatcher<ProjectId> {
```
armiol: We usually use `NoOp...` spelling, not `Noop`. I can see just a single `Noop` usage in the project (see `NoopQueryProcessor`), which also needs to be addressed (maybe you could help?)

server/src/main/java/io/spine/server/BoundedContextBuilder.java
```diff
@@ -302,6 +338,84 @@ boolean hasRepository(Repository<?, ?> repository) {
...
+                                                              .equals(entityClass));
+        return result;
+    }
+
+    /**
+     * Adds the passed command dispatcher to the dispatcher registration list which will be
+     * processed after the Bounded Context is created.
+     */
+    @CanIgnoreReturnValue
+    public BoundedContextBuilder addCommandDispatcher(CommandDispatcher<?> commandDispatcher) {
```
armiol: The tricky business here is that repositories are also `CommandDispatcher`s and `EventDispatcher`s. 

Let's make it impossible to use `[add | remove | has ]...Dispatcher(...)` and ` API with the `...Repository`-typed arguments.

dmitrykuzmin:
@armiol PTAL again.

dmitrykuzmin:
@armiol PTAL again.

PR #1101, opened 2019-06-26
Add entity class name to system events

dmdashenkov:
The entity class name is at handy when it comes to diagnostic reports.


dmdashenkov:
@armiol, PTAL.

PR #1104, opened 2019-06-28
Kill dispatching exceptions

dmdashenkov:
Before this PR, all the errors caused by message dispatching were thrown as exceptions. Exception handling was distributed between various `onError` methods as well as between different levels of abstraction, from buses to transactions.

Now, all the dispatching errors are converted into system events. Users may choose to subscribe to those events for diagnostics and recovery.

## System Events

There are 4 new types of system events:
 1. `HandlerFailedUnexpectedly` is the most common one. It is emitted at any generic error which occurs during a message handling, including runtime exceptions in handler methods, errors in the framework code, and errors in integrating user code and the framework code. Non-entity constraint violations also lead to `HandlerFailedUnexpectedly` events (while entity constraint violations cause `ConstraintViolated` events to be emitted).
 2. `RoutingFailed` event is emitted when the user code which routes messages to entity instances fails with an exception. The main difference between this event and `HandlerFailedUnexpectedly` is that the framework does not know the ID of the target entity; thus, the data included in the event payload is a bit different.
 3. `CannotDispatchEventTwice` and 4. `CannotDispatchCommandTwice` are events emitted when a duplicate event or command is dispatched to an entity. These events can be emitted at the `Inbox` level, i.e. before the delivery.

## Message Propagation 

Previously, when propagating a message through the system, the control flow heavily relied on the exceptions to be thrown when something does wrong. Now, a method dispatching a message would not throw an exception even if dispatching fails. That's why we introduce `PropagationOutcome`, a structure which aggregates an outcome of a message propagation, successful or otherwise.

A successful propagation may result in events, a rejection, or commands. Those results are posted into the respective buses on the `MessageEndpoint` level. Previously, commands used to be posted from within the entity level. Now, they are processed side-by-side with the events.

An unsuccessful propagation is described with a `spine.core.Error`. `Error`s are converted into system events and posted into the system Event Bus.

When many events are processed in a single transaction, one of them may fail and halt processing. Events which were not sent for processing because another event had failed result in `interrupted` outcomes.

Some messages may be forced on an entity by measures beyond normal propagation (such as catch-up processes). If an event does not belong to an entity, the entity may ignore it. Such processing is reflected in the `PropagationOutcome` as the `ignored` state.

### `MessageDispatcher` contract

For quite a long time, `MessageDispathcer`s returned the IDs of the target entities after dispatching a message. Those values were never used in production code and sometimes had to be artificially manufactured (e.g. in abstract message handlers). Now, `MessageDispatcher.dispatch` does not return anything.

`MessageDispatcher.onError` was deleted, since all the errors are now broadcast via system Event Bus.


dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL.

I'm planning on improving the coverage in further PRs.

armiol:
@dmdashenkov 
```
There are 4 new types of system events:
```
But there are three points on the list.

armiol:
@dmdashenkov please rephrase this one:
```
 The events which were not processed because of the premature handling abortion result in the interrupted outcomes.
```

alexander-yevsyukov:
`PropagationOutcome` also needs renaming. We already have a good common word for commands and events — `Signal`. Why don't we use it in this context? How about `SignalOutcome` or `SignalResult`, or `SignalDispatchingResult`?

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -135,7 +131,7 @@ protected BoundedContext(BoundedContextBuilder builder) {
      */
     protected final void init() {
         eventBus.init(this);
-        tenantIndex.registerWith(this);
+        tenantIndex.init(this);
```
alexander-yevsyukov: This reads as if `TenantIndex` initializes the `BoundedContext.

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -201,7 +197,7 @@ public static BoundedContextBuilder multitenant(String name) {
     public <I, E extends Entity<I, ?>> void register(Repository<I, E> repository) {
         checkNotNull(repository);
         Security.allowOnlyFrameworkServer();
-        repository.injectContext(this);
+        initialize(repository);
```
alexander-yevsyukov: How about just `init()`?

server/src/main/java/io/spine/server/ContextAware.java
```diff
@@ -0,0 +1,67 @@
...
+     * Initializes this instance as a part of the given Bounded Context.
+     *
+     * <p>This method should be only called once. However, it is allowed to initialize a single
+     * instance many times with the <strong>same</strong> Bounded Context. If the same instance is
+     * initialized with different Contexts, a runtime exception is thrown.
+     *
+     * @param context
+     *         the Context to which this instance belongs
+     */
+    void init(BoundedContext context);
```
alexander-yevsyukov: Once again, it would be better to name it `injectContext()`, or `register()`, or (got forbid) `setContext()`. It would read better on the usage side.

server/src/main/java/io/spine/server/VisibilityGuard.java
```diff
@@ -71,7 +72,7 @@ private void checkNotAlreadyRegistered(Class<? extends Message> stateClass) {
         if (alreadyRegistered != null) {
             throw newIllegalStateException(
                     "A repository for the state class %s already registered: %s",
```
alexander-yevsyukov: Please add backticks for the code related parts of the message and the period at the end of the sentence.

server/src/test/java/io/spine/server/model/contexts/tasks/CreationRetry.java
```diff
@@ -26,16 +26,17 @@
...
+import static com.google.common.collect.Sets.newHashSet;
+
 /**
  * A {@link io.spine.server.command.Commander Commander} which retries the task
  * creation if it fails.
  */
 public final class CreationRetry extends AbstractCommander {
 
-    private static final Set<TaskId> rejectedTasks = new HashSet<>();
+    private static final Set<TaskId> rejectedTasks = newHashSet();
```
alexander-yevsyukov: Google considers these static factory method as deprecated since Java 7.

dmdashenkov:
@alexander-yevsyukov, PTAL again. Let's discuss the leftover comments vocally.

PR #1105, opened 2019-07-01
Simplify server environment config

alexander-yevsyukov:
This PR continues improvements for the configuration of server environment features of an app.

In particular:
 * `EventStore` made an interface, instances of which a `StorageFactory` creates. Default implementation creates `DefaultEventStore`, which is based on `Repository`.
 * `TransportFactory` is not configured via `ServerEnvironment`.
 * `EventBus` creation no longer requires setting `EventStore` or `StreamExecutor`.


server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -85,11 +86,23 @@
      */
     private @Nullable StorageFactory productionStorageFactory;
 
+    /**
+     * The storage factory for tests.
+     *
+     * <p>If not configured {@link InMemoryStorageFactory} will be used.
```
armiol:
```
If not configured,
```

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -253,35 +265,85 @@ public void configureTracing(TracerFactory tracerFactory) {
...
      * @throws NullPointerException
      *         if the production {@code StorageFactory} was not
-     *         {@linkplain #configureProductionStorage(StorageFactory) configured} prior to the call
+     *         {@linkplain #configureStorage(StorageFactory) configured} prior to the call
      */
     public StorageFactory storageFactory() {
-        if (Environment.getInstance().isTests()) {
-            return InMemoryStorageFactory.newInstance();
+        if (environment().isTests()) {
+            if (storageFactoryForTests != null) {
```
armiol: I would say
```java
            if (storageFactoryForTests == null) {
                this.storageFactoryForTests = InMemoryStorageFactory.newInstance();
            }
            return storageFactoryForTests;
```
That's one line shorter.

server/src/test/java/io/spine/server/ServerEnvironmentTest.java
```diff
@@ -205,6 +210,67 @@ void testsFactory() {
...
+
+        @Test
+        @DisplayName("throw NPE if not configured")
+        void throwsIfNotConfigured() {
+            assertThrows(NullPointerException.class, serverEnvironment::transportFactory);
+        }
+
+        @Test
+        @DisplayName("do not allow setting `InMemoryTransportFactory`")
+        void noInMemProdTransport() {
```
armiol: Actually, we may use in-memory transport in production. That works flawlessly if the bounded contexts are deployed on the same node.

Let's discuss in person.

PR #1107, opened 2019-07-05
Move `InboxStorage` to `StorageFactory`

armiol:
This PR makes the `InboxStorage` instances to be produced by `StorageFactory`. 

Now, by default, the `Delivery` is initialized with the `InboxStorage` created by the server's `StorageFactory`. The framework users are still able to provide the `Delivery` with a custom `InboxStorage` via `Delivery.Builder` though.

To make the Google Cloud-specific and JDBC-backed implementations of `InboxStorage` easier, the API of `InboxStorage` was simplified and the intermediate `ShardedStorage` class was eliminated as redundant.

As a side note, the circular dependency between `Delivery` and `StorageFactory` (one wanting another upon the initialization) was resolved.


server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -174,7 +175,10 @@ public void configureDelivery(Delivery delivery) {
      * <p>Unless {@linkplain #configureDelivery(Delivery) updated manually}, returns
      * a {@linkplain Delivery#local() local implementation} of {@code Delivery}.
      */
-    public Delivery delivery() {
+    public synchronized Delivery delivery() {
+        if(delivery == null) {
```
dmdashenkov: Please fix code alignment.

server/src/main/java/io/spine/server/storage/StorageFactory.java
```diff
@@ -79,6 +80,13 @@
     createProjectionStorage(ContextSpec context,
                             Class<? extends Projection<I, ?, ?>> projectionClass);
 
+    /**
+     * Creates a new {@link InboxStorage}.
+     *
+     * @param multitenant whether the created storage should be multi-tenant.
```
dmdashenkov: Please remove the `.` at the end.

PR #1108, opened 2019-07-05
Configure `BoundedContext` only through builder

alexander-yevsyukov:
This PR restricts the ways a `BoundedContext` can be configured, making it possible only via `BoundedContextBuilder`.

Notable changes:
  * calls to `BoundedContext.registerXxx()` methods (which are now `@Internal`) are now only allowed for server-side framework classes. This includes testing  utilities. Calling from outside the framework or from non-server framework classes will result in `SecurityException`.
  * `EventEnricher`, bus listeners and filters are now configured via `BoundedContextBuilder`.
  * It is no longer possible to configure `CommandBus` and `EventBus` separately and pass it to `BoundedContextBuilder`.
  * Security check calls migrated to `io.spine.security.InvocationGuard`. `io.spine.server.security.InvocationGuard` was simplified to handle only server-related security aspects (and renamed to `Security`).
  * `Bus` classes made `@Internal`.


server/src/main/java/io/spine/server/BoundedContextBuilder.java
```diff
@@ -254,6 +218,27 @@ public BoundedContextBuilder addCommandDispatcher(CommandDispatcher<?> commandDi
...
+     */
+    public BoundedContextBuilder addCommandListener(Consumer<CommandEnvelope> listener) {
+        checkNotNull(listener);
+        commandBus.addListener(listener);
+        return this;
+    }
+
     /**
      * Adds the passed event dispatcher to the dispatcher registration list which will be processed
      * after the Bounded Context is created.
```
armiol: The issue is located below, but I am not able to leave my comment there.

IDEA detects some code duplicates with this piece:
```java
        if (eventDispatcher instanceof Repository) {
            return add((Repository<?, ?>) eventDispatcher);
        }
        eventDispatchers.add(eventDispatcher);
        return this;
```

server/src/main/java/io/spine/server/entity/Repository.java
```diff
@@ -270,8 +268,7 @@ protected final BoundedContext context() {
     }
 
     /**
-     * The callback called by a {@link BoundedContext} during the {@linkplain
-     * BoundedContext#register(Repository) registration} of the repository.
+     * The callback called by a {@link BoundedContext} when adding the repository.
```
armiol: `is invoked` would look better to me.

server/src/main/java/io/spine/server/event/EventBus.java
```diff
@@ -371,4 +375,23 @@ public UnsupportedEventException handle(EventEnvelope event) {
             return exception;
         }
     }
+
+    /**
+     * Precondition check for objects depending on a reference to {@code EventBus}.
```
armiol: Please start the description with a verb in its 3rd form.

server/src/main/java/io/spine/server/security/Security.java
```diff
@@ -0,0 +1,79 @@
...
+package io.spine.server.security;
+
+import io.spine.code.java.PackageName;
+import io.spine.server.Server;
+import io.spine.system.server.SystemContext;
+
+import static java.lang.String.format;
+
+/**
+ * Controls which class can call a method.
```
armiol: `is allowed` would be closer in meaning. Every class _can_ call a method, but hardly every is allowed to.

server/src/test/java/com/example/OutsideClass.java
```diff
@@ -0,0 +1,39 @@
...
+import io.spine.server.security.GivenRestrictedApi;
+import io.spine.server.security.Security;
+
+/**
+ * A test environment class from outside the framework package structure for testing
+ * {@link Security#allowOnlyFrameworkServer()}.
+ *
+ * @see io.spine.server.security.InvocationGuardTest.ServerFramework#prohibitingFromOutside()
+ */
+public final class OutsideClass {
```
armiol: For these classes, I would use `Foreign` prefix.

server/src/test/java/io/spine/server/event/EventBusStandIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
...
+        Stand stand = context.stand();
+
+        eventClasses().forEach(eventClass ->
+                assertThat(eventBus.dispatchersOf(eventClass))
+                        .doesNotContain(stand)
+        );
+    }
+
+    @Test
+    @DisplayName("has `Stand` as the dispatcher of events of the registered repository")
```
armiol: `have`. Since it goes after `should`.

server/src/main/java/io/spine/server/security/Security.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public final class Security extends SecurityManager {
```
armiol: Just in case, `base` has `io.spine.security.CallerProvider`, which is also a `SecurityManager`. 

Please consider making that class `public` and `@Internal` and extend it in `core-java`.

server/src/main/java/io/spine/server/bus/BusBuilder.java
```diff
@@ -78,9 +79,9 @@ protected BusBuilder() {
      * <p>The order of appending the filters to the builder is the order of the filters in
      * the resulting bus.
      *
-     * @param filter the filter to append
+     * @param filter the filter to add
      */
-    public final B appendFilter(BusFilter<E> filter) {
+    public final B addFilter(BusFilter<E> filter) {
```
dmdashenkov: Append seems more descriptive to me. While we can `add` elements to a set, `append` implies ordering.

server/src/main/java/io/spine/server/bus/Listeners.java
```diff
@@ -51,4 +52,9 @@ public void accept(E envelope) {
             }
         });
     }
+
+    @VisibleForTesting
+    boolean contains(Consumer<E> listener) {
+        return listeners.contains(listener);
```
dmdashenkov: Can we create a distinct interface for a listener? Maybe not the scope of this PR though.

A `Consumer` seems rather raw and generic. `interface Listener<E> extends Consumer<E>` would at least give us a place to explain the purpose of a listener.

server/src/main/java/io/spine/server/security/Security.java
alexander-yevsyukov: I'd rather not expose security aspects for the security reasons :)

PR #1110, opened 2019-07-09
Delete unit test framework

dmdashenkov:
This PR removes the entity unit test framework in favour of the `BlackBoxBoundedContext`-based tests.

In `core-java`, we only used the framework to check `MirrorProjection` behaviour. Now we use the black box approach instead.

This PR unloads the changes from #1104.


dmdashenkov:
@armiol, PTAL.

PR #1115, opened 2019-07-16
System feature configuration

dmdashenkov:
This PR introduces API for configuring System context features.

Previously, all the capabilities of the System context used to be mandatory for all users.
While all of the features of the System context may be useful to some users, other users may live completely without them. That said, users should be cautious to turn off system features, as they might still need the information provided by a given system capability later.

Users may configure:
 - storing system events (_OFF by default_);
 - writing a `CommandLog` for each domain command (_OFF by default_);
 - mirroring domain Aggregates (_ON by default_).

System events might be required later to catch up domain projections (for instance, if they rely on entity state updates).
Command logs might be required for audits or debugging. This is the only place in the system where the actual commands, and not their outcome, are stored.
Mirrors allow querying domain aggregates as efficiently as they were projections, i.e. without replaying recent events. Without mirrors, domain aggregates cannot be queried via `QueryService`.

The reason to turn these features off is to improve the performance of the configured `Delivery` and to reduce the amount of stored data.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/system/server/SystemFeatures.java
```diff
@@ -0,0 +1,161 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.system.server;
+
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import io.spine.annotation.Internal;
+
+/**
+ * System bounded context feature configuration.
```
armiol: Please rephrase so that the relations are clear.

server/src/main/java/io/spine/system/server/SystemFeatures.java
```diff
@@ -0,0 +1,161 @@
...
+     *
+     * @return self for method chaining
+     */
+    public SystemFeatures forgetEvents() {
+        this.storeEvents = false;
+        return this;
+    }
+
+    /**
+     * Obtains the Command log setting.
```
armiol: Let's have the spelling of `Command log` (`@link CommandLog` above, etc) consistent.

server/src/main/java/io/spine/system/server/SystemFeatures.java
```diff
@@ -0,0 +1,161 @@
...
+     */
+    private SystemFeatures() {
+    }
+
+    /**
+     * Obtains the default configuration.
+     *
+     * <p>By default, the system context:
+     * <ol>
+     *     <li>Stores Aggregate mirrors for querying.
```
armiol: Let's speak of this as of `Enables querying of the latest {@code Aggregate} states.`

And I would name the `aggregateMirrors` field in the same manner.

dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@alexander-yevsyukov, PTAL.

server/src/main/java/io/spine/server/ContextSpec.java
```diff
@@ -38,32 +40,34 @@
 
     private final BoundedContextName name;
     private final boolean multitenant;
+    private final boolean storeEvents;
 
-    private ContextSpec(BoundedContextName name, boolean multitenant) {
+    private ContextSpec(BoundedContextName name, boolean multitenant, boolean storeEvents) {
```
alexander-yevsyukov: The `storeEvents` parameter looks weird and not native for a regular `BoundedContext`. What a context does not store events? It can be equally named `isSystem`... :) This looks like the idea of context hierarchy and its properties leaked into here.

server/src/main/java/io/spine/server/storage/system/SystemAwareStorageFactory.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Creates a new {@link EventStore}.
+     *
+     * <p>If the given context does not store events, returns an {@link EmptyEventStore}. Otherwise,
+     * delegates to the wrapped instance.
+     */
+    @Override
+    public EventStore createEventStore(ContextSpec context) {
+        return context.storesEvents()
```
alexander-yevsyukov: Once again, there is a conflict between the name `SystemAware...` and non-system-awareness in the `storesEvents()`. Why not just have `isSystem()` then?

server/src/main/java/io/spine/system/server/SystemFeatures.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Obtains the system events persistence setting.
+     *
+     * @return {@code true} if system events should be stored, {@code false} otherwise
+     */
+    @Internal
+    public boolean includePersistentEvents() {
+        return storeEvents;
+    }
+}
```
alexander-yevsyukov: `equals()`, `hashCode()`, `toString()`?

alexander-yevsyukov: On the other note, if we were building a server environment in another language, could it be a `Message`-based type? Don't we describe a `System` context after all?

PR #1116, opened 2019-07-16
Command reaction on external events

dmitrykuzmin:
This PR enables command reaction on the external events for the commander entities.

The external command reaction method can be declared as follows:
```java
@Command(external = true)
StartProject handle(ProjectScheduled event) {
    // ...
}
```

The "external" command substitution methods, although technically possible as a side effect, are caught by the `ModelVerifier` and counted as errors. This is done to avoid confusion, as currently there are no "external" commands in the system.

So the method like
```java
@Command(external = true)
StartProject handle(ScheduleProject command) {
    // ...
}
```
won't pass the `Model` verification, instead, the `ExternalCommandReceiverMethodError` error will be thrown.

### DIFF
--- a/server/src/main/java/io/spine/server/command/model/CommanderClass.java
+++ b/server/src/main/java/io/spine/server/command/model/CommanderClass.java
@@ -102,7 +102,7 @@ public final class CommanderClass<C extends Commander>
     }
 
     /**
-     * Makes sure no command substitution methods are marked as
+     * Ensures no command substitution methods are marked as
      * {@linkplain io.spine.server.command.Command#external()} external} in the class.
      *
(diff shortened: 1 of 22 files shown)


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/command/model/CommanderClass.java
```diff
@@ -96,4 +100,26 @@ public boolean producesCommandsOn(EventClass eventClass) {
         SetView<CommandClass> result = union(commandOutput(), delegate.producedTypes());
         return result;
     }
+
+    /**
+     * Makes sure no command substitution methods are marked as
```
armiol: `Ensures`

armiol:
@dmitrykuzmin please see my comment.

PR #1117, opened 2019-07-17
Query service behaviour

dmdashenkov:
This PR addresses 2 issues with the query-side behaviour:
 1. When sending a query with no filters and without specifying to load _all_ the entities explicitly, the response is now an error. Previously, we loaded all the entities.
 1. When sending a query for all the entities, the ordering and result count limit are now regarded. Previously, they were ignored.

This PR also brings a change in the `Query` internal structure.


dmdashenkov:
@armiol, PTAL.
@alexander-yevsyukov, PTAL.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -236,29 +245,10 @@ protected E findOrCreate(I id) {
...
-     *
-     * <p>Note: The storage must be assigned before calling this method.
-     *
-     * @param ids
-     *         entity IDs to search for
      * @param fieldMask
-     *         mask to apply on entities
-     * @return all the entities in this repository with the IDs contained in the given {@code ids}
-     * @see #loadAll(Iterable)
+     *         the entity state fields to load; note that the resulting state must be valid,
```
armiol: I would move the clause about the resulting state expected to be valid to a new paragraph in the method-level documentation.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -269,37 +259,28 @@ protected E findOrCreate(I id) {
...
-        Iterator<EntityRecord> records = loadAllRecords();
-        Function<EntityRecord, E> toEntity = storageConverter().reverse();
-        Iterator<E> result = transform(records, toEntity::apply);
-        return result;
-    }
-
     /**
      * Obtains iterator over all present {@linkplain EntityRecord entity records}.
      *
+     * <p>The resulting entity states have only the specified in the {@code mask} fields. However,
```
armiol: As long as this method does not take `{@code mask}` argument, let's rephrase.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -269,37 +259,28 @@ protected E findOrCreate(I id) {
...
     }
 
     /**
      * Finds the entities passing the given filters and applies the given {@link FieldMask}
-     * to the results. A number of elements to retrieve can be limited by {@link Pagination}.
-     * OrderBy in which to look for and return results in is specified by the {@link OrderBy}.
+     * to the results.
+     *
+     * <p>A number of elements to retrieve can be limited to a certain number. Order in
+     * which to look for and return results in is specified by the {@link OrderBy}.
```
armiol: Please remove `in` (it's the second one in this sentence and therefore redundant).

PR #1119, opened 2019-07-19
Remove generic parameter from `MessageDispatcher`

dmdashenkov:
Since recently, `MessageDispatcher.dispatch(..)` does not return anything.

This allows us to remove one of the generic parameters of the `MessageDispatcher`.

This PR removes the generic parameter from `MessageDispatcher` and all its descendants.


dmdashenkov:
@armiol, PTAL.

PR #1120, opened 2019-07-22
Post mutual requests for external messages

dmdashenkov:
This PR resolves the issue with the `IntegrationBus`. Previously, when two repositories were registered sequentially in respective (distinct) contexts, only the first one received the `RequestForExternalMessages` from the second, but not vice-versa. Consequently, only events from the first to second repositories could be dispatched.

Since this PR, `IntegrationBus` sends a `RequestForExternalMessages` when it received such a request from another context. This enables contexts to exchange their required types.

Also, in this PR we tune the in-mem implementation of transport, so that all the memory structures holding references to different context are cleared on `close()`.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/integration/ConfigurationBroadcast.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Notifies other Bounded contexts about a change in the requested messages.
+     *
+     * <p>If the given {@code needs} are the same as previous ones, i.e. the needs did not change,
+     * the request is not sent.
+     *
+     * @param needs
+     *         the new needs of current context
+     */
+    synchronized void onNeedsUpdated(Set<ChannelId> needs) {
```
armiol: This is unclear. Why would channel IDs be the needs of the Bounded Context? To me, the needs of a context are the types of `Message`s. I.e. "I am `Projects` context and I need `UserCreated` messages".

Could you please rename?

server/src/main/java/io/spine/server/integration/ConfigurationChangeObserver.java
```diff
@@ -27,21 +27,31 @@
...
+    private final IntegrationBus integrationBus;
     private final BoundedContextName boundedContextName;
     private final Function<Class<? extends Message>, BusAdapter<?, ?>> adapterByClass;
 
+    /**
+     * Names of Bounded Contexts already known to this observer.
+     *
+     * <p>If a context is unknown, the observer publishes a {@code RequestForExternalMessages}.
+     */
+    private final Set<BoundedContextName> knownContexts = synchronizedSet(new HashSet<>());
```
armiol: Let's define the `hashCode()` for the `BoundedContextName`.

dmdashenkov:
@armiol, PTAL again.

server/src/main/java/io/spine/server/transport/MessageChannel.java
```diff
@@ -42,4 +39,20 @@
...
+
+    /**
+     * Obtains the type of the messages transferred through this channel.
+     */
+    default TypeUrl targetType() {
+        ChannelId channelId = id();
+        return TypeUrl.parse(channelId.getTargetType());
+    }
+
+    static ChannelId channelIdFor(TypeUrl messageType) {
```
armiol: Please document.

PR #1121, opened 2019-07-22
Performance Improvements

armiol:
This PR improves the performance of the `core-java` modules basing on the speed of `DeliveryTest` integration tests.

Highlights:

- Some of the most called but redundant `vBuild()` calls switched to `build()`.
- The most often validated messages are made `FieldAwareMessage`s and their `readValue(..)` was implemented to avoid reflective access during the validation.
- Guava's `Cache` was used to reduce the CPU and memory footprint for the commonly asked/created data (e.g. `BoundedContextName` for a raw `Class`, which was an expensive lookup).


core/src/main/java/io/spine/core/CommandIdMixin.java
```diff
@@ -20,16 +20,26 @@
...
+interface CommandIdMixin extends SignalId, FieldAwareMessage, CommandIdOrBuilder {
 
     @Override
     default String value() {
         return getUuid();
     }
+
+    @Override
+    default Object readValue(Descriptors.FieldDescriptor field) {
+        if(0 == field.getIndex()) {
```
dmdashenkov: Please apply autoformatting.

core/src/main/java/io/spine/core/CommandMixin.java
```diff
@@ -88,4 +91,20 @@ default boolean isScheduled() {
         }
         return false;
     }
+
+    @Override
+    default Object readValue(Descriptors.FieldDescriptor field) {
+        switch (field.getIndex()) {
+            case 0:
```
dmdashenkov: Can we use named constants for field numbers instead?

core/src/main/java/io/spine/core/EventIdMixin.java
```diff
@@ -20,16 +20,26 @@
...
+interface EventIdMixin extends SignalId, FieldAwareMessage, EventIdOrBuilder {
 
     @Override
     default String value() {
         return getValue();
     }
+
+    @Override
+    default Object readValue(Descriptors.FieldDescriptor field) {
+        if(0 == field.getIndex()) {
```
dmdashenkov: Please apply autofiormatting.

server/src/main/java/io/spine/server/aggregate/AggregateEndpoint.java
```diff
@@ -177,7 +180,7 @@ private DispatchOutcome applyProducedEvents(A aggregate, DispatchOutcome command
             event.getContextBuilder()
                  .setVersion(signalVersion);
         }
-        return correctedCommandOutcome.vBuild();
+        return correctedCommandOutcome.build();
```
dmdashenkov: Please use `buildPartial` here and in similar cases. We should explicitly state the absence of validation.

server/src/main/java/io/spine/server/dispatch/ProducedCommandsMixin.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.protobuf.Descriptors.FieldDescriptor;
+import io.spine.annotation.GeneratedMixin;
+import io.spine.validate.FieldAwareMessage;
+
+/**
+ * A mixin interface for the {@link ProducedCommands} message type.
+ */
+@GeneratedMixin
+public interface ProducedCommandsMixin extends ProducedCommandsOrBuilder, FieldAwareMessage {
```
dmdashenkov: Please make generated mixins package-private when possible.

server/src/main/proto/spine/server/dispatch/dispatching.proto
```diff
@@ -133,6 +139,7 @@ message Interruption {
 
 // Ignored signal handling.
 message Ignore {
+    option (is).java_type = "IgnoreMixin";
```
dmdashenkov: Let's consider making this a convention and removing the need to add the option if the message type implements an interface named in the same way as the type itself but with the `Mixin` suffix.

If I remember correctly, we already pass the whole classpath to our `protoc` plugin. We might as well check if there is such an interface.

Although, this is, most likely, not a v1.0 feature.

dmdashenkov: Also, generating messages to implement `FieldAwareMessage` is fairly easy at this point.

server/src/main/java/io/spine/server/aggregate/AggregateEndpoint.java
armiol: I view this differently. Let's discuss in person.

alexander-yevsyukov: The absence of `v` prefix does this.

core/src/main/java/io/spine/core/CommandMixin.java
armiol: We would have to depend on the descendants (i.e. on the classes, which are marked by this mixin interface), as the `...OrBuilder` has no such notion. So, unfortunately, no.

server/src/main/java/io/spine/server/dispatch/ProducedCommandsMixin.java
armiol: Done.

server/src/main/proto/spine/server/dispatch/dispatching.proto
armiol: As discussed in person, we would rather invest in a generation of a fully-blown in-place validation.

PR #1124, opened 2019-07-24
Assert `Query` execution result in `BlackBoxBoundedContext`

dmitrykuzmin:
This PR adds an `assertQueryResult(Query)` method to the `BlackBoxBoundedContext`.

The method runs a query within the context and then provides a `Subject` for assessing its results.

The provided subject (`QueryResultSubject`) by itself is an `IterableOfProtosSubject` where the object being tested is an `Iterable` of received entity states (in the form of `Message`s).

The subject thus can do the checks as follows:

```
context.assertQueryResult(query)
       .comparingExpectedFieldsOnly()
       .containsExactly(expectedState1, expectedState2)
```

The `QueryResultSubject` also provides some number of convenience methods for checking response status, received entity versions, etc.


dmitrykuzmin:
@armiol PTAL.

testutil-server/src/main/java/io/spine/testing/server/blackbox/verify/query/QueryResultSubject.java
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Creates a new instance of the subject.
+     *
+     * <p>Unlike other {@code Subject}s, the {@code QueryResultSubject} does not accept
+     * {@code null} arguments, as {@code null} {@code QueryResponse} always indicates an error.
+     */
+    public static
+    QueryResultSubject assertQueryResult(QueryResponse queryResponse) {
+        checkNotNull(queryResponse, "`QueryResponse` must never be `null`");
```
armiol: We put periods at the end of the error messages which are sentences.

PR #1125, opened 2019-07-25
Unify query and storage names of EntityColumns

dmdashenkov:
Previously, an entity column had 2 different names:
 - a stored name — the name used to persist the column value in a database;
 - a query name — the name used to reference the column in a query.

This approach had its drawbacks. Primarily, there was a need in conversion between the two names when querying an entity.

Now, each entity column has a single name for both querying and persistence.


dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL.

PR #1126, opened 2019-07-26
Assert incoming `Subscription` updates in `BlackBoxBoundedContext`

dmitrykuzmin:
This PR adds an `assertSubscriptionUpdate(Topic, Consumer<ProtoSubject>)` method to the `BlackBoxBoundedContext`.

The method allows to test the updates which are received for the specified `topic`. The second method argument is a consumer of an `entityState`/`eventMessage` check done with the `ProtoSubject` means which is run for every received update.

The example of method usage:
```java
VerifyingCounter updateCounter =
        context.assertSubscriptionUpdates(
                topic,
                assertEachReceived -> assertEachReceived.comparingExpectedFieldsOnly()
                                                        .isEqualTo(expected)
        );
context.receivesCommand(createProject); // Some command creating the `expected`.
updateCounter.verifyEquals(1);
```

Please note that the return value allows to verify the number of updates received. It may be ignored, but then receiving no updates will be considered a valid result.


dmitrykuzmin:
@armiol PTAL.

Issue #925, opened 2018-12-21
Event from reaction is routed before the inititial event is dispatched to target


mdrachuk:
This leads to an issue, when there is an event (`UserSignedIn`) which creates another event as a reaction (`UserConsentRequested`) and they both are dispatched to a single entity (`SessionProjection`) with the first event (`UserSignedIn`) actually creating it, and the second event requiring a repository (`SessionRepository`) query to be routed to the correct target.

As of now, after routing an event to a projection, a `DispatchEventToSubscriber` command is posted to the System Bounded Context, after which the next event from queue gets dispatched. 

This means that upon dispatching two events from the `DispatchQueue`, when it comes to the second  event, the first one is not dispatched, but scheduled to be dispatched in the System Bounded Context.

alexander-yevsyukov:
@armiol, do you think this is a bug?

armiol:
@dmitrykuzmin can you please check if this is still reproducible?

armiol:
@dmitrykuzmin maybe we need an integration test on this matter.

dmitrykuzmin:
The issue can no longer be reproduced. 

A test for the described scenario was added in #1128.

armiol:
This behavior is now reproducible again. However, it does not look an outstanding issue, as the read-side by contract is eventually consistent. While in this scenario the code relies not only upon the ordering of the events dispatched to the projection (which is guaranteed by the framework) but onto the projection state changes that presumably should have taken place.

alexander-yevsyukov:
Closing this as a non-bug. There is a chance that a reaction event got into the delivery window earlier that the originating event. There is no generic remedy against this case. The code consuming the events should take into account the eventual consistency.

PR #1128, opened 2019-07-29
Add a test for event reaction product routing

dmitrykuzmin:
This PR addresses #925.

With the current Spine version the problem is not reproducible anymore.

This PR adds a test for the scenario described in the issue, to make sure the problem won't return again.

### DIFF
--- a/server/src/test/java/io/spine/server/route/EventRoutingIntegrationTest.java
+++ b/server/src/test/java/io/spine/server/route/EventRoutingIntegrationTest.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2019, TeamDev. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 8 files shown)


dmitrykuzmin:
@armiol PTAL.

server/src/test/java/io/spine/server/route/EventRoutingIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+                .setSessionId(sessionId)
+                .build();
+        Session session = Session
+                .newBuilder()
+                .setId(sessionId)
+                .setUserId(userId)
+                .setUserConsentRequested(true)
+                .build();
+
+        // If the routing of `UserConsentRequested` event is done before its origin
```
armiol: Please move this clause to the method-level documentation.

server/src/test/proto/spine/test/route/user/user.proto
```diff
@@ -0,0 +1,52 @@
...
+
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package="io.spine.test.event";
+option java_multiple_files = true;
+
+import "spine/core/user_id.proto";
+
+message User {
```
armiol: Let's name the messages with prefixes reflecting the test suite name. I.e. `RUser` (standing for "routing").

PR #1129, opened 2019-07-30
Fail on posting unsupported commands in `BlackBoxBoundedContext`

### LINKED ISSUE #1008: `BlackBoxBoundedContext` should fail on posting generated unsupported command
Now it simply swallows the fact. We do post a stacktrace of an exception to the console, like:

```
Caused by: java.lang.IllegalStateException: Command posting failed with status: error {
  type: "spine.core.CommandValidationError"
  code: 1
  message: "Commands of the type `spine.testing.server.blackbox.BbAssignScrumMaster` are not supported."
  attributes {
    key: "commandType"
    value {
      string_value: "spine.testing.server.blackbox.BbAssignScrumMaster"
    }
  }
}
.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:589)
	at io.spine.server.commandbus.CommandSequence.checkSent(CommandSequence.java:233)
	at io.spine.server.commandbus.CommandSequence.post(CommandSequence.java:197)
	at io.spine.server.commandbus.CommandSequence.postAll(CommandSequence.java:163)
```

But with a big number of tests, such diagnostics is not sufficient because surrounding tests usually do pass.

dmitrykuzmin:
This PR fixes #1008.

The `BlackBoxBoundedContext` will now fail on posting unsupported commands, both directly to the `receivesCommand(...)`/`receivesCommands(...)` methods and generated from the commanding entities (`@Command`).

For this, the `BlackBoxBoundedContext` subscribes to the `CommandErrored` event posted by the system.

The `CommandErrored` event unifies various kinds of errors that can happen to a command. Currently, we fail the test only in case of unsupported command, while leaving the rest of error types as some kind of valid test cases.

### DIFF
--- a/testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
+++ b/testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
@@ -21,6 +21,7 @@
 package io.spine.testing.server.blackbox;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.truth.extensions.proto.ProtoSubject;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.protobuf.Message;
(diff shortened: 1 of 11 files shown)


testutil-server/src/main/java/io/spine/testing/server/blackbox/UnsupportedGuard.java
```diff
@@ -0,0 +1,82 @@
...
+                    commandType));
+        commandType = null;
+    }
+
+    private static boolean isUnsupportedError(Error error) {
+        return COMMAND_VALIDATION_ERROR_TYPE.equals(error.getType())
+                && error.getCode() == UNSUPPORTED_COMMAND_VALUE;
+    }
+}
+
```
armiol: This line is redundant.

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
```diff
@@ -145,11 +152,19 @@
 
     private final MemoizingObserver<Ack> observer;
 
+    /**
+     * A guard that verifies that unsupported commands do not get posted to the
+     * {@code BlackBoxBoundedContext}.
+     */
+    private final UnsupportedGuard unsupportedGuard;
```
armiol: Right now it looks like a guard, which isn't supported for some reason.

Let's rename the class to include `Command`. E.g. `UnsupportedCommandGuard`.

Issue #984, opened 2019-03-08
Eliminate Mockito dependency in Testutil


alexander-yevsyukov:
We cannot make the methods of the `Repository` final:
  * `idClass()`,
  * `entityClass()`,
  * `entityStateType()`.

The Mockito dependency is bad by itself, but there are more problems we're having. For example, the class `AggregateMessageDispatcher` does this:
```java
    private static <I, A extends Aggregate<I, ?, ?>> AggregateRepository<I, A> mockRepository() {
        TestAggregateRepository mockedRepo = mock(TestAggregateRepository.class);
        when(mockedRepo.lifecycleOf(any())).thenCallRealMethod();
        when(mockedRepo.idClass()).thenReturn(Object.class);
        return mockedRepo;
    }
```
Returning `Object.class` for an ID is violation of our contract of identifiers.

alexander-yevsyukov:
@armiol, FYI.

armiol:
Resolved in #1132.

PR #1132, opened 2019-07-31
Remove Mockito dependency from Testutil

dmitrykuzmin:
This PR addresses #984.

The Mockito dependency and its usages were eliminated from the `testutil-core`, `testutil-client` and `testutil-server` modules.


dmitrykuzmin:
@armiol PTAL.

PR #1133, opened 2019-08-02
Add a test for `AggregateStorage` index uniqueness

dmitrykuzmin:
This PR adds a test which checks the uniqueness of aggregate IDs received from the `AggregateStorage.index()` method.

The test will be run for every `AggregateStorage` implementation.

The `InMemoryAggregateStorage` already works correctly and doesn't require any modifications.


dmitrykuzmin:
@armiol PTAL.

PR #1134, opened 2019-08-04
Update Subscription API

armiol:
This PR changes the API of subscriptions.

Previously, an active subscription was passing the updates of the entities which had matched the subscription target. However, it wasn't possible to tell, whether an entity had changed in a way it had stopped to match — in this case, the subscription callback was silent.

This PR brings a `bool` property telling that the entity no longer matches the target criteria, while it was matching those previously:

```proto
// The update happened to the entity of interest.
message EntityStateUpdate {

    // The entity ID packed as Any.
    google.protobuf.Any id = 1 [(required) = true];

    // The update may be one of two things.
    oneof kind {

        // The state of entity has been updated and packed here as `Any`.
        //
        // If the field mask is set in `Topic`, this value is masked.
        //
        google.protobuf.Any state = 2;

        // The state of the entity has been changed, so that previously
        // it matched the subscription `Target` filters, and now it does not.
        bool no_longer_matching = 3;
    }
}
```

In addition, the entity column API was marked `@Experimental`. The behavior of queries and subscriptions involving columns is still inconsistent. So this is decided to re-work in the future. See #1135.

Also, as a side change, `EntityStateChanged` now holds the "old" entity state in addition to the "new" one.

To avoid random build failures, the behavior of `DeliveryTest` was made deterministic.


server/src/main/java/io/spine/server/entity/storage/Column.java
```diff
@@ -40,9 +42,12 @@
  * <p>If there are repeated column names within an {@code Entity},
  * the exception will be raised when a repository serving the entity is added to
  * its {@code BoundedContext}.
+ *
+ * <p>Please note this feature is experimental.
```
dmdashenkov: How come? It's been here for years. What is it in entity columns that we're not sure about?

server/src/main/java/io/spine/server/entity/storage/ColumnRecords.java
```diff
@@ -32,6 +34,8 @@
  * A utility for dealing with the {@linkplain EntityRecordWithColumns} and
  * the {@linkplain EntityColumn entity columns}.
  */
+@Experimental
+@Internal
```
dmdashenkov: It is at least `SPI`. This utility is required by the storage implementations.

server/src/main/java/io/spine/server/stand/UpdateHandler.java
```diff
@@ -25,100 +25,153 @@
...
 import static io.spine.base.FieldPaths.getValue;
 import static io.spine.server.storage.OperatorEvaluator.eval;
 import static io.spine.util.Exceptions.newIllegalArgumentException;
 
 /**
- * Matches the incoming event against a subscription criteria.
- *
- * <p>The class descendants decide on how to turn the event envelope into a set of data pieces
- * suitable for filtering.
+ * Tells which {@linkplain SubscriptionUpdate subscription updates} an {@code Event}s occurred
```
dmdashenkov: Please review the grammar. Also, please consider simplifying this sentence. Maybe it's because of all the Javadoc tags, but I had to read it a couple of times to get the meaning.

server/src/main/java/io/spine/server/stand/UpdateHandler.java
```diff
@@ -25,100 +25,153 @@
...
-                .getFilterList()
-                .stream()
-                .allMatch(f -> checkPasses(message, f));
-        return result;
+    boolean includeAll() {
+        return target().getIncludeAll();
+    }
+
+    /**
+     * "Activates" this callback with a given action.
```
dmdashenkov: I'm concerned with the `Handler` name. Maybe callback would be better.

dmdashenkov: Why use quotation marks? Also, why not to call this method `activate`?

server/src/main/java/io/spine/server/stand/EntityUpdateHandler.java
```diff
@@ -0,0 +1,172 @@
...
+import io.spine.protobuf.AnyPacker;
+import io.spine.server.type.EventEnvelope;
+import io.spine.system.server.event.EntityStateChanged;
+
+import java.util.Optional;
+
+/**
+ * The update handler of {@code Subscription}s for {@code Entity} state updates.
+ */
+class EntityUpdateHandler extends UpdateHandler {
```
alexander-yevsyukov: Can we make it `final`?

server/src/main/java/io/spine/server/stand/EventUpdateHandler.java
```diff
@@ -20,27 +20,66 @@
...
+import io.spine.type.TypeUrl;
+
+import java.util.Optional;
 
 /**
- * Updates an event subscription based on the incoming event.
+ * The update handler of {@code Subscription}s for {@code Event}s.
  */
-final class EventSubscriptionCallback extends SubscriptionCallback {
+class EventUpdateHandler extends UpdateHandler {
```
alexander-yevsyukov: It used to be `final`, and I cannot find classes derived from it. Can we make it `final` too?

server/src/main/java/io/spine/server/entity/storage/Column.java
armiol: Please see #1135.

server/src/main/java/io/spine/server/entity/storage/ColumnRecords.java
armiol: Same as above. We don't want SPI-users to rely on it at this point.

server/src/main/java/io/spine/server/stand/UpdateHandler.java
armiol: There is a `callback` in this task already. Make sure you have reviewed the code.

Please suggest something more meaningful, if you can. I will think of it as well, but so far that's the best from simple names.

PR #1138, opened 2019-08-07
`AggregateStorage.index()` using `MirrorRepository`

dmitrykuzmin:
This PR optimizes the work of `AggregateStorage.index()` method.

From now on it will try to get aggregate IDs from the `MirrorRepository`, where they are already stored in a convenient form for reading.

If, for some reason, the aggregate mirror is inaccessible, the `index()` method will fall back to the old implementation, where the distinct aggregate IDs are extracted from the huge amount of event records.

The version is now `1.0.1-SNAPSHOT`. The `base` and `time` versions used are `1.0.1`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.0.0`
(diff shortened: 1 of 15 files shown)


dmitrykuzmin:
@armiol PTAL.

PR #1139, opened 2019-08-07
Synchronize subscription registry

dmdashenkov:
This PR tackles the issue appearing in the `MultitenantSubscriptionRegistry`. Despite declaring all its methods as `synchronized`, the registry was _not_ thread-safe.

In this PR we:
 1. Make the `MultitenantSubscriptionRegistry` thread-safe via concurrent collections and thoughtful synchronization.
 2. Start logging errors which occur while dispatching a message in a Bus, since that is a potential insertion.
 3. Start logging errors which occur while delivering an entity update to subscribers, for the same reason.

The framework version is set to `1.0.1`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -457,7 +457,7 @@
  The dependencies distributed under several licenses, are used according their commercial-use-friendly license.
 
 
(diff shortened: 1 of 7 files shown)


dmdashenkov:
@armiol, PTAL.

PR #1140, opened 2019-08-12
Optimize `MirrorRepository` routing

### LINKED ISSUE #988: Optimize `MirrorRepository` routing
`MirrorRepository` manages a system `Mirror` projection. Each `Mirror` represents the latest state of an aggregate in a respective domain bounded context.

While routing incoming events (such as `EntityStateChanged`, `EntityDeleted` etc) `MirrorRepository` tries to understand whether the update concerns an entity of `Aggregate` type. It uses a `String`-typed `typeUrl` field of an entity ID passed deep inside of incoming events. Then a message descriptor is loaded by a type URL value, and an `(entity).kind` the option value is read. And if the entity is not an aggregate, the event is ignored.

Such a lookup is very inefficient. 

Instead, an instance of `MirrorRepository` should be configured with type URLs of `Aggregate`s, which are being registered at runtime. Perhaps, an `AggregateRegistered` system event will suite the needs.

In this case, a routing procedure for each event will simply be a `HashMap.contains()` call.

dmitrykuzmin:
This PR resolves #988.

From now on, all mirrored entity types are "registered" to the `MirrorRepository` during the corresponding entity repository registration. No additional `proto` definition lookup is needed to perform event routing.

As a side effect, the logic of deciding whether an entity type should have mirror or not is moved from the `MirrorRepository` to the corresponding entity `Repository` instance.

Spine version is bumped to `1.0.2-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/aggregate/AggregateStorage.java
```diff
@@ -90,13 +90,16 @@ protected AggregateStorage(boolean multitenant) {
...
+     *         if the state type of a stored {@code Aggregate} is not mirrored by the
+     *         {@code MirrorRepository}
      */
-    void configureMirror(MirrorRepository mirrorRepository,
-                         AggregateClass<? extends Aggregate<I, ?, ?>> aggregateClass) {
-        Mirror.configure(mirrorRepository, aggregateClass, isMultitenant())
-              .ifPresent(m -> aggregateMirror = m);
+    void configureMirror(MirrorRepository mirrorRepository, TypeUrl stateType) {
+        checkState(mirrorRepository.isMirroring(stateType),
+                   "An aggregate type %s is not mirrored by a `MirrorRepository`.", stateType);
```
armiol:
```
An aggregate of type `%s` is not mirrored by a `MirrorRepository`.
```

server/src/main/java/io/spine/system/server/event/EntityLifecycleEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.google.errorprone.annotations.Immutable;
+import io.spine.annotation.GeneratedMixin;
+import io.spine.annotation.Internal;
+import io.spine.base.EventMessage;
+import io.spine.core.MessageId;
+import io.spine.type.TypeUrl;
+
+/**
+ * A common base for events related to the {@link io.spine.server.entity.Entity Entity} lifecycle.
```
armiol: Let's say "describing the ... lifecycle".

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -350,7 +350,31 @@ private void registerIfAware(Object contextPart) {
...
+     * </ol>
+     *
+     * @param entityStateClass
+     *         the class of the state of the entity managed by the resulting repository
+     * @return the requested repository or {@link Optional#empty()} if it's not found or not
+     *         visible
+     * @see VisibilityGuard
+     */
+    @Internal
+    public Optional<Repository> findRepository(Class<? extends Message> entityStateClass) {
```
armiol: Maybe we could have a method telling whether the repository is registered? In this way, the API would be much more consistent. 

The`repositoryFor` is very strict and assumes the repository *is* registered, but there is no way to tell whether it is. So I would fill this gap instead of duplicating the API call with a very small lexical difference.

dmitrykuzmin:
@armiol PTAL again.

PR #1141, opened 2019-08-21
Hide `Given` and test dispatchers

dmdashenkov:
In this PR we hide the old test utilities for creating entities and dispatching signals to them.

The utilities are preserved in order to be used in the framework tests. 

The framework users should not run tests on such a low level. Instead, `BlackBoxBoundedContext` is the tool to build unit/integration tests for Spine-based entities.


dmdashenkov:
@armiol, PTAL.

PR #1149, opened 2019-08-28
Selective handlers

alexander-yevsyukov:
This PR improves the `server.model` package in the following ways:
  * Method and class names were improved.
  * Now only one _type_ of `Throwable` is allowed for a handler method. There can be several `Throwable`s that share the same super-type.
  * `MethodExceptionCheck` now throws prohibited exception classes instead of throwing `ISE`.


server/src/main/java/io/spine/server/model/MethodSignature.java
```diff
@@ -78,24 +78,30 @@ protected MethodSignature(Class<? extends Annotation> annotation) {
...
      */
-    protected abstract ImmutableSet<Class<?>> validReturnTypes();
+    protected abstract ImmutableSet<Class<?>> returnTypes();
 
     /**
-     * Obtains the set of allowed exceptions that method may declare to throw.
+     * Obtains the type of a {@code Throwable} which a method can declare.
+     *
+     * <p>A handler method may declare more than one {@code Throwable}, but they must be of
+     * the same super-type.
```
dmdashenkov: I don't get this. Does this paragraph mean that we have to provide the broadest exception type?

PR #1150, opened 2019-08-28
Update the version to `1.0.2`

dmitrykuzmin:
This PR does the release of the `core` library `1.0.2`.

Although the actual version is `1.0.4-SNAPSHOT` already, this PR (temporarily) sets the version to `1.0.2` to do the `1.0.2` release, which is needed by the Spine `bootstrap` plugin and other libs dependent on `core`.


dmitrykuzmin:
@armiol PTAL. The build should pass once https://github.com/SpineEventEngine/time/pull/68 is merged.

PR #1151, opened 2019-08-29
Better parameters

alexander-yevsyukov:
This PR improves work with method parameters.
  * `MethodAttribute` was renamed to `Attribute`, preparing to use this interface not only for methods, but for parameters as well (for supporting `@Where` parameter filter).
  * `MethodParams` which used to be a utility, now is the first-class Java class.
  * Some method names were also improved.

### DIFF
--- a/server/src/main/java/io/spine/server/aggregate/model/AllowImportAttribute.java
+++ b/server/src/main/java/io/spine/server/aggregate/model/AllowImportAttribute.java
@@ -22,7 +22,7 @@ package io.spine.server.aggregate.model;
 
 import com.google.errorprone.annotations.Immutable;
 import io.spine.server.aggregate.Apply;
-import io.spine.server.model.MethodAttribute;
+import io.spine.server.model.Attribute;
 
 import java.lang.reflect.Method;
 
@@ -37,7 +37,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
  * @see io.spine.server.aggregate.Apply#allowImport()
  */
 @Immutable
-public enum AllowImportAttribute implements MethodAttribute<Boolean> {
+public enum AllowImportAttribute implements Attribute<Boolean> {
 
     ALLOW(true),
 
@@ -50,12 +50,12 @@ public enum AllowImportAttribute implements MethodAttribute<Boolean> {
     }
 
     @Override
-    public String getName() {
+    public String parameter() {
         return "allowImport";
     }
 
     @Override
-    public Boolean getValue() {
+    public Boolean value() {
         return value;
     }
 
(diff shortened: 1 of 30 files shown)


server/src/main/java/io/spine/server/aggregate/model/AllowImportAttribute.java
```diff
@@ -64,14 +64,17 @@ public Boolean getValue() {
...
      * @return the value of {@code AllowImportAttribute} for the given {@code Method}
-     * @throws IllegalArgumentException if the passed method is not properly annotated
+     * @throws IllegalArgumentException
+     *         if the passed method is not properly annotated
      */
     public static AllowImportAttribute of(Method method) {
         checkNotNull(method);
         Apply annotation = method.getAnnotation(Apply.class);
-        checkArgument(annotation != null, "The method `%s` is not annotated with `@Apply`", method);
+        checkArgument(annotation != null, "The method `%s` is not annotated with `@Apply`.",
```
dmdashenkov: Please split all 3 arguments onto separate lines.

server/src/main/java/io/spine/server/event/model/EventAcceptingMethodParams.java
```diff
@@ -95,14 +92,14 @@
     private final ImmutableList<Class<?>> expectedParameters;
     private final boolean awareOfCommandType;
 
-    EventAcceptingMethodParams(ImmutableList<Class<?>> parameters, boolean type) {
-        expectedParameters = parameters;
-        awareOfCommandType = type;
+    EventAcceptingMethodParams(boolean awareOfCommandType, Class<?>... parameters) {
```
dmdashenkov: While we're here, please consider replacing this `boolean` flag with an enum for better readability.

server/src/main/java/io/spine/server/model/Attribute.java
```diff
@@ -20,32 +20,24 @@
...
+ *
+ * @param <V>
+ *         the type of the value of the attribute. As it would typically be an annotation field
+ *         value, its type cannot be restricted to anything except for {@link Object}.
+ * @see HandlerMethod#attributes()
  */
-@Internal
 @Immutable(containerOf = "V")
-public interface MethodAttribute<V> {
+public interface Attribute<V> {
```
dmdashenkov: Maybe, it's a `Parameter`. The doc hints on that.

Also, I don't understand why the method is called `parameter` and not `name`.

server/src/main/java/io/spine/server/model/MethodParams.java
```diff
@@ -20,91 +20,90 @@
...
-     *         the method params to check
-     * @param expectedTypes
-     *         the expected types
-     * @return {@code true} if the method parameters are of expected types, {@code false} otherwise
+     * Verifies if these parameters match the passed types.
      */
-    public static boolean consistsOfTypes(Class<?>[] methodParams,
-                                          List<Class<?>> expectedTypes) {
-        if (methodParams.length != expectedTypes.size()) {
+    public boolean match(Class<?>... types) {
```
dmdashenkov: Maybe, instead of having this overload, expand the signature of `are`  to 
```java
public boolean are(Class<?> type1, Class<?> type2, Class<?>... rest)
```

Guava uses that trick in many places, e.g. `ImmutableList.of(...)`.

PR #1152, opened 2019-08-30
Import third-party events

dmdashenkov:
In this PR we introduce new API for posting events into a Bounded Context manually.

### Mitivation

In real-life projects, often there are upstream third-party systems to integrate with. Such systems may emit events. However, constructing an instance of `io.spine.core.Event` and posting it into the corrent `EventBus` may be cumbersome, as well as error-prone, since the `Event` type is the internal API of the framework.

### API

`ThirdPartyContext` is an adapter for external systems. `ThirdPartyContext` represents an external system as if it was just another Bounded Context. The adapter posts submitted by the user events to other Bounded Contexts via the common transport.

Typical usage looks as follows:
```java
ThirdPartyContext gitHub = ThirdPartyContext.multitenant("GitHub");
ActorContext actor = ActorContext
        .newBuilder()
        .setUserId(/* ... */)
        .setTimestamp(/* ... */)
        .setTenantId(/* ... */) // A tenant ID is required for a multitenant Context.
        .vBuild();
EventMessage event = IssueClosed
        .newBuilder()
        // ...
        .vBuild();
gitHub.emittedEvent(actor, event);
```

### Further Improvements

The next step for this feature is to allow events going through the `IntegrationBus` be posted into the domestic `EventBus` so that they are stored in the domestic `EventStore` for catch up and audit.

### Other Changes

This PR also re-enabled the 3 disabled tests in the `server` module. 2 of them used to fail due to unimplemented functionality and the third one is just too slow. All the tests now pass. The slow one is marked as `SlowTest` so that it can be skipped locally by running `fastTest` Gradle task instead of `test`.


dmdashenkov:
@alexander-yevsyukov, PTAL.
@dmitrykuzmin, PTAL.

server/src/main/java/io/spine/server/event/funnel/AbstractFunnel.java
```diff
@@ -0,0 +1,166 @@
...
+import static io.spine.base.Identifier.newUuid;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * The implementation base for {@link EventFunnel}s.
+ *
+ * <p>Posts the given events into a {@link Bus}.
+ *
+ * @param <M>
+ *         type of accepted by the target bus
```
dmitrykuzmin: I think there is a word `messages` missing here.

server/src/main/java/io/spine/server/event/funnel/AbstractFunnel.java
```diff
@@ -0,0 +1,166 @@
...
+     * Obtains a copy of this funnel but with the given {@code Ack} observer.
+     *
+     * @return new instance of the funnel
+     */
+    abstract AbstractFunnel<M> copyWithObserver(StreamObserver<Ack> resultObserver);
+
+    @Override
+    public final EventFunnel with(StreamObserver<Ack> resultObserver) {
+        checkNotNull(resultObserver);
+        return resultObserver.equals(this.resultObserver)
```
dmitrykuzmin: Why do we have this optimization here, but not for `forTenant(TenantId)` method?

server/src/main/java/io/spine/server/event/funnel/AimEvents.java
```diff
@@ -0,0 +1,75 @@
...
+
+    AimEvents(BoundedContext context, UserId actor) {
+        this.context = checkNotNull(context);
+        this.actor = checkNotNull(actor);
+    }
+
+    /**
+     * Specifies that the events should be posted to an Aggregate via the {@link ImportBus}.
+     *
+     * <p>If the Aggregate imports the events successfully, they are posted into the Event Bus of
```
dmitrykuzmin: Maybe we should also mention what will happen if there is no Aggregate in the system which imports a posted event.

server/src/main/java/io/spine/server/event/funnel/EventFunnel.java
```diff
@@ -0,0 +1,61 @@
...
+
+import io.grpc.stub.StreamObserver;
+import io.spine.base.EventMessage;
+import io.spine.core.Ack;
+import io.spine.core.TenantId;
+
+/**
+ * A funnel which posts events from an external source into a Bounded Context.
+ *
+ * <p>This if a part of the fluent API for manually posting events.
```
dmitrykuzmin:
```suggestion
 * <p>This is a part of the fluent API for manually posting events.
```

server/src/main/java/io/spine/server/event/funnel/EventFunnel.java
```diff
@@ -0,0 +1,61 @@
...
+     * <p>Multiple invocations of this method override the tenant value.
+     *
+     * @return a new {@code EventFunnel} with the given observer
+     */
+    EventFunnel forTenant(TenantId tenantId);
+
+    /**
+     * Posts the given events into the target Bounded Context.
+     *
+     * @param events one or more events to post
```
dmitrykuzmin: Please, fix the formatting.

alexander-yevsyukov:
> by definition, this event would not be domestic.

What if we migrate existing code to a new solution, which is DDD-based? Then, we'd have a new Context and a legacy system which generates events for us. They are domestic in the sense it's the same context.

server/src/main/java/io/spine/server/event/funnel/PostEvents.java
```diff
@@ -0,0 +1,85 @@
...
+        this.context = checkNotNull(context);
+    }
+
+    /**
+     * Specifies the actor who produces the events.
+     *
+     * <p>An actor may be a user of the system or a component inside the system. The {@code UserId}
+     * specified as the actor should be helpful when finding out the origin of the event.
+     *
+     * <p>If the actor if a software component or a third-party system, rather than a human user,
```
dmitrykuzmin:
```suggestion
     * <p>If the actor is a software component or a third-party system, rather than a human user,
```

alexander-yevsyukov:
We need to avoid `boradcast()` and `toAggregate()` details being exposed to the party which integrates with us.

server/src/test/java/io/spine/server/event/funnel/given/DocumentAggregate.java
```diff
@@ -0,0 +1,155 @@
...
+import io.spine.server.event.funnel.EditText;
+import io.spine.server.event.funnel.OpenOfficeDocumentUploaded;
+import io.spine.server.event.funnel.PaperDocumentScanned;
+import io.spine.server.event.funnel.TextEdited;
+import io.spine.server.tuple.Pair;
+import io.spine.time.LocalDateTime;
+import io.spine.time.Now;
+
+/**
+ * An test aggregate representing a {@code Document}.
```
dmitrykuzmin:
```suggestion
 * A test aggregate representing a {@code Document}.
```

dmdashenkov:
@alexander-yevsyukov, @dmitrykuzmin, PTAL again.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.base.Identifier.newUuid;
+import static io.spine.base.Time.currentTime;
+import static io.spine.protobuf.AnyPacker.pack;
+import static io.spine.server.BoundedContextBuilder.notStoringEvents;
+
+/**
+ * An external upstream system which is not implemented in Spine.
```
dmitrykuzmin: Maybe it's better to use `non-Spine-based` or `not based on Spine` if I got the meaning correctly.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+public final class ThirdPartyContext implements AutoCloseable {
+
+    private final BoundedContext context;
+    private final Any producerId;
+
+    /**
+     * Creates a new single-tenant instance of {@code ThirdPartyContext} with the given name.
+     *
+     * @param name
+     *         name of the third-party system
```
alexander-yevsyukov: What we create is our representation of a _part_ of a 3rd party system, as if it was a Bounded Context. So, I'd say this: “A name of a Context which represents a part of a third-party system.“ Anyway, it should say that it's not the whole system.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+     */
+    public static ThirdPartyContext singleTenant(String name) {
+        checkNotNull(name);
```
alexander-yevsyukov: Can we check for an empty string too, please?

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Creates a new multitenant instance of {@code ThirdPartyContext} with the given name.
+     *
+     * @param name
+     *         name of the third-party system
+     */
+    public static ThirdPartyContext multitenant(String name) {
+        checkNotNull(name);
```
alexander-yevsyukov: Can we check for an empty string too, please?

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+        return newContext(name, true);
+    }
+
+    private static ThirdPartyContext newContext(String name, boolean multitenant) {
+        BoundedContext context = notStoringEvents(name, multitenant).build();
+        context.integrationBus()
+               .notifyOfCurrentNeeds();
```
alexander-yevsyukov: We need to name this method so that it tells about the fact that Contexts need to speak to each other, maybe again, because a new context appeared in the whole picture of the system.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+     * this Context is multitenant.
+     *
+     * @param actorContext
+     *         the info about the actor, a user or a software component, who emits the event
+     * @param eventMessage
+     *         the event
+     */
+    public void emittedEvent(ActorContext actorContext, EventMessage eventMessage) {
+        checkNotNull(actorContext);
+        checkNotNull(eventMessage);
```
alexander-yevsyukov: Maybe also `checkNotDefault()` for both? It may save new users some time on finding the fact that dispatching does not work for default messages.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+        checkTenant(actorContext, eventMessage);
+
+        EventId id = EventId
```
alexander-yevsyukov: Why `Events.generateId()` does not work here?

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+                .newBuilder()
+                .setValue(newUuid())
+                .build();
+        EventContext eventContext = EventContext
```
alexander-yevsyukov: Please have `EventFactory` to avoid generating event IDs and assembling `Event` instances. We need to encapsulate these details of creating events even.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
```diff
@@ -0,0 +1,190 @@
...
+     * <p>This overload may only be used for single-tenant third-party contexts. If this Context is
+     * multitenant, this method throws an exception.
+     *
+     * @param userId
+     *         the ID of the actor, a user or a software component, who emits the event
+     * @param eventMessage
+     *         the event
+     */
+    public void emittedEvent(UserId userId, EventMessage eventMessage) {
+        checkNotNull(userId);
```
alexander-yevsyukov: Please consider for checking for non-default too.

server/src/main/java/io/spine/server/model/HandlerMap.java
```diff
@@ -220,14 +223,26 @@ public H handlerOf(M messageClass) {
...
-        checkState(count == 1,
-                   "Unexpected number of handlers for messages of class %s: %s.%n%s",
-                   targetType, count, handlers);
-        H result = handlers
-                .stream()
-                .findFirst()
-                .get();
-        return result;
+        if (count != 1) {
+            logWrongMethodCount(count, targetType);
```
alexander-yevsyukov: Please test this branch.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
dmdashenkov: We create `Event`s with an `EventFactory` now. All the necessary event validation is encapsulated in there.

server/src/main/java/io/spine/server/integration/ThirdPartyContext.java
dmdashenkov: It is checked as a part of `ActorContext` validation.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

server/src/main/java/io/spine/server/event/model/EventReactorSignature.java
```diff
@@ -32,12 +33,13 @@
 /**
  * The signature of {@link EventReactorMethod}.
  */
-final class EventReactorSignature extends EventAcceptingSignature<EventReactorMethod> {
+@Internal
+public final class EventReactorSignature extends EventAcceptingSignature<EventReactorMethod> {
```
alexander-yevsyukov: Can we hide this class back?

server/src/main/java/io/spine/server/integration/ConfigurationChangeObserver.java
```diff
@@ -89,7 +89,7 @@ public void handle(ExternalMessage value) {
         clearStaleSubscriptions(request.getRequestedMessageTypeList(), origin);
         if (!knownContexts.contains(origin)) {
             knownContexts.add(origin);
-            integrationBus.notifyOfCurrentNeeds();
+            integrationBus.introduceSelf();
```
alexander-yevsyukov: This doesn't sound right. The bus is already introduced to the system. It's the new context which introduces itself. Or, if you want the bus to do the job the method should be named differently.

server/src/main/java/io/spine/server/model/HandlerMap.java
```diff
@@ -220,14 +223,34 @@ public H handlerOf(M messageClass) {
...
+            );
+        } else {
+            /*
+              The map should have found all the duplicates during construction.
+              This is a fail-safe execution branch which ensures that no changes in the `HandlerMap`
+              implementation corrupt the model.
+            */
+            _error().log(
+                    "There are %d handler methods found for the type `%s`." +
+                            "Expected only one. Model is corrupt.",
```
alexander-yevsyukov: > Model is corrupt

This doesn't tell the reader much of info. It just scares him. Tell the reader what to do about the problem.
Please include method names into diagnostics.

server/src/main/java/io/spine/server/model/HandlerMap.java
```diff
@@ -220,14 +223,34 @@ public H handlerOf(M messageClass) {
...
-                .get();
-        return result;
+        if (count != 1) {
+            throw wrongMethodCount(handlers, targetType);
+        }
+        H handler = Iterables.getOnlyElement(handlers);
+        return handler;
+    }
+
+    private IllegalStateException wrongMethodCount(Collection<H> handlers, M targetType) {
```
alexander-yevsyukov: Please make this a state-check method and name it `checkOnlyOne()`.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

PR #1153, opened 2019-08-30
Delivery improvements

armiol:
In this changeset, several improvements to the `Delivery` mechanism are made.

**Consistent shard indexes**

The default sharding strategy now relies on Murmur3 hashing algorithm, rather than on the native Java `hashCode()`. It allows determining the shard index consistently across different JVMs.

**Delivery stages and monitoring**

As long as the messages during the delivery are read page-by-page, the `Delivery` may now be configured with the page size. Additionally, an API user now can control whether the delivery should be continued after delivering another page. In environments, such as AppEngine Standard, there are time restrictions on the max duration of the request processing. 

The usage of `DeliveryMonitor` allows to obtain more fine-grained control and stop the execution before the instance gets killed.

**Batch dispatching**

In case the consequent messages in the same shard are headed to the same target, they are now grouped in a batch and dispatched together. Also, instead of reading and writing the `Entity` per-message, the `RepositoryCache` is used to cache `load` and `store` calls. 

It starts caching before the batch of `InboxMessage`s is delivered and stops the caching afterward, propagating all the updates to the underlying storage.

**Cleaning `ShardedWorkRegistry`**

When an application node picks up a shard, a respective record is created in the `ShardedWorkRegistry`, telling that the node picked up some shard at a certain time. However, as the application nodes may get killed, some of the records may contain stale data and prevent other nodes from picking the shards.
The new `ShardedWorkRegistry.releaseExpiredSessions(ShardIndex, Duration)` call allows to manually clear the registry from the nodes which have picked up their sessions a long time ago (specified by the `duration` parameter).

Additionally, some polishing has been done in the API of `Endpoint`s removing the parameter discrepancy and streamlining the execution.


server/src/main/java/io/spine/server/entity/RepositoryCache.java
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * A function which stores the {@code Entity} state to its real repository.
+     *
+     * @param <I>
+     *         the type of {@code Entity} identifiers
+     * @param <E>
+     *         the type of entity
+     */
+    public interface Store<I, E extends Entity<I, ?>> extends Consumer<E> {
```
dmdashenkov: Why do we need `I`?

server/src/main/java/io/spine/server/entity/RepositoryCache.java
```diff
@@ -0,0 +1,139 @@
...
+
+    public E load(I id) {
+        IdInTenant<I> idInTenant = idInTenant(id);
+        if (!idsToCache.contains(idInTenant)) {
+            return loadFn.apply(idInTenant.value());
+        }
+
+        if (!cache.containsKey(idInTenant)) {
+            E entity = loadFn.apply(idInTenant.value());
+            cache.put(idInTenant, entity);
```
dmdashenkov: Can the cache grow too quick and fill the memory?

server/src/test/java/io/spine/server/delivery/InboxIdsTest.java
```diff
@@ -0,0 +1,46 @@
...
+    InboxIdsTest() {
+        super(InboxIds.class);
+    }
+
+    @Override
+    protected void configure(NullPointerTester tester) {
+        super.configure(tester);
+        tester.setDefault(InboxId.class, InboxId.getDefaultInstance())
+              .setDefault(TypeUrl.class, TypeUrl.of(Calc.class));
+        tester.testStaticMethods(getUtilityClass(), NullPointerTester.Visibility.PACKAGE);
```
dmdashenkov: The `configure` method is not meant to execute tests.

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+ * all the messages are being dispatched the "busy" {@code Entity}. Depending on the number
+ * of messages to deliver, it may take a lot of time.
+ *
+ * <p>Environments such as Google AppEngine Standard imply restrictions on the duration of the
+ * request processing. Therefore, it may be necessary to stop the message delivery for this shard
+ * at some point, and re-schedule the shard processing again by sending another request and
+ * thus resetting the processing clock.
+ */
+@SPI
+public class DeliveryMonitor {
```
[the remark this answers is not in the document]
armiol: I am having more hopes for it in upcoming catch-up implementation, as we need to monitor how the instances of `Projection`s are progressing with their event playing.

So in future I am planning to arm it with more listeners and maybe with more controlling stuff such as the prioritization of catch-up over the regular delivery etc.

At first, I was going to name it `Controller`, but there are some unwanted connotations. We will always be able to rename later, but to me `Monitor` is a term general enough, while `CircuitBreaker` limits its use.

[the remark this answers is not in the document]
armiol: So it's not going to be a functional interface as well.

server/src/main/java/io/spine/server/entity/RepositoryCache.java
armiol: Not in our use cases. The maximum number of cached entities is the total number of shards — and that's in case there is just one node serving the whole app. It's because each shard delivery is a single-thread process, and the only entity is cached at a time by the delivery.  

End-users will hardly ever have 100k of shards and just a single instance processing them all.

I don't want to use Guava's cache for such a simple use case. And it's much slower than the concurrent `HashMap`.

I have added the note on usage though.

PR #1156, opened 2019-09-04
Dispatch external events through `EventBus`

dmdashenkov:
### `IntegrationBus`

In this PR we refactor the mechanism of integrating Bounded Contexts.

In order for two Contexts to communicate, they emit events and subscribe to events of one another. The handlers which accept events from outside of a Bounded Context (e.g. from other Contexts) are marked as `external`.

Previously, a Bounded Context had an `IntegrationBus` which:
 1. published domestic events to outside subscribers;
 2. dispatched external events to the external subscribers/reactors/commanders inside its own Context.

Now, `IntegrationBus` is substituted with `IntegrationBroker`.

Just like `IntegrationBus`, `IntegrationBroker` publishes domestic events to outside by tapping into the domestic `EventBus`.

Just like `IntegrationBus`, `IntegrationBroker` communicates with brokers of other contexts in order to find out which event to publish.

Just like `IntegrationBus`, `IntegrationBroker` seats on top of a `TransportFactory`, which instantiates message channels for inter-context communication.

Unlike `IntegrationBus`, `IntegrationBroker` is not a `Bus`. A broker does not bother dispatching events to targets (such as entity repositories). Instead, it posts external events into the `EventBus` and the `EventBus` performs the dispatching.

Unlike `IntegrationBus`, `IntegrationBroker` does not hold a registry of all external dispatchers in its Context for the reason stated in the previous paragraph.

Unlike `IntegrationBus`, `IntegrationBroker` is not a public API of a `BoundedContext`. Users should not interact with it directly. SPI users should interact with `TransportFactory` and related concepts.

### Other API changes

This PR changes the visibility of `Command`, `Event`, and related types. Previously, the types were exposed to the end user. Now, `Command`, `Event`, `Commands`, and `Events` are marked as `SPI` and the associated generated mixins are package-private.

### Version

This PR marks the release of `1.1.0`.


dmdashenkov:
@alexander-yevsyukov, @dmitrykuzmin, PTAL.

server/src/main/java/io/spine/server/model/SignalOriginMismatchError.java
```diff
@@ -17,21 +17,23 @@
...
+ * An error which signifies that the {@code external} attribute was misplaced on a handler method.
  */
-enum ExternalMessageValidator implements EnvelopeValidator<ExternalMessageEnvelope> {
-    INSTANCE;
+public final class SignalOriginMismatchError extends ModelError {
+
+    private static final long serialVersionUID = 0L;
+
+    private static final String ERROR_TEMPLATE =
+            "Mismatch of `external` value for the handler method %s. " +
```
dmitrykuzmin: Maybe it's better to say that we received external/non-external signal as an input to the non-external/external method, because it's not immediately apparent, what does "misplaced" mean.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -352,7 +352,7 @@ protected final void onRoutingFailed(SignalEnvelope<?, ?, ?> envelope, Throwable
 
     @Override
     public Set<EventClass> domesticEvents() {
-        return aggregateClass().domesticEvents();
+        return aggregateClass().events();
```
dmitrykuzmin: Why does the `domesticEvents()` method of `AggregateRepository` delegates to the (all) `events()` method of aggregate class? If this is not a typo, could you please clarify vocally or in doc?

dmdashenkov: Renamed to `events()`.

dmdashenkov:
@dmitrykuzmin, PTAL again.

PR #1157, opened 2019-09-04
Release only non-empty sessions

dmdashenkov:
Previously, when force-releasing old shards, we cleaned up all the shards which have been picked last time before a certain timestamp. That set included the shards which had no active sessions, i.e. were not being processed by any node at the time.

Now, we only release the shards which have an active session, as they are the only ones which can possibly be "stuck".

This approach prevents us from race conditions in a distributed/multithreaded environment. Also, this gives the user a clearer picture of which shards were released forcefully.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.0.3`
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #1158, opened 2019-09-05
Abstract base for `ShardedWorkRegistry`

dmdashenkov:
In this PR we extract common implementation parts from `SessionWorkRegistry` implementations into an abstract class. The in-mem implementation now does, and other implementations should, extend `AbstractWorkRegistry` in order to avoid duplicate code.


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #1159, opened 2019-09-05
Filter event subscription by parameter annotation

alexander-yevsyukov:
This PR introduces `@Where` annotation for the first parameter of an event subscriber method.

`@Subscribe.filter()` was deprecated. The annotation interface `ByField` was not marked as deprecated to avoid numerous warnings on import statements in the code which still uses this interface (for backward compatibility).

This PR depends on [this one from `base`](https://github.com/SpineEventEngine/base/pull/468).

### DIFF
--- a/core/src/main/java/io/spine/core/Where.java
+++ b/core/src/main/java/io/spine/core/Where.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2019, TeamDev. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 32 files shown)


server/src/test/java/io/spine/server/projection/given/cls/FilteringProjection.java
```diff
@@ -47,14 +47,14 @@ private FilteringProjection(String id) {
     }
 
     /** This method does not use the value passed in the event. */
-    @Subscribe(filter = @ByField(path = VALUE_FIELD_PATH, value = SET_A))
-    void onReserved(StringImported event) {
+    @Subscribe
+    void onReserved(@Where(field = VALUE_FIELD_PATH, equals = SET_A)StringImported event) {
```
dmdashenkov:
```suggestion
    void onReserved(@Where(field = VALUE_FIELD_PATH, equals = SET_A) StringImported event) {
```

server/src/test/java/io/spine/system/server/given/diagnostics/ViolationsWatch.java
```diff
@@ -62,11 +61,11 @@ void onInvalidText(ConstraintViolated event) {
...
 
-    @Subscribe(filter = @ByField(
-            path = LAST_MESSAGE_TYPE_PATH,
-            value = "type.spine.io/spine.system.server.test.StartVerification"
-    ))
-    void onInvalidVerification(ConstraintViolated event) {
+    @Subscribe
+    void onInvalidVerification(
+            @Where(field = LAST_MESSAGE_TYPE_PATH,
+                    equals = "type.spine.io/spine.system.server.test.StartVerification")
```
dmdashenkov: Please fix the code alignment.

core/src/main/java/io/spine/core/Where.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    String field();
+
+    /**
+     * The expected value of the field.
+     *
+     * <p>The value converted with help of {@link io.spine.string.Stringifier Stringifier}s into
+     * the type of the actual value of the message field.
+     */
+    String equals();
```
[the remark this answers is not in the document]
alexander-yevsyukov: As discussed vocally, the current version is shorter and works well.

PR #1162, opened 2019-09-09
Test if `Bus` is open before posting

alexander-yevsyukov:
This PR adds the ability to see if a `Bus` is open so that the state can be checked prior to posting (instead of catching `IllegalStateException`).

* `io.spine.server.Closable` interface was introduced for the classes that need to check the status of openness.

## Build changes
 AppVeyor configuration was removed in favor of GitHub Action which performs the Windows build in this repository.

## Deprecations
* `BundedContext.isClosed()` is deprecated in favor of newly `isOpen()`.


server/src/main/java/io/spine/server/bus/Bus.java
```diff
@@ -397,3 +408,4 @@ private FilterChainSupplier(BusBuilder<?, T, E, ?, ?> builder) {
         }
     }
 }
+
```
dmdashenkov: Please remove this line.

server/src/main/java/io/spine/server/integration/IntegrationBus.java
```diff
@@ -146,15 +147,15 @@ public void registerWith(BoundedContext context) {
     }
 
     @Override
-    public boolean isRegistered() {
+    public synchronized boolean isRegistered() {
```
dmdashenkov: It seems to be an overkill to make all these methods synchronized.

Let's discuss on a per-case basis.

PR #1164, opened 2019-09-09
Allow `InMemoryStorageFactory` to be used in production

dmitrykuzmin:
This PR allows the usage of `InMemoryStorageFactory` as a production storage factory.

Previously, it was forbidden to specify `InMemoryStorageFactory` as an argument for `ServerEnvironment.configureStorage(...)`.

It led to an over-complication of simple Spine-based applications, including Spine examples where we had to create stub storage implementations.

Now, the usage of `InMemoryStorageFactory` is allowed. It still needs to be configured by the user explicitly.


dmitrykuzmin:
@dmdashenkov PTAL.

PR #1174, opened 2019-09-26
Avoid using real gRPC server in tests

dmitrykuzmin:
This PR rewrites the gRPC-related tests that were failing from time to time locally and on CI.

Now they will use a stub instead of the real server and the failures should stop occurring.

This PR also mentions version [1.1.2](https://github.com/SpineEventEngine/core-java/releases/tag/v1.1.2) as the latest stable Spine version in the project-level README.


dmitrykuzmin:
@armiol PTAL.

PR #1175, opened 2019-10-02
Enforce the rules applied to the handler method annotations

armiol:
This changeset updates the rules applied to the signatures of handler methods. 

In particular, methods annotated with `@Assign`, `@React`, `@Command` and `@Subscribe` are carefully checked for the requirements set for each of the annotations. This behavior has been documented and covered with unit tests.

Special `DoNothing` command message has been introduced. It should be used as a return result of `@Command`-ing methods if they declare an option not to emit any command in response to the incoming event:

```java
@Command
EitherOf3<PauseTask, StopTask, DoNothing> on(TaskUnassigned event) {...}
```

Such a command message is a counterpart of `Nothing` event message. Similarly, it is ignored by the framework once emitted.

The framework version is set to `1.1.6`.


server/src/main/java/io/spine/server/command/model/CommandAcceptingSignature.java
```diff
@@ -79,7 +88,7 @@
         MESSAGE {
             @Override
             public boolean matches(MethodParams params) {
-                return params.is(CommandMessage.class);
+                return params.is(CommandMessage.class) && params.is(GeneratedMessageV3.class);
```
dmdashenkov: Why `GeneratedMessageV3`? I suppose, `Message` provides a sufficient interface. And `CommandMessage` extends `Message`.

server/src/main/java/io/spine/server/entity/EntityVisibility.java
```diff
@@ -84,22 +87,43 @@ private EntityVisibility(Visibility value) {
...
+     *         into this method.
+     *         We don't want to define {@code (entity)} option for the {@code Event} message,
+     *         therefore this method handles {@code Event.class} as a special case and returns
+     *         {@code NONE} visibility level for it.
      */
-    public static EntityVisibility of(Class<? extends Message> stateClass) {
+    public static Optional<EntityVisibility> of(Class<? extends Message> stateClass) {
         checkNotNull(stateClass);
+
+        if(Event.class.equals(stateClass)) {
```
dmdashenkov: Please fix formatting.

server/src/main/java/io/spine/server/event/model/StateSubscriberSpec.java
```diff
@@ -68,7 +70,11 @@ public boolean matches(MethodParams params) {
         }
         @SuppressWarnings("unchecked") // Checked above.
         Class<? extends Message> firstParameter = (Class<? extends Message>) params.type(0);
-        EntityVisibility visibility = EntityVisibility.of(firstParameter);
+        Optional<EntityVisibility> visibilityOption = EntityVisibility.of(firstParameter);
+        if(!visibilityOption.isPresent()) {
```
dmdashenkov: Please fix formatting.

server/src/main/java/io/spine/server/model/MethodParams.java
```diff
@@ -113,6 +113,16 @@ public boolean is(Class<?> type) {
...
             return false;
         }
+        return firstIs(type);
+    }
+
+    /**
+     * Verifies if the first parameter of the method is of the passed type.
+     */
+    public boolean firstIs(Class<?> type) {
+        if(size() == 0) {
```
dmdashenkov: Same here.

server/src/test/java/io/spine/server/model/MethodSignatureTest.java
```diff
@@ -55,11 +60,57 @@ protected final void testValid(Method method) {
     @ParameterizedTest
     @MethodSource("invalidMethods")
     protected final void testInvalid(Method method) {
-        assertThrows(SignatureMismatchException.class, () -> wrap(method));
+
+        //TODO:2019-09-24:alex.tymchenko: a workaround to detect invalid `@Command`-er methods.
```
dmdashenkov: Please resolve.

server/src/test/java/io/spine/server/model/MethodSignatureTest.java
```diff
@@ -55,11 +60,57 @@ protected final void testValid(Method method) {
...
+        try {
+            Optional<? extends HandlerMethod> result = wrap(method);
+            if(result.isPresent()) {
```
dmdashenkov: Please fix formatting.

armiol:
@dmdashenkov comments addressed. PTAL again.

server/src/main/java/io/spine/server/command/Command.java
```diff
@@ -171,7 +171,9 @@
  * <pre>
  *
  * {@literal @}Command
- *  CreateProject on(Rejections.CannotCreateProject rejection) { {@literal // change the parameters and try again...} }
+ *  CreateProject on(Rejections.CannotCreateProject rejection) {
```
dmdashenkov: Please check identation here as well.

server/src/main/java/io/spine/server/command/model/CommandReactionSignature.java
```diff
@@ -122,7 +121,7 @@ public boolean matches(MethodParams params) {
             @Override
             public boolean matches(MethodParams params) {
                 return params.are(EventMessage.class, EventContext.class)
-                        && params.firstIs(GeneratedMessageV3.class);
+                        && params.declaredAsClasses();
```
dmdashenkov: Can we check that all params are declared as classes in a single place? Now, we have to copy this check for each signature.

PR #1183, opened 2019-10-23
Bump Base

alexander-yevsyukov:
This PR:
 1. Addresses deprecations introduced in recent versions of `base`.
 2. Fixes visibility and documentation of public API related to composition of `Query` and `Topic`.
     Previously, the abstract base common to `QueryBuilder` and `TopicBuilder` wasn't a part of public API but was referenced in the public API documentation.


client/src/main/java/io/spine/client/QueryBuilder.java
```diff
@@ -96,18 +96,16 @@ public QueryBuilder orderBy(String column, OrderBy.Direction direction) {
...
     public QueryBuilder limit(int count) {
         checkLimit(count);
         this.limit = count;
         return self();
     }
 
-    @SuppressWarnings("PMD.UnusedPrivateMethod")  /* See https://github.com/pmd/pmd/issues/770. */
-    private static void checkLimit(Number count) {
-        checkArgument(count.longValue() > 0, "A Query limit must be more than 0.");
+    static void checkLimit(long count) {
```
yuri-sergiichuk: Let's add at least a one-liner doc here.

core/src/main/java/io/spine/core/Events.java
```diff
@@ -93,10 +93,11 @@ public static EventMessage ensureMessage(Message eventOrMessage) {
...
      * @throws IllegalArgumentException
      *         if the ID string value is empty or blank
+     * @deprecated please use {@link io.spine.util.Preconditions2#checkNotDefaultArg(Message)}
      */
+    @Deprecated
     public static EventId checkValid(EventId id) {
-        checkNotNull(id);
-        checkNotEmptyOrBlank(id.getValue(), "event ID");
+        checkNotDefaultArg(id);
         return id;
```
yuri-sergiichuk: I believe we can just: `return checkNotDefaultArg(id);` here. But while it's deprecated already, not sure if it matters.

client/src/main/java/io/spine/client/TargetBuilder.java
```diff
@@ -78,7 +78,7 @@
  * @param <B>
  *         a type of the builder implementations
  */
-abstract class AbstractTargetBuilder<T extends Message, B extends AbstractTargetBuilder> {
+public abstract class TargetBuilder<T extends Message, B extends TargetBuilder> {
```
yuri-sergiichuk: So why is it public now?

server/src/main/java/io/spine/server/entity/storage/Columns.java
```diff
@@ -114,17 +114,17 @@ static EntityColumn findColumn(Class<? extends Entity<?, ?>> entityClass, String
     }
 
     static void checkColumnName(String columnName) {
-        checkNotEmptyOrBlank(columnName, "entity column name");
+        checkNotEmptyOrBlank(columnName);
```
yuri-sergiichuk: `Entity column name should not be empty or blank`?

server/src/main/java/io/spine/server/tenant/CurrentTenant.java
```diff
@@ -76,8 +74,7 @@ static TenantId ensure() throws IllegalStateException {
      * @param tenantId a non-null and non-default instance of {@code TenantId}
      */
     static void set(TenantId tenantId) {
-        checkNotNull(tenantId);
-        checkArgument(isNotDefault(tenantId));
+        checkNotDefaultArg(tenantId);
```
yuri-sergiichuk: `"Tenant ID should not be default"`?

server/src/test/java/io/spine/server/bus/BusesTest.java
```diff
@@ -68,6 +71,12 @@ void passNullToleranceCheck() {
                                               .setCode(1)
                                               .build())
                 .setDefault(RejectionEnvelope.class, defaultRejection)
+                .setDefault(Message.class, StringValue.newBuilder()
```
yuri-sergiichuk: Will `TestValues.newUuidValue()` work here?

testutil-core/src/main/java/io/spine/testing/core/given/GivenTenantId.java
```diff
@@ -52,9 +52,8 @@ public static TenantId generate() {
      * @param value must be non-null, not empty, and not-blank
      * @return new {@code TenantId}
      */
-    @SuppressWarnings("DuplicateStringLiteralInspection")
     public static TenantId of(String value) {
-        checkNotEmptyOrBlank(value, "value");
+        checkNotEmptyOrBlank(value);
```
yuri-sergiichuk: `"Tenant ID value should not be empty or blank."`?

client/src/main/java/io/spine/client/TargetBuilder.java
alexander-yevsyukov: Because we reference the class in Javadocs.

Issue #1135, opened 2019-08-04
Introduce `Entity` column option for Protobuf `Entity` state definitions


armiol:
Currently, the entity columns are defined in Java with `@Column` annotation.

While this is convenient, it's completely invisible for those who use them for querying or subscribing to `Entity` state updates.

Therefore it is proposed to introduce a `(column)` option.

```proto
message UserProfile {
    option (entity).kind = PROJECTION;

    UserId id = 1;

    // ...

    // When this user was registered in the system.
    google.protobuf.Timestamp when_registered = 7 [(required) = true];
    
    // A year of user's registration in the system.
    int32 year_of_registration = 8 [(column) = true];    
}
```

A presence of `[(column) = true]` implies optionality of the marked field.

End-users will have two options to set the value for the column:

1. Use Spine-generated interface when declaring their entity. E.g.

```java
interface UserProfileWithColumns extends EntityWithColumns {
    
    int getYearOfRegistration();
}

// ...
class UserProfileProjection 
            extends Projection<UserId, UserProfile, UserProfile.Builder>
            implements UserProfileWithColumns { ... }
```
In which case this method becomes mandatory for implementation. The framework routines will then be able to call the methods in all `T extends EntityWithColumns` entity classes.

2. Fill the value in manually when updating the `Entity` state.

In both cases, the `year_of_registration` column value should be finalized by the moment of transaction commit. If an end-user decides to leave it empty — that's what the framework will count as an intentional action.

In this way, the dictionary of the language shared between a server and its clients will have a clear notion of what can be used for querying and subscribing.

dmitrykuzmin:
Closed with [base#492](https://github.com/SpineEventEngine/base/pull/492), #1184.

PR #1184, opened 2019-10-28
The `(column)` option

dmitrykuzmin:
This PR addresses #1135.

The entity columns will now be defined in Protobuf with the help of `(column)` option:
```proto
message UserProfile {
    option (entity).kind = PROJECTION;
    // ...
    int32 year_of_registration = 8 [(column) = true];    
}
```
The column value can be updated in two ways:

1. In handler methods of the entity like any other part of entity state.
2. With the help of Spine-generated interface:
```java
interface UserProfileWithColumns extends EntityWithColumns {
    
    int getYearOfRegistration();
}

// ...
class UserProfileProjection 
            extends Projection<UserId, UserProfile, UserProfile.Builder>
            implements UserProfileWithColumns { ... }
```
The fields which are updated in both ways simultaneously will always have the value from the interface-inherited getter (i.e. the no.2 above).

The value from the interface is finalized by the moment of transaction commit and propagated to the entity state (i.e. the `EntityRecord`, `EntityStateChangedEvent`, etc. will all store the updated column values).

The `archived`, `deleted` and other system fields became the system columns which are exceptions to the "ordinary" columns and are processed in a similar way as previously.

For now, entity columns can only be applied to entities of `PROJECTION` and `PROCESS_MANAGER` types. For aggregates and generic messages the presence of `(column)` option is ignored.

Spine version advances to `1.1.13`.


dmitrykuzmin:
@armiol, @dmdashenkov PTAL.

server/src/test/proto/spine/test/stand/cafeteria.proto
```diff
@@ -20,6 +20,8 @@ message Menu {
 
     MenuId id = 1;
     repeated Dish dish = 2;
```
armiol: Can you please either have empty lines between all three fields, or no empty lines at all.

server/src/main/java/io/spine/server/entity/Transaction.java
```diff
@@ -363,16 +366,28 @@ public final void commit() throws InvalidEntityStateException, IllegalStateExcep
...
 
+    /**
+     * Propagates entity column values to the entity state and commits a transaction.
+     */
+    private void updateColumnsAndCommit() {
+        S stateWithColumns = stateWithColumns();
+        if (initialState.equals(stateWithColumns)) {
+            commitUnchangedState();
+        } else {
+            commitChangedState(stateWithColumns, false);
```
dmdashenkov: This is not obvious. Why do we not `updateColumns` when the columns have been changed?

server/src/main/java/io/spine/server/entity/storage/AbstractColumnMapping.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * The basic implementation of mapping rules for entity {@linkplain Column columns}.
+ *
+ * <p>Since entity columns are proto-based an have a limited amount of possible types, this class
```
dmdashenkov:
```suggestion
 * <p>Since entity columns are proto-based and have a limited amount of possible types, this class
```

server/src/main/java/io/spine/server/entity/storage/AbstractColumnMapping.java
```diff
@@ -0,0 +1,212 @@
...
+                    ImmutableMap.builder();
+
+            setupCustomMapping(builder);
+
+            customMapping = builder.build();
+        }
+        return customMapping;
+    }
+
+    private Optional<ColumnTypeMapping<?, ? extends R>> customMappingFor(Class<?> aClass) {
```
dmdashenkov: Please name the parameter so that its value is clear.

server/src/main/java/io/spine/server/entity/storage/AbstractColumnMapping.java
```diff
@@ -0,0 +1,212 @@
...
+                customMapping().keySet()
+                               .stream()
+                               .filter(cls -> cls.isAssignableFrom(aClass))
+                               .map(customMapping()::get)
+                               .findFirst()
+                               .map(rule -> (ColumnTypeMapping<?, ? extends R>) rule);
+        return result;
+    }
+
+    private Optional<ColumnTypeMapping<?, ? extends R>> standardMappingFor(Class<?> aClass) {
```
dmdashenkov: Same here.

server/src/main/java/io/spine/server/entity/storage/Introspector.java
```diff
@@ -0,0 +1,188 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.code.proto.ColumnOption.columnsOf;
+import static io.spine.reflect.Methods.setAccessibleAndInvoke;
+import static io.spine.util.Exceptions.newIllegalStateException;
+
+/**
+ * An extractor of entity columns.
+ */
+final class Introspector {
```
dmdashenkov: An **intro**spector examines its own properties. This is an `Inspector`, or maybe a `Scanner`.

server/src/main/java/io/spine/server/entity/storage/Introspector.java
```diff
@@ -0,0 +1,188 @@
...
+                                                        Class<?> clazz,
+                                                        NoSuchMethodException e) {
+        throw newIllegalStateException(
+                e,
+                "Expected to find a getter with name `%s` in class `%s` according to the " +
+                        "declaration of column `%s`.",
+                getterName, clazz.getCanonicalName(), field.name());
+    }
+
+    private static class ColumnData {
```
dmdashenkov: Please document this class and its factory methods.

server/src/main/java/io/spine/server/storage/memory/EntityQueryMatcher.java
```diff
@@ -112,63 +111,61 @@ private boolean columnValuesMatch(EntityRecordWithColumns record) {
...
-            Optional<MemoizedValue> columnValue = columnValue(record, filter.getKey());
-            if (columnValue.isPresent()) {
-                boolean matches = checkSingleParameter(filter.getValue(), columnValue.get());
-                if (matches) {
-                    return true;
-                }
+        for (Map.Entry<Column, Filter> filter : filters.entries()) {
+            Column column = filter.getKey();
+            if (!record.hasColumn(column.name())) {
+                continue;
```
dmdashenkov: Please try to get rid of the `continue`. I'd use the stream API here instead.

server/src/main/java/io/spine/server/storage/memory/EntityRecordComparator.java
```diff
@@ -22,28 +22,29 @@
...
-import static io.spine.util.Exceptions.newIllegalArgumentException;
+import static io.spine.util.Exceptions.newIllegalStateException;
+import static io.spine.util.Preconditions2.checkNotDefaultArg;
 
 /**
  * A comparator for sorting the contents of {@link TenantRecords}
  * in a provided {@link OrderBy order}.
+ *
+ * @implNote While more sophisticated storage implementations can order records by
+ *         non-{@link Comparable} fields like {@link com.google.protobuf.Message message}-type
```
dmdashenkov: How exactly does a storage sort records by a message field?

PR #1185, opened 2019-10-30
Empty context

alexander-yevsyukov:
This PR adds `EmptyContext` type for complementing message types that do not have a context (such as entity state) where bi-consuming APIs require both message and its context.

Version updates:
  * `spine-base` -> `1.1.9`
  * Gradle -> 5.6.3

The version advanced to `1.1.12`.


core/src/main/proto/spine/core/empty_context.proto
```diff
@@ -0,0 +1,44 @@
...
+
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.core";
+option java_multiple_files = true;
+option java_outer_classname = "EmptyContextProto";
+
+// An empty context is used in combination with message types that do not have a context
+// co conform APIs that require both message and its context.
```
armiol: `to`.

PR #1186, opened 2019-11-04
Client api

alexander-yevsyukov:
This PR introduces high-level client API. Until this PR the client API was based on `ActiveRequestFactory` which allowed to create requests. Placing requests to the backend required dealing with gPRC stubs of `CommandService`, `QueryService`, and `SubscriptionService`, which would require significant learning curve of gRPC from the users of the framework. New Client API requires only knowing connection details to the remote server. In-process implementation is also provided with the primary purpose of writing functional tests.

The Client API wraps around `ActorRequestFactory` providing fluent API for:
  * posting commands;
  * running queries;
  * subscribing to entity states;
  * subscribing to events.

When subscribing the Client API accepts an implementation of a functional interface (not a `StreamObserver` which is not convenient or relevant in majority of client-side cases). Streaming or handling errors can be handled again via specific functional interfaces `ErrorHandler` and `ConsumerErrorHandler` correspondingly.

### Breaking changes
  * `GrpcContainer` moved into the `io.spine.server` package. Previously, the class was placed under the `io.spine.server.transport` package by mistake.


client/src/main/java/io/spine/client/MultiEventConsumers.java
```diff
@@ -0,0 +1,189 @@
...
+import static com.google.protobuf.TextFormat.shortDebugString;
+import static io.spine.client.DelegatingConsumer.toRealConsumer;
+
+/**
+ * An association of event types to their consumers which also delivers events.
+ *
+ * <p>A consumer of an event can accept {@linkplain EventConsumer event message and its context}
+ * or {@linkplain Consumer only event message}.
+ */
+final class MultiEventConsumers implements Logging {
```
dmdashenkov: Looks like an `EventBus`.

client/src/main/java/io/spine/client/QueryRequest.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * <p>Filtering by field values (via {@link #where(Filter...)} and
+ * {@link #where(CompositeFilter...)} methods) can be composed using the {@link Filters}
+ * utility class.
+ *
+ * @param <S>
+ *         the type of the queried entity states
+ * @see Filters
+ */
+public final class QueryRequest<S extends EntityState>
```
dmdashenkov: It's strange that the API of the `FilteringRequest` hierarchy mirrors the API of `TargetBuilders` hierarchy.

client/src/main/java/io/spine/client/TargetBuilder.java
```diff
@@ -42,32 +42,28 @@
  * <p>The {@link Target} matching the builder configuration is created with {@link #buildTarget()},
  * while the {@link FieldMask} is composed with {@link #composeMask()}.
  *
- * <p>The public API of this class is inspired by SQL syntax.
- * <pre>
- *     {@code
+ * <p>The public API of this class is inspired by SQL syntax:
+ * <pre>{@code
```
dmdashenkov: Do we still use `{@code }` for multiline snippets? According to @armiol, it might not be necessary. Let's discuss.

client/src/main/java/io/spine/client/package-info.java
```diff
@@ -19,9 +19,34 @@
...
+ *
+ * <dl>
+ *   <dt><strong>High-level {@link io.spine.client.Client} API</strong></dt>
+ *   <dd>is meant for client-side Java applications that would communicate with backend
+ *   services via a gRPC connection. An instance of the {@link io.spine.client.Client} class
+ *   establishes this connection and serves as a fluent API gateway for composing and posting
+ *   requests.
+ *   </dd>
+ *
+ *   <dt><strong>Low-level {@link io.spine.client.ActorRequestFactory} API</strong></dt>
```
dmdashenkov: Do we need `strong` if we've already goit `dt`?

server/src/main/java/io/spine/server/ConnectionBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+ * <p>A gRPC server can be exposes at a given port or in-process with a given name. In-process
+ * arrangements, while being fully-featured are used primarily for testing.
+ *
+ * @see io.grpc.inprocess.InProcessChannelBuilder
+ * @see io.grpc.inprocess.InProcessServerBuilder
+ */
+public abstract class ConnectionBuilder {
+
+    private @MonotonicNonNull Integer port;
+    private @MonotonicNonNull String serverName;
```
dmdashenkov: Looks like these two fields are really `final` and `@Nullable`.

server/src/main/java/io/spine/server/stand/UpdateHandler.java
```diff
@@ -87,13 +77,14 @@ void handle(EventEnvelope event) {
...
     private void checkActive(EventEnvelope event) {
-        checkState(
-                isActive(),
-                "Dispatched an event of type `%s` to the non-active subscription with the ID `%s`.",
-                event.typeUrl(),
-                subscription.getId()
-                            .getValue()
-        );
+        if (!isActive()) {
+            throw newIllegalStateException(
```
dmdashenkov: Why? `checkState` seems more obvious (as it does not have a negation).

client/src/main/java/io/spine/client/CompositeFilterMixin.java
```diff
@@ -0,0 +1,50 @@
...
+@GeneratedMixin
+interface CompositeFilterMixin extends Message, CompositeMessageFilter<Message> {
+
+    List<Filter> getFilterList();
+    CompositeOperator getOperator();
+
+    @Override
+    default List<MessageFilter<Message>> filters() {
+        List<Filter> list = getFilterList();
+        return ImmutableList.copyOf(list);
```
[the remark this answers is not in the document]
alexander-yevsyukov: The point was to convert the type of the returned `List`. But I'll do it another way.

client/src/main/java/io/spine/client/SubscriptionObserver.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void onNext(SubscriptionUpdate value) {
+        SubscriptionUpdate.UpdateCase updateCase = value.getUpdateCase();
+        switch (updateCase) {
+            case ENTITY_UPDATES:
+                value.getEntityUpdates()
+                     .getUpdateList()
+                     .stream()
+                     .map(EntityStateUpdate::getState)
+                     .map(any -> (M) unpack(any))
```
dmitrykuzmin: The `EntityStateUpdate` may not store a state but have `no_longer_matching` set to `true` instead. Looks like this line will fail in this case.

In general, it seems that we lack the way of telling the client that some entity state that previously matched his subscription criteria does not anymore.

client/src/main/java/io/spine/client/MessageFilter.java
```diff
@@ -0,0 +1,31 @@
...
+package io.spine.client;
+
+import com.google.protobuf.Message;
+
+import java.util.function.Predicate;
+
+/**
+ * Tells whether or not a message satisfies the specified criteria.
+ */
+interface MessageFilter<M extends Message> extends Predicate<M> {
```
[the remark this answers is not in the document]
alexander-yevsyukov: It's not meant to be.

server/src/main/java/io/spine/server/stand/UpdateHandler.java
alexander-yevsyukov: To avoid calculation of `TypeUrl` all the time.

client/src/main/java/io/spine/client/TargetBuilder.java
alexander-yevsyukov: It's not just pre-formatted something (like console example, for example). It is code.

client/src/main/java/io/spine/client/QueryRequest.java
alexander-yevsyukov: We're building another layer of API. That's why it's done this way.

client/src/main/java/io/spine/client/package-info.java
alexander-yevsyukov: We do. This way the tagged text comes in bold, which makes it look like a header. Otherwise, it's the same font as the rest of the text. The `dt/dd` markup is just semantics, and in Javadoc we're limited with how we apply styles.

I've tried headers, and they are way too big for this docs. Have a look like Guava does package docs.

client/src/main/java/io/spine/client/SubscribingRequest.java
```diff
@@ -0,0 +1,121 @@
...
+    public Subscription post() {
+        Topic topic = builder().build();
+        StreamObserver<W> observer = createObserver();
+        return subscribe(topic, observer);
+    }
+
+    /**
+     * Subscribes to receive all messages of the specified type.
+     */
+    public Subscription all() {
```
dmitrykuzmin: Why do we need this method? It looks like it does completely the same as `post()` but discards all previous filtering calls. But how would that be used in the client code?

client/src/main/java/io/spine/client/SubscriptionObserver.java
alexander-yevsyukov: Well spotted! I've added the filtering condition, but this raises a question: “What do we do at the client, when no-longer-matching thing happen?”

alexander-yevsyukov: I'll file an issue for this because it's not a trivial thing, and I don't want to hold the client API because of this matter.

alexander-yevsyukov: Here's the issue on this matter: #1211

client/src/main/java/io/spine/client/CompositeFilterMixin.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+@GeneratedMixin
+interface CompositeFilterMixin extends CompositeFilterOrBuilder, CompositeMessageFilter<Message> {
+
+    @Override
+    default List<MessageFilter<Message>> filters() {
+        List<Filter> filterList = getFilterList();
+        @SuppressWarnings("unchecked") /* The cast below is safe (see
+         https://docs.oracle.com/javase/tutorial/java/generics/subtyping.html for details).
+         `Filter` does implement `MessageFilter<Message>` via `FilterMixin` */
```
dmitrykuzmin:
```suggestion
         `Filter` does implement `MessageFilter<Message>` via `FilterMixin`. */
```

client/src/main/java/io/spine/client/Consumers.java
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * <p>There are two types of errors that may occur when delivering messages to the consumers:
+ * <ol>
+ *     <li><strong>Streaming errors</strong> occur on the
+ *      {@linkplain StreamObserver#onError(Throwable) level of gRPC}. When such an error occur,
+ *      no further messages will be delivered. To handle such an error, please
+ *      supply a streaming {@link Builder#onStreamingError(ErrorHandler) ErrorHandler}.
+ *      If no handler is configured, default implementation will log the error.
+ *
+ *     <li><strong>Consuming errors</strong> occur when one of a consumer throws. Such an error
```
dmitrykuzmin:
```suggestion
 *     <li><strong>Consuming errors</strong> occur when one of the consumers throws. Such an error
```

client/src/main/java/io/spine/client/LoggingConsumerErrorHandler.java
```diff
@@ -0,0 +1,61 @@
...
+
+package io.spine.client;
+
+import com.google.common.flogger.FluentLogger;
+import com.google.protobuf.Message;
+
+import static io.spine.client.DelegatingConsumer.toRealConsumer;
+
+/**
+ * Logs the fact of the error using the {@linkplain FluentLogger#atSevere() server} level
```
dmitrykuzmin:
```suggestion
 * Logs the fact of the error using the {@linkplain FluentLogger#atSevere() severe} level
```

client/src/main/java/io/spine/client/LoggingErrorHandler.java
```diff
@@ -0,0 +1,47 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.client;
+
+import com.google.common.flogger.FluentLogger;
+import com.google.protobuf.Message;
+
+/**
+ * Logs the fact of the error using the {@linkplain FluentLogger#atSevere() server} level
```
dmitrykuzmin:
```suggestion
 * Logs the fact of the error using the {@linkplain FluentLogger#atSevere() severe} level
```

client/src/main/java/io/spine/client/MessageConsumer.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.google.protobuf.Message;
+import io.spine.base.MessageContext;
+
+import java.util.function.BiConsumer;
+
+/**
+ * Functional interface for consumers of messages.
+ *
+ * @param <M>
```
dmitrykuzmin: Please fix the formatting here and below.

.idea/codeStyles/Project.xml
```diff
@@ -30,12 +30,6 @@
     </XML>
     <codeStyleSettings language="Dart">
       <option name="RIGHT_MARGIN" value="100" />
-      <option name="KEEP_BLANK_LINES_IN_CODE" value="1" />
```
dmdashenkov: Please try to pull `config`  and see if this is reverted.

PR #1187, opened 2019-11-06
`EntityState` marker interface

dmitrykuzmin:
This PR applies the newly [generated](https://github.com/SpineEventEngine/base/pull/494) `EntityState` marker interface instead of generic `Message` where appropriate across the code base.

Spine version advances to `1.2.0`.


dmitrykuzmin:
@armiol, @dmdashenkov PTAL. 

The build will pass once https://github.com/SpineEventEngine/base/pull/494 and https://github.com/SpineEventEngine/config/pull/92 are merged.

server/src/main/java/io/spine/server/entity/storage/InterfaceBasedColumn.java
```diff
@@ -96,7 +96,7 @@
     }
 
     @Override
-    public @Nullable Object valueIn(Message entityState) {
+    public @Nullable Object valueIn(EntityState entityState) {
```
alexander-yevsyukov: How about naming the parameter just `state`?

server/src/main/java/io/spine/server/entity/storage/SimpleColumn.java
```diff
@@ -66,7 +66,7 @@
     }
 
     @Override
-    public @Nullable Object valueIn(Message entityState) {
+    public @Nullable Object valueIn(EntityState entityState) {
```
alexander-yevsyukov: How about naming the parameter just `state`?

server/src/main/java/io/spine/server/stand/InMemoryTypeRegistry.java
```diff
@@ -68,8 +68,10 @@ static TypeRegistry newInstance() {
         TypeUrl entityType = repository.entityStateType();
 
         if (repository instanceof RecordBasedRepository) {
+            RecordBasedRepository<I, E, ? extends EntityState> recordBasedRepository
+                    = (RecordBasedRepository<I, E, ? extends EntityState>) repository;
             typeToRepositoryMap.put(entityType,
-                                    (RecordBasedRepository<I, E, ? extends Message>) repository);
+                                    recordBasedRepository);
```
alexander-yevsyukov: The 2nd parameter can be now put on the same line with `entityType`.

server/src/test/proto/spine/test/entity/natural_number.proto
```diff
@@ -33,6 +33,7 @@ option java_multiple_files = true;
 // Zero is considered a natural number.
 //
 message NaturalNumber {
+    option (.entity).kind = ENTITY;
```
alexander-yevsyukov: Do we really need to have the `.` before the `entity`? I know this makes it topmost reference, but we normally don't do this in the type declarations. Why do it here?

dmitrykuzmin: There is a [known issue](https://github.com/protocolbuffers/protobuf/issues/5223) in proto where package name clashes with the option name. In such cases we have to use the dot. 

For now I just renamed the test package, so it looks "cleaner".

PR #1188, opened 2019-11-06
Identity utils

alexander-yevsyukov:
This PR encapsulates creation of identity based on `MessageId` into a utility class so that finding such cases is easier, and returned values can be documented.

It does not address the “mystery” of having `Empty` type URL in the returned values. And that should be discussed and addressed separately.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.1.13`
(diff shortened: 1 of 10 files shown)


dmdashenkov:
@alexander-yevsyukov, from what I see, we could ditch the idea of having an ID of `Empty` and allow system events to identify their producer in some other way.

Issue #1191, opened 2019-11-09
Validate filtering client queries on the client


alexander-yevsyukov:
Suppose I want to subscribe to messages of a certain type filtering by the value of a field.
Suppose I made a mistake in the name of the field. The mistake would result in server-side error because we don't validate the reference when the subscription is posted. This is:
 1) too late;
 2) hard to debug.

We need to have client-side validation of field references before the request is submitted to the server.

alexander-yevsyukov:
It can be made via a mix-in interface. We know the name of the type, we can obtain the descriptor of the message. Get the field reference and see if the field is there.

PR #1197, opened 2019-11-13
Validate the subscription topic on client

dmitrykuzmin:
This PR addresses #1191.

It introduces the `TargetMixin` which allows to check if the target is valid.

By "valid" we assume the target is of `EntityState`/`EventMessage` type and has filters which target existing message fields.

By default, the check is run in `TopicBuilder` on topic creation. Thus, all "standard" API for topic creation will run the check. But, if the topic is manually created and passed to the `SubscriptionService`, it should be checked also manually.

For queries, the target validness check is not run, as the mistake in declared type/filters will be revealed at once, removing the need to double check it on the client.

Spine version advances to `1.2.3`.

### DIFF
--- a/client/src/main/java/io/spine/client/FilterMixin.java
+++ b/client/src/main/java/io/spine/client/FilterMixin.java
@@ -97,7 +97,7 @@ public interface FilterMixin extends FilterOrBuilder {
      */
     default void checkFieldIsColumnIn(Descriptor message) {
         checkNotNull(message);
-        checkFieldIsTopLevel();
+        checkFieldAtTopLevel();
         if (!fieldIsColumnIn(message)) {
             throw newIllegalArgumentException(
                     "The entity column `%s` is not found in entity state type `%s`. " +
@@ -109,7 +109,7 @@ public interface FilterMixin extends FilterOrBuilder {
     /**
      * Checks if the target field is a top-level field.
      */
-    default boolean fieldIsTopLevel() {
+    default boolean fieldAtTopLevel() {
         return !field().isNested();
     }
 
@@ -119,8 +119,8 @@ public interface FilterMixin extends FilterOrBuilder {
      * @throws IllegalArgumentException
      *         if the field is not a top-level field
      */
-    default void checkFieldIsTopLevel() {
-        if (!fieldIsTopLevel()) {
+    default void checkFieldAtTopLevel() {
+        if (!fieldAtTopLevel()) {
             throw newIllegalArgumentException(
                     "The entity filter contains a nested entity column `%s`. " +
                             "Nested entity columns are currently not supported.",
--- a/client/src/main/java/io/spine/client/FilterMixin.java
+++ b/client/src/main/java/io/spine/client/FilterMixin.java
@@ -21,14 +21,21 @@
 package io.spine.client;
 
 import com.google.protobuf.Descriptors.Descriptor;
+import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.Message;
+import io.spine.annotation.GeneratedMixin;
 import io.spine.base.EntityState;
 import io.spine.base.Field;
 import io.spine.base.FieldPath;
+import io.spine.code.proto.ColumnOption;
+import io.spine.code.proto.FieldDeclaration;
 import io.spine.type.TypeUrl;
+
+import java.util.Optional;
 
 import static io.spine.util.Exceptions.newIllegalArgumentException;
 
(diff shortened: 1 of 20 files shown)


dmitrykuzmin:
@armiol, @dmdashenkov PTAL. The build depends on https://github.com/SpineEventEngine/base/pull/498.

client/src/main/java/io/spine/client/FilterMixin.java
```diff
@@ -0,0 +1,147 @@
...
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * Extends the {@link Filter} with validation routines.
+ */
+@GeneratedMixin
+public interface FilterMixin {
+
+    @SuppressWarnings("override") // Implemented in the generated code.
+    FieldPath getFieldPath();
```
armiol: Instead of this, you could extend the interface from `FilterOrBuilder`.

client/src/main/java/io/spine/client/FilterMixin.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    default boolean fieldPresentIn(Descriptor message) {
+        checkNotNull(message);
+        Field field = field();
+        boolean result = field.presentIn(message);
+        return result;
+    }
+
+    /**
+     * Verifies that the target field is present in the passed message type.
```
armiol: Let's also say that it may throw an exception.

client/src/main/java/io/spine/client/FilterMixin.java
```diff
@@ -0,0 +1,147 @@
...
+                    "The entity column `%s` is not found in entity state type `%s`. " +
+                            "Please check the field exists and is marked with `(column)` option.",
+                    field(), message.getFullName());
+        }
+    }
+
+    /**
+     * Checks if the target field is a top-level field.
+     */
+    default boolean fieldIsTopLevel() {
```
armiol: Maybe, `fieldAtTopLevel`?

Same below.

client/src/main/java/io/spine/client/FilterMixin.java
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Validates a filter against the queried type.
+     *
+     * <p>Makes sure the target field is a valid entity column in case the {@link EntityState} is
+     * queried and is a valid message field in all other cases.
+     *
+     * @throws IllegalArgumentException
+     *         if the target field is not present in the type or doesn't satisfy the constraints
+     */
+    default void validateAgainst(TypeUrl targetType) {
```
armiol: This is way too broad. It looks closer to `checkCanFilter(TypeUrl targetType)`. It would be even better if `checkCanApplyTo(Target target)`, but I am not sure that's possible.

client/src/main/java/io/spine/client/TargetMixin.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Collection;
+
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * Extends the {@link Target} with validation routines.
+ */
+@GeneratedMixin
+@SuppressWarnings("override") // Methods are implemented in the generated code.
```
armiol: See my comment above on this matter.

dmitrykuzmin:
@armiol PTAL again.

client/src/test/proto/spine/test/events/events.proto
```diff
@@ -0,0 +1,20 @@
...
+package spine.test.events;
+
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.test.client";
+option java_outer_classname = "EventsProto";
+option java_multiple_files = true;
+
+message ClCreateProject {
```
dmdashenkov: It's misleading that a file called `events.proto` has one command and one value type in it.

client/src/main/java/io/spine/client/TargetMixin.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.Collection;
+
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * Extends the {@link Target} with validation routines.
+ */
+@GeneratedMixin
+public interface TargetMixin extends TargetOrBuilder {
```
dmdashenkov: Please make the mixin interfaces package-private. If they are `public` by design, please document the reasons, as usually that is not the case.

PR #1198, opened 2019-11-15
Allow `Version` in ordering filters

dmitrykuzmin:
This PR adds `io.spine.core.Version` to the types that can be used in the ordering filters ("greater than", "less than or equals", etc.).

It is reasonable because in most storages, including our own ones, the `Version` instances are mapped to values which can be reasonably ordered.

Spine version advances to `1.2.4`.


dmitrykuzmin:
@armiol PTAL.

PR #1199, opened 2019-11-15
`AggregateHistoryCorrupted` event

dmdashenkov:
Sometimes, when an `Apply` method logic changes, or when the infrastructure fails, causing double delivery, Aggregates may obtain a faulty history. In this case, an event (or event**s**) which has already been applied to the Aggregate cause an error during a replay. This renders the Aggregate unable to update its own state beyond the erroneous event.

Starting now, the behaviour of such Aggregates changes:
1. An Aggregate with a corrupted history cannot handle signals.
2. A special system event — `AggregateHistoryCorrupted` — is emitted each time this Aggregate receives signal.

Note that signals dispatched to an Aggregate with a corrupted history are considered delivered, even though the Aggregate never received them. This way, we avoid retrying the delivery each time when a new worker instance picks up the shard to which the Aggregate belongs.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.2.1`
+# Dependencies of `io.spine:spine-client:1.2.2`
 
 ## Runtime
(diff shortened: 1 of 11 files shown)


dmdashenkov:
@armiol, PTAL.

PR #1200, opened 2019-11-19
Enable multi-threading in `Bus`es

armiol:
Previously, all the `Bus`es allowed a single-threaded execution only. They used a synchronized `DispatchingQueue` in order to guarantee that the only message was being processed at a time. In a local environment that solved some of potential ordering issues.

However, now, as `Delivery` mechanisms are responsible for ordering and de-duplication, a single-threaded `Bus` becomes a bottle-neck. It makes impossible to implement a multi-threaded delivery and significantly increases the time from when a message is posted to the Bounded Context until it's written to the `Inbox` of its target entity.

This PR addresses such an issue. In particular, it
* removes the synchronization in `Bus`es;
* introduces local versioning in `Inbox` and thus guarantees that the messages emitted in the same server instance will be written to the `Inbox` and then delivered in the order of appearance;
* tweaks the timing of the notifications sent upon a shard update (`ShardObserver.onMessage(...)`) until all of the message dispatchers receive this new message to their `Inboxes`.

The latter is important. Before this changeset, the observers were notified of a new `Inbox` message right after it was written to the `Inbox`. That was fine for commands, but for the events, it did not really work as intended. 

An event may potentially be dispatched to many dispatchers. That's why we need to make sure that each of them receives this event in its `Inbox`. Only then can we notify the outer world to start processing the shard content. Otherwise, it is not possible to guarantee the ordering of the events dispatched to a particular dispatcher.

In order to achieve this, a `MulticastBus` notifies the `Delivery` when a particular message starts and finishes its dispatching. Upon the completion of this process, the shard observers are notified of changes.

As a result of the changes, the `Bus`es stopped to guarantee that the only message is being dispatched at a time in a local environment. Therefore, a #925 has been reopened. The previous behavior would not be consistent in the distributed behavior anyway, so such a regression does not look critical. In future we may need to find a solution for such an eventual consistency (see the issue for more details).

**Side changes**

A proper type checks were added to the message routing phase. Previously, it was possible to pass a wrong type of an ID to the routing and get away with that until the framework was forced into an exceptional state during the dispatching. Now `RoutingFailed` is emitted instead.

Delivery now checks for the "late" messages that may have arrived after all the shard messages were delivered and before the lock on the shard has been released. This is a very rare case, currently reproduced only in in-memory running mode. However, it may happen under heavy load in a production mode.

`DeliveryTest` made `public` in order to enable other Spine libraries to run these tests with different storage providers.

The framework version becomes `1.2.5`.


server/src/main/java/io/spine/server/bus/Bus.java
```diff
@@ -317,14 +319,61 @@ public boolean hasListener(Listener<E> listener) {
...
+            try {
+                dispatch(envelope);
+            } catch (Throwable t) {
+                _error().withCause(t)
+                        .log("Error when dispatching %s[ID: %s].",
+                             envelope.messageClass(),
+                             signalId);
+                throw t;
+            } finally {
+                onDispatched(signalId);
```
dmdashenkov: I'm not sure if we have to do the same thing (e.g. delivery) if dispatching fails.

server/src/main/java/io/spine/server/entity/Repository.java
```diff
@@ -373,6 +376,36 @@ public final boolean isOpen() {
...
+     * the identifiers} served by this repository.
+     *
+     * <p>An {@code IllegalStateException} is thrown otherwise.
+     *
+     * @param result the result of routing
+     * @param <R> the type of the result
+     */
+    private <R> void checkMatchesIdType(R result) {
+        Class routingResultType = null;
+        if(result instanceof Collection) {
```
dmdashenkov: Please fix code alignment here.

server/src/main/java/io/spine/server/entity/Repository.java
```diff
@@ -373,6 +376,36 @@ public final boolean isOpen() {
...
+            Collection asCollection = (Collection) result;
+            Iterator iterator = asCollection.iterator();
+            if(iterator.hasNext()) {
```
dmdashenkov: Please use `Iterables.getFirst(..)` instead.

server/src/main/java/io/spine/server/storage/memory/InMemoryInboxStorage.java
```diff
@@ -77,38 +78,62 @@ TenantInboxRecords createSlice() {
...
 
     @Override
-    public void write(InboxMessage message) {
+    public synchronized Optional<InboxMessage> oldestMessageToDeliver(ShardIndex index) {
+        TenantInboxRecords storage = multitenantStorage.currentSlice();
+        Optional<InboxMessage> result =
+                storage.readAll()
+                       .stream()
+                       .filter((r) -> index.equals(r.getShardIndex()) &&
+                               r.getStatus() == InboxMessageStatus.TO_DELIVER)
```
dmdashenkov: Please extract into a method or a variable with a lambda so that the alignment could be better.

server/src/main/java/io/spine/server/storage/memory/InMemoryInboxStorage.java
```diff
@@ -77,38 +78,62 @@ TenantInboxRecords createSlice() {
...
+
+    @Override
+    public synchronized void write(InboxMessage message) {
         multitenantStorage.currentSlice()
                           .put(message.getId(), message);
     }
 
     @Override
-    public void writeAll(Iterable<InboxMessage> messages) {
+    public synchronized void writeAll(Iterable<InboxMessage> messages) {
```
dmdashenkov: Please document why are the methods `synchronized`.

server/src/main/java/io/spine/server/storage/memory/TenantAggregateRecords.java
```diff
@@ -82,7 +84,7 @@
      *
      * @return immutable list
      */
-    List<AggregateEventRecord> historyBackward(AggregateReadRequest<I> request) {
+    synchronized List<AggregateEventRecord> historyBackward(AggregateReadRequest<I> request) {
```
dmdashenkov: Same as above.

server/src/main/java/io/spine/server/storage/memory/TenantInboxRecords.java
```diff
@@ -103,8 +105,11 @@ public int compare(InboxMessage first, InboxMessage second) {
             if (timeComparison != 0) {
                 return timeComparison;
             }
-            int hashCodeComparison = Integer.compare(first.hashCode(), second.hashCode());
-            return hashCodeComparison;
+            int versionComparison = Integer.compare(first.getVersion(), second.getVersion());
+            if(versionComparison == 0) {
```
dmdashenkov: Please fix code alignment.

server/src/test/java/io/spine/server/delivery/given/TaskAssignment.java
```diff
@@ -0,0 +1,64 @@
...
+import io.spine.server.procman.ProcessManager;
+import io.spine.server.procman.ProcessManagerRepository;
+import io.spine.server.route.EventRoute;
+import io.spine.server.route.EventRouting;
+import io.spine.test.delivery.DTaskAssigned;
+import io.spine.test.delivery.DTaskAssignment;
+import io.spine.test.delivery.DTaskCreated;
+import io.spine.testing.core.given.GivenUserId;
+
+
```
dmdashenkov: Please remove the empty line.

server/src/main/java/io/spine/server/bus/Bus.java
armiol: As discussed vocally, we must tell that the dispatching is completed to set the potentially "waiting" listeners free.

server/src/main/java/io/spine/server/storage/memory/InMemoryInboxStorage.java
armiol: All of the mutating methods were `synchronized` already except for this one. The corresponding explanation added at the top of the class.

server/src/main/java/io/spine/server/delivery/InboxPart.java
```diff
@@ -182,4 +185,33 @@ private boolean duplicate(InboxMessage message) {
...
+     * A counter providing the version for the incoming {@code InboxMessage}s.
+     *
+     * <p>Serves to distinguish the messages arrived at the very same millisecond, even from
+     * different threads.
+     *
+     * <p>Hard-coded to reset the value to zero every {@link #MAX_VERSION} times. This allows
+     * to avoid any number overflows, but still is sufficient for realistic use-cases.
+     */
+    @ThreadSafe
+    private static final class Version {
```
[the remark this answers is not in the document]
armiol: I don't want fancy naming here. Renamed to `VersionCounter` and moved to the package level.

server/src/main/java/io/spine/server/delivery/VersionCounter.java
```diff
@@ -0,0 +1,53 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.delivery;
+
+import javax.annotation.concurrent.ThreadSafe;
```
dmdashenkov: How come we even have this? This is a part of JSR 305. I thought we decided to get rid of it.

server/src/main/java/io/spine/server/delivery/VersionCounter.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@ThreadSafe
+final class VersionCounter {
+
+    private static final int MAX_VERSION = 10_000;
+    private static final VersionCounter instance = new VersionCounter();
+
+    private final AtomicInteger counter = new AtomicInteger();
+
+    private synchronized int getNextValue() {
```
dmdashenkov: Why have an `AtomicInteger` if we only use it in `synchronized` code?

server/src/main/java/io/spine/server/delivery/VersionCounter.java
dmdashenkov: I've checked and we depend on `com.google.code.findbugs:jsr305:3.0.2`. Maybe it's time to retire it in favour of the Checker Framework and others. Let's discuss.

server/src/main/java/io/spine/server/delivery/Delivery.java
```diff
@@ -475,4 +518,60 @@ protected void onShardUpdated(InboxMessage message) {
...
+     */
+    private class DeliveryDispatchListener implements MulticastDispatchListener {
+
+        private final Multimap<SignalId, InboxMessage> pending =
+                synchronizedListMultimap(MultimapBuilder.hashKeys()
+                                                        .arrayListValues()
+                                                        .build());
+
+
+        private final Set<SignalId> currentlyDispatching = synchronizedSet(new HashSet<>());
```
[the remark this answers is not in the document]
armiol: As discussed vocally, we use a `Synchronized` instead of a `Concurrent` since the concurrent-one allows simultaneous access. Meaning that writes may happen at the very same time as reads, and write results may yet not be visible. What we want is to make sure that everything written is visible for reading.

PR #1215, opened 2019-12-16
Fix enum columns propagation

dmitrykuzmin:
This PR fixes an issue with the enum-type columns propagation.

For the enumerated types, the `Message.Builder.setField(...)` method requires a `EnumValueDescriptor` as an input rather than the raw enum value. 

This changeset takes that into account when updating the entity column values in a transaction.

Spine version advances to `1.2.10`.

### DIFF
--- a/server/src/main/java/io/spine/server/entity/Transaction.java
+++ b/server/src/main/java/io/spine/server/entity/Transaction.java
@@ -479,9 +479,9 @@ public abstract class Transaction<I,
      * getter to the entity state.
      *
      * <p>The enum-typed columns require an additional conversion as the
-     * {@link Message.Builder#setField(Descriptors.FieldDescriptor, Object) setField(...)} method
-     * should receive a {@link com.google.protobuf.Descriptors.EnumValueDescriptor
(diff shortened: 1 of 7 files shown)


dmitrykuzmin:
@armiol PTAL.

The build depends on https://github.com/SpineEventEngine/base/pull/501.

dmitrykuzmin:
@dmdashenkov PTAL.

server/src/main/java/io/spine/server/entity/Transaction.java
```diff
@@ -467,14 +469,28 @@ private S stateWithColumns() {
...
     }
 
+    /**
+     * Propagates a column value which is obtained with the help of a manually implemented column
+     * getter to the entity state.
+     *
+     * <p>The enum-typed columns require an additional conversion as the
+     * {@link Message.Builder#setField(Descriptors.FieldDescriptor, Object)} method should receive
+     * a {@link com.google.protobuf.Descriptors.EnumValueDescriptor EnumValueDescriptor} instance
+     * per Protobuf rules.
```
dmdashenkov:
```suggestion
     * by the Protobuf rules.
```

PR #1219, opened 2020-01-02
Integrate validation code generation

dmdashenkov:
In this PR we migrate core modules to `base` and `time` of version `1.3.2`, which include the generated validation code.

Code generation for core types is turned on.

The version is updated to `1.3.2`.


dmdashenkov:
@armiol, PTAL.

PR #1221, opened 2020-01-09
Projection Catch-up

armiol:
This PR introduces the `Projection` catch-up and re-structures the `Delivery` routines.

## Catch-up

A `Projection` catch-up is a replay of the historical events from the event log on the requested projection instances, starting a certain point in the past.

During the catch-up, the previous state of the catching-up projection instance is killed. The new state then starts to continuously build-up from the historical events, starting from the point in time specified.

**Use-cases**

Typical scenarios in which a catch-up may be helpful are listed below.

* A new `Projection` type is registered and the instances of this type need to be populated with the historical data.
* The business logic of an existing `Projection` has changed and its instances must be re-built.
* A technical failure has occurred somewhere in the past, breaking the state of some of the `Projection` instances.

**Starting a catch-up**

The `ProjectionRepository` now exposes an API to start a catch-up of its instances:

```java

OrderViewProjectionRepository repo = new OrderViewProjectionRepository();

// skipping the registration in the Bounded Context.

Timestamp catchUpSince = ...;
OrderId myOrderId = ...;
OrderId anotherOrderId = ...;

repo.catchUp(catchUpSince, ImmutableSet.of(myOrderId, anotherOrderId));
```

It is impossible to catch-up the same instance more than once at a time. A respective `CatchUpAlreadyStartedException` is then thrown.

In case a new repository has been registered and needs to be filled with the projections according to the event history, another API endpoint is available:

```java
QuarterlyReportRepository repo = new QuarterlyReportRepository();

// skipping the registration in the Bounded Context.

Timestamp catchUpSince = ...;

repo.catchUpAll(catchUpSince);
```

Once the catch-up is completed, a special `CatchUpCompleted` event is emitted by the framework.

During the catch-up, the live messages dispatched to the catching-up entities are not delivered to them. When the catch-up is completed, the dispatching of the live events continues in a normal fashion. The framework guarantees there will be no events lost during the catch-up.

**Configuring the catch-up through `DeliveryBuilder`**

From now on, in addition to delivering the messages sent in a real-time, `Delivery` dispatches the historical events sent to the catching-up projections. These events are dispatched through the same shards as the live messages. A special `CatchUpStation` is responsible for handling this use-case. See more on that in the respective section.

To control how many historical events are read and put into shards, the end-users may configure the maximum number of messages read from the history per at a time.

```java
Delivery.newBuilder()
        .setCatchUpPageSize(42)
        // ... the rest of the configuration
```

This is helpful to balance the per-shard throughput so that the live messages are still dispatched through the same shards in a reasonable time.

The statuses of the ongoing catch-up processes are stored in a dedicated `CatchUpStorage`, which can also be customized either through the `StorageFactory` or via the `DeliveryBuilder`:

```java
Delivery.newBuilder()
        .setCatchUpStorage(myCustomStorage)
        // ... the rest of the configuration
```
The latter is useful, if the creation of a custom `CatchUpStorage` requires more application-specific context.

## Changes in `Delivery`

As previously the delivery process for each shard index is now split into `DeliveryStage`. In the scope of
each stage, a certain number of messages is read from the respective shard of the `Inbox`.

In the scope of each `DeliveryStage`, the page of the `InboxMessage`s is placed to the newly introduced `Conveyor`, which is responsible for tracking the status of each message. The conveyor instance is then run through the pipeline of stations, each modifying the state of the messages in the conveyor.
At the end of the pipeline, the changes made to the messages are committed to the underlying
`InboxStorage` in a bulk. Such an approach allows minimizing the number of requests sent to the storage.

`InboxMessage`s now have a new `Timestamp keep_until` field, which tells for how long the message should be kept in its storage after it is delivered. In this way, the already delivered messages are used as a source for deduplication. As previously, the framework users may configure this value
via the `DeliveryBuilder`:
```java 
Delivery.newBuilder()
        .setIdempotenceWindow(Durations.fromSeconds(15))
        // ... the rest of the configuration
```        

Below is the list of the conveyor stations in the current `Delivery` pipeline.

1. Catch-up station

This station is responsible for dispatching the historical events in
`TO_CATCH_UP` status to the respective targets. Also, while the target entity is under a catch-up, 
all the live messages headed to it are removed by this station. 

Once the catch-up is completed, this station handles the transition period, in which the last batch of the historical events and live messages are dispatched together. See the `CatchUpStation` API documentation for more details.

2. Live delivery station

This station is responsible for dispatching the messages sent in a real-time by filtering the conveyor contents for the messages in `TO_DELIVER` Another responsibility of this station is to set for how long the delivered messages should be kept according to the idempotence window settings. See the `LiveDeliveryStation` API documentation for more details.

3. Cleanup station

This station removes the messages which are already delivered and are no longer needed for the
deduplication. See the `CleanupStation` for the description.

**Deduplication**

During the dispatching, the `Conveyor` keeps track of the delivered messages. The stations performing the actual message dispatching rely on this knowledge and deduplicate the messages prior to calling the target's endpoint.

Additionally, the `Delivery` now provides a `DeliveredMessages`, which is a cache of recently delivered messages.  Each instance of the `Conveyor` has access to it and uses it in deduplication procedures.

**Shard maintenance**

In this changeset, a special `ShardMaintenanceProcess` has been introduced. Via it, the `Delivery` performs the maintenance of the shards. In the scope of this PR, a special `ShardProcessingRequested` event is handled by this process. It is guaranteed to be dispatched to the maintenance process via the selected shard, thus triggering the `Delivery` to handle the messages from it. It is required to ensure that all historical events are delivered to all their targets. See more on that in the description of the `CatchUpProcess` down below.

## How `CatchUpProcess` works

The process is started by `CatchUpStarter` composing a request received from the `ProjectionRepository`'s `catchUp(..)`/ `catchUpAll(..)` methods. A corresponding `CatchUpRequested` event is emitted. 

In its lifecycle, this process moves through several statuses.

**Not started**

The process is created in this status upon receiving a `CatchUpRequested` event. The actions
include:

 * A `CatchUpStarted` event is emitted. The target projection repository listens to this event and kills the state of the matching entities.
 * The catch-up process in moved to the `IN_PROGRESS` status.

**`IN_PROGRESS`**

When the process is in this status, the event history is read and the matching events are sent to the `Inbox`es of the corresponding projections.

The catch-up maintains this status until history is read until the point in time, which is very close to the `Time.currentTime()`.

At this stage, the actions are as follows.

 * The historical event messages are read from the `EventStore` respecting the time range requested and the time of the last read operation performed by this process. The maximum number of the events read is determined by `DeliveryBuilder.setCatchUpPageSize(int)`.

 * Depending on the targets requested for the catch-up, the events are posted to the corresponding entities.

 * Unless the timestamps of the events are getting close to the current time, an `HistoryEventsRecalled` is emitted, leaving the process in the `IN_PROGRESS` status and triggering the next round similar to this one.

 * If the timestamps of the events read on this step are as close to the current time as `CatchUpProcessBuilder.withTurbulencePeriod(Duration)` value — so-called  "turbulence" period, which by default is `500` ms — the `HistoryFullyRecalled` is emitted.

**`FINALIZING`**

The process moves to this status when the event history has been fully recalled and the corresponding `HistoryFullyRecalled` is received. 

At this stage the `Delivery` stops the propagation of the events to the catch-up messages, waiting for this process to populate the inboxes with the messages arriving to be dispatched during the "turbulence" period.

Potentially, the inboxes will contain the duplicates produced by both the live users and this process. To deal with it, a deduplication is performed by the `Delivery`.
See `CatchUpStation` API documentation for more details on that.

The actions are as follows.

 * All the remaining messages of the matching event types are read `EventStore`. In this operation, the read limits set by the `Delivery` are NOT used, since the goal is to read the remainder of the events.

 * If there were no events read, the `CatchUpCompleted` is emitted and the process is moved to the `COMPLETED` status.

 * If there were some events read, the `LiveEventsPickedUp` event is emitted. This allows some time for the posted events to become visible to the `Delivery`. The reacting handler of the `LiveEventsPickedUp` completes the process, once this event is delivered to this process.

**`COMPLETED`**

Once the process moves to the `COMPLETED` status, the corresponding `Delivery` routines deduplicate, reorder and dispatch the "paused" events. After that, normal live delivery flow is resumed.

To ensure that all the shards in which the "paused" historical events reside are processed, this process additionally emits the "maintenance" `ShardProcessingRequested` events for each shard involved. By dispatching them, the system guarantees that the `Delivery` observes the `COMPLETED` status of this process and delivers the remainder of the messages.

## Other changes

**`AbstractStatefulReactor`**

The `CatchUp` process manager implemented within this changeset should have had access to the `EventStore` of the Domain Bounded Context. Therefore it was not possible to have it registered in the System counterpart. However, it is impossible to have a process manager with a certain state more than in one Bounded Context.

Therefore, a special `AbstractStatefulReactor` was introduced. It is an `AbstractEventReactor` with state. Unlike conventional `AbstractEventReactor`s, the instances of this one have their own `Inbox`es and are served with the messages through `Delivery`. The `CatchUpProcess` is one good example of such an event reactor.

**`Limit` for `EventStore` queries**

It is now possible to specify the maximum number of results when reading from an `EventStore`:
```java
EventStreamQuery.Limit limit =  EventStreamQuery.Limit
                    .newBuilder()
                    .setValue(value)
                    .vBuild();
EventStreamQuery query = EventStreamQuery
                    .newBuilder()
                    .setAfter(longPast)
                    .setBefore(future)
                    .setLimit(limit)
                    .vBuild();
 eventStore.read(query, observer);                    
 ```

**`lastHandledEventTime`**

A per-`ProjectionRepository` property called `lastHandledEventTime` has been removed as obsolete.

**Idempotence window**

In order to be closer in the meaning, the `idempotenceWindow` configuration property has been renamed to `deduplicationWindow`:
```java
Delivery.newBuilder()
        .setDeduplicationWindow(Durations.fromSeconds(10));
```

## Version update

 The framework version is set to `1.4.6`.


alexander-yevsyukov:
> 1.4.5
@armiol, it's a big feature change. I'd say we need to put it as `1.5`.

server/src/main/java/io/spine/server/delivery/CatchUpStarter.java
```diff
@@ -0,0 +1,238 @@
...
+                .collect(toList());
+        boolean alreadyCatchingUp = hasIntersections(active, ids);
+        if (alreadyCatchingUp) {
+            throw new CatchUpAlreadyStartedException(projectionStateType, ids);
+        }
+    }
+
+    private static ActorRequestFactory requestFactory(UserId actor, boolean multitenant) {
+        TenantFunction<ActorRequestFactory> function =
+                new TenantFunction<ActorRequestFactory>(multitenant) {
```
dmdashenkov: Please extract into a helper class.
Also, I'm not sure if this is good enough. Can this ever be called from a custom thread (i.e. via an `Executor`)? Maybe, we should pass the tenant ID as a parameter to the builder.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,252 @@
...
+    @SuppressWarnings({"MethodWithMultipleLoops", "OverlyComplexMethod", "OverlyNestedMethod"})
+    @Override
+    public final Result process(Conveyor conveyor) {
+        Map<DispatchingId, InboxMessage> dispatchToCatchUp = new HashMap<>();
+
+        for (InboxMessage message : conveyor) {
+            for (CatchUp job : jobs) {
+                CatchUpStatus jobStatus = job.getStatus();
+
+                if (matches(job, message)) {
```
dmdashenkov: Please split and/or add inline comments.
This probably can be a few stations instead of just one.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,252 @@
...
+     * Tells whether the job matched the passed {@code InboxMessage}.
+     *
+     * <p>To match, two conditions must be met:
+     *
+     * <ol>
+     *     <li>the target entity type of the job and the message must be the same;</li>
+     *
+     *     <li>the identifier of the message target must be included into the list of the
+     *     identifiers specified in the job OR the job matches all the targets of the entity type.
+     *     </li>
```
dmdashenkov: Please remove `</li>`s.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,252 @@
...
+                                     .getId();
+        return targets.stream()
+                      .anyMatch((t) -> t.equals(rawEntityId));
+    }
+
+    /**
+     * The comparator which sorts the messages chronologically, but ensures that if there is
+     * a {@link CatchUpStarted} event in the sorted batch, it goes on top.
+     */
+    private static final class CatchUpMessageComparator
```
dmdashenkov: Please extract to upper level.

server/src/main/java/io/spine/server/delivery/InboxPart.java
```diff
@@ -117,70 +106,49 @@ private InboxSignalId signalIdFrom(M envelope, I targetId) {
...
+
+        @SuppressWarnings("unchecked")    // Only IDs of type `I` are stored.
+                I unpackedId = (I) InboxIds.unwrap(message.getInboxId());
+        TenantAwareRunner
+                .with(envelope.tenantId())
+                .run(() -> call.invoke(endpoint, unpackedId, envelope));
+    }
+
+    /**
+     * Determines how to perform the call to the endpoint.
```
dmdashenkov:
```suggestion
     * Passes a message to an endpoint.
```

server/src/main/java/io/spine/server/delivery/LiveDeliveryStation.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Dispatches the messages from the conveyor.
+     *
+     * @param conveyor
+     *         the conveyor on which the messages are travelling
+     * @return how many messages were delivered and whether there were any errors during the
+     *         dispatching
+     */
+    @Override
+    public final Result process(Conveyor conveyor) {
```
dmdashenkov: Please split into several helper methods.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,252 @@
...
+     * </ol>
+     *
+     * @param job
+     *         the catch-up job
+     * @param message
+     *         the message to match to the job
+     * @return {@code true} if the message matches the job, {@code false} otherwise
+     */
+    @VisibleForTesting
+    static boolean matches(CatchUp job, InboxMessage message) {
```
alexander-yevsyukov: This looks like a candidate for a mixin method for `CatchUpMixin` (which I previously commented as probably not used).

server/src/main/java/io/spine/server/event/AbstractStatefulReactor.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * An abstract base for {@link AbstractEventReactor}s which have state and require
+ * {@linkplain io.spine.server.delivery.Delivery delivering} the messages through an
+ * {@link io.spine.server.delivery.Inbox Inbox}.
+ */
+public abstract class AbstractStatefulReactor<I, S extends Message, B extends ValidatingBuilder<S>>
+        extends AbstractEventReactor {
+
+    private final Inbox<I> inbox;
+    private final Object endpointLock = new Object();
```
dmdashenkov: Please use a `ReentrantLock` instead.

server/src/main/java/io/spine/server/tenant/IdInTenant.java
```diff
@@ -88,13 +89,21 @@ public boolean equals(Object o) {
...
+                Objects.equals(tenantId, another.tenantId);
     }
 
     @Override
     public int hashCode() {
         return Objects.hash(id, tenantId);
     }
+
+    @Override
+    public String toString() {
```
dmdashenkov: Please use `ToStringHelper` in this case and in similar cases in the change set.

server/src/main/proto/spine/server/catchup/catch_up.proto
```diff
@@ -0,0 +1,95 @@
...
+    //
+    int32 current_round = 5;
+
+    // The total number of shards in the environment
+    // at the moment of the last dispatching of the historical messages.
+    //
+    int32 total_shards = 6;
+
+    // The indexes of the shards through which the historical messages were dispatched.
+    repeated int32 affected_shard = 7;
```
dmdashenkov: Why not `ShardId`s?

server/src/test/java/io/spine/server/delivery/given/ConsecutiveNumberProcess.java
```diff
@@ -0,0 +1,57 @@
...
+import io.spine.server.command.Assign;
+import io.spine.server.procman.ProcessManager;
+import io.spine.server.tuple.Pair;
+import io.spine.test.delivery.ConsecutiveNumber;
+import io.spine.test.delivery.EmitNextNumber;
+import io.spine.test.delivery.NegativeNumberEmitted;
+import io.spine.test.delivery.PositiveNumberEmitted;
+
+/**
+ * @author Alex Tymchenko
```
dmdashenkov: Is this a settings fluke?

server/src/test/proto/spine/test/delivery/delivery_events.proto
```diff
@@ -46,3 +46,17 @@ message DTaskAssigned {
 
     spine.core.UserId assignee = 2;
 }
+
+message PositiveNumberEmitted {
+
+    string id = 1;
+
+    int32 value = 2 [(min).value = "0"];
```
dmdashenkov: Please add `exclusive = true` for this field and for `NegativeNumberEmitted. value`.

server/src/test/java/io/spine/server/delivery/CatchUpTest.java
```diff
@@ -0,0 +1,355 @@
...
+        public Timestamp currentTime() {
+            Instant now = Instant.now();
+            Timestamp result = Timestamp.newBuilder()
+                                        .setSeconds(now.getEpochSecond())
+                                        .setNanos(now.getNano())
+                                        .build();
+            return result;
+        }
+    }
+
```
dmdashenkov: Please remove this line.

server/src/test/java/io/spine/server/delivery/given/CounterView.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public static final class Repository
+            extends ProjectionRepository<String, CounterView, DCounter> {
+
+        @OverridingMethodsMustInvokeSuper
+        @Override
+        protected void setupEventRouting(EventRouting<String> routing) {
+            super.setupEventRouting(routing);
+            routing.route(NumberAdded.class, (EventRoute<String, NumberAdded>) (message, context) ->
```
dmdashenkov: I suggest:
```java
routing.route(NumberAdded.class,
              (message, context) -> withId(message.getCalculatorId()));
```

server/src/main/java/io/spine/server/delivery/ShardMaintenanceProcess.java
```diff
@@ -0,0 +1,98 @@
...
+
+import static io.spine.util.Exceptions.newIllegalStateException;
+
+/**
+ * The framework-internal process performing the maintenance of delivery shards.
+ *
+ * <p>Has its own {@link Inbox}, so the messages arriving to it are dispatched
+ * by the {@link Delivery}.
+ */
+class ShardMaintenanceProcess extends AbstractEventReactor {
```
alexander-yevsyukov: Can we make this class final?

server/src/main/java/io/spine/server/delivery/ShardMaintenanceProcess.java
```diff
@@ -0,0 +1,98 @@
...
+     * By handling this event, guarantees that the messages in the
+     * {@linkplain ShardProcessingRequested#getId() specified shard} have been processed
+     * by the {@link Delivery}.
+     */
+    @React
+    Nothing on(ShardProcessingRequested event) {
+        return nothing();
+    }
+
+    @CanIgnoreReturnValue
```
alexander-yevsyukov: I don't think we need to have this annotation here and in the parent class because the method returns void.

server/src/main/java/io/spine/server/storage/memory/TenantCatchUpRecords.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.Collections.synchronizedMap;
+
+/**
+ * The catch-up processes for for a tenant.
+ */
+public class TenantCatchUpRecords implements TenantStorage<CatchUpId, CatchUp> {
```
alexander-yevsyukov: Can we make this class package-private?

server/src/main/java/io/spine/server/delivery/ShardMaintenanceProcess.java
alexander-yevsyukov: Also, as discussed vocally, let's make this class derive from `AbstractEventSubscriber`.

server/src/main/proto/spine/server/event/event_stream_query.proto
```diff
@@ -43,14 +43,33 @@ message EventStreamQuery {
...
+
+    // Set the limit on the maximum number of events returned in scope of a single query.
+    //
+    // If this parameter is not specified, the resulting set of events is not restricted in size.
+    //
+    Limit limit = 4;
+
+    // Definition of the query `LIMIT`.
+    //
+    // Sets the maximum number of events to return from an EventStore.
```
alexander-yevsyukov: Please backtick `EventStore`.

server/src/test/java/io/spine/server/delivery/CatchUpStationTest.java
```diff
@@ -0,0 +1,287 @@
...
+            ImmutableSet<InboxMessage> messages = messagesToTargetOneOf(type);
+            CatchUp job = catchUpJob(type, STARTED, currentTime(), null);
+            assertMatchesEvery(messages, job);
+        }
+
+        private void assertMatchesEvery(ImmutableSet<InboxMessage> messages, CatchUp job) {
+            for (InboxMessage message : messages) {
+                assertThat(CatchUpStation.matches(job, message))
+                        .isTrue();
+
```
alexander-yevsyukov: Please remove this empty line.

server/src/test/java/io/spine/server/delivery/CatchUpStationTest.java
```diff
@@ -0,0 +1,287 @@
...
+            ImmutableSet<InboxMessage> messages = messagesToTargetOneOf(type);
+            CatchUp job = catchUpJob(type, STARTED, currentTime(), ImmutableSet.of(targetTwo));
+            assertMatchesNone(messages, job);
+        }
+
+        private void assertMatchesNone(ImmutableSet<InboxMessage> messages, CatchUp job) {
+            for (InboxMessage message : messages) {
+                assertThat(CatchUpStation.matches(job, message))
+                        .isFalse();
+
```
alexander-yevsyukov: Please remove this empty line.

server/src/test/java/io/spine/server/delivery/CatchUpTest.java
```diff
@@ -0,0 +1,355 @@
...
+    public void tearDown() {
+        super.tearDown();
+        Time.resetProvider();
+    }
+
+    /**
+     * Uses the default wall clock provider, which now emulates the nanoseconds by default.
+     */
+    @Nested
+    @DisplayName(", given the time is provided with nanosecond resolution,")
```
alexander-yevsyukov: Please remove prefixing commas here and below. They look confusing in the test report. It looks as if there was some text before which was accidentally dropped.

server/src/test/proto/spine/test/delivery/delivery_events.proto
alexander-yevsyukov: Why cannot it be just "1" then?

server/src/main/proto/spine/server/catchup/catch_up.proto
armiol: This is done to save up the space taken by this record in the storage with max BLOB size.  There may be lots of shards involved into the message dispatching, and Cloud Datastore [allows us to store `1,048,572` bytes per record at most](https://cloud.google.com/datastore/docs/concepts/limits). Basically, if there are 1000 shards,  we would take at least 4000 bytes more + the overhead for Proto message format, which I believe takes at least as  much.

server/src/main/java/io/spine/server/delivery/CatchUpStarter.java
```diff
@@ -0,0 +1,238 @@
...
+                                                  .setActor(actor)
+                                                  .setTenantId(id)
+                                                  .build();
+                    }
+                };
+        return function.execute();
+    }
+
+    @SuppressWarnings("MethodWithMultipleLoops")
+    private static boolean hasIntersections(Iterable<CatchUp> ongoing, @Nullable Set<?> ids) {
```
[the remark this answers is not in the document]
armiol: As discussed, it's hardly doable in this case.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
armiol: I have tried to reduce the number of classes in the `.delivery` package as it's already huge. That's why it's here.

server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
```diff
@@ -0,0 +1,531 @@
...
+ * each shard involved. By dispatching them, the system guarantees that the {@code Delivery}
+ * observes the {@code COMPLETED} status of this process and delivers the remainer of the messages.
+ *
+ * @implNote Technically, the instances of this class are not
+ *         {@linkplain io.spine.server.procman.ProcessManager process managers}, since it is
+ *         impossible to register the process managers with the same state in
+ *         a multi-{@code BoundedContext} application.
+ */
+@SuppressWarnings({"ClassWithTooManyMethods", "OverlyCoupledClass"})    // It does a lot.
+public final class CatchUpProcess<I>
```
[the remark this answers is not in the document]
armiol: Documented with more love.

server/src/main/java/io/spine/server/delivery/Delivery.java
```diff
@@ -72,24 +57,61 @@
...
+ * the same shards as the live messages. A special {@link CatchUpStation} is responsible for
+ * handling this use-case. See more on that in the respective section.
+ *
+ * <p>To control how many historical events are read and put into shards, the end-users may
+ * configure the {@linkplain DeliveryBuilder#setCatchUpPageSize(int) maximum number of messages}
+ * read from the history at a time. This is helpful to balance the per-shard throughput, so
+ * that the live messages are still dispatched through the same shards in a reasonable time.
+ *
+ * <p>The statuses of the ongoing catch-up processes are stored in a dedicated
+ * {@link CatchUpStorage}. The {@code DeliveryBuilder} {@linkplain
```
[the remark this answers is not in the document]
armiol: Addressed by using the alt-text properly.

server/src/main/java/io/spine/server/event/AbstractStatefulReactor.java
armiol: As discussed, `synchronized` is good enough. It cannot be misused.

server/src/main/java/io/spine/server/delivery/LiveDeliveryStation.java
armiol: It's short enough and the station-level documentation should help.
Splitting it into more methods will make it worse.

armiol:
@alexander-yevsyukov @dmdashenkov I will still need to merge with the today-merged PR, but please feel free to review again. Not much is expected to change upon the back-merge from `master`.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,211 @@
...
+     *         there were any errors in that
+     */
+    @SuppressWarnings({"MethodWithMultipleLoops", "OverlyComplexMethod", "OverlyNestedMethod"})
+    @Override
+    public final Result process(Conveyor conveyor) {
+        Map<DispatchingId, InboxMessage> dispatchToCatchUp = new HashMap<>();
+
+        for (InboxMessage message : conveyor) {
+            for (CatchUp job : jobs) {
+                CatchUpStatus jobStatus = job.getStatus();
```
alexander-yevsyukov: We probably don't have to do this if the job does not matches the messag.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,211 @@
...
+
+                if (job.matches(message)) {
```
alexander-yevsyukov: How about having this:

```java
if (!job.matches(message)) {
     continue;
}

File: server/src/main/java/io/spine/server/delivery/CatchUpStation.java (hunk 4 above)
Comment (alexander-yevsyukov) [reply server/src/main/java/io/spine/server/delivery/CatchUpStation.java]: I think we'll need to revisit the `delivery` package soon. Presumably, when we get down to working on new Buses. The `process()` method in this class really needs refactoring, and we need a Method Object there. This Method Object would be really linked with the `CatchUpStation` class, while this Comparator is a more separate thing. It can be easier understood being at the top level of the package. It won't hurt. We have many `CatchUp`-classes there. One more there, won't hurt if the price for this would be the `process(` method which is easier to understand.

File: server/src/main/java/io/spine/server/delivery/LiveDeliveryStation.java (hunk 6 above)
Comment (alexander-yevsyukov) [reply server/src/main/java/io/spine/server/delivery/LiveDeliveryStation.java]: We have a case similar to `CatchUpStation` here. Only the author of the code can understand what's going on in this method and why. The Javadoc for this method does not help much because it says that it dispatches messages. It's understood already by the name of the method `process()` and its parameter `Conveyor`. But how does it do it? Why do we have logic which is _that_ complex? 

I think we need another Method Object here. Please consider.

File: server/src/main/java/io/spine/server/delivery/CatchUpProcess.java [hunk 27]
```diff
@@ -0,0 +1,661 @@
...
+    List<ShardProcessingRequested> on(CatchUpCompleted ignored) {
+        int shardCount = builder().getTotalShards();
+        List<Integer> affectedShards = builder().getAffectedShardList();
+        List<ShardProcessingRequested> events = toShardEvents(shardCount, affectedShards);
+        return events;
+    }
+
+    /*
+     * Utilities and re-usable actions
+     ************************/
```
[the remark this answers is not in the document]
armiol: @alexander-yevsyukov from what I can see from your reaction, the sections are a bad idea at all. 

Going to remove these here, and let's discuss [our code style wiki page](https://github.com/SpineEventEngine/core-java/wiki/Java-Code-Style#comment-separators-for-sections-of-the-class).

server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
```diff
@@ -0,0 +1,661 @@
...
+
+    /**
+     * The type URL of the process state.
+     *
+     * <p>Used for the message routing.
+     */
+    private static final TypeUrl TYPE = TypeUrl.from(CatchUp.getDescriptor());
+
+    /**
+     * The duration of the "turbulence" period, counting back from the current time.
```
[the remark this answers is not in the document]
armiol: @alexander-yevsyukov there is a decent description of the turbulence period in the class-level documentation. I have added it to address your previous comment a few days ago.

Either you have missed it (which seems more likely), or you really want this concept to be introduced at full speed. I think that's going to be an overdesign, but let's try.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,211 @@
...
+                    DispatchingId dispatchingId = new DispatchingId(message);
+                    if (jobStatus == IN_PROGRESS) {
```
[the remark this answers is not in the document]
armiol: @alexander-yevsyukov This piece is explained by the class-level doc. The same applies the `LiveDeliveryStation`.

This whole class _is_ itself a method object, as is the `LiveDeliveryStation`. 

I will try to extract pieces one more time. It's may make it look better designed. But the extra call chain will make it harder to follow.
And, to avoid creating extra objects, I will probably have to modify the variables by references. 

Let's see if it makes the code easier for non-authors :)

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -338,40 +336,148 @@ public final boolean canDispatch(EventEnvelope event) {
...
+     * Preconditions.checkArgument(boolean, String, Object)}.
+     *
+     * <p>Allows to postpone the resource-heavy stringification of the argument until that is
+     * really required.
+     *
+     * @param value
+     *         the value to be used as an argument
+     * @return the object to use as an argument in the message-formatting API
+     */
+    private static Object lazyArg(Object value) {
```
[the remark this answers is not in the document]
alexander-yevsyukov: Or, perhaps, we may want to have our custom precondition utilities that would accept `LazyArg` (or something similar). To see what I mean, please see how lazy logging is done in Flogger in `com.google.common.flogger.LogContext.logImpl()` method. 

This isn't a small, but useful change. We may have some other checks that accept time consuming operations in other places in our code. We can improve that altogether,  and _then_ change this piece too. If you concur, please file an issue referring to this method.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -338,40 +336,148 @@ public final boolean canDispatch(EventEnvelope event) {
...
+     *         the routing schema
+     * @return the set of the entity identifiers, which actually received the dispatched event
+     * @see CatchUpEndpoint
+     */
+    private Set<I> sendToCatchingUp(Event event, @Nullable Set<I> restrictToIds) {
+        EventEnvelope envelope = EventEnvelope.of(event);
+        Set<I> catchUpTargets;
+        if (envelope.message() instanceof CatchUpSignal) {
+            catchUpTargets = restrictToIds == null
+                             ? ImmutableSet.copyOf(index())
```
alexander-yevsyukov: This operation is potentially big and long (assuming we go for the `index()` to the storage). Can we run it once per repository under the catch up operation?

server/src/main/java/io/spine/server/delivery/InboxMessageComparator.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public int compare(InboxMessage m1, InboxMessage m2) {
+        int timeComparison = Timestamps.compare(m1.getWhenReceived(), m2.getWhenReceived());
+        if (timeComparison != 0) {
+            return timeComparison;
+        }
+        int versionComparison = Integer.compare(m1.getVersion(), m2.getVersion());
+        if(versionComparison != 0) {
```
dmdashenkov: Please add a space.
I'd also consider using a `Comparator.comparing(..)` chain instead of implementing the interface directly.

server/src/main/java/io/spine/server/delivery/ShardMaintenanceProcess.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public void handle(EventEnvelope event) {
+        ShardEvent message = (ShardEvent) event.message();
+        inbox.send(event)
+             .toSubscriber(message.getId());
+    }
+
+    private class EventEndpoint implements MessageEndpoint<ShardIndex, EventEnvelope> {
```
dmdashenkov: Please make it `final` and document.

server/src/main/java/io/spine/server/delivery/CatchUpStation.java
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * <h1>Overview</h1>
+ *
+ * <p>Matches the messages on the passed {@link Conveyor} to the known {@link CatchUp} jobs basing
+ * on the target entity type and the identifier of the entity to which the message is sent.
+ *
+ * <p>Depending on the status of the job, the matched messages are processed accordingly. See
+ * more on that below.
+ *
+ * <b>1. Catch-up {@code IN_PROGRESS}.</b>
```
alexander-yevsyukov: Please have these headers coming in separate paragraphs. Now they come as a part of a proceeding paragraph.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
armiol: The idea here is to have the most recent `index()`, so we shouldn't cache it.

In the real use cases, it is going to be used only if a catch-up for all the repository entities is requested.  If this repository is a new one and requires catching up, the index is going to be empty. And if the repository entities are all messed up and require the complete re-catch-up,
it will and should take long. So the expectations are managed.

server/src/main/java/io/spine/server/delivery/CatchUpMixin.java
```diff
@@ -0,0 +1,66 @@
...
+     *     <li>the identifier of the message target must be included into the list of the
+     *     identifiers specified in the job OR the job matches all the targets of the entity type.
+     * </ol>
+     *
+     * @param message
+     *         the message to match to the job
+     * @return {@code true} if the message matches the job, {@code false} otherwise
+     */
+    @VisibleForTesting
+    default boolean matches(InboxMessage message) {
```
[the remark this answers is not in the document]
armiol: I have removed this annotation and marked the whole class as `Internal`.

It originally was `public`, as this interface was in the different package.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -49,32 +52,41 @@
...
 public abstract class ProjectionRepository<I, P extends Projection<I, S, ?>, S extends EntityState>
         extends EventDispatchingRepository<I, P, S> {
 
     private @MonotonicNonNull Inbox<I> inbox;
 
+    private @MonotonicNonNull CatchUpProcess<I> catchUpProcess;
+
     private @MonotonicNonNull RepositoryCache<I, P> cache;
 
     /**
```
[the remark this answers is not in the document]
armiol: Appended the docs.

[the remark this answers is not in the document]
armiol: The piece on the state update routing makes sense, as discussed vocally.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -101,7 +113,17 @@ public void registerWith(BoundedContext context) throws IllegalStateException {
         super.registerWith(context);
         ensureDispatchesEvents();
         initCache(context.isMultitenant());
-        initInbox();
+        Delivery delivery = ServerEnvironment.instance()
+                                             .delivery();
+        initInbox(delivery);
+        initCatchUp(context, delivery);
```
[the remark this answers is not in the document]
armiol: Appended the documentation.

server/src/main/java/io/spine/server/projection/ProjectionRepository.java
```diff
@@ -338,40 +336,148 @@ public final boolean canDispatch(EventEnvelope event) {
...
         inbox().send(EventEnvelope.of(event))
                .toSubscriber(id);
     }
 
-    @Internal
-    public void writeLastHandledEventTime(Timestamp timestamp) {
-        checkNotNull(timestamp);
-        projectionStorage().writeLastHandledEventTime(timestamp);
+    /**
+     * Repeats the dispatching of the events from the event log to the requested entities
```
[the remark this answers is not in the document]
armiol: Done.

PR #1222, opened 2020-01-15
Better routing error message in Black Box

dmdashenkov:
## Error messages

When posting an event to `BlackBoxBoundedContext`, it is easy to forget to specify a producer ID. In this case, the Black Box used to set a `UserId` as a producer. If the event routing is not configured for some entities, the routing was likely to cause an error when trying to determine event targets. Such errors can be pretty confusing if the context contains entities, which actually use `UserId` as their ID.

Now, the default producer ID is set to an instance of `BlackBoxId`. This way, the developer who sees the error message will rightfully attribute the error to the `BlackBoxBoundedContext` misconfiguration.

## Limits and orderings

When generating validation code, our compiler warns us that the use of `(goes).with` on `limit` field is incorrect, since `limit` is a number field and cannot be checked for presence. This was always the case, however, the runtime validation was unable to detect such mistakes. 

The option is now removed from the definition. The `QueryValidatior` is extended to check for the correct use or `limit`.


dmdashenkov:
@armiol, PTAL.

testutil-server/src/main/proto/spine/testing/server/black_box.proto
```diff
@@ -0,0 +1,22 @@
...
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.testing.server";
+option java_outer_classname = "BlackBoxProto";
+option java_multiple_files = true;
+
+import "spine/core/bounded_context.proto";
+
+// ID of a `BlackBoxBoundedContext`.
+//
+// Identifies events posted directly into a `BlackBoxBoundedContext`.
```
armiol: `Identifies the producer of the events...`

PR #1224, opened 2020-01-17
Less console output on errors

dmdashenkov:
As much as we value a verbose log message, it might sometimes get too long, and thus, hard to process.
In this PR we reduce a particularly long error message.

The information is still accessible to the API user, if they subscribe to a system event.


dmdashenkov:
@armiol, PTAL.

PR #1227, opened 2020-01-22
Concurrency issues in reflective calls

dmdashenkov:
In this PR, we migrate from using `java.lang.reflect.Method`s to using `java.lang.invoke.MethodHandle`s for obtaining values of entity columns.
With `MethodHandle`s, we don't have to change the accessibility parameter upon each invocation (but only once — upon building the `MethodHandle` instance).

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -415,7 +415,7 @@
  The dependencies distributed under several licenses, are used according their commercial-use-friendly license.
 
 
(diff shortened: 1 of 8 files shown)


dmdashenkov:
@armiol, PTAL.
The build requires [base/507](https://github.com/SpineEventEngine/base/pull/507).

Issue #1212, opened 2019-12-05
Strongly-typed columns in query/subscription client API


armiol:
At the moment when composing a query or subscribing to some updates an end-user refers to the fields by their string representation. E.g.

```java
client.asGuest()
      .subscribe(ProjectDetails.class)
      .where(eq("project_name", value))
      .post()
```
A potential typo in `"project_name"` would be invisible at the first glance.

Also, there is a difference in language. As the Java definitions use the camel-cased representation when Proto fields are snake-cased.

What would be nice to do is to introduce the generated enumerations, such as `FieldName` and `ColumnName` for events and entities respectively. Each listing the available Protobuf fields and making the code snippet above to transform into something like this:

```java
client.asGuest()
      .subscribe(ProjectDetails.class)
      .where(eq(ProjectDetails.Columns.projectName, value))
      .post()
```
Or, for the event subscriptions:
```java
client.asGuest()
      .subscribe(PizzaDelivered.class)
      .where(eq(PizzaDelivered.Fields.overdue, value))
      .post()
```

This effectively will make `eq` methods not accepting `String`s but strong-typed objects, depending on the type of the target object.

armiol:
@alexander-yevsyukov FYI, that's what we have discussed vocally.

PR #1229, opened 2020-01-28
Strongly-typed columns in query/subscription client API

dmitrykuzmin:
This PR addresses #1212.

It takes advantage of the strongly-typed fields and columns generated by [base#508](https://github.com/SpineEventEngine/base/pull/508) and embeds them in the high-level client API.

### New Client API

The old `where(...)` methods that are accepting plain `Filter` and `CompositeFilter` instances are now deprecated in favor of new overloads which accept the typed filters as arguments. The typed filters form a hierarchy which mirrors the hierarchy of query and entity/event subscription requests, and are created with help of generated columns and fields.

An example of the new API:
```java

public QueryRequest<S> where(QueryFilter... filter) {
    builder().where(extractFilters(filter));
    return this;
}

public QueryRequest<S> where(CompositeQueryFilter... filter) {
    builder().where(extractFilters(filter));
    return this;
}
```

The `QueryFilter` and `CompositeQueryFilter` are the typed wrappers for the `Filter` and `CompositeFilter`  proto messages respectively. The wrapper creation is regulated by the generated field and column types, not allowing to specify `Field`-based filter to a query request and vice-versa.

### Event subscriptions
The event subscription request also accepts strongly-typed filters, as follows:
```java
public EventSubscriptionRequest<E> where(EventFilter... filter) {
    builder().where(extractFilters(filter));
    return self();
}

public EventSubscriptionRequest<E> where(CompositeEventFilter... filter) {
    builder().where(extractFilters(filter));
    return self();
}
```
The `EventFilter` can be created in multiple ways to enable filtering events by the event context:
```java
public static EventFilter eq(EventMessageField field, Object value) {
    checkNotNull(field);
    checkNotNull(value);
    return new EventFilter(field, value, EQUAL);
}

public static EventFilter eq(EventContextField field, Object value) {
    checkNotNull(field);
    checkNotNull(value);
    return new EventFilter(field, value, EQUAL);
}

// Same goes for all other operators.
// ...
```
The `EventContextField` instance may be obtained from the `EventContext.Field` class, in the same way as all other strongly-typed fields and columns. So an example of event subscription request could be:
```java
 client().asGuest()
         .subscribeToEvent(UserLoggedIn.class)
         .where(all(eq(UserLoggedIn.Field.user(), userId),
                    eq(EventContext.Field.pastMessage()
                                         .actorContext()
                                         .actor(), guestUser)))
                    // ...
                    .post();
```
On the Protobuf level filtering by event context still works by appending the `"context."` prefix to a field path in the filter.

### Additional Model Compiler configurations
To enable the behavior described above, the additional Model Compiler configurations are specified:
```gradle
modelCompiler {
    fields {
        generateFor "spine.core.Event", markAs("io.spine.base.EntityStateField")
        generateFor "spine.core.EventContext", markAs("io.spine.core.EventContextField")
    }
}
```
Basically, they:
1. Enable the strongly-typed fields generation for `spine.core.Event`, as it's a subscribable entity type (see `EEntity`).
2. Enable the strongly-typed field generation for `spine.core.EventContext` with a custom marker superclass, to enable the typed event filter creation described above.

### Other
The `orderBy(...)` method of `QueryRequest` is also switched to the new columns usage (the old version is deprecated).

Please note, that the described changes affect only the high-level Client API. The lower-level API like `TargetBuilder` is still capable of accepting column/field names as plain strings.

### Version
The Spine version advances to `1.4.5`.


dmitrykuzmin:
@armiol, @dmdashenkov PTAL. The build should pass once the corresponding changes from `base` and `config` are merged.

armiol:
@dmitrykuzmin please also mention the subscription to events — i.e. how it is affected (or not) by this changeset.

alexander-yevsyukov:
@dmitrykuzmin, please compose a short document which explains using this feature. You can create a Google Doc, and I'll convert it into Markdown placing in the site hierarchy.

dmitrykuzmin:
@armiol, @dmdashenkov PTAL again.

dmitrykuzmin:
@alexander-yevsyukov The short document with feature usage explanation can be found [here](https://docs.google.com/document/d/1RPSNJOqGpVKqnrxvpgJLQuFHOA2jeFLhmTcJmRcFm6M/edit?usp=sharing).

PR #1235, opened 2020-02-18
Root command ID of an Event

dmdashenkov:
The current methods `Event.rootCommandId()` and `Event.rootMessage()` do not account for:
 - events imported via `ImportBus`;
 - events which do not have `Origin` in there `EventContext`, i.e. the old format events.

The method `Event.rootCommandId()` is being deprecated and the method `Event.rootMessage()` is being adjusted to support old-format events.

### DIFF
--- a/core/src/main/java/io/spine/core/EventContextMixin.java
+++ b/core/src/main/java/io/spine/core/EventContextMixin.java
@@ -29,6 +29,7 @@ import io.spine.time.InstantConverter;
 import io.spine.validate.FieldAwareMessage;
 
 import java.time.Instant;
+import java.util.Optional;
 
 import static io.spine.util.Exceptions.newIllegalStateException;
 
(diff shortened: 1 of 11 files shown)


dmdashenkov:
@armiol, @yuri-sergiichuk, PTAL.

core/src/main/java/io/spine/core/EventContextMixin.java
```diff
@@ -89,6 +90,39 @@ default ActorContext actorContext() {
...
+     * @see Event#rootMessage()
+     */
+    @SuppressWarnings("deprecation") // For backward compatibility.
+    default Optional<MessageId> rootMessage() {
+        EventContext.OriginCase origin = getOriginCase();
+        switch (origin) {
+            case PAST_MESSAGE:
+                return Optional.of(getPastMessage().root());
+            case IMPORT_CONTEXT:
+                return Optional.empty();
```
yuri-sergiichuk: I docs you mention that the root can be an `event` if the event was imported into the system, but here, we return `empty` for imported origins.

core/src/main/java/io/spine/core/EventContextMixin.java
```diff
@@ -89,6 +90,39 @@ default ActorContext actorContext() {
...
+            case EVENT_CONTEXT:
+            case COMMAND_CONTEXT:
+            case ORIGIN_NOT_SET:
+            default:
+                @SuppressWarnings("DuplicateStringLiteralInspection") // Coincidence.
+                MessageId id = MessageId
```
yuri-sergiichuk: In the docs, you state that you'll return `empty` for root messages themselves. I assume that only the root message should have the `origin` as `NOT_SET`, right?

core/src/main/java/io/spine/core/EventMixin.java
```diff
@@ -59,10 +59,11 @@ default Timestamp time() {
...
     default MessageId rootMessage() {
-        EventContext.OriginCase origin = context().getOriginCase();
-        return origin == PAST_MESSAGE
-               ? context().getPastMessage().root()
-               : messageId();
+        return isImported()
+               ? messageId()
+               : context()
+                       .rootMessage()
+                       .orElseThrow(IllegalStateException::new);
```
yuri-sergiichuk: Are you sure it's a good idea to throw an exception here? I'd better return `Optional`. It's pretty uncommon to throw an exception from Protobuf message.

dmdashenkov:
@yuri-sergiichuk, PTAL again.

PR #1237, opened 2020-02-20
Improvements to Java client

dmdashenkov:
This PR brings a few minor improvements to various client-facing APIs.

1. When importing an event, the `Timestamp` from the `ActorContext` is now used for the event, instead of always using the current time. This is helpful if the event has happened in the past.

2. The entity states are now unpacked from the `QueryResult` more efficiently.


dmdashenkov:
@armiol, PTAL.

PR #1239, opened 2020-02-27
Updates to `InboxMessage` behavior

armiol:
This changeset modifies the way we store `InboxMessage`s.

* The timestamp of `InboxMessage` now has microsecond precision. On storage level, it allows using the native data types, which typically support microseconds at most.
* The identifier of `InboxMessage` now includes the `ShardIndex` which previously was a part of `InboxMessage` body. Such an approach allows storing `InboxMessage`s as a hierarchy of objects. In turn, in Google Datastore that opens a possibility to query these messages in a transaction.

Please note that existing binaries of `InboxMessage` become **incompatible** with this code version.
So make sure to deliver or purge the existing messages of this type prior to migration.

As a minor change, the `CatchUpTest` has been made open for configuration in the dependant libraries, such as `spine-gcloud-java`.

The library version is updated to `1.4.9`.

### DIFF
--- a/server/src/test/java/io/spine/server/delivery/CatchUpTest.java
+++ b/server/src/test/java/io/spine/server/delivery/CatchUpTest.java
@@ -77,7 +77,7 @@ import static org.junit.jupiter.api.Assertions.fail;
  * millisecond precision. It is required to test the catch-up in the scenarios close to the legacy
  * applications, as at that time there were no emulation of the nanosecond time resolution.
  *
- * <p>As the downstream libraries, such as Spine Google Cloud library would want to run the same
+ * <p>As the downstream libraries, such as Spine Google Cloud library, would want to run the same
  * tests under their specific conditions, the big and slow catch-up tests are made {@code public}.
  * In this way such tests may be overridden and disabled, if needed.
  */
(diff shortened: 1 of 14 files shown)


server/src/test/java/io/spine/server/delivery/CatchUpTest.java
```diff
@@ -66,8 +67,23 @@
 import static java.util.stream.Collectors.toList;
 import static org.junit.jupiter.api.Assertions.fail;
 
+/**
+ * Tests on {@linkplain io.spine.server.projection.ProjectionRepository#catchUp(Timestamp, Set)
```
dmdashenkov:
```suggestion
 * Tests for the {@linkplain io.spine.server.projection.ProjectionRepository#catchUp(Timestamp, Set)
```

server/src/test/java/io/spine/server/delivery/CatchUpTest.java
```diff
@@ -66,8 +67,23 @@
...
+ * projection catch-up} functionality.
+ *
+ * <p>The test routines are designed to check both small and big use-cases, including
+ * the full catch-up. To deal with the different wall-clock providers, some of the tests
+ * configure the {@linkplain Time#currentTime() time provider} to return the values with
+ * millisecond precision. It is required to test the catch-up in the scenarios close to the legacy
+ * applications, as at that time there were no emulation of the nanosecond time resolution.
+ *
+ * <p>As the downstream libraries, such as Spine Google Cloud library would want to run the same
```
dmdashenkov:
```suggestion
 * <p>As the downstream libraries, such as Spine Google Cloud library, would want to run the same
```

PR #1241, opened 2020-02-28
Migration API for record-based entities

dmitrykuzmin:
This PR extends the `RecordBasedRepository` with API to perform the migration of stored entities to account for the model changes.

### Migration API
The repository is extended with `applyMigration(ID, Migration)` and `applyMigration(Set<ID>, Migration)` methods.

The `Migration` is a transformation operation which accepts entities one-by-one and modifies their state, lifecycle flags or can even delete the entity record. This class (or rather its descendants, `ProjectionMigration` and `ProcessManagerMigration`) can be inherited to create custom user-defined migration operations.

The applied entity modification can be configured by implementing the method `apply(entityState)`:
```java
class ProjectMigration extends ProjectionMigration<...> {

    // A mapper from the "old" to a "new" state.
    public Project apply(Project state) {
        return state.toBuilder()
                    .setName("new_name")
                    .build();
    }
}
```
The new state returned from `apply` will then be injected into the entity under migration.

In `apply(...)`, it's also possible to call additional modification methods like `markArchived()` or `removeFromStorage()`. See `Migration` for details.

Each migration is preceded by the entity lookup by ID and is finalized by either saving the modified entity back to the repository or deleting the entity record, if it is configured by the `Migration`.

All changes are applied to entity under an active `Transaction`, with the last step of migration being a transaction commit. This means all entity lifecycle events are triggered normally, enabling subscription updates and other system-events-dependent features. 

The posted events will have an instance of newly introduced `MigrationApplied` system event as a producing message.

### Columns update
The standard migrations provided by Spine include the `UpdateProjectionColumns` and `UpdatePmColumns` migrations for projections and process managers respectively. When applied, such migration triggers the recalculation of interface-based columns of an entity, propagating the new values to the storage fields as well as to the serialized entity state.

The column update can be applied as follows:
```java
// Projection.
projectionRepository.applyMigration(ImmutableSet.of(id1, id2), new UpdateProjectionColumns<>());

// Process manager.
pmRepository.applyMigration(ImmutableSet.of(id1, id2), new UpdatePmColumns<>());
```

### Other migrations
The framework also provides `Mark(Projection/Pm)Archived`, `Mark(Projection/Pm)Deleted` and `Remove(Projection/Pm)FromStorage` migrations which can mark entity archived, mark entity deleted and physically remove an entity record from the storage respectively. The migrations are available for both projections and process managers under the respective `migration` packages.

### Other
This PR also fixes an issue with `Nullable` interface-based columns propagation and improves the description of `Columns.valuesIn(...)` operation.

### Version
Spine version advances to `1.4.11`.


dmitrykuzmin:

server/src/main/java/io/spine/server/entity/Migration.java
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * Returns {@code true} if the entity under migration is
+     * {@linkplain Entity#isDeleted() deleted}.
+     */
+    protected final boolean isDeleted() {
+        return currentOperation().isDeleted();
+    }
+
+    final boolean physicallyRemoveRecord() {
```
yuri-sergiichuk: Should we add some docs to this method and the one below?

server/src/main/java/io/spine/server/entity/Migration.java
```diff
@@ -0,0 +1,277 @@
...
+        return currentOperation().physicallyRemoveRecord();
+    }
+
+    final void finishCurrentOperation() {
```
yuri-sergiichuk: Looks like this method should be called by the caller code to finalize the operation. Am I right? If so, it worth documenting the contract.

server/src/main/java/io/spine/server/entity/RecordBasedRepository.java
```diff
@@ -121,6 +128,90 @@ public void store(E entity) {
...
+        checkNotNull(migration);
+        checkEntityIsTransactional();
+
+        E entity = findOrThrow(id);
+        migration.applyTo((T) entity, (RecordBasedRepository<I, T, S>) this);
+        if (migration.physicallyRemoveRecord()) {
+            delete(id);
+            migration.finishCurrentOperation();
+        } else {
+            store(entity);
```
yuri-sergiichuk: It looks a bit odd to me that we `finish` the operation only if the migration removes a record. Why is it so?

server/src/test/java/io/spine/server/projection/given/Migrations.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package io.spine.server.projection.given;
+
+import io.spine.base.EntityState;
+import io.spine.protobuf.ValidatingBuilder;
+import io.spine.server.projection.Projection;
+import io.spine.server.projection.ProjectionMigration;
+
+public final class Migrations {
```
yuri-sergiichuk: I'd say these migrations could be part of the Framework standard migrations. Worth considering moving them to the sources?

dmitrykuzmin:
@yuri-sergiichuk PTAL again.

server/src/main/java/io/spine/server/entity/EntityLifecycle.java
```diff
@@ -568,14 +585,19 @@ private void postHandlerFailed(MessageId handledSignal, Error error) {
         postEvent(systemEvent);
     }
 
-    protected void postEvent(EventMessage event, Origin explicitOrigin) {
+    @CanIgnoreReturnValue
+    protected Optional<Event> postEvent(EventMessage event, Origin explicitOrigin) {
```
yuri-sergiichuk: What about docs for these methods?

server/src/main/java/io/spine/server/entity/EntityLifecycle.java
```diff
@@ -316,6 +318,21 @@ public final void onStateChanged(EntityRecordChange change,
         postIfRestored(change, messageIds);
     }
 
+    /**
+     * Posts the {@link MigrationApplied} event.
+     *
+     * @return the event or an empty {@code Optional} if the posting was blocked by the
+     *         {@link #eventFilter}
+     */
+    public final Optional<Event> onMigrationApplied() {
```
yuri-sergiichuk: That's a bit strange that we're able to filter out the migration event, but on the other hand, in the `Migration` we throw an `IllegalStateException` in case of such filtering.

server/src/main/java/io/spine/server/procman/migration/UpdateColumns.java
```diff
@@ -0,0 +1,57 @@
...
+import io.spine.server.entity.Migration;
+import io.spine.server.entity.Transaction;
+import io.spine.server.procman.ProcessManager;
+import io.spine.server.procman.ProcessManagerMigration;
+
+/**
+ * A migration operation that does the update of interface-based columns of a
+ * {@link ProcessManager}.
+ *
+ * <p>When applied to an entity, this operation will trigger the recalculation of entity storage
```
armiol: Please append a paragraph explaining when such migration may be handy. I.e. describe a use-case scenario.

server/src/main/java/io/spine/server/projection/migration/UpdateColumns.java
```diff
@@ -0,0 +1,56 @@
...
+import io.spine.protobuf.ValidatingBuilder;
+import io.spine.server.entity.Migration;
+import io.spine.server.entity.Transaction;
+import io.spine.server.projection.Projection;
+import io.spine.server.projection.ProjectionMigration;
+
+/**
+ * A migration operation that does the update of interface-based columns of a {@link Projection}.
+ *
+ * <p>When applied to an entity, this operation will trigger the recalculation of entity storage
```
armiol: Same as above.

server/src/main/java/io/spine/server/projection/migration/UpdateColumns.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @implNote The operation relies on the fact that column values are automatically calculated and
+ *         propagated to the entity state on a transaction {@linkplain Transaction#commit() commit}
+ *         and thus does not change the entity state itself in {@link #apply(EntityState)}.
+ *
+ * @see io.spine.server.entity.storage.InterfaceBasedColumn
+ * @see io.spine.server.entity.RecordBasedRepository#applyMigration(Object, Migration)
+ */
+@Experimental
+public final class UpdateColumns<I,
```
armiol: Can we have a single `UpdateColumns` class, which would be created using a factory method? 

E.g. `UpdateColumns.of(MyProjectView.class)` or `UpdateColumns.of(MyProcess.class)`. 

We then would detect if it is a projection or a process manager and create a corresponding instance.

It would be cool to avoid two _public_ classes with the same name.

dmitrykuzmin:
@armiol PTAL again.

PR #1242, opened 2020-03-02
Custom log site and new black box API

### LINKED ISSUE #1226: Add an ability to set the ActorId in `BlackBoxBoundedContext`
Currently, the `BlackBoxBoundedContext` sets the `ActorId` on its own and we can't modify it.

But for testing purposes, we may need to emulate commands from different actors.

The way to achieve this is to add an ability to set the desired `ActorId` to the `BlackBoxBoundedContext`. From my perspective the API can be the following:
``` Java
context().receivesCommandFrom(UserId actor, CommandMessage c);
```

dmdashenkov:
## Log site

Flogger library provides API for changing the header of log messages via `LogSite` instances. In this PR we add new logging API for all the entities. In a handler method, a user may obtain a preconfigured logger as follows:
```java
class TaskAggregate extends Aggregate<...> implements LoggingEntity {

    @React
    AssigneeRemoved on(UserDeleted event, EventContext ctx) {
        _fine().log("Reacting to an event");
        // ...
    }
}
```

The resulting log would be similar to:
```
org.example.TaskAggregate on(UserDeleted, EventContext)
FINE: Reacting to an event
```

### Implementation details

A new interface `HandlerLifecycle` is introduced. This is a set of callbacks regarding a message handler method invocation.

All the entities which extend `AbstractEntity` now implement this interface. The framework users may choose to implement it in other message handlers, namely `AbstractEventReactor`, `AbstractEventSubscriber`, `AbstractCommander`, and `AbstractCommandHandler` subclasses.

The framework users may employ the interface for their own needs, such as tracking the performance of the handler methods, adding extra logging, etc. 

## Custom actor ID for Black Box Bounded Context

With this PR, we add API for configuring the ID and time zone used to produce actor requests in the Black Box Bounded Context.

In some domains, entities and repositories may rely on the details from the `ActorContext`, such as the actor ID. One common use case for this is routing commands by the actor ID.

In order to simulate real-life actors in black-box tests, we add the API for changing the actor how posts commands via `BlackBoxBoundedContext`:
```java
BlackBoxBoundedContext
        .from(builder)
        .withActor(actorId)
        .receivesCommand(..)
        ...
```

Also, users may change the time zone ID and offset via the `in` method:
```java
BlackBoxBoundedContext
        .from(builder)
        .in(ZoneIds.of("Europe/Kyiv"), ZoneOffsets.ofHours(2))
        .receivesCommand(..)
        ...
```
Or do both at the same time via the `withActorIn` method:
```java
BlackBoxBoundedContext
        .from(builder)
        .withActorIn(actorId, ZoneIds.of("Europe/Kyiv"), ZoneOffsets.ofHours(2))
        .receivesCommand(..)
        ...
```

Resolves #1226.


dmdashenkov:
@armiol, PTAL.

alexander-yevsyukov:
@dmdashenkov, would log trancing improvements work for underscored methods like `_debug()`?

dmdashenkov:
@alexander-yevsyukov, no, they won't. These logging configurations only apply to the logs created via the `log().at(...)` construct. The method is deliberately named `log()` and not `logger()`  to avoid a naming clash if a user wants to use `Logging` in their entity.

server/src/test/java/io/spine/server/log/given/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+@CheckReturnValue
+@ParametersAreNonnullByDefault
+package io.spine.server.log.given;
```
armiol: Please document this package.

server/src/test/proto/spine/test/library/commands.proto
```diff
@@ -0,0 +1,22 @@
...
+
+import "spine/test/library/card.proto";
+
+message BorrowBooks {
+    LibraryCardId card = 1 [(required) = true, (validate) = true];
+    repeated Isbn book_id = 2 [(required) = true, (validate) = true];
+}
+
+message ReturnBook {
+    LibraryCardId card = 1 [(required) = true, (validate) = true];
```
armiol: Do we intend to validate `LibraryCardId.reader`? If so, then `LibraryCardId` should also say the `reader` is required.

dmdashenkov:
@armiol, PTAL again.

PR #1245, opened 2020-03-05
Reduce `CommandBus`

dmdashenkov:
Previously, the `CommandBus` checked each command by attempting to route it with the default routing strategy — by the first field.

If a client wants to send a command and it not aware of the receiver, this approach can go wrong. For example, if the first field of a command message happens to be a `repeated`, the validation fails.

At the same time, command's target can override routing (e.g. a `Repository`) or not require one at all (e.g. `AbstractCommandHandler` descendants). It is not possible, however, to pass this information to the `CommandBus` level.

In this PR we remove this validation step.

After this change, if a command routing fails, we will still get an error and a system event. However, not the users will have more flexibility when defining command messages.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -415,7 +415,7 @@
  The dependencies distributed under several licenses, are used according their commercial-use-friendly license.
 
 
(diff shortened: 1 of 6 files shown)


dmdashenkov:
@armiol, PTAL.

Issue #1244, opened 2020-03-05
Add support for strongly-typed columns in `Filters` utility


yuri-sergiichuk:
The new [strongly-typed columns](https://github.com/SpineEventEngine/base/pull/508) bring clarity to the filtering in the query service and they are supported well in the new client API, but unfortunately, the standard query API is not supported.

The current usage forces us to have smth like:
```java
String orderIdColumn = ImageItem.Column
        .orderId()
        .name()
        .toString();
Query query = factory(userId)
        .query()
        .select(ImageItem.class)
        .where(Filters.eq(orderIdColumn, orderId))
        .build();
```

It'd be great if we have overloads in the `Filters` utility class that accept `EntityColumn`, e.g.:
```java
    public static Filter eq(EntityColumn column, Object value) {
        checkNotNull(column);
        checkNotNull(value);
        return createFilter(column.name().asPath(), value, EQUAL);
    }
```

yuri-sergiichuk:
Closed with #1246.

PR #1246, opened 2020-03-05
Extend `Filters` API with strongly-typed columns support

dmitrykuzmin:
This PR addresses #1244.

The `Filters` utility is extended with methods that allow to create `Filter` instances from strongly-typed columns and fields.

This should be convenient for users that do not rely on high-level `Client` API and use plain `QueryService`, `SubscriptionService`, etc. instead.

Spine version advances to `1.4.13`.


dmitrykuzmin:

armiol:
@dmitrykuzmin please make sure you don't submit the same version as @dmdashenkov in #1245.

PR #1249, opened 2020-03-16
Improvements for testing utilities (part 1)

alexander-yevsyukov:
Changes introduced by this PR:
  * Outdated `Verify`-based API is removed. This includes both `client` and `server` packages.
     Remaining bits related to `VerifyAcknowledgements` is going to be removed in the following PR.
  * `ContextAwareTest` abstract test suite base is introduced.

Improvements in `BlackBoxBoundedContext`:
  * It now can write events to `EventStore` so the user can arrange a test environment this way.
  * It no longer exposes `eventBus()` and `commandBus()`.
  * `subscribe(Topic)` now returns `SubscriptionFixture` which can assert both entity states and event messages via `IterableOfProtosSubject`.

Changes to `BlackBoxBoundedContext` and removal of old Verify-based API are breaking. It's more or less safe to do since the number of users are small, and efforts for arranging full deprecating cycle will not pay off.


testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
```diff
@@ -246,7 +228,7 @@ public static MultitenantBlackBoxContext multiTenant(EventEnricher enricher) {
     /**
      * Creates a multitenant instance with the specified name and enricher.
      */
-    public static MultitenantBlackBoxContext multiTenant(String name, EventEnricher enricher) {
+    private static MultitenantBlackBoxContext multiTenant(String name, EventEnricher enricher) {
```
armiol: I assume that the rest of similar factory methods will become `private` or even get removed in part 2.

testutil-server/src/main/java/io/spine/testing/server/blackbox/ContextAwareTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private @Nullable BlackBoxBoundedContext<?> context;
+
+    /**
+     * Creates a new builder for the Bounded Context under the test.
+     */
+    protected abstract BoundedContextBuilder contextBuilder();
+
+    @BeforeEach
+    void createContext() {
```
armiol: I would make this `protected` and mark as `@OverridingMethodsMustInvokeSuper`. The reason is that in the test cases there may be more actions before the Bounded Context is created, such as setting up the mocks.

testutil-server/src/main/java/io/spine/testing/server/blackbox/ContextAwareTest.java
```diff
@@ -0,0 +1,82 @@
...
+        BoundedContextBuilder contextBuilder = contextBuilder();
+        checkNotNull(contextBuilder,
+                     "`contextBuilder()` must return a non-null `BoundedContextBuilder`.");
+        context = BlackBoxBoundedContext.from(contextBuilder);
+    }
+
+    @AfterEach
+    void closeContext() {
```
armiol: Similar to the comment above, I would make this `protected` and `@OverridingMethodsMustInvokeSuper`.

testutil-server/src/main/java/io/spine/testing/server/blackbox/SubscriptionFixture.java
```diff
@@ -0,0 +1,177 @@
...
+import static io.spine.client.SubscriptionUpdate.UpdateCase.EVENT_UPDATES;
+
+/**
+ * Allows to assert updates received on a subscription.
+ */
+public final class SubscriptionFixture {
+
+    private final BoundedContext context;
+    private final Topic topic;
+    private final List<SubscriptionUpdate> updates = new ArrayList<>();
```
armiol: Let's make is a synchronized `List`. Not a `Concurrent` one.

The reason is that the subscription updates in tests may arrive from different threads. The `Concurrent` collections are "eventually consistent", so we want to enable the plain old synchronization.

testutil-server/src/main/java/io/spine/testing/server/blackbox/SubscriptionFixture.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        @Override
+        public void onCompleted() {
+            // Do nothing.
+        }
+    }
+
+    /**
+     * Activates the first received {@link Subscription}.
```
armiol: Judging on the implementation, it activates every subscription it receives. Not sure which one is the correct one — the implementation or the documentation.

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxBoundedContext.java
alexander-yevsyukov: Yes, I intend to remove them at all in part 2. I'll make the deprecated in this part.

alexander-yevsyukov: As discussed vocally, I'll revert it and address them all in p2.

PR #1252, opened 2020-03-18
Improvements for testing utils (Part 3)

alexander-yevsyukov:
This PR introduces the following improvement:

* **Breaking change:** Most of the `@Internal` methods of `BoundedContext` moved to its internal class `InternalAccess` instance of which is available via the `internalAccess()` method. 
The method is available only to the server-side framework code.
* **Breaking change:**`BlackBoxBoundedContext` was renamed to `BlackBoxContext`.
* Creation of `BlackBoxContext` from `BoundedContextBuilder` was simplified.
* **Breaking change:** Simplified combinations of `UserId` and `ZoneId` parameters of `BlackBoxContext`.
* `TestActorRequestFactory` no longer requires `ZoneOffset`. It's calculated from `ZoneId`.
* Removed previously deprecated methods of `BlackBoxCountext`.
* Removed deprecated factory methods of `TestEventFactory`.


testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
```diff
@@ -618,26 +367,48 @@ public T receivesEventsProducedBy(Object producerId, EventMessage first, EventMe
      *         a list of domain event to be dispatched to the Bounded Context
      * @return current instance
      */
-    private T receivesEvents(Collection<EventMessage> domainEvents) {
+    private BlackBoxContext receivesEvents(Collection<EventMessage> domainEvents) {
```
dmitrykuzmin: Maybe it makes sense to include all these methods accepting a `Collection` of events and commands into public API of `BlackBoxContext`?

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -82,19 +81,36 @@
 @SuppressWarnings({"OverlyCoupledClass", "ClassWithTooManyMethods"})
 public abstract class BoundedContext implements Closeable, Logging {
 
+    /** Basic features of the context. */
     private final ContextSpec spec;
+
+    /** The Command Bus of this context. */
     private final CommandBus commandBus;
```
dmdashenkov: Do we really need this kind of documentation? We give no new information to the reader.

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -539,4 +458,117 @@ public String toString() {
...
+    public InternalAccess internalAccess() {
+        Security.allowOnlyFrameworkServer();
+        return this.internalAccess;
+    }
+
+    /**
+     * Provides access to features of {@link BoundedContext} used internally by the framework.
+     */
+    @Internal
+    public class InternalAccess {
```
dmdashenkov: Please make it `final` and move to the upper level. If needed, I'd rather have some `private` API turn into package-private than have `BoundedContext` more bloated.

Also, this looks like a `RepositoryRegistry`.

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -539,4 +458,117 @@ public String toString() {
...
+
+    /**
+     * Provides access to features of the context used internally by the framework.
+     *
+     * @throws SecurityException
+     *         if called from outside the framework
+     */
+    @Internal
+    public InternalAccess internalAccess() {
+        Security.allowOnlyFrameworkServer();
```
dmdashenkov: We must allow read-only access to the users. I'd even consider making it a non-internal API. It is often used when trying to bind together repositories, e.g. for routing purposes.

server/src/main/java/io/spine/server/BoundedContext.java
```diff
@@ -539,4 +458,117 @@ public String toString() {
...
+        /** Obtains an {@link IntegrationBroker} of this {@code BoundedContext}. */
+        public IntegrationBroker broker() {
+            return self().broker;
+        }
+
+        /**
+         * Obtains a tenant index of this Bounded Context.
+         *
+         * <p>If the context is single-tenant returns
+         * {@linkplain TenantIndex#singleTenant() null-object} implementation.
```
dmdashenkov: I think, it's not a null-object per se. Just like an `ImmutableList.of(singleElement)` is not a null-object.

server/src/test/java/io/spine/server/projection/ProjectionTest.java
```diff
@@ -178,14 +178,15 @@ void handleStateUpdates() {
     @Test
     @DisplayName("throw ISE if no handler is present for event")
     void throwIfNoHandlerPresent() {
-        ProjectionRepository<String, ?, ?> repository =
+        @SuppressWarnings("unchecked") ProjectionRepository<String, ?, ?> repository =
```
dmdashenkov: Please fix the alignment.

testutil-client/src/main/java/io/spine/testing/client/TestActorRequestFactory.java
```diff
@@ -50,24 +50,30 @@ public TestActorRequestFactory(UserId actor, ZoneId zoneId) {
...
                       .newBuilder()
                       .setTenantId(tenantId)
                       .setActor(actor)
-                      .setZoneOffset(zoneOffset)
+                      .setZoneOffset(toOffset(zoneId))
                       .setZoneId(zoneId)
         );
     }
 
+    @VisibleForTesting
```
dmdashenkov: Do we need to specify `@VisibleForTesting`?  The whole `TestActorRequestFactory` class is `@VisibleForTesting`.

version.gradle
```diff
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def spineVersion = '1.5.2'
+final def spineVersion = '1.5.3'
```
dmdashenkov: There are many breaking changes in this version. I suggest we make a new minor update — `1.6.0`. As always, we can first publish this as `1.5.3` and check that it works and then advance to `1.6.0`.

server/src/main/java/io/spine/server/BoundedContext.java
alexander-yevsyukov: I added this as a sort of “warm-up” for the reader. The class is long. It simply prepares the mental context of a programmer who sees this code for the first time.

server/src/main/java/io/spine/server/BoundedContext.java
alexander-yevsyukov: Users can always arrange dependencies between the repositories before they are added to `BoundedContextBuilder`. Allowing this as a general practice encourages Big Balls of Muds.

server/src/main/java/io/spine/server/BoundedContext.java
alexander-yevsyukov: Single tenant means there is no tenancy in this context. And it does not contains a reference to the _real_ tenant ID value. In this case, it's just a constant, which says “there are no tenants here”. That's why it's a null object.

version.gradle
alexander-yevsyukov: We will surely have 1.6.0 once other breaking changes are ready. @armiol is working on the other set of such, which is going to be way bigger. Changes in this PR are not that big, comparing with new storages API coming soon.

We surely do not abide to semantic versioning. But assuming that there are still not so many users of the framework, we'd better update quicker and easier rather than follow an artificial ballet with proper deprecation cycles.

testutil-client/src/main/java/io/spine/testing/client/TestActorRequestFactory.java
alexander-yevsyukov: It's a remnant of previous step. I'll surely remove the annotation, but later we'll need to [move the method into the Time library](https://github.com/SpineEventEngine/time/issues/87).

server/src/main/java/io/spine/server/BoundedContext.java
alexander-yevsyukov: Yes, surely `final`. But it's not just a repository registry. Methods there are not only related to the repositories. I wanted to minimize the number of internally exposed things in `BoundedContext` itself. Maybe, we'll be able to refine it later. For now it's just easing on the `@Internal` API visible for the user of the framework.

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
alexander-yevsyukov: Let's wait till there is a real need. We can easily add it later should we see such a need. For now we have a method which access only one `Event` or message. The API of `BlackBoxContext` allows chaining, so if we need a few of `Event`s, they can be passed sequentially. Also, a test suite can always perform a streaming API call (on the collection of events or commands) referencing corresponding method of `BlackBoxContext`.

PR #1256, opened 2020-03-26
Update GitHub pages publishing

dmdashenkov:
In this PR we migrate from the local script for publishing Javadoc into GitHub Pages to the new script in `config`.

This changes the structure of the published doc. After we update the site, the `gh-pages` branch will have to be manually cleaned up from the previous version of the doc.


dmdashenkov:
@armiol, PTAL.

PR #1257, opened 2020-03-30
System events performance improvement

### LINKED ISSUE #779: `CommandBus` performance
After introducing system commands, posting a command into a `CommandBus` has become a significantly time-consuming operation.

This matter requires some further investigation.
As an idea, we may try to post system commands asynchronously.

dmdashenkov:
In this PR we allow system events to be posted in parallel. This improves performance in certain conditions.

The parallel publishing is enabled by default for production use and disabled for tests, since some tests rely on eventual consistency. It can be disabled for separate (or all) Bounded Contexts:
```java
BoundedContextBuilder builder = BoundedContext.singleTenant("example");
// ...
builder.systemFeatures().disableParallelPosting();
```

### Performance
For commands, we ran a test with 2000 commands posted from a single thread.
When system events are emitted synchronously, the test runs in 200-210 ms. After enabling parallelism, the runtime is shortened to 190-200 ms. 

For commands and imported events, the time of a synchronous run is close to 1030 ms, and for a paralleled run — 990 ms.

If the `CommandLog` system feature is enabled, the test runs in 970 ms in synch mode in paralleled mode — 970 ms.

Fixes #779.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -415,7 +415,7 @@
  The dependencies distributed under several licenses, are used according their commercial-use-friendly license.
 
 
-This report was generated on **Mon Mar 30 20:17:37 EEST 2020** using [Gradle-License-Report plugin](https://github.com/jk1/Gradle-License-Report) by Evgeny Naumenko, licensed under [Apache 2.0 License](https://github.com/jk1/Gradle-License-Report/blob/master/LICENSE).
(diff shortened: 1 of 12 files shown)


dmdashenkov:
@armiol, PTAL.

PR #1258, opened 2020-04-01
Propagate all events to subscriptions

dmdashenkov:
Previously, the events produced by non-entity objects (e.g. `AbstractEventReactor` descendants) were not propagated to the subscriptions of the `SubscriptionService`.

Now, we use a `Listener` in the `EventBus` for propagating events and updates to subscribers.

Also, in this PR we relax the check for types when subscribing on events. Previously, if an event was not declared explicitly via a tuple, there was no way of subscribing to it via `SubscriptionService`, since it did not reach the event type registry in `Stand`. Now, the registry serves as a mere heuristic. If the event type is not there, we still create a subscription. However, in such a case, there would be a subscription in each Bounded Context instead of just a single subscription is a single Bounded Context.


dmdashenkov:
@armiol, PTAL.

server/src/main/java/io/spine/server/stand/EventTap.java
```diff
@@ -0,0 +1,50 @@
...
+
+package io.spine.server.stand;
+
+import io.spine.server.bus.Listener;
+import io.spine.server.type.EventEnvelope;
+import io.spine.type.TypeUrl;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+final class EventTap implements Listener<EventEnvelope> {
```
armiol, line 29: Please document the class.

PR #1261, opened 2020-04-06
Improve event and state assertions

alexander-yevsyukov:
This PR adds assertion methods for events and entity states to `BlackBoxContext`. 

Entity and entity state assertion methods now accept an ID as the first parameter. Methods that used to have the ID as the second parameter were deprecated in favour of ones with the new order.

### DIFF
--- a/testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
+++ b/testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
@@ -166,6 +166,7 @@ public abstract class BlackBoxContext implements Logging {
      * Sets the given {@link UserId} as the actor ID for the requests produced by this context.
      */
     public final BlackBoxContext withActor(UserId user) {
+        checkNotNull(user);
         this.actor = this.actor.withId(user);
         return this;
     }
@@ -174,6 +175,7 @@ public abstract class BlackBoxContext implements Logging {
      * Sets the given time zone parameters for the actor requests produced by this context.
      */
     public final BlackBoxContext in(ZoneId zoneId) {
+        checkNotNull(zoneId);
         this.actor = this.actor.in(zoneId);
         return this;
     }
@@ -212,6 +214,7 @@ public abstract class BlackBoxContext implements Logging {
      */
     @CanIgnoreReturnValue
     public final BlackBoxContext receivesCommand(CommandMessage domainCommand) {
+        checkNotNull(domainCommand);
         return receivesCommands(singletonList(domainCommand));
     }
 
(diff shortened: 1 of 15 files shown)


testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
```diff
@@ -492,58 +505,130 @@ public void close() {
...
+                         .comparingExpectedFieldsOnly();
+        return stateAssertion;
+    }
+
+    /**
+     * Asserts that there is an entity with the passed ID and the passed state.
+     *
+     * <p>The method compares only fields in the passed state.
+     */
+    public final <I, S extends EntityState> void assertState(I id, S entityState) {
```
dmdashenkov, line 589:
```suggestion
    public final void assertStateEquals(Object id, EntityState entityState) {
```

I

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
```diff
@@ -492,58 +505,130 @@ public void close() {
...
         return repository;
     }
 
+    /**
+     * Asserts that the is an entity with the passed ID and the passed type of state.
+     *
+     * @return assertion that compares only expected fields
+     */
+    public final <I, S extends EntityState> ProtoFluentAssertion
+    assertState(I id, Class<S> stateClass) {
```
dmdashenkov, line 574: I suggest we get rid of unnecessary type parameters. In this case, since neither `<I>` nor `<S>` is used in the return type, they both are unnecessary.

Also, there used to be a style check for that. Is it off now?

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
```diff
@@ -492,58 +505,130 @@ public void close() {
...
     EntitySubject assertEntity(Class<E> entityClass, I id) {
-        @Nullable Entity<I, ?> found = findEntity(entityClass, id);
+        return assertEntity(id, entityClass);
+    }
+
+    /**
+     * Obtains a Subject for an entity of the passed class with the given ID.
+     */
+    public final <I, E extends Entity<I, ? extends EntityState>>
+    EntitySubject assertEntity(I id, Class<E> entityClass) {
```
dmdashenkov, line 521: Why do we change the order of parameters? One could argue that specifying an entity type first and then its ID is more natural from the user's perspective.

alexander-yevsyukov: I mentioned this in one of the commit comments. Here's why. 1) When we assert against an entity state, it's a big block of code. Having the ID parameter below this block makes it poorly visible.

2) A reference to an object usually comes first in a procedure that deals with an object. We use IDs as reference to entities.

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxContext.java
alexander-yevsyukov: Indeed, they are not used. But they are consistent with the rest of assertion methods. Here having `<I>` as a generic parameter we hint on the fact that an object which can be passed to it can be of one of the identifier types, not any `Object`.

PR #1268, opened 2020-04-27
`(set_once)` in entities

dmdashenkov:
Previously, the `(set_once)` constraint in entities was ignored. Since this PR, we will check entity state changes and produce `ConstraintViolated` system event if an entity state change violates a `(set_once)` constraint.


dmdashenkov:
@armiol, PTAL.

PR #1270, opened 2020-04-30
Handle error acks

alexander-yevsyukov:
This extends the client API with an ability to handle an error returned when posting a command or another message for asynchronous processing:
 * `ServerErrorHandler` interface was introduced. It accepts `io.spine.base.Error` obtained from the server-side in an `Ack`.
 * The `Client` got methods for customising error handlers
    * `onStreamingError()` — for handing errors obtained from gRPC.
    *  `onServerError()` — for handling `Error` obtained as a part of an `Ack`.
   These handlers will be passed as default values for `ClientRequest`s produced by the client.
 * The `ClientRequest` class also got `onStreamingError()` and `onServerError()` methods for fine tuning of handlers that may be necessary for a specific request.

### Other changes
  * `LoggingHandler` interface is now more general and does not contain an association with a type. `LoggingHandlerWithType` is introduced for handlers that do need such an association.
 * Removed redundant methods for obtaining a tenant from a signal.
 * The `Signal` interface got the `type()` method which obtains the type of the enclosed message.
 * Added `WithTime` interface for encapsulating obtaining `Timestamp`, `Instant` and time range checking.
 *  Added `WithActor` interface for encapsulating obtaining `ActorContext` and its properties.
 * Added `ContextWithActor` interface which is common base for `CommandContext` and `EventContext`.
 * `CommandContext` got a mix-in so that it's easier to obtain a user or a tenant of a command.
 * `Response` got a mix-in for easier checking of its type.
 * The internal interface `MessageRejection` was renamed to `MessageError` to avoid confusion with real rejections. It was also moved under the `server` package.
 * The internal `MessageInvalid` interface was moved to the `server` package.

### Deprecations
 * `@ByField` was deprecated because `@Subscribe.filter()` which uses it was deprecated earlier.
 * `Command.time()` was deprecated in favour of `WithTime.timestamp()`.


client/src/main/java/io/spine/client/PostingErrorHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.function.BiConsumer;
+
+/**
+ * A handler for an error occurred when posting a message for asynchronous processing
+ * at the server side.
+ *
+ * <p>The handler accepts a message which caused an error, and an {@link io.spine.base.Error Error}
+ * obtained from the server.
+ */
+public interface PostingErrorHandler extends BiConsumer<Message, Error> {
```
dmdashenkov, line 35: `Posting` tells that this error handler posts something. I suggest `MessageErrorHandler` or `ServerErrorHandler`.

core/src/main/java/io/spine/core/CommandContextMixin.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.errorprone.annotations.Immutable;
+import io.spine.annotation.GeneratedMixin;
+
+/**
+ * Mix-in interface extending {@link CommandContext}.
+ */
+@GeneratedMixin
+@Immutable
+public interface CommandContextMixin extends CommandContextOrBuilder, ContextWithActor {
```
dmdashenkov, line 31: Please make the interface package-private.

core/src/main/java/io/spine/core/ContextWithActor.java
```diff
@@ -0,0 +1,31 @@
...
+package io.spine.core;
+
+import com.google.errorprone.annotations.Immutable;
+import io.spine.base.MessageContext;
+
+/**
+ * A context of a message initiated by a user.
+ */
+@Immutable
+public interface ContextWithActor extends MessageContext, WithActor {
```
dmdashenkov, line 30: Maybe `SignalContext`?

core/src/main/java/io/spine/core/ResponseMixin.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package io.spine.core;
+
+import io.spine.base.Error;
+
+/**
+ * Mixin interface for the {@link Response} objects.
+ */
+public interface ResponseMixin extends ResponseOrBuilder {
```
dmdashenkov, line 28: I think, this should be package-private and a `@GeneratedMixin`. And probably `@Immutable` as well.

client/src/main/java/io/spine/client/PostingErrorHandler.java
alexander-yevsyukov: That's a difficult one. The problem with the `Server` prefix is that in combination with `Handler` it sounds as if we can handle the error that happened at the server-side.

dmdashenkov: Don't we?

client/src/main/java/io/spine/client/CommandRequest.java
```diff
@@ -106,9 +113,35 @@
...
+    public CommandRequest onConsumingError(ConsumerErrorHandler<EventMessage> handler) {
+        checkNotNull(handler);
+        eventConsumers.onConsumingError(handler);
+        return this;
+    }
+
+    /**
+     * Assigns a handler for an error occurred on the server-side (such as validation error)
+     * in response to posting a command.
+     */    @OverridingMethodsMustInvokeSuper
```
armiol, line 140: Please fix the layout.

client/src/main/java/io/spine/client/CommandRequest.java
```diff
@@ -132,28 +173,85 @@ public CommandRequest onStreamingError(ErrorHandler handler) {
...
+            Status status = ack.getStatus();
+            switch (status.getStatusCase()) {
+                case OK:
+                    return Optional.ofNullable(subscriptions)
+                                   .orElse(ImmutableSet.of());
+                case ERROR:
+                    cancelVoidSubscriptions();
+                    reportErrorWhenPosting(status);
+                    return ImmutableSet.of();
+                case REJECTION:
```
armiol, line 214: Do we ever return a response in this status?

client/src/main/java/io/spine/client/DeliveringMultiEventObserver.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.google.common.collect.ImmutableMap;
+import io.grpc.stub.StreamObserver;
+import io.spine.base.EventMessage;
+import io.spine.core.Event;
+import io.spine.logging.Logging;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Delivers an event to all consumers subscribed to events of this type.
```
armiol, line 31: There does not seem to be `this` type. I would rephrase with `of a particular type.`

server/src/test/java/io/spine/client/CommandRequestTest.java
```diff
@@ -98,4 +110,105 @@ void rejections() {
...
+
+            assertThat(returnedError)
+                    .isNotNull();
+            assertThat(postedMessage)
+                    .isInstanceOf(Command.class);
+            Command expected = Command
+                    .newBuilder()
+                    .setMessage(AnyPacker.pack(commandMessage))
+                    .build();
+            ProtoTruth.assertThat(postedMessage)
```
armiol, line 209: I think, you can use the `static import` here as well. It should get resolved among similar methods.

client/src/main/java/io/spine/client/CommandRequest.java
alexander-yevsyukov: I don't think so. It does not make sense in the current async. implementation, and I could not find such cases. Still, we still have this option, so I've covered this as a branch.

We can possibly deprecate it in the `Status` type, and address it in the code gradually.

client/src/main/java/io/spine/client/PostingErrorHandler.java
alexander-yevsyukov: Agreed. Going to rename it.

client/src/main/java/io/spine/client/CommandRequest.java
alexander-yevsyukov: After some further investigation, I've found out that in some cases a `Status` can hold a `rejection`. It cannot happen when the `Status` is a part of acknowledgement for a command. So, the branch is valid, and we cannot deprecate the field.

PR #1272, opened 2020-05-04
gRPC client tests

dmdashenkov:
When subscribing to an Event type, we sometimes cannot tell which Bounded Context it belongs to. To walk around, we are putting subscriptions into each available Context with a hope that one of them does fire the event.

While doing so, we used to call gRPC's stream observer callbacks too many times, breaking the gRPC unary call contract. In this PR we fix the callback invocations and add gRPC tests which make sure such requests do not cause errors.


dmdashenkov:
@alexander-yevsyukov, PTAL.

Issue #615, opened 2017-11-03
TransportFactorySwitch for test and production mode


dmytro-kashcheiev:
Right now spine provides `StorageFactorySwitch` to provide different implementations of storage factory for test and production modes. 

It would be great if we could have the same switch for `TransportFactory` used within `IntegrationBus`. As far as in most cases test and production implementation of the `IntegrationBus` transport will be different.

PR #1274, opened 2020-05-19
Per-environment configuration for `ServerEnvironment` parts

serhii-lekariev:
#### `EnvSetting`

This PR introduces a new class for values that may differ between environments - `EnvSetting<P>`, which allows users to configure their value so that it's different between environments like so:

```java
EnvSetting<StorageFactory> storage = new EnvSetting<>();
StorageFactory factory = InMemoryStorageFactory.newInstance();

// `Tests` is a type exposed by `base`. 
storage.use(factory, Tests.type());

Optional<StorageFactory> productionStorage = storage.value(Production.type());
assertThat(productionStorage).isEmpty();

Optional<StorageFactory> storageForTests = storage.value(Tests.type());
assertThat(storageForTests).isPresent();
assertThat(storageForTests.get()).isSameInstanceAs(factory);
```

#### `ServerEnvironment` changes

Previously, `ServerEnvironment` allowed to use different `StorageFactory` instances for different environments.

This functionality has been expanded to also allow configuration of:

- `Delivery`: defaults to `Delivery.local()` if no value has been configured;
- `TracerFactory`: does not default to any value, it's accessor still returns an `Optional<TracerFactory>`;
- `TransportFactory`: defaults to an `InMemoryTransportFactory` for tests, still requires explicit configuration for all other environments.

#### Deprecation notes

The following `ServerEnvironment` methods have been deprecated:

- `configureStorageForTests`: prefer `use(storage, Tests.class)`;
- `configureStorage`: prefer `use(storage, Production.class); // or any other env type`;
- `configureDelivery`: prefer `use(delivery, Production.class); // or any other env type`.

Fixes https://github.com/SpineEventEngine/core-java/issues/615.


server/src/test/java/io/spine/server/storage/system/SystemAwareStorageFactoryTest.java
```diff
@@ -65,25 +65,27 @@ void nulls() {
...
+                   .reset();
     }
 
     @Test
     @DisplayName("wrap test storage")
     void wrapTestStorage() {
         ServerEnvironment serverEnv = ServerEnvironment.instance();
         StorageFactory testStorage = InMemoryStorageFactory.newInstance();
-        serverEnv.configureStorageForTests(testStorage);
+        serverEnv.configureStorage().tests(testStorage);
```
armiol, line 88: I would rather have something like 
```java
 serverEnv.useStorageFactory(factory).inTests();
```
That's a lot closer to the way we speak.

server/src/main/java/io/spine/server/EnvironmentDependantValue.java
```diff
@@ -0,0 +1,209 @@
...
+ * </pre>
+ * <p>If no wrapping function is specified, the value is never wrapped.
+ *
+ * <p>{@code EnvironmentDependantValues} do not determine the environment themselves: it's up to the
+ * caller to ask for the appropriate one.
+ *
+ * @param <P>
+ *         the type of value
+ */
+public class EnvironmentDependantValue<P> {
```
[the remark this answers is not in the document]
serhii-lekariev: `ConfigurableValue` is too generic, but sort of true?

It's a wrapper around two values, both of which may be absent, and those values have "production" and "testing" semantics, which I would call two different environments.

server/src/main/java/io/spine/server/EnvironmentDependantValue.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private @Nullable P productionValue;
+    private final UnaryOperator<P> productionWrappingFunction;
```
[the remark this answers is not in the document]
armiol: Why is this resolved?

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -232,75 +228,92 @@ public void configureTracing(TracerFactory tracerFactory) {
...
     private static Environment environment() {
         return Environment.instance();
     }
 
     /**
-     * Assigns {@code TransportFactory} for the production mode of the application.
+     * Returns the configuration object to change the transport factory.
+     *
+     * <p>Callers may configure the testing factory via
+     * {@code configureTransport().tests(testFactory)} and the production factory via
```
armiol, line 271: Is it still valid?

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -232,75 +228,92 @@ public void configureTracing(TracerFactory tracerFactory) {
...
+        return result;
     }
 
     /**
      * This is test-only method required for cleaning of the server environment instance in tests.
      */
     @VisibleForTesting
     public void reset() {
-        this.transportFactory = null;
+        this.transportFactories.nullifyBoth();
```
armiol, line 314: Let's call this method `reset()` or `clear()`.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -313,11 +326,17 @@ public void close() throws Exception {
         if (tracerFactory != null) {
             tracerFactory.close();
         }
-        if (transportFactory != null) {
-            transportFactory.close();
+        if (transportFactories.production()
```
armiol, line 329: To elimitate the repeating `if (... .production().isPresent()) { ...}` please introduce the corresponding API to the `EnvironmentDependantValue` itself. 

I can see at least four cases only in this class.

server/src/main/java/io/spine/server/EnvironmentDependantValue.java
```diff
@@ -0,0 +1,241 @@
...
+ * </pre>
+ * <p>If no wrapping function is specified, the value is never wrapped.
+ *
+ * <p>{@code EnvironmentDependantValues} do not determine the environment themselves: it's up to the
+ * caller to ask for the appropriate one.
+ *
+ * @param <P>
+ *         the type of value
+ */
+public final class EnvironmentDependantValue<P> {
```
armiol, line 57: `EnvSetting`?

server/src/main/java/io/spine/server/EnvironmentDependantValue.java
```diff
@@ -0,0 +1,241 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A mutable value that may differ for the production and testing environments.
+ *
+ * <p>Allows configuring a wrapping function to adjust mutations as such:
+ * <pre>
+ * {@code
+ * EnvironmentDependantValue<StorageFactory> storages = EnvironmentDependantValue
```
armiol, line 36: I don't understand why you call it `storages`. It's `storageFactory`;

serhii-lekariev:
@armiol, PTAL.

serhii-lekariev:
@armiol, PTAL again.

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -0,0 +1,127 @@
...
+     *         value to assign to one of environments
+     */
+    public void use(P value, EnvironmentType type) {
+        checkNotNull(value);
+        checkNotNull(type);
+        this.settingValue.put(type, value);
+    }
+
+    /**
+     * Represents an operation over a value that returns no result and may finish with an error.
```
armiol, line 117: If this is an operation, let's call it so.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -72,30 +76,24 @@
      * initialized with the {@linkplain Delivery#local() local} delivery by default.
      */
     @SuppressWarnings("FieldAccessedSynchronizedAndUnsynchronized")
-      // No synchronization in `reset()` as it is test-only.
+    // No synchronization in `reset()` as it is test-only.
     private @MonotonicNonNull Delivery delivery;
```
armiol, line 80: Let's try to make the `Delivery` specific to the environment as well.

To do so, `configureDelivery` should be deprecated and migrated to `use(...)`.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -198,109 +196,127 @@ public void configureDeployment(Supplier<DeploymentType> supplier) {
...
+    public ServerEnvironment use(TracerFactory tracerFactory, EnvironmentType envType) {
+        checkNotNull(envType);
+        environment().register(envType);
+        this.tracerFactory.use(tracerFactory, envType);
+        return this;
     }
 
     /**
-     * Assigns {@code TracerFactory} to this server environment.
+     * Configures the specified transport factor to be used under the specified environment type.
```
armiol, line 225: `factory`.

I would say
```
Configures the specified transport factory for the selected type of environment.`
```

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -198,109 +196,127 @@ public void configureDeployment(Supplier<DeploymentType> supplier) {
...
+     *
+     * @return this instance of {@code ServerEnvironment} this instance
```
armiol, line 227: Let's kill the duplication.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -198,109 +196,127 @@ public void configureDeployment(Supplier<DeploymentType> supplier) {
...
+        EnvironmentType currentType = environment().type();
+        return this.tracerFactory.value(currentType);
     }
 
     /**
-     * Obtains production {@code StorageFactory} previously associated with the environment.
+     * Obtains the storage factory for the current environment.
+     *
+     * <p>For tests, if the value was not set, defaults to a new {@code InMemoryStorageFactory}.
+     * <p>For other environments, if the value was not set, throws a {@code IllegalStateException}.
```
armiol, line 249: Please append a line above.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -198,109 +196,127 @@ public void configureDeployment(Supplier<DeploymentType> supplier) {
...
+     * Assigns the specified {@code TransportFactory} for the specified application environment.
      *
-     * @see #configureStorageForTests(StorageFactory)
+     * @return this instance of {@code ServerEnvironment}
      */
-    public void configureStorage(StorageFactory productionStorageFactory) {
-        checkNotNull(productionStorageFactory);
-        this.productionStorageFactory = wrap(productionStorageFactory);
+    @CanIgnoreReturnValue
+    public ServerEnvironment use(StorageFactory storage, EnvironmentType envType) {
```
armiol, line 204: Let's refer to this and to the rest of similar parameters as to `factory`.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -198,109 +196,127 @@ public void configureDeployment(Supplier<DeploymentType> supplier) {
...
+        checkNotNull(envType);
+        environment().register(envType);
+        this.storageFactory.use(wrap(storage), envType);
+        return this;
     }
```
armiol, line 209: You could extract the common part from each of these methods (and check for `null` the remaining parameter):
```java
    private static <V> void use(V value, EnvSetting<V> setting, EnvironmentType type) {
        checkNotNull(value);
        checkNotNull(type);
        checkNotNull(setting);
        environment().register(type);
        setting.use(value, type);
    }
```

The public methods will then look like this:
```java
    @CanIgnoreReturnValue
    public ServerEnvironment use(StorageFactory factory, EnvironmentType envType) {
        SystemAwareStorageFactory wrapped = wrap(factory);
        use(wrapped, this.storageFactory, envType);
        return this;
    }
```

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -0,0 +1,127 @@
...
+            operation.accept(settingValue);
+        }
+    }
+
+    /**
+     * If the value for the specified environment is set, just returns it.
+     *
+     * <p>If it is not set, runs the specified supplier, configures and returns the supplied value.
+     */
+    P assignOrDefault(Supplier<P> defaultValue, EnvironmentType type) {
```
armiol, line 87: This method either reads the value or writes and reads it. The contract is confusing to me and so is the name, as it tells it assigns, but in many cases, it just reads.

This code needs more love. I will try to come up with something better than this.

testutil-server/src/main/java/io/spine/testing/server/tenant/TenantAwareTest.java
```diff
@@ -34,14 +35,13 @@
 /**
  * Abstract base for test suites that test tenant-aware functionality.
  *
- * <p>This class must be used only from {@linkplain Environment#isTests() test execution context}.
+ * <p>This class must be used only from {@linkplain io.spine.base.Tests# test execution environment}.
```
armiol, line 38: The hash symbol seems redundant to me.

server/src/main/java/io/spine/server/EnvSetting.java
serhii-lekariev: I disagree. 

`java.util.function.Consumer` Javadoc also says that it's an `operation`. 

Moreover, I think the name immediately tells a developer what this is: it's a regular consumer, except it can `throw`.

server/src/main/java/io/spine/server/ServerEnvironment.java
serhii-lekariev: Just something I've found: for any non-`Tests` environment, if we want to configure a delivery for it, we also need to configure a storage factory for it first. Should we specify it in Javadocs?

serhii-lekariev:
@armiol, PTAL again.

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -0,0 +1,129 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A mutable value that may differ between {@linkplain EnvironmentType environment types}.
+ *
+ * <p>For example:
+ * <pre>
+ *
+ * {@literal EnvSetting <StorageFactory>} storageFactory = new EnvSetting<>();
```
armiol, line 39: Let's kill that extra space character.

server/src/main/java/io/spine/server/EnvSetting.java
armiol: This name is way too generic for this particular case. That's why I proposed to name it more specifically.

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -0,0 +1,129 @@
...
+            operation.accept(settingValue);
+        }
+    }
+
+    /**
+     * If the value for the specified environment is set, just returns it.
+     *
+     * <p>If it is not set, runs the specified supplier, configures and returns the supplied value.
+     */
+    V assignOrDefault(Supplier<V> defaultValue, EnvironmentType type) {
```
armiol, line 89: Now:

```java
private final EnvSetting<MyFactory> setting = new EnvSetting<>();

//...


return setting.assignOrDefault(    // this method says it writes, but in fact it reads — and sometimes writes.
        () -> create();
        type;
);
```


Proposed change:


```java
private final EnvSetting<MyFactory> setting = 
                new EnvSetting<>(type, () -> create());  // default `fallback` value for `type`.

private final EnvSetting<AnotherFactory> anotherSetting = new EnvSetting<>();  // no fallbacks.



//...

MyFactory value = setting.value();  // mandatory setting, e.g. `StorageFactory`; throws ISE if `empty()`.

Optional<AnotherFactory> optionalValue = anotherSetting.optionalValue();  // optional setting, e.g. `Tracer`.                 
```

serhii-lekariev:
@armiol, PTAL again. Build fails because this branch depends on the `base` `1.5.18`. I can either decrease the version here, or wait until `base` is merged and published.

server/src/main/java/io/spine/server/EnvSetting.java
armiol: The extra symbol is still there.

armiol: Try to generate Javadocs themselves. Let's see if `new EnvSetting<>();` portion renders as expected.

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * <p>This implementation does <b>not</b> perform any synchronization, thus, if different threads
+ * {@linkplain #use(Object, Class) configure} and {@linkplain #value(Class) read the value},
+ * no effort is made to ensure any consistency.
+ *
+ * @param <V>
+ *         the type of value
+ */
+@Internal
+public final class EnvSetting<V> {
```
armiol, line 94: I don't understand why this class is public.

server/src/main/java/io/spine/server/delivery/Delivery.java
```diff
@@ -334,6 +338,10 @@ public static Delivery local() {
      * <p>The returned instance of {@code Delivery} is configured to use
      * {@linkplain UniformAcrossAllShards#singleShard() the single shard}.
      *
+     * <p>To construct a {@code Delivery} instance, a {@code StorageFactory} is needed.
+     * If it was not configured in the {@code ServerEnvironment}, uses a new {@code
```
armiol, line 342: Who uses?

server/src/main/java/io/spine/server/delivery/DeliveryBuilder.java
```diff
@@ -221,6 +225,9 @@ public DeliveryBuilder setInboxStorage(InboxStorage inboxStorage) {
      *
      * <p>If none set, the storage is initialized by the {@code StorageFactory} specific for
      * this {@code ServerEnvironment}.
+     *
+     * <p>If no {@code StorageFactory} is present in the {@code ServerEnvironment}, a new
+     * {@code InMemoryStorage} is used.
```
armiol, line 230: There is no such thing as `InMemoryStorage`. Please review your changeset for this copy-paste mistake.

server/src/test/java/io/spine/server/ServerEnvironmentTest.java
```diff
@@ -147,41 +152,56 @@ void receivesStandalone() {
...
+        void deprecatedMethod() {
+            environment.setTo(Production.class);
+
+            InMemoryStorageFactory storageFactory = InMemoryStorageFactory.newInstance();
+            serverEnvironment.configureStorage(storageFactory);
+            assertThat(serverEnvironment.storageFactory())
+                    .isInstanceOf(SystemAwareStorageFactory.class);
+            assertThat(((SystemAwareStorageFactory) serverEnvironment.storageFactory()).delegate())
+                    .isSameInstanceAs(storageFactory);
+
```
armiol, line 203: Please kill this line.

server/src/test/java/io/spine/server/ServerEnvironmentTest.java
```diff
@@ -198,14 +218,65 @@ void resetEnvironment() {
...
+    @Nested
+    @DisplayName("configure `StorageFactory` for a custom environment")
+    class LocalStorageFactoryConfig {
+
+        @BeforeEach
+        void reset() {
+            serverEnvironment.reset();
+            Environment.instance()
+                       .register(new Local());
```
armiol, line 248: Not sure if this piece is migrated to the latest `base` API yet, but you may easily use a ctor instead of a new instance here.

server/src/test/java/io/spine/server/ServerEnvironmentTest.java
```diff
@@ -198,14 +218,65 @@ void resetEnvironment() {
...
+            Environment.instance()
+                       .setTo(new Local());
```
armiol, line 250: See how weird this is? You have several pieces of the same inconvenient code in this test. How much easier that would be if you could treat `setTo` as registration as well?

version.gradle.kts
```diff
@@ -25,6 +25,6 @@
  * as we want to manage the versions in a single source.
  */
 
-val spineBaseVersion: String by extra("1.5.12")
+val spineBaseVersion: String by extra("1.5.18")
 val spineTimeVersion: String by extra("1.5.12")
```
armiol, line 29: Let's migrate to the latest `time` once it's published.

Also, let's set the `core-java` version to be the same as `base`.

serhii-lekariev:
@armiol, PTAL again.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -47,6 +47,29 @@
 
 /**
  * The server conditions and configuration under which the application operates.
+ *
+ * <h1>Configuration</h1>
+ * <p>Some parts of the {@code ServerEnvironment} can be customized based on the {@code
```
armiol, line 52: Please have an empty line above.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -47,6 +47,29 @@
...
+ * EnvironmentType}. To do so, one of the overloads of the {@code use} method can be called.
+ * Two environment types exist out of the box: {@link Tests} and {@link Production}. For example:
+ * <pre>
+ *     ServerEnvironment.use(productionStorageFactory, Production.class)
+ *                      .use(testingStorageFactory, Tests.class)
+ *                      .use(memoizingTracerFactory, Production.class)
+ * </pre>
+ * A custom environment type may also be used:
```
armiol, line 60: Add `<p>` and one more empty line above.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -47,6 +47,29 @@
...
+ *     final class StagingEnvironment extends EnvironmentType {
+ *         ...
+ *     }
+ *
+ *     // Can also be
+ *     // `ServerEnvironment.use(inMemoryStorageFactory, new Staging(stagingServiceDiscovery));`,
+ *     // if `Staging` expects dependencies to its constructor.
+ *     ServerEnvironment.use(inMemoryStorageFactory, Staging.class);
+ * </pre>
+ * If {@code Staging} is {@link Environment#type() enabled}, the specified value is going to be
```
armiol, line 71: Start with `<p>` and add an empty line above.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -47,6 +47,29 @@
 
 /**
  * The server conditions and configuration under which the application operates.
+ *
+ * <h1>Configuration</h1>
+ * <p>Some parts of the {@code ServerEnvironment} can be customized based on the {@code
+ * EnvironmentType}. To do so, one of the overloads of the {@code use} method can be called.
+ * Two environment types exist out of the box: {@link Tests} and {@link Production}. For example:
+ * <pre>
```
armiol, line 55: Let's have an empty line above.

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -45,27 +45,29 @@
...
- * configured} internally.
+ * <p>Fallback values are calculated once on first {@linkplain #value(Class) access} for the
+ * specified environment. Every subsequent access returns the cached value.
  *
  * <pre>
  *      // This `Supplier` is calculated only once.
  *     {@literal Supplier<StorageFactory>} fallbackStorage = InMemoryStorageFactory::newInstance;
  *
- *     {@literal EnvSetting<StorageFactory>} setting = {@literal new EnvSetting<>(Tests.class, fallbackStorage);}
+ *     {@literal EnvSetting<StorageFactory>} setting =
```
armiol, line 69: Can we fit this into a single literal and in a single line?
```java
{@literal EnvSetting<StorageFactory> setting = new EnvSetting<>(Tests.class, fallbackStorage);}
```

yuri-sergiichuk:
@serhii-lekariev so `ServerEnvironment.configureTransport` is removed completely. Maybe worth mention that? That's kind of a breaking change (a tiny one, but still).

serhii-lekariev:
> @serhii-lekariev so `ServerEnvironment.configureTransport` is removed completely. Maybe worth mention that? That's kind of a breaking change (a tiny one, but still).

I've updated the PR description.

PR #1276, opened 2020-05-22
Fix tupe package doc

alexander-yevsyukov:
This PR improves the description of the `tuple` package.

The dependency on `base` was reverted to `1.5.10` to fix the error caused by incompatibility with Kotlin-based build scripts.

The version advances to `1.5.13`.


alexander-yevsyukov:
@armiol, PTAL.

server/src/main/java/io/spine/server/tuple/package-info.java
```diff
@@ -19,42 +19,57 @@
  */
 
 /**
- * This package provides tuples for return values from
- * {@linkplain io.spine.server.command.Assign command handling} or
- * {@linkplain io.spine.server.command.Command commanding} methods.
+ * This package provides classes for return values from message handling methods.
```
armiol, line 22: I would say `... from the message-handling methods.` It's easier to comprehend, as it becomes an adjective and a noun instead of a noun-adjective-noun.

server/src/main/java/io/spine/server/tuple/package-info.java
```diff
@@ -19,42 +19,57 @@
...
+ * sometime later. In this case combining returned messages does not reflect the domain language.
+ * Returning a tuple tells a part of the story in terms of the domain language right in the method
+ * signature. Details of this story are obtained from the method body.
+ *
+ * <p>It should be re-iterated that the purpose of this package is limited to the scenarios
+ * of handling messages. Programmers are strongly discouraged from using this package for
+ * other purposes.
+ *
+ * <h1>Two groups of classes</h1>
+ * This package provides two groups of classes:
```
armiol, line 59: I would have an empty line above. So that the Javadoc in its raw form reads better in the editor.

server/src/main/java/io/spine/server/tuple/package-info.java
```diff
@@ -64,11 +79,28 @@
...
+ * </pre>
  *
  * <h1>Alternatives</h1>
  *
+ * <p>In some cases it is needed to return one message from a limited set of possible options.
+ * For example, the following method issues a command {@code MoveToTrash} if no work
+ * was reported on the task, and {@code CancelTask} if some work has been already logged.
+ * <pre>
+ *    {@literal @}Command
+ *     EitherOf2&lt;MoveToTrash, CancelTask&gt; handle(RemoveTask cmd) { ... }
```
dmdashenkov, line 101: Can we use `{@literal }` instead of escape sequences?

alexander-yevsyukov: I tried. It becomes too wordy specially when it comes to links or code samples.

PR #1277, opened 2020-05-22
Downgrade version of `base`

dmdashenkov:
Since a recent change, the version `1.5.11` of `base` was overridden by a new artifact. This caused a build failure in this repository due to an incompatibility in the Gradle plugin test utils.

In this PR we downgrade the version of `base` to `1.5.10` so that the build would pass again.


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #1278, opened 2020-05-28
Kotlin Gradle scripts

dmdashenkov:
In this PR we migrate all Gradle scripts to Kotlin.

We also start checking that the current version of the `core` library is not yet published to the Maven repository. As a rule, we only need to check one module per Git repository.


dmdashenkov:
@armiol, PTAL.

build.gradle.kts
```diff
@@ -0,0 +1,256 @@
...
+    tasks.register("javadocJar", Jar::class) {
+        from("$projectDir/build/docs/javadoc")
+        archiveClassifier.set("javadoc")
+
+        dependsOn(tasks.javadoc)
+    }
+
+    idea {
+        module {
+            generatedSourceDirs.add(file(generatedJavaDir))
```
armiol, line 207: I assume they don't have `addAll()`, don't they?

buildSrc/src/main/kotlin/io/spine/gradle/internal/CheckVersionIncrement.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.gradle.api.internal.AbstractTask
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.TaskAction
+import java.net.URL
+
+/**
+ * A task which verifies that the current version of the library has not been published to the given
+ * Maven repository yet.
+ */
+open class CheckVersionIncrement : AbstractTask() {
```
armiol, line 36: Do we have to carry this with each of the repositories, or we can copy it over from `config` and not have it in the GitHub repo itself?

version.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * The versions of the libraries used.
+ *
+ * This file is used in both module `build.gradle.kts` scripts and in the integration tests,
+ * as we want to manage the versions in a single source.
+ */
+
+project.extra.apply {
+    this["spineBaseVersion"] = "1.5.12"
```
armiol, line 29: This is going to be pretty hard to parse from the auto-updating script.

Is there anything we could do to make it more object-oriented, maybe? Or less `String`-dependent?

buildSrc/src/main/kotlin/io/spine/gradle/internal/CheckVersionIncrement.kt
dmdashenkov: We have to copy it because CI needs it as well. As a rule, we don't do `pull` on CI since `config` used in a repository may be older (intentionally) or use a non-master branch.

dmdashenkov:
@armiol, PTAL again.

PR #1279, opened 2020-06-18
Update `spine-base` and `spine-time` dependencies

yuri-sergiichuk:
In this PR I have bumped the `base` and `time` dependencies to their up-to-date versions (v1.5.19).

As a side change, I have backported the behavior of the publishing script (same as in https://github.com/SpineEventEngine/config/pull/124) that allows us to fail the build in case of an issue with the publishing.


yuri-sergiichuk:
@armiol, PTAL.

PR #1284, opened 2020-08-06
Add warning logs in `ServerEnvironment`

yuri-sergiichuk:
In this PR I have added warning logs on using `InMemoryStorageFactory` in `Production` environment. I guess it should be a good hint for developers trying to figure out why queries return nothing in non-test environments.

Answering the question "how it can happen that we have an in-memory config for Production": it is very tempting to start working with the application using in-memory configurations and then replace them with the actual ones. From my perspective, that's a valid case while developing applications.


yuri-sergiichuk:
@armiol PTAL.

yuri-sergiichuk:
Closing the PR in favor of the #1285 issue. We may need to give the issue more thought and come up with a more decent solution.

PR #1292, opened 2020-08-14
Post and forget

alexander-yevsyukov:
This PR adds an ability to post a command without subscribing to events. 

Prior to this PR, it was possible to call `CommandRequest.post()` ignoring the result of the method. Now the contract is tightened:
  * It is required to subscribe to at least one event before calling `post()`. Otherwise, `IllegalStateException` will be thrown suggesting to call `observe()` or `postAndForget()` instead.
 * If no events are of interest for the client-side code, `postAndForget()` should be called. If this method is called and at least one event subscription were requested, the method throws `IllegalStateException` suggesting to call `post()` instead.
 * The value returned by `post()` cannot be ignored.

### DIFF
--- a/client/src/main/java/io/spine/client/CommandRequest.java
+++ b/client/src/main/java/io/spine/client/CommandRequest.java
@@ -171,10 +171,25 @@ public final class CommandRequest extends ClientRequest implements Logging {
      *         client code "knows" how many of them it expects. Also, some events may not arrive
      *         because of communication or business logic reasons. That's why the returned
      *         subscriptions should be cancelled by the client code when it no longer needs it.
+     * @see #postAndForget()
      */
     public ImmutableSet<Subscription> post() {
         PostOperation op = new PostOperation();
         return op.perform();
+    }
+
+    /**
+     * Posts the command without subscribing to events that may be generated because of
+     * handling the command.
+     *
+     * @throws IllegalStateException
+     *   if {@link #observe(Class, EventConsumer)} or {@link #observe(Class, Consumer)} were
+     *   called in the command request configuration chain before calling this method
+     * @see #post()
+     */
+    public void postAndForget() throws IllegalStateException {
+        PostOperation op = new PostOperation();
+        op.performWithoutSubscriptions();
     }
 
     @VisibleForTesting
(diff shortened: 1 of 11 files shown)


client/src/main/java/io/spine/client/CommandRequest.java
```diff
@@ -171,13 +171,27 @@ public CommandRequest onServerError(ServerErrorHandler handler) {
...
      */
-    @CanIgnoreReturnValue
     public ImmutableSet<Subscription> post() {
         PostOperation op = new PostOperation();
         return op.perform();
     }
 
+    /**
+     * Posts the command without subscribing to events that may be generated because of
+     * handling the command.
```
yuri-sergiichuk, line 183: Maybe like this?

```suggestion
     * Posts the command without subscribing to events that may be generated during
     * the command handling.
```

client/src/main/java/io/spine/client/CommandRequest.java
```diff
@@ -171,13 +171,27 @@ public CommandRequest onServerError(ServerErrorHandler handler) {
...
+     *
+     * @throws IllegalStateException
+     *   if {@link #observe(Class, EventConsumer)} or {@link #observe(Class, Consumer)} were
+     *   called in the command request configuration chain before calling this method
```
yuri-sergiichuk, line 187: Is it the new formatting style? We usually have the description aligned with the exception name

client/src/main/java/io/spine/client/CommandRequest.java
```diff
@@ -201,9 +215,41 @@ private PostOperation() {
...
             subscribeToEvents();
             Ack ack = client().post(command);
             Status status = ack.getStatus();
+            return handleStatus(status);
+        }
+
+        private void performWithoutSubscriptions() {
+            if (!consumers.isEmpty()) {
+                throw newIllegalStateException(
+                        "Subscriptions to events were requested. Please call `post()` instead.");
```
yuri-sergiichuk, line 236: Let's move the closing `)` to the next line. It looks a bit prettier this way.

client/src/main/java/io/spine/client/CommandRequest.java
```diff
@@ -201,9 +215,41 @@ private PostOperation() {
...
+            }
+            Ack ack = client().post(command);
+            Status status = ack.getStatus();
+            handleStatus(status);
+        }
+
+        private void subscribeToEvents() {
+            Client client = client();
+            this.subscriptions = subscribe(client, command, consumers, streamingErrorHandler());
+            checkNotNull(subscriptions);
```
yuri-sergiichuk, line 246: Maybe worth having some words here as well?

Issue #1291, opened 2020-08-13
`server` lacks proper Custom Environments support


yuri-sergiichuk:
With the introduction of the Custom Environments (see PR [#539](https://github.com/SpineEventEngine/base/pull/539)), we're now able to create custom environment types as `Beta` or `Staging`.

The feature itself is nice, but there are still places in the framework that treat `Production` in a very special way that is not portable for the usage of custom environment types.

Let's consider we have an application that is deployed in Production and Staging environments. Now, let's glance at the following parts of the framework where we should reconsider the current behavior.

-----

`ServerEnvironment#close`: 

```java ServerEnvironment.java
    @Override
    public void close() throws Exception {
        tracerFactory.ifPresentForEnvironment(Production.class, AutoCloseable::close);
        transportFactory.ifPresentForEnvironment(Production.class, AutoCloseable::close);
        storageFactory.ifPresentForEnvironment(Production.class, AutoCloseable::close);
    }
```

The code above will just not free the resources for the Staging environment even though it is the desired behavior.

-----

`StateSubscriberMethod#contextOf`:

```java StateSubscriberMethod.java
    private BoundedContextName contextOf(Class<?> cls) {
        Model model = Model.inContextOf(cls);
        BoundedContextName name = model.contextName();
        if (Environment.instance().is(Production.class) && name.equals(assumingTests())) {
            _warn().log(
                    "The class `%s` belongs to the Bounded Context named `%s`," +
                    " which is used for testing. As such, it should not be used in production." +
                    " Please see the description of `%s` for instructions on" +
                    " annotating packages with names of Bounded Contexts of your application.",
                    cls.getName(), assumingTests().getValue(), BoundedContext.class.getName()
            );
        }
        return name;
    }
```

We'll end up deploying code without `@BoundedContext` annotation up to the Production environment and we'll not notice the issue while running on the Staging.

-----

`SystemSettings#defaults`:

```java SystemSettings.java
    public static SystemSettings defaults() {
        SystemSettings settings = new SystemSettings()
                .disableCommandLog()
                .enableAggregateQuerying()
                .forgetEvents();
        if (Environment.instance().is(Production.class)) {
            settings.enableParallelPosting();
        } else {
            settings.disableParallelPosting();
        }
        return settings;
    }
```

The system events will be posted in parallel only in Production, but that should be the desired behavior on Staging as well.

yuri-sergiichuk:
[...]

PR #1293, opened 2020-08-18
Improve custom environments support

dmitrykuzmin:
This PR fixes issue #1291.

Parts of the framework where the `Production` environment usage was hard-coded are changed to support the custom environment types.

Please note, that the `Tests` environment is still treated specially in some cases. For example, the parallel posting of system events is disabled for tests as some of them rely on the eventual consistency of the data.

The library version advances to `1.5.25`.

### DIFF
--- a/server/src/main/java/io/spine/server/EnvSetting.java
+++ b/server/src/main/java/io/spine/server/EnvSetting.java
@@ -137,7 +137,7 @@ final class EnvSetting<V> {
     }
 
     /**
-     * Applies the passed operation to this setting regardless of the current environment.
+     * Applies the passed operation to this setting regardless of current environment.
      *
      * <p>This means the operation is applied to all passed setting {@linkplain #environmentValues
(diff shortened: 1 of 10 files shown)


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -86,7 +85,6 @@
  * @param <V>
  *         the type of value
  */
-@Internal
```
armiol, line 89: Not sure why this class is not internal anymore. Do we expect the framework users to use it directly?

dmitrykuzmin: The class became `package-private` at one point, so I thought this annotation presence is just a typo.

armiol: OK, now it makes sense.

PR #1295, opened 2020-08-28
Improve support of rejections thrown from the `CommandBus` filters

dmitrykuzmin:
This PR partially addresses https://github.com/SpineEventEngine/web/issues/6.

Currently, the `Ack` received from a command filter can contain either `OK` status, `Error`, or rejection `Event`.

While the first two cases are well-supported in the current implementation, the `rejection` field is not really expected to be populated.

Per vocal discussion, it was decided that receiving a rejection in a command filter is actually a valid use-case, and thus this PR tweaks the `core` to support such a case. The system context will post the corresponding system events when the command is rejected this way, and the provided rejection event will be posted to the `EventBus`.

### Other changes

* `BusFilter` interface is expanded with shortcut methods for the convenient `Ack` creation.
* The command and event filters are included in the bounded context `testingCopy()` and can thus be specified on `BlackBoxContext` creation.
* The `AbstractCommandHandler` will post the corresponding events when its handler method invocation leads to a rejection.

The library version advances to `1.5.27`.


dmitrykuzmin:
@armiol PTAL.

server/src/main/java/io/spine/server/bus/AckFactory.java
```diff
@@ -35,34 +35,33 @@
 import static io.spine.util.Preconditions2.checkNotDefaultArg;
 
 /**
- * A utility for working with {@link Bus buses} and related data types.
+ * A utility for producing {@link Ack} instances.
  */
 @Internal
-public final class Buses {
+public final class AckFactory {
```
armiol, line 41: I would call that `Acks`.

server/src/main/java/io/spine/server/bus/BusFilter.java
```diff
@@ -40,18 +47,98 @@
...
      * </ul>
      *
-     * @param envelope the envelope with the message to filter
+     * @param envelope
+     *         the envelope with the message to filter
      * @return {@code Optional.empty()} if the message passes the filter,
      *         {@linkplain Ack posting result} with either status otherwise
      */
-    Optional<Ack> accept(E envelope);
+    Optional<Ack> doFilter(E envelope);
```
armiol, line 65: Not sure why rename that.

The idea of having a method like `doFilter()` is to avoid the clash with `filter()` in the public API. We don't have `filter()` here.

server/src/main/java/io/spine/server/commandbus/AckRejectionPublisher.java
```diff
@@ -0,0 +1,66 @@
...
+    public void onNext(Ack value) {
+        Status status = value.getStatus();
+        if (status.getStatusCase() == REJECTION) {
+            eventBus.post(status.getRejection());
+        }
+    }
+
+    @Override
+    public void onError(Throwable t) {
+        // NO-OP.
```
armiol, line 59: Shouldn't we rethrow it?

server/src/main/java/io/spine/server/commandbus/CommandAckMonitor.java
```diff
@@ -71,20 +65,27 @@ private CommandAckMonitor(Builder builder) {
...
      */
     @Override
     public void onNext(Ack value) {
-        super.onNext(value);
         postSystemEvent(value);
     }
 
+    @Override
+    public void onError(Throwable t) {
+        // NO-OP.
```
armiol, line 78: Same as above. Shouldn't we rethrow the exception?

server/src/main/java/io/spine/server/commandbus/CommandBus.java
```diff
@@ -92,6 +95,18 @@
     @LazyInit
     private @MonotonicNonNull CommandValidator commandValidator;
 
+    /**
+     * An observer which processes the {@linkplain io.spine.base.ThrowableMessage rejections}
+     * thrown by the bus {@linkplain BusFilter filters}.
+     *
+     * <p>Is NO-OP at bus creation. Once an {@link EventBus} is
```
armiol, line 102: `When the bus is first created, this observer does nothing. Once an {@link EventBus} is ...`

armiol (changes requested):
@dmitrykuzmin LGTM in general. Please see my comments.

dmitrykuzmin:
@armiol PTAL again.

armiol (approved):

Issue #1228, opened 2020-01-23
Add a way to create a `Pair` with an already created `Optional<>`


nick-dolgiy:
Spine allows having an `Optional` as the second argument of `Pair` but the only way to create it is call `Pair.withNullable()` which allows the second argument to be `null`, but there is no option to create a `Pair` if I have a message that is already wrapped in `Optional`.

I see the method signature like this:
``` Java
public static  <M extends Message, O extends Optional<? extends Message>> Pair<M, O> withOptional(M a, O b)
```

The resulting API will be:
``` Java
Optional<Message> optional = Optional.of(...);

Pair<Message, Optional<Message>> = Pair.withOptional(message, optional);
```

serhii-lekariev:
Reassigned to @armiol.

alexander-yevsyukov:
Closed by #1296.

PR #1296, opened 2020-09-01
Pair on `Optional`

alexander-yevsyukov:
This PR adds a factory method to create `Pair` using `Optional` as the second parameter (see #1228).

Other minor changes:

  * **Fixed style of package description headers.** Previously they were H1, as demanded by IDEA style check for Javadoc. 
     It turned out that this...
          1) was incorrect from the HTML point of view which assumes only one H1 header per page;
          2) did not meet header nesting in the generated HTML which has package description section as H2.
  * **Improved code of precondition checks**. Each `Tuple` class now has a companion-like section of code which avoids 
       repeating passing the same class name for static precondition checks exposed by the `Tuple` class.


server/src/main/java/io/spine/server/tuple/Pair.java
```diff
@@ -94,4 +118,18 @@ public A getA() {
     public B getB() {
         return value(this, 1);
     }
+
+    /*
+     * Companion code
+     *******************/
```
dmdashenkov, line 124: I think the divider is not really necessary.

server/src/main/java/io/spine/server/tuple/Quartet.java
```diff
@@ -125,4 +126,28 @@ public C getC() {
     public D getD() {
         return value(this, 3);
     }
+
+    /*
+     * Companion code
+     *******************/
```
dmdashenkov, line 132: Same as above.

server/src/main/java/io/spine/server/tuple/Quintet.java
```diff
@@ -153,4 +154,28 @@ public D getD() {
     public E getE() {
         return value(this, 4);
     }
+
+    /*
+     * Companion code
+     *******************/
```
dmdashenkov, line 160: Same as above.

server/src/main/java/io/spine/server/tuple/Triplet.java
```diff
@@ -104,4 +105,28 @@ public B getB() {
     public C getC() {
         return value(this, 2);
     }
+
+    /*
+     * Companion code
+     *******************/
```
dmdashenkov, line 111: Same as above.

dmdashenkov (approved):
@alexander-yevsyukov, LGTM with a single suggestion (to 4 files).

PR #1298, opened 2020-09-02
Fix how the `Migration`s with custom state update are applied

### LINKED ISSUE #1297: The `Migration` result is not stored if it is equal to `entityStateWithColumns`
## The issue

A custom `Migration` does not modify the entity state if the modified state is equal to the `entityStateWithColumns`.

## Expected result

Any result of a custom `Migration` should modify the entity state.

## More details
If one has written a custom migration that for some reason produced the same result as the `entityStateWithColumns`, the migration result is not applied.

Consider the following Process to be migrated:
```proto
message Project {
    option (entity).kind = PROCESS_MANAGER;

    ProjectId id = 1;
    string name = 2 [(column) = true];
    repeated Task task = 3;

    enum Status {
        UNDEFINED = 0;
        CREATED = 1;
        STARTED = 2;
        DONE = 3;
        CANCELLED = 4;
    }
    Status status = 4;

    string id_string = 5 [(column) = true];
}
```

Let's assume that we have an oldish process state in the storage and want to perform its migration.
While doing the migrating, we set to `id_string` to the `id` field value:

```java
    @Override
    public Project apply(Project project) {
        ProjectId id = id();
        Project result = project
                .toBuilder()
                .setIdString(id.toString())
                .vBuild();
        return result;
    }
```

And in the process manager itself, we have the `id_string` column defined as:
```java

    @Override
    public String getIdString() {
        return id().toString();
    }
```

When we apply the migration, the `id_string` value is set to the ID field value. Then, when the migration transaction is going to be committed the `Transaction#updateColumns` method is triggered, but `stateWithColumns` value is now just equal to the migrated one and thus the entity state is not updated:

```java
private void updateColumns() {
        S stateWithColumns = stateWithColumns();
        if (!stateWithColumns.equals(entity.state())) {
            entity.updateState(stateWithColumns);                 <----- This updateState is never called
        }
    }
```

See the test case here: https://github.com/SpineEventEngine/core-java/commit/75b579495774e270f3ac2ad924f7587ec13f06bc#diff-04de199876e461313a5f827e087e59edR692

dmitrykuzmin:
This PR fixes #1297.

The root cause of the bug is a discrepancy between entity and underlying transaction versions in `Migration` when updating the entity state.

This PR thus adds the missing `setVersion(...)` call.

It also applies some refactoring to the `Migration` class. Its current logic is confusing, with entity updates being propagated partially in-place, partially on transaction commit. This leads to strange code paths and observed values when debugging.

After the refactoring, all entity changes will be propagated on transaction commit at the very end of `Migration.applyTo(...)`.


dmitrykuzmin:

server/src/test/java/io/spine/server/procman/ProcessManagerRepositoryTest.java
```diff
@@ -346,9 +348,10 @@ void events() {
                     .comparingExpectedFieldsOnly()
                     .isEqualTo(event.messageId());
             PmProjectStarted eventMessage = (PmProjectStarted) event.enclosedMessage();
-            assertThat(unpack(systemEvent.getEntity().getId()))
-                      .comparingExpectedFieldsOnly()
-                      .isEqualTo(eventMessage.getProjectId());
+            assertThat(unpack(systemEvent.getEntity()
+                                         .getId()))
+                    .comparingExpectedFieldsOnly()
+                    .isEqualTo(eventMessage.getProjectId());
```
yuri-sergiichuk, line 354: Why do we compare only the IDs here and below? Can we just hand-craft the expected message and compare the whole message instead?

server/src/test/java/io/spine/server/procman/ProcessManagerRepositoryTest.java
```diff
@@ -648,6 +655,74 @@ void whenCreating() {
...
+
+        // Check the entity is now found by the provided filters.
+        Iterator<TestProcessManager> afterMigration =
+                repository.find(targetFilters, ResponseFormat.getDefaultInstance());
+        assertThat(afterMigration.hasNext()).isTrue();
+
+        // Check the column value is propagated to the entity state.
+        TestProcessManager entityWithColumns = afterMigration.next();
+        assertThat(entityWithColumns.state()
+                                    .getIdString()).isEqualTo(id.toString());
```
yuri-sergiichuk, line 686: I'd actually compare `state` and not the ID only.

server/src/test/java/io/spine/server/projection/given/SetTestProjectionName.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package io.spine.server.projection.given;
+
+import io.spine.server.projection.ProjectionMigration;
+import io.spine.test.projection.Project;
+import io.spine.test.projection.ProjectId;
+
+/**
+ * A migration which sets the projection name to a predefined {@linkplain #NEW_NAME value}.
```
yuri-sergiichuk, line 28:
```suggestion
 * Sets the projection {@code name} to a predefined {@linkplain #NEW_NAME value}.
```

server/src/test/java/io/spine/server/procman/given/repo/SetTestProcessName.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package io.spine.server.procman.given.repo;
+
+import io.spine.server.procman.ProcessManagerMigration;
+import io.spine.test.procman.Project;
+import io.spine.test.procman.ProjectId;
+
+/**
+ * A migration which sets the process name to a predefined {@linkplain #NEW_NAME value}.
```
yuri-sergiichuk, line 28:
```suggestion
 * Sets the process {@code name} to a predefined {@linkplain #NEW_NAME value}.
```

server/src/test/java/io/spine/server/procman/given/repo/SetTestProcessName.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public final class SetTestProcessName
+        extends ProcessManagerMigration<ProjectId, TestProcessManager, Project, Project.Builder> {
+
+    public static final String NEW_NAME = "Migrated project";
```
yuri-sergiichuk, line 33: Maybe `Migrated process`? You have `Migrated projection` in the projection migration below.

server/src/test/java/io/spine/server/procman/ProcessManagerRepositoryTest.java
```diff
@@ -738,7 +815,8 @@ void deleteEntityViaMigration() {
 
         Optional<TestProcessManager> found = repository().find(id);
         assertThat(found.isPresent()).isTrue();
```
yuri-sergiichuk, line 817: There's a whole set of `Truth8` assertions that should allow having a bit nicer syntax for `Optional`s. Could you please check it out and consider using them instead?

server/src/main/java/io/spine/server/entity/Migration.java
```diff
@@ -238,6 +239,7 @@ private void warnOnNoSystemEventsPosted() {
         private boolean physicallyRemoveRecord;
 
         private final E entity;
+        private @MonotonicNonNull Transaction<I, E, S, ?> tx;
         private final RecordBasedRepository<I, E, S> repository;
 
         private @MonotonicNonNull Event systemEvent;
```
yuri-sergiichuk, line 245: Same for `@LazyInit`. It looks like we should add the annotation to the `systemEvent` as well.

server/src/main/java/io/spine/server/entity/Migration.java
```diff
@@ -238,6 +239,7 @@ private void warnOnNoSystemEventsPosted() {
         private boolean physicallyRemoveRecord;
```
yuri-sergiichuk, line 239: In the `Operation` docs it states that it's the operation done on an entity instance. In fact, we apply the migration to the `tx().builder()` now, right?

I guess it really worth adding some more details on how the migration operation is actually being done while you're in the context now.

yuri-sergiichuk (changes requested):
@dmitrykuzmin PTAL at the comments below.

dmitrykuzmin:
@yuri-sergiichuk PTAL again.

server/src/main/java/io/spine/server/entity/Migration.java
```diff
@@ -75,8 +76,9 @@
     private @Nullable Operation<I, S, E> currentOperation;
 
     /**
-     * Applies the migration to a given entity, starting a new migration
-     * {@linkplain Operation operation}.
+     * Applies the migration {@linkplain Operation operation} to a given entity.
+     *
+     * @see Operation Migration.Operation for details
```
yuri-sergiichuk, line 81: This `see` section looks redundant. You have the very same link as for `operation`. I'd better remove it.

server/src/main/java/io/spine/server/entity/Migration.java
```diff
@@ -75,8 +76,9 @@
     private @Nullable Operation<I, S, E> currentOperation;
 
     /**
-     * Applies the migration to a given entity, starting a new migration
-     * {@linkplain Operation operation}.
+     * Applies the migration {@linkplain Operation operation} to a given entity.
```
yuri-sergiichuk, line 79:
```suggestion
     * Applies the migration {@linkplain Operation operation} to a given {@code entity}.
```

server/src/test/java/io/spine/server/bus/BusFilterTest.java
```diff
@@ -71,15 +72,15 @@ void passNullToleranceCheck() {
     void letPass() {
         BusFilter<CommandEnvelope> filter = new BusFilters.Accepting();
         Optional<Ack> ack = filter.filter(commandEnvelope);
-        assertThat(ack.isPresent()).isFalse();
+        Truth8.assertThat(ack).isEmpty();
```
yuri-sergiichuk, line 75: Please import `assertThat` statically here and in the other places.

server/src/test/java/io/spine/server/procman/ProcessManagerRepositoryTest.java
```diff
@@ -675,7 +754,8 @@ void updateColumns() {
 
         // Check the column value is propagated to the entity state.
         TestProcessManager entityWithColumns = afterMigration.next();
-        assertThat(entityWithColumns.state().getIdString()).isEqualTo(id.toString());
+        assertThat(entityWithColumns.state()
```
yuri-sergiichuk, line 757: Can we please have expected state here as well?

server/src/test/java/io/spine/server/procman/ProcessManagerRepositoryTest.java
```diff
@@ -723,8 +803,9 @@ void archiveEntityViaMigration() {
         repository().applyMigration(id, new MarkPmArchived<>());
 
         Optional<TestProcessManager> found = repository().find(id);
-        assertThat(found.isPresent()).isTrue();
-        assertThat(found.get().isArchived()).isTrue();
+        Truth8.assertThat(found).isPresent();
```
yuri-sergiichuk, line 806: Please use static imports.

server/src/test/java/io/spine/server/procman/ProcessManagerRepositoryTest.java
```diff
@@ -773,6 +855,7 @@ private static TargetFilters targetFilters(QueryFilter first, QueryFilter... res
     }
 
     private static boolean hasId(TestProcessManager projection, ProjectId id) {
-        return projection.id().equals(id);
+        return projection.id()
```
yuri-sergiichuk, line 858: Let's flip the params. You'll then be able to have them on the same line.

server/src/test/java/io/spine/server/projection/ProjectionRepositoryTest.java
```diff
@@ -548,6 +551,79 @@ void notRegisterIfSubscribedToNothing() {
...
+
+        // Check the entities are now found by the provided filters.
+        Iterator<TestProjection> foundAfterMigration =
+                repository.find(filters, ResponseFormat.getDefaultInstance());
+
+        ImmutableList<TestProjection> results = ImmutableList.copyOf(foundAfterMigration);
+        assertThat(results).hasSize(2);
+        assertThat(results)
+                .comparingElementsUsing(idCorrespondence())
+                .containsExactly(id1, id2);
```
yuri-sergiichuk, line 624: Can we test states here as well?

server/src/test/java/io/spine/server/projection/ProjectionRepositoryTest.java
```diff
@@ -623,7 +699,7 @@ void archiveEntityViaMigration() {
         repository().applyMigration(id, new MarkProjectionArchived<>());
 
         Optional<TestProjection> found = repository().find(id);
-        assertThat(found.isPresent()).isTrue();
+        Truth8.assertThat(found).isPresent();
```
yuri-sergiichuk, line 702: Please use static imports.

yuri-sergiichuk (changes requested):
@dmitrykuzmin PTAL at the comments.

dmitrykuzmin:
@yuri-sergiichuk PTAL again.

server/src/test/java/io/spine/server/procman/ProcessManagerRepositoryTest.java
```diff
@@ -648,6 +655,78 @@ void whenCreating() {
...
+
+        // Check the entities are now found by the provided filters.
+        Iterator<TestProcessManager> foundAfterMigration =
+                repository.find(filters, ResponseFormat.getDefaultInstance());
+
+        ImmutableList<TestProcessManager> results = ImmutableList.copyOf(foundAfterMigration);
+        assertThat(results).hasSize(2);
+        assertThat(results)
+                .comparingElementsUsing(stateWithPropagatedName())
+                .containsExactly(pm1, pm2);
```
yuri-sergiichuk, line 727: So the migration does nothing? We're comparing the results to the same PMs we've manually stored.

yuri-sergiichuk: OK, now I see that inside the correspondence we actually manually change the state of the PMs to match the actual one. It looks a bit too complicated...

yuri-sergiichuk (approved):
@dmitrykuzmin in general LGTM.

I'd prefer to have a bit more straight-forward states comparison.

PR #1305, opened 2020-10-08
[2.x-jdk8] New Storage API and massive cleanup

armiol:
This changeset brings a number of significant API changes, including the support of a new Query features (see [the respective PR in `base`](https://github.com/SpineEventEngine/base/pull/564)).

## Unification in Storage implementations

- From now on, all storage implementations are unified, and use a configured `RecordStorage` underneath by delegating all of their operations to it. Please see the updated Javadocs of the `StorageFactory` for more details.

- `EntityState<I>` is now parameterized with the type of the entity identifier.

The framework is configured to use interface to mark the Java classes generated from the Proto messages. Therefore, the concrete values of `<I>` entity identifiers are detected for each proto-message declaring the entity state; Spine Proto plugin inserts them, binding the Java class of each Entity state to the type of its identifier. This makes the bounds between the types of entity, its repository, storage, and state much more strict, as each of them is now typed with the same `<I>` type of entity ID. 

Many use cases in Spine tests were fixed to satisfy these criteria. Interestingly, it all worked fine previously, as the tests were dealing with a limited scope of the framework functionality in each case.

## Storage interfaces

- `Storage` interface was simplified. As a result, it has lost its third generic parameter, which previously was declared as `R extends ReadRequest<I>`. 

- The `AbstractStorage` implementation of a `Storage` interface now has `isClosed` method working through `!isOpen()`. The `isClosed()` method became `final`. 

- All methods producing the storages existing in scope of a Bounded Context now take `ContextSpec` as a first parameter. It allows to use this information when configuring the settings of an underlying DBMS. E.g. set the names of tables or namespaces in RDBMS.

## Entity-releated changes

- The Spine-generated interface-based columns are removed. End-users should declare their columns as a part of the entity state in the respective Protobuf message. Also, before the commit of each transaction, an overridable `TransactionalEntity.onBeforeCommit()` is executed. It allows the end-users to insert a common logic on updating the calculated fields by interacting with Entity's `builder()`. Here is an exterpt from our unit tests (see `MenuProjection`):

```java

public final class MenuProjection
        extends Projection<MenuId, Menu, Menu.Builder> {

    public static final String UUID_COLUMN = "uuid";

    // `@Subscribe` methods here ...

   
    @Override
    protected void onBeforeCommit() {

    	// Updates the `UUID` column of the state.
        builder().setUuid(id().getUuid());
    }
}

```

The types describing the `Column`s and their parts were removed, as they mostly repeated the language definitions from `base`. The latter types are used instead.

- Entity lifecycle columns are now strictly defined via an enumeration instead of a plain `String` as previously. See `EntityRecordColumn`, `HasLifecycleColumns` and `HasVersionColumn`.

- All `Aggregate`, `Projection` and `ProcessManager` repositories are now `QueryableRepository`. It's a contract defining to query each of the repositories for the `EntityRecord`s.

- `EEntity` is removed in favor of storing the `Event`s as message records along with their columns.

- Entity.modelClass() is made `public` in order to allow the discovery of `Entity` attributes to the storage routines.

## Changes in data storage

- Aggregate mirroring now happens inside the `AggregateStorage`. All mirror-related functionality was removed from the System context. The instruction on the data migration is pending. See the `AggregateStorage` Javadocs for more details.

- Events are no longer persisted with a help of an `EEntity`. The latter type is removed. Instead, they are stored as message records.

- Similar to events, Tenants are also stored as message records, and not `Entity`-s. 

## Client and Testing

- It is now possible to specify several `order_by` statements for a `Query` sent from a client-side.

- `ClientRequest` now works with `EntityQuery`.

- `BlackBoxContext` was renamed to `BlackBox`. 

- `QueryFilter` and `CompositeQueryFilter` have been deprecated in favour of entity query language.

## Bug fixes

**Aggregate snapshot creation now works properly**
	
Previously, the snapshot was created on top of the latest state of the Aggregate, not on top the one which corresponded to the moment, when the event count exceeded the snapshot trigger count.

Say, a snapshot trigger is set to five. An Aggregate emits seven events after handling a command. According to the rules, the framework should persist five events, then a snapshot which reflects the Aggregate state after five events. Then persist two of the remaining events. However, before this changeset, the framework always persisted the snapshot based on the last known Aggregate state. I.e. the one after seven events, not after five.

Also, `Snapshot` Proto message now includes the lifecycle flags of an `Aggregate` at the moment of snapshotting. It allows to retain the state of an aggregate after the history replay, if it was archived or deleted right prior to snapshotting. Previously, we were losing the archived/deleted flags at all in such cases.

**Massive amount of non-working or obsolete test code was killed**

Not much to tell. It's just nice to maintain less code.

## Other changes

- `AggregateRepository`, `AggregatePartRepository` and `AggregateStorage` got one more generic parameter, being the type of the Aggregate state. This is needed in order to bound the type of the stored aggregate states.

- The default `Delivery` settings now operate with `CatchUpStorage` and `InboxStorage` in a single-tenant mode.

- Projection catch-up process has been made more stable in complex cases of concurrent event dispatching to the instances-under-catch-up. In this regard, the structure of the stored `CatchUp` messages was changed in an incompatible way with the data stored by the previous versions of the framework.

- `SystemClient` no longer provides an ability to search for a repository of an Entity registered in the System context. This is no longer needed, as the Aggregate mirroring is now located solely on the Domain context part of the application. The `SystemReadSide.readDomainAggregate(..)` has met the same destiny for the same reason.

- `SystemSettings` have lost all methods related to the `Aggregate` querying. Now, the ability to query aggregates is determined by its visibility level only.

- `EitherOfN` types are appended with `hasA()`, `hasB()`, ..., `hasZ()` methods. They tell if the N-th element of the tuple is set.

The version of the library has been set to the next snapshot in `2.x-jdk8` branch.

### DIFF
--- a/client/src/main/java/io/spine/client/Filters.java
+++ b/client/src/main/java/io/spine/client/Filters.java
@@ -30,7 +30,6 @@ import io.spine.base.EventMessageField;
 import io.spine.base.Field;
 import io.spine.base.FieldPath;
 import io.spine.client.CompositeFilter.CompositeOperator;
-import io.spine.code.proto.FieldName;
 import io.spine.core.Event;
 import io.spine.core.EventContextField;
 import io.spine.core.Version;
@@ -112,9 +111,11 @@ public final class Filters {
      *         the column to filter by
      * @param value
      *         the requested value
+     * @param <V>
+     *         the type of column values
      * @return a new instance of {@code Filter}
      */
-    public static Filter eq(Column<?, ?> column, Object value) {
+    public static <V> Filter eq(Column<?, V> column, V value) {
         checkNotNull(column);
         checkNotNull(value);
         return createFilter(column.name(), value, EQUAL);
@@ -195,9 +196,11 @@ public final class Filters {
      *         the entity column to filter by
      * @param value
      *         the requested value
+     * @param <V>
+     *         the type of column values
      * @return a new instance of {@code Filter}
      */
-    public static Filter gt(Column<?, ?> column, Object value) {
+    public static <V> Filter gt(Column<?, V> column, V value) {
         checkNotNull(column);
         checkNotNull(value);
         return createFilter(column.name(), value, GREATER_THAN);
@@ -284,9 +287,11 @@ public final class Filters {
      *         the entity column to filter by
      * @param value
      *         the requested value
+     * @param <V>
+     *         the type of column values
      * @return a new instance of {@code Filter}
      */
-    public static Filter lt(Column<?, ?> column, Object value) {
+    public static <V> Filter lt(Column<?, V> column, V value) {
         checkNotNull(column);
         checkNotNull(value);
         return createFilter(column.name(), value, LESS_THAN);
@@ -373,9 +378,11 @@ public final class Filters {
      *         the entity column to filter by
      * @param value
      *         the requested value
+     * @param <V>
+     *         the type of column values
      * @return a new instance of {@code Filter}
      */
-    public static Filter ge(Column<?, ?> column, Object value) {
+    public static <V> Filter ge(Column<?, V> column, V value) {
         checkNotNull(column);
         checkNotNull(value);
         return createFilter(column.name(), value, GREATER_OR_EQUAL);
@@ -462,9 +469,11 @@ public final class Filters {
      *         the entity column to filter by
      * @param value
      *         the requested value
+     * @param <V>
+     *         the type of column values
      * @return a new instance of {@code Filter}
      */
-    public static Filter le(Column<?, ?> column, Object value) {
+    public static <V> Filter le(Column<?, V> column, V value) {
         checkNotNull(column);
         checkNotNull(value);
         return createFilter(column.name(), value, LESS_OR_EQUAL);
@@ -576,7 +585,7 @@ public final class Filters {
      * @return new instance of {@link CompositeFilter}
      * @see #either(Filter, Filter...) for the public API equivalent
      */
-     static CompositeFilter either(Collection<Filter> filters) {
+    static CompositeFilter either(Collection<Filter> filters) {
         checkNotNull(filters);
         checkArgument(!filters.isEmpty(),
                       "Composite filter must contain at least one plain filter in it.");
@@ -594,7 +603,7 @@ public final class Filters {
      *
      * @param filters
      *         the aggregated filters
-     * @return new instance of {@link CompositeFilter}
+     * @return new instance of {@code CompositeFilter}
      * @see #all(Filter, Filter...) for the public API equivalent
      */
     static CompositeFilter all(Collection<Filter> filters) {
@@ -604,39 +613,57 @@ public final class Filters {
         return composeFilters(filters, ALL);
     }
 
-    static Filter createFilter(String fieldPath, Object value, Operator operator) {
-        Field field = Field.parse(fieldPath);
-        return createFilter(field, value, operator);
-    }
-
+    /**
+     * Creates a new filter which targets a certain column by its name comparing the actual
+     * column values to the one passed value according to the passed operator.
+     *
+     * @param columnName
+     *         the name of the column
+     * @param value
+     *         the value to filter by
+     * @param operator
+     *         the operator to use in comparison of actual values to the passed one
+     * @return a new filter instance
+     */
     static Filter createFilter(ColumnName columnName, Object value, Operator operator) {
         String fieldPath = columnName.value();
         return createFilter(fieldPath, value, operator);
     }
 
-
-    static Filter createFilter(FieldName fieldName, Object value, Operator operator) {
-        FieldPath fieldPath = fieldName.asPath();
-        return createFilter(fieldPath, value, operator);
-    }
-
+    /**
+     * Creates a new filter which targets a certain {@code Message} field by comparing the actual
+     * field values to the one passed value according to the passed operator.
+     *
+     * @param field
+     *         the declaration of a field in a {@code Message} instances of which are filtered
+     * @param value
+     *         the value to filter by
+     * @param operator
+     *         the operator to use in comparison of actual values to the passed one
+     * @return a new filter instance
+     */
     static Filter createFilter(Field field, Object value, Operator operator) {
+        checkNotNull(field);
         FieldPath fieldPath = field.path();
         return createFilter(fieldPath, value, operator);
     }
 
-    static Filter createFilter(FieldPath path, Object value, Operator operator) {
-        Any wrappedValue = toAny(value);
-        Filter filter = Filter
-                .newBuilder()
-                .setFieldPath(path)
-                .setValue(wrappedValue)
-                .setOperator(operator)
-                .build();
-        return filter;
-    }
-
+    /**
+     * Creates a filter for {@link Event} instances which is applied to one of the fields in
+     * the {@link io.spine.core.EventContext EventContext} corresponding to the {@code Event}.
+     *
+     * @param field
+     *         the declaration of the field in the {@code EventContext}
+     * @param value
+     *         the value to filter by
+     * @param operator
+     *         the operator to use in filtering
+     * @return a new instance of {@code Filter}
+     */
     static Filter createContextFilter(Field field, Object value, Operator operator) {
+        checkNotNull(field);
+        checkNotNull(value);
+        checkNotNull(operator);
         FieldPath fieldPath = Event.Field.context()
                                          .getField()
                                          .nested(field)
@@ -644,7 +671,20 @@ public final class Filters {
         return createFilter(fieldPath, value, operator);
     }
 
+    /**
+     * Creates a new composite filter out of the passed {@code Filter} instances joining them
+     * with the passed composite operator.
+     *
+     * @param filters
+     *         the filters to include into the composite filter
+     * @param operator
+     *         the operator to use in evaluation of the intermediate evaluation results
+     *         of each of the aggregated filters
+     * @return a new instance of {@code CompositeFilter}
+     */
     static CompositeFilter composeFilters(Iterable<Filter> filters, CompositeOperator operator) {
+        checkNotNull(filters);
+        checkNotNull(operator);
         CompositeFilter result = CompositeFilter
                 .newBuilder()
                 .addAllFilter(filters)
@@ -653,26 +693,64 @@ public final class Filters {
         return result;
     }
 
+    /**
+     * Obtains the {@link Filter}s from the passed array of {@code Filter} wrappers and returns
+     * them as new array.
+     *
+     * @param filters
+     *         the filter wrappers
+     * @return a new array of the {@code Filter} instances
+     */
     static Filter[] extractFilters(TypedFilter<?>[] filters) {
+        checkNotNull(filters);
         return stream(filters)
                 .map(TypedFilter::filter)
                 .toArray(Filter[]::new);
     }
 
+    /**
+     * Obtains the {@link Filter}s from the passed collection of {@code Filter} wrappers and returns
+     * them as a new immutable list.
+     *
+     * @param filters
+     *         the filter wrappers
+     * @param <M>
+     *         the type of the filtered messages specific to the passed filter wrappers
+     * @return an immutable list of the {@code Filter}s
+     */
     static <M extends Message> ImmutableList<Filter>
     extractFilters(Collection<? extends TypedFilter<M>> filters) {
+        checkNotNull(filters);
         return filters.stream()
                       .map(TypedFilter::filter)
                       .collect(toImmutableList());
     }
 
+    /**
+     * Obtains the {@link CompositeFilter}s from the passed array of {@code CompositeFilter}
+     * wrappers and returns them as new array.
+     *
+     * @param filters
+     *         the composite filter wrappers
+     * @return a new array of the {@code CompositeFilter} instances
+     */
     static CompositeFilter[] extractFilters(TypedCompositeFilter<?>[] filters) {
+        checkNotNull(filters);
         return stream(filters)
                 .map(TypedCompositeFilter::filter)
                 .toArray(CompositeFilter[]::new);
     }
 
+    /**
+     * Ensures that the instances of the passed type are supported for ordering comparison.
+     *
+     * <p>Throws an {@link IllegalArgumentException} if the passed type does not qualify for that.
+     *
+     * @param cls
+     *         the type to check
+     */
     static void checkSupportedOrderingComparisonType(Class<?> cls) {
+        checkNotNull(cls);
         Class<?> dataType = Primitives.wrap(cls);
         boolean supported = isSupportedNumber(dataType)
                 || Timestamp.class.isAssignableFrom(dataType)
@@ -700,5 +778,24 @@ public final class Filters {
     public static Predicate<Event> toEventFilter(CompositeFilter filterData) {
         checkNotNull(filterData);
         return new CompositeEventFilter(filterData);
+    }
+
+    private static Filter createFilter(FieldPath path, Object value, Operator operator) {
+        checkNotNull(path);
+        checkNotNull(value);
+        checkNotNull(operator);
+        Any wrappedValue = toAny(value);
+        Filter filter = Filter
+                .newBuilder()
+                .setFieldPath(path)
+                .setValue(wrappedValue)
+                .setOperator(operator)
+                .build();
+        return filter;
+    }
+
+    private static Filter createFilter(String fieldPath, Object value, Operator operator) {
+        Field field = Field.parse(fieldPath);
+        return createFilter(field, value, operator);
     }
 }
--- a/client/src/main/java/io/spine/client/QueryFilter.java
+++ b/client/src/main/java/io/spine/client/QueryFilter.java
@@ -35,7 +35,11 @@ import static io.spine.client.Filters.createFilter;
 
 /**
  * A query filter which targets a {@linkplain EntityColumn column} of an entity.
+ *
+ * <p>This type is deprecated. Use {@linkplain io.spine.query.EntityQuery entity queries} instead.
+ * See {@code io.spine.query} package documentation for more details on usage.
  */
+@Deprecated
 public final class QueryFilter extends TypedFilter<EntityState<?>> {
 
     private static final long serialVersionUID = 0L;
--- a/client/src/main/java/io/spine/client/QueryFilter.java
+++ b/client/src/main/java/io/spine/client/QueryFilter.java
@@ -113,7 +113,7 @@ public final class QueryFilter extends TypedFilter<EntityState<?>> {
      * @param <V>
      *         the type of the column values
      */
-    public static <V> QueryFilter ge(EntityColumn<?, ?> column, V value) {
+    public static <V> QueryFilter ge(EntityColumn<?, V> column, V value) {
         checkNotNull(column);
         checkNotNull(value);
         checkSupportedOrderingComparisonType(value.getClass());
--- a/client/src/main/java/io/spine/client/TargetBuilder.java
+++ b/client/src/main/java/io/spine/client/TargetBuilder.java
@@ -25,7 +25,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.protobuf.FieldMask;
 import com.google.protobuf.Message;
-import io.spine.annotation.Internal;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.Set;
@@ -286,7 +285,6 @@ public abstract class TargetBuilder<T extends Message, B extends TargetBuilder<T
      * <p>If the passed set is empty, does nothing.
      */
     @CanIgnoreReturnValue
-    @Internal
     B where(ImmutableSet<CompositeFilter> predicates) {
         checkNotNull(predicates);
         if (predicates.size() > 0) {
--- a/server/src/main/java/io/spine/server/aggregate/Aggregate.java
+++ b/server/src/main/java/io/spine/server/aggregate/Aggregate.java
@@ -430,7 +430,7 @@ public abstract class Aggregate<I,
         Version version;
         LifecycleFlags lifecycle;
         if (isTransactionInProgress()) {
-            AggregateTransaction tx = (AggregateTransaction) tx();
+            AggregateTransaction<?, ?, ?> tx = (AggregateTransaction<?, ?, ?>) tx();
             state = builder().buildPartial();
             version = tx.currentVersion();
             lifecycle = tx.lifecycleFlags();
--- a/server/src/main/java/io/spine/server/aggregate/UncommittedHistory.java
+++ b/server/src/main/java/io/spine/server/aggregate/UncommittedHistory.java
@@ -35,7 +35,7 @@ import static java.util.stream.Collectors.toList;
 /**
  * Uncommitted events and snapshots created for this aggregate during the dispatching.
  *
- * <p>Watches the how the events {@linkplain Aggregate#invokeApplier(EventEnvelope) are sent}
+ * <p>Watches how the events {@linkplain Aggregate#invokeApplier(EventEnvelope) are sent}
  * to the {@link Aggregate} applier methods. Remembers all such events as uncommitted.
  *
  * <p>Once an aggregate is loaded from the storage, the {@code UncommittedHistory}
--- a/server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
+++ b/server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
@@ -25,6 +25,8 @@ import com.google.protobuf.Timestamp;
 import io.spine.query.RecordColumn;
 import io.spine.query.RecordColumns;
 
+import static io.spine.query.RecordColumn.create;
+
 /**
  * The columns stored for {@link CatchUp} states.
  */
@@ -36,21 +38,20 @@ public class CatchUpColumn {
      * Stores the status of the catch-up process.
      */
     public static final RecordColumn<CatchUp, CatchUpStatus> status =
-            new RecordColumn<>("status", CatchUpStatus.class, CatchUp::getStatus);
+            create("status", CatchUpStatus.class, CatchUp::getStatus);
 
     /**
      * Stores the time when the history has been last read by the catch-up process.
      */
     public static final RecordColumn<CatchUp, Timestamp> when_last_read =
-            new RecordColumn<>("when_last_read", Timestamp.class, CatchUp::getWhenLastRead);
+            create("when_last_read", Timestamp.class, CatchUp::getWhenLastRead);
 
     /**
      * Stores the type URL of the projection-under-catch-up.
      */
     public static final RecordColumn<CatchUp, String> projection_type =
-            new RecordColumn<>("projection_type",
-                               String.class, (m) -> m.getId()
-                                                     .getProjectionType());
+            create("projection_type", String.class, (m) -> m.getId()
+                                                            .getProjectionType());
 
     /**
      * Prevents this type from instantiation.
--- a/server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
+++ b/server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
@@ -37,21 +37,21 @@ public class CatchUpColumn {
     /**
      * Stores the status of the catch-up process.
      */
-    public static final RecordColumn<CatchUp, CatchUpStatus> status =
-            create("status", CatchUpStatus.class, CatchUp::getStatus);
+    public static final RecordColumn<CatchUp, CatchUpStatus>
+            status = create("status", CatchUpStatus.class, CatchUp::getStatus);
 
     /**
      * Stores the time when the history has been last read by the catch-up process.
      */
-    public static final RecordColumn<CatchUp, Timestamp> when_last_read =
-            create("when_last_read", Timestamp.class, CatchUp::getWhenLastRead);
+    public static final RecordColumn<CatchUp, Timestamp>
+            when_last_read = create("when_last_read", Timestamp.class, CatchUp::getWhenLastRead);
 
     /**
      * Stores the type URL of the projection-under-catch-up.
      */
-    public static final RecordColumn<CatchUp, String> projection_type =
-            create("projection_type", String.class, (m) -> m.getId()
-                                                            .getProjectionType());
+    public static final RecordColumn<CatchUp, String>
+            projection_type = create("projection_type", String.class, (m) -> m.getId()
+                                                                              .getProjectionType());
 
     /**
      * Prevents this type from instantiation.
--- a/server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
+++ b/server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
@@ -547,12 +547,11 @@ public final class CatchUpProcess<I>
             if (observedJob.getStatus() == FINALIZING) {
                 int indexOfFinalized = event.getIndex()
                                             .getIndex();
-                List<Integer> finalizedShards = builder.getFinalizedShardList();
-                if (!finalizedShards.contains(indexOfFinalized)) {
+                if (!builder.getFinalizedShardList().contains(indexOfFinalized)) {
                     builder.addFinalizedShard(indexOfFinalized);
                 }
                 List<Integer> affectedShards = builder.getAffectedShardList();
-                if (finalizedShards.containsAll(affectedShards)) {
+                if (builder.getFinalizedShardList().containsAll(affectedShards)) {
                     CatchUpCompleted completed = completeProcess(builder.getId());
                     return EitherOf3.withA(completed);
                 }
--- a/server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
+++ b/server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
@@ -137,7 +137,6 @@ import static java.util.stream.Collectors.toSet;
  *     <li>The status if the catch-up process is set to {@link CatchUpStatus#IN_PROGRESS
  *     IN_PROGRESS}.
  * </ol>
-
  *
  * <h3>{@link CatchUpStatus#IN_PROGRESS IN_PROGRESS}</h3>
  *
--- a/server/src/main/java/io/spine/server/delivery/GroupByTargetAndDeliver.java
+++ b/server/src/main/java/io/spine/server/delivery/GroupByTargetAndDeliver.java
@@ -63,7 +63,7 @@ final class GroupByTargetAndDeliver implements DeliveryAction {
                 delivery.deliver(deliveryPackage);
             } catch (RuntimeException exception) {
                 errors.addException(exception);
-            } catch (@SuppressWarnings("ErrorNotRethrown") /* False positive */ ModelError error) {
+            } catch (@SuppressWarnings("ErrorNotRethrown") /* False-positive */ ModelError error) {
                 errors.addError(error);
             }
         }
--- a/server/src/main/java/io/spine/server/delivery/InboxColumn.java
+++ b/server/src/main/java/io/spine/server/delivery/InboxColumn.java
@@ -37,56 +37,56 @@ final class InboxColumn {
     /**
      * Stores the identifier of the signal packed into this inbox message.
      */
-    static final RecordColumn<InboxMessage, InboxSignalId> signal_id =
-            create("signal_id", InboxSignalId.class, InboxMessage::getSignalId);
+    static final RecordColumn<InboxMessage, InboxSignalId>
+            signal_id = create("signal_id", InboxSignalId.class, InboxMessage::getSignalId);
 
     /**
      * Stores the identifier of the parent inbox.
      */
-    static final RecordColumn<InboxMessage, InboxId> inbox_id =
-            create("inbox_id", InboxId.class, InboxMessage::getInboxId);
+    static final RecordColumn<InboxMessage, InboxId>
+            inbox_id = create("inbox_id", InboxId.class, InboxMessage::getInboxId);
 
     /**
      * Stores the index of the shard in which this inbox message resides.
      */
-    static final RecordColumn<InboxMessage, ShardIndex> inbox_shard =
-            create("inbox_shard", ShardIndex.class, InboxMessage::shardIndex);
+    static final RecordColumn<InboxMessage, ShardIndex>
+            inbox_shard = create("inbox_shard", ShardIndex.class, InboxMessage::shardIndex);
 
     /**
      * Stores {@code true} if this inbox message hold an event; stores {@code false} otherwise.
      */
-    static final RecordColumn<InboxMessage, Boolean> is_event =
-            create("is_event", Boolean.class, InboxMessage::hasEvent);
+    static final RecordColumn<InboxMessage, Boolean>
+            is_event = create("is_event", Boolean.class, InboxMessage::hasEvent);
 
     /**
      * Stores {@code true} if this inbox message hold a command; stores {@code false} otherwise.
      */
-    static final RecordColumn<InboxMessage, Boolean> is_command =
-            create("is_command", Boolean.class, InboxMessage::hasCommand);
+    static final RecordColumn<InboxMessage, Boolean>
+            is_command = create("is_command", Boolean.class, InboxMessage::hasCommand);
 
     /**
      * Stores the label of this inbox message.
      */
-    static final RecordColumn<InboxMessage, InboxLabel> label =
-            create("label", InboxLabel.class, InboxMessage::getLabel);
+    static final RecordColumn<InboxMessage, InboxLabel>
+            label = create("label", InboxLabel.class, InboxMessage::getLabel);
 
     /**
      * Stores the status of the delivery for this inbox message.
      */
-    static final RecordColumn<InboxMessage, InboxMessageStatus> status =
-            create("status", InboxMessageStatus.class, InboxMessage::getStatus);
+    static final RecordColumn<InboxMessage, InboxMessageStatus>
+            status = create("status", InboxMessageStatus.class, InboxMessage::getStatus);
 
     /**
      * Stores the time when the inbox message has been received and placed into the inbox.
      */
-    static final RecordColumn<InboxMessage, Timestamp> received_at =
-            create("received_at", Timestamp.class, InboxMessage::getWhenReceived);
+    static final RecordColumn<InboxMessage, Timestamp>
+            received_at = create("received_at", Timestamp.class, InboxMessage::getWhenReceived);
 
     /**
      * Stores the version of the inbox message.
      */
-    static final RecordColumn<InboxMessage, Integer> version =
-            create("version", Integer.class, InboxMessage::getVersion);
+    static final RecordColumn<InboxMessage, Integer>
+            version = create("version", Integer.class, InboxMessage::getVersion);
 
     /**
      * Prevents this type from instantiation.
--- a/server/src/main/java/io/spine/server/delivery/InboxColumn.java
+++ b/server/src/main/java/io/spine/server/delivery/InboxColumn.java
@@ -25,6 +25,8 @@ import com.google.protobuf.Timestamp;
 import io.spine.query.RecordColumn;
 import io.spine.query.RecordColumns;
 
+import static io.spine.query.RecordColumn.create;
+
 /**
  * Columns stored along with each {@link InboxMessage}.
  */
@@ -36,55 +38,55 @@ final class InboxColumn {
      * Stores the identifier of the signal packed into this inbox message.
      */
     static final RecordColumn<InboxMessage, InboxSignalId> signal_id =
-            new RecordColumn<>("signal_id", InboxSignalId.class, InboxMessage::getSignalId);
+            create("signal_id", InboxSignalId.class, InboxMessage::getSignalId);
 
     /**
      * Stores the identifier of the parent inbox.
      */
     static final RecordColumn<InboxMessage, InboxId> inbox_id =
-            new RecordColumn<>("inbox_id", InboxId.class, InboxMessage::getInboxId);
+            create("inbox_id", InboxId.class, InboxMessage::getInboxId);
 
     /**
      * Stores the index of the shard in which this inbox message resides.
      */
     static final RecordColumn<InboxMessage, ShardIndex> inbox_shard =
-            new RecordColumn<>("inbox_shard", ShardIndex.class, InboxMessage::shardIndex);
+            create("inbox_shard", ShardIndex.class, InboxMessage::shardIndex);
 
     /**
      * Stores {@code true} if this inbox message hold an event; stores {@code false} otherwise.
      */
     static final RecordColumn<InboxMessage, Boolean> is_event =
-            new RecordColumn<>("is_event", Boolean.class, InboxMessage::hasEvent);
+            create("is_event", Boolean.class, InboxMessage::hasEvent);
 
     /**
      * Stores {@code true} if this inbox message hold a command; stores {@code false} otherwise.
      */
     static final RecordColumn<InboxMessage, Boolean> is_command =
-            new RecordColumn<>("is_command", Boolean.class, InboxMessage::hasCommand);
+            create("is_command", Boolean.class, InboxMessage::hasCommand);
 
     /**
      * Stores the label of this inbox message.
      */
     static final RecordColumn<InboxMessage, InboxLabel> label =
-            new RecordColumn<>("label", InboxLabel.class, InboxMessage::getLabel);
+            create("label", InboxLabel.class, InboxMessage::getLabel);
 
     /**
      * Stores the status of the delivery for this inbox message.
      */
     static final RecordColumn<InboxMessage, InboxMessageStatus> status =
-            new RecordColumn<>("status", InboxMessageStatus.class, InboxMessage::getStatus);
+            create("status", InboxMessageStatus.class, InboxMessage::getStatus);
 
     /**
      * Stores the time when the inbox message has been received and placed into the inbox.
      */
     static final RecordColumn<InboxMessage, Timestamp> received_at =
-            new RecordColumn<>("received_at", Timestamp.class, InboxMessage::getWhenReceived);
+            create("received_at", Timestamp.class, InboxMessage::getWhenReceived);
 
     /**
      * Stores the version of the inbox message.
      */
     static final RecordColumn<InboxMessage, Integer> version =
-            new RecordColumn<>("version", Integer.class, InboxMessage::getVersion);
+            create("version", Integer.class, InboxMessage::getVersion);
 
     /**
      * Prevents this type from instantiation.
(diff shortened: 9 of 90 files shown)


client/src/main/java/io/spine/client/Filters.java
```diff
@@ -112,12 +109,12 @@ public static Filter eq(String fieldPath, Object value) {
...
      *
      * @param column
-     *         the entity column to filter by
+     *         the column to filter by
      * @param value
      *         the requested value
      * @return a new instance of {@code Filter}
      */
-    public static Filter eq(EntityColumn column, Object value) {
+    public static Filter eq(Column<?, ?> column, Object value) {
```
alexander-yevsyukov, line 117: Now, as we have the `Column` type parameterised with the type of values (`V`( as the second parameter, can we bind the type of values passed to this and similar methods below?

client/src/main/java/io/spine/client/Filters.java
```diff
@@ -565,6 +562,27 @@ public static CompositeFilter either(Filter first, Filter... rest) {
         return composeFilters(asList(first, rest), EITHER);
```
alexander-yevsyukov, line 562: A question, similar to methods accepting `Column<?, ?>`. This part of code was not modified by this PR, but the methods accepting `EntityStateField` also accept `Object` as the second parameter.

Can we have `EntityStateField` and `SubscribableField` extended (in one of the following PRs in `base`) so that they have a generic parameter with the type of the referenced field? It would allow to make query- and subscription- API more type-safe.

client/src/main/java/io/spine/client/Filters.java
```diff
@@ -591,6 +609,12 @@ static Filter createFilter(String fieldPath, Object value, Operator operator) {
         return createFilter(field, value, operator);
     }
 
+    static Filter createFilter(ColumnName columnName, Object value, Operator operator) {
+        String fieldPath = columnName.value();
+        return createFilter(fieldPath, value, operator);
+    }
+
+
     static Filter createFilter(FieldName fieldName, Object value, Operator operator) {
```
alexander-yevsyukov, line 618: Please document package-private methods of this class.

client/src/main/java/io/spine/client/QueryFilter.java
```diff
@@ -36,11 +36,11 @@
 /**
  * A query filter which targets a {@linkplain EntityColumn column} of an entity.
  */
-public final class QueryFilter extends TypedFilter<EntityState> {
+public final class QueryFilter extends TypedFilter<EntityState<?>> {
```
alexander-yevsyukov, line 39: What if we have generic parameter `V` for this class? The static factories below do accept it. Why cannot we make it a bit more save on the level of the class declaration too?

client/src/main/java/io/spine/client/TargetBuilder.java
```diff
@@ -277,6 +279,22 @@ public B where(CompositeFilter... predicate) {
...
     }
 
+    /**
+     * Performs the same action as {@link #where(CompositeFilter...)}, but acts as a shortcut
+     * overload for internal use.
+     *
+     * <p>If the passed set is empty, does nothing.
+     */
+    @CanIgnoreReturnValue
+    @Internal
```
alexander-yevsyukov, line 289: I guess we can do without the `@Internal` annotation since the method is package-private.

client/src/main/proto/spine/client/query.proto
```diff
@@ -121,8 +121,8 @@ message ResponseFormat {
     google.protobuf.FieldMask field_mask = 1;
 
     // An order for entities returned by the Query.
-    OrderBy order_by = 2 [(validate) = true,
-                          (if_invalid).msg_format = "Invalid query order"];
+    repeated OrderBy order_by = 2 [(validate) = true,
```
alexander-yevsyukov, line 124: This is an important extension of our query language. Please don't forget to include it into the release notes.

alexander-yevsyukov (changes requested):
Please see my comments. I'm at 37/582. 

Let's discuss the matter with generic parameters because it's a significant one. I'm not suggesting to address it all in this PR, but we need to address it soon.

dmdashenkov (commented):
I'm roughly 40% done. Please see the comments.

server/src/main/java/io/spine/server/aggregate/TruncateOperation.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Runs the history truncation.
+     *
+     * <p>The history records are truncated starting from the most recent one and deep into the
+     * history. The operation is performed until all the following conditions are true:
+     *
+     * <ul>
+     *     <li>the number of aggregate snapshots among the deleted history records is less than
+     *     a given number passed as {@code snapshotIndex};</li>
```
alexander-yevsyukov, line 62: If memory serves, `<li>` does not need to have the closing tag in Javadoc.

server/src/main/java/io/spine/server/aggregate/UncommittedHistory.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.List;
+import java.util.function.Supplier;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Uncommitted events and snapshots created for this aggregate during the dispatching.
+ *
+ * <p>Watches the how the events {@linkplain Aggregate#invokeApplier(EventEnvelope) are sent}
```
alexander-yevsyukov, line 38: I think “the” before “how” is not needed.

server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Stores the status of the catch-up process.
+     */
+    public static final RecordColumn<CatchUp, CatchUpStatus> status =
+            new RecordColumn<>("status", CatchUpStatus.class, CatchUp::getStatus);
+
+    /**
+     * Stores the time when the history has been last read by the catch-up process.
+     */
+    public static final RecordColumn<CatchUp, Timestamp> when_last_read =
```
alexander-yevsyukov, line 42: We probably need to have an annotation for column fields in Java classes to explain why they are named following Protobuf, rather than Java, conventions.

server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Stores the type URL of the projection-under-catch-up.
+     */
+    public static final RecordColumn<CatchUp, String> projection_type =
+            new RecordColumn<>("projection_type",
+                               String.class, (m) -> m.getId()
+                                                     .getProjectionType());
+
+    private CatchUpColumn() {
```
alexander-yevsyukov, line 53: Similarly to annotating fields, please consider annotating such custom column classes with a documentation-only annotation. It would allow people viewing the code understand a bigger story behind such a class.

alexander-yevsyukov: The documentation for the class annotation can tell that such a class should have a private constructor and a static method that returns `ImmutableList` of columns.

alexander-yevsyukov (changes requested):
Please see my questions and comments (74/582)

server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
```diff
@@ -374,33 +464,107 @@ CatchUpStarted handle(CatchUpRequested e, EventContext ctx) {
...
+        CatchUpId id = builder().getId();
+        Optional<CatchUp> stateVisibleToDelivery = findJob(id, event.getRunInfo());
+        if (stateVisibleToDelivery.isPresent()) {
+            CatchUp observedJob = stateVisibleToDelivery.get();
+            if (observedJob.getStatus() == FINALIZING) {
+                int indexOfFinalized = event.getIndex()
+                                            .getIndex();
+                if (!builder().getFinalizedShardList()
+                              .contains(indexOfFinalized)) {
+                    builder().addFinalizedShard(indexOfFinalized);
```
alexander-yevsyukov, line 543: Please extract `builder()` into a variable.

dmdashenkov (changes requested):
@armiol, what's the proposed way to migrate the `Mirror`s to the new Aggregate storage?

server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
```diff
@@ -374,33 +464,107 @@ CatchUpStarted handle(CatchUpRequested e, EventContext ctx) {
...
+                    return EitherOf3.withA(completed);
+                }
+                return EitherOf3.withC(nothing());
+            }
+        }
+        ShardProcessingRequested stillRequested = shardProcessingRequested(id, event.getIndex());
+        return EitherOf3.withB(stillRequested);
+    }
+
+
```
alexander-yevsyukov, line 558: Please remove extra empty lines.

server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
```diff
@@ -374,33 +464,107 @@ CatchUpStarted handle(CatchUpRequested e, EventContext ctx) {
...
+
+    private static Optional<CatchUp> findJob(CatchUpId id, DeliveryRunInfo deliveryInfo) {
+        Optional<CatchUp> stateVisibileToDelivery =
+                deliveryInfo.getCatchUpJobList()
+                               .stream()
+                               .filter((job) -> job.getId()
+                                                   .equals(id))
```
alexander-yevsyukov, line 565: Consider flipping `equals()`. It would make it one-line instead of two.

alexander-yevsyukov (changes requested):
Please see my comments (75/582)

server/src/main/java/io/spine/server/storage/memory/TenantDataStorage.java
```diff
@@ -29,7 +29,7 @@
  * @param <I> the type of entity IDs
  * @param <R> the type of stored records
  */
-interface TenantStorage<I, R> {
+interface TenantDataStorage<I, R> {
```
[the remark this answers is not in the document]
armiol: @dmdashenkov the thing is that this _is_ a storage for any tenant data. If you have any ideas on renaming, please be my guest :)

server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
```diff
@@ -105,7 +113,7 @@
  *
  * <p><b>{@linkplain CatchUpStatus#CUS_UNDEFINED Not started}</b>
```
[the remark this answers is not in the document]
alexander-yevsyukov: As discussed vocally, the piece about the “turbulence” belongs to the introduction part, and does not need a separate section.

server/src/main/java/io/spine/server/delivery/InboxColumn.java
```diff
@@ -0,0 +1,97 @@
...
+    static final RecordColumn<InboxMessage, Integer> version =
+            new RecordColumn<>("version", Integer.class, InboxMessage::getVersion);
+
+    private InboxColumn() {
+    }
+
+    /**
+     * Returns all the column definitions.
+     */
+    static ImmutableList<RecordColumn<InboxMessage, ?>> definitions() {
```
alexander-yevsyukov, line 93: What if we have `RecordColumns<InboxMessage>` that would return `ImmutableList` when needed?

alexander-yevsyukov (changes requested):
Please see my comments. This bit suggest how we can arrange a common interface for column definitions. Have a look. (85/582)

server/src/main/java/io/spine/server/aggregate/Aggregate.java
```diff
@@ -392,15 +420,33 @@ protected final String missingTxMessage() {
...
+     *
      * @return new snapshot
      */
     final Snapshot toSnapshot() {
-        Any state = AnyPacker.pack(state());
+        S state;
+        Version version;
+        LifecycleFlags lifecycle;
+        if (isTransactionInProgress()) {
+            AggregateTransaction tx = (AggregateTransaction) tx();
```
alexander-yevsyukov, line 433: Please use generic parameters having: 

```java
            AggregateTransaction<?, ?, ?> tx = (AggregateTransaction<?, ?, ?>) tx();
```
It would eliminate the IDEA's warning on generics.

server/src/main/java/io/spine/server/entity/storage/AsEntityRecordColumn.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.errorprone.annotations.Immutable;
+import io.spine.query.Column;
+import io.spine.query.RecordColumn;
+import io.spine.server.entity.EntityRecord;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.util.Exceptions.newIllegalStateException;
+
+/**
+ * Utility which creates views on the {@linkplain Column columns}.
```
alexander-yevsyukov, line 32: Which columns? All, or some specific ones?

alexander-yevsyukov (changes requested):
Please see my comments and questions (122/582).

server/src/main/java/io/spine/server/entity/storage/EntityRecordStorage.java
```diff
@@ -0,0 +1,339 @@
...
+            }
+            if(hasDeletedColumn) {
+                builder.where(deleted.asRecordColumn(Boolean.class)).is(false);
+            }
+            result = builder.build();
+        }
+        return result;
+    }
+
+    private static boolean isLifecycleColumn(Column<?, ?> column) {
```
alexander-yevsyukov, line 324: I think we should consider having `LifecycleColumns` or something like that. Yes, we have only two of them, but just in case we need yet another one in the future, we'd need to gather the knowledge of such columns in may places in the code. It would be good to consolidate it under a class.

For example, `EntityRecordWithColumns.create(I id, EntityRecord record)` deals with two columns placing them into a `Map`. This class has at least two places where it checks for `archived` and then `deleted` columns, see constructor and `onlyActive()`. Can we consolidate this knowledge under one class?

server/src/main/java/io/spine/server/entity/storage/EntityRecordWithColumns.java
```diff
@@ -21,188 +21,141 @@
...
+     *         the type of the identiiers
+     * @return a new instance of {@code EntityRecordWithColumns}
      */
-    public ImmutableSet<ColumnName> columnNames() {
-        return ImmutableSet.copyOf(storageFields.keySet());
+    public static <I> EntityRecordWithColumns<I> create(I id, EntityRecord record) {
+        checkNotNull(id);
+        checkNotNull(record);
+        LifecycleFlags flags = record.getLifecycleFlags();
+        ImmutableMap<ColumnName, Object> lifecycleValues =
```
alexander-yevsyukov, line 97: In relation to this fragment, please see my earlier comment in `EntityStorageRecord` regarding the chances of consolidating the knowledge about lifecycle columns under a class.

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+    @VisibleForTesting
+    public static <I, R extends Message>
+    RecordWithColumns<I, R> of(I identifier, R record, Map<ColumnName, Object> storageFields) {
+        return new RecordWithColumns<>(identifier, record, storageFields);
+    }
+
+    /**
+     * Returns the identifier of the record.
+     */
+    public I id() {
```
alexander-yevsyukov, line 113: Can we make it `final`?

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+        return id;
+    }
+
+    /**
+     * Returns the message of the record.
+     */
+    public R record() {
```
alexander-yevsyukov, line 120: Can we make it `final`?

server/src/main/java/io/spine/server/entity/storage/EntityRecordWithColumns.java
```diff
@@ -21,188 +21,141 @@
...
+        return record().isDeleted();
     }
 
     @Override
     public boolean isActive() {
-        return record.isActive();
-    }
-
-    @Override
-    public boolean equals(Object o) {
```
alexander-yevsyukov, line 191: Don't we need `hashCode()` and `equals()` for this class? `RecordWithColumns` is a bit restrictive on `equals()`.

server/src/main/java/io/spine/server/entity/storage/FindActiveEntites.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param <I>
+     *         the type of the queried entity identifiers
+     * @param <S>
+     *         the type of the entity states
+     */
+    static final class Builder<I, S extends EntityState<I>>
+            extends EntityQueryBuilder<I, S, Builder<I, S>, FindActiveEntites<I, S>> {
+
+        private final boolean hasArchivedColumn;
```
alexander-yevsyukov, line 66: Once again we have a pair of fields that tells the same thing about having this pair of columns. We had it in `EntityRecordStorage`. Perhaps, we can consolidate this knowledge. Maybe not in this PR, but later?

alexander-yevsyukov (changes requested):
Please see my comments and questions (139/583).

server/src/main/java/io/spine/server/entity/storage/EntityRecordWithColumns.java
```diff
@@ -21,188 +21,141 @@
...
+import static io.spine.server.entity.storage.EntityRecordColumn.deleted;
+import static java.util.Collections.emptyMap;
 
 /**
- * A value of {@link EntityRecord} associated with the values of its {@linkplain Column columns}.
+ * A value of {@link EntityRecord} associated with the values
+ * of its {@linkplain io.spine.query.Column columns}.
  */
-public final class EntityRecordWithColumns implements WithLifecycle {
+@Internal
```
alexander-yevsyukov, line 47: Please add `@apiNote` on why this class is internal.

alexander-yevsyukov: Or, maybe it's `@SPI`?

server/src/main/java/io/spine/server/entity/storage/FindActiveEntites.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Creates a new builder for {@code FindActiveEntites}.
+     *
+     * @param <I>
+     *         the type of identifiers of the queries entities
+     * @param <S>
+     *         the type of entity states
+     * @return a new builder instance
+     */
+    public static <I, S extends EntityState<I>> Builder<I, S>
```
alexander-yevsyukov, line 49: Why do we want it `public` in a package-private class?

server/src/main/java/io/spine/server/entity/storage/FindActiveEntites.java
```diff
@@ -0,0 +1,99 @@
...
+    newBuilder(Class<I> idType, Class<S> stateType,
+               boolean hasArchivedColumn, boolean hasDeletedColumn) {
```
alexander-yevsyukov, line 51: Again, we have a knowledge of two columns passed as two boolean parameters. Earlier, they were fields in a class. It looks like we're missing some abstraction.

alexander-yevsyukov (changes requested):
Please see my comments and questions (159/583).

server/src/main/java/io/spine/server/storage/MessageRecordSpec.java
```diff
@@ -0,0 +1,109 @@
...
+     * A method object to extract the record identifier, once such a record is passed.
+     */
+    private final ExtractId<R, I> extractId;
+
+    /**
+     * The columns to store along with the record itself.
+     */
+    private final ImmutableMap<ColumnName, RecordColumn<R, ?>> columns;
+
+    public MessageRecordSpec(Class<I> idType,
```
alexander-yevsyukov, line 70: Please document.

server/src/main/java/io/spine/server/storage/MessageRecordSpec.java
```diff
@@ -0,0 +1,109 @@
...
+                             Class<R> recordType,
+                             ExtractId<R, I> extractId,
+                             Iterable<RecordColumn<R, ?>> columns) {
+        super(idType, recordType);
+        this.columns = stream(columns).collect(toImmutableMap(RecordColumn::name, (c) -> c));
+        this.extractId = extractId;
+    }
+
+    public MessageRecordSpec(Class<I> idType, Class<R> recordType, ExtractId<R, I> extractId) {
```
alexander-yevsyukov, line 79: Please document.

server/src/main/java/io/spine/server/storage/MessageRecordSpec.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public Optional<Column<?, ?>> findColumn(ColumnName name) {
+        RecordColumn<R, ?> result = columns.get(name);
+        return Optional.ofNullable(result);
+    }
+
+    @Immutable
+    @FunctionalInterface
+    public interface ExtractId<R extends Message, I> extends Function<R, I> {
```
alexander-yevsyukov, line 107: Please document.

server/src/main/java/io/spine/server/storage/MessageStorage.java
```diff
@@ -0,0 +1,137 @@
...
+     * <p>The identifier and the columns for the written records are extracted via
+     * the {@linkplain #recordSpec() record specification}.
+     *
+     * @param messages
+     *         the batch of the messages to write
+     * @throws IllegalStateException
+     *         if the storage is already closed
+     */
+    protected void writeBatch(Iterable<M> messages) {
+        List<RecordWithColumns<I, M>> records = stream(messages).map(this::toRecord)
```
alexander-yevsyukov, line 113: Please consider moving `stream()` to the next line.

server/src/main/java/io/spine/server/storage/QueryConverter.java
```diff
@@ -0,0 +1,309 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+import static io.spine.util.Exceptions.newIllegalStateException;
+import static java.lang.String.join;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Converts the queries defined in Protobuf into the language of {@code io.spine.query} package.
+ */
+@Internal
```
alexander-yevsyukov, line 55: Please document why this class is `@Internal`. Can it be `@SPI`?

server/src/main/java/io/spine/server/storage/QueryConverter.java
```diff
@@ -0,0 +1,309 @@
...
+        Class<R> recordType = spec.recordType();
+        RecordQueryBuilder<I, R> builder = RecordQuery.newBuilder(idType, recordType);
+        fieldMask(builder, format);
+        orderByAndLimit(builder, spec, format);
+        return builder.build();
+    }
+
+    private static <I, R extends Message> void
+    filters(RecordQueryBuilder<I, R> builder, RecordSpec<I, R, ?> spec, TargetFilters filters) {
+        if (filters.getFilterCount() > 0) {
```
alexander-yevsyukov, line 125: Why do we want to check the count? The generated code uses `List` instance anyway. What if we just inline the `compositeFilters` variable and remove the `if` check? It would make the whole this method not that much tested.

If you wanted to make it more obvious that nothing happens if the list is empty, why not just do it at the beginning of the method and return right away if so?

server/src/main/java/io/spine/server/storage/QueryConverter.java
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * Converts the passed composite filter into a set of conditions and adds them to the passed
+     * builder according to the record specification.
+     */
+    @CanIgnoreReturnValue
+    private static <I, R extends Message> RecordQueryBuilder<I, R>
+    convert(CompositeFilter filter, RecordSpec<I, R, ?> spec, RecordQueryBuilder<I, R> builder) {
+        if (filter.getFilterCount() > 0) {
+            List<Filter> childFilters = filter.getFilterList();
+            for (Filter childFilter : childFilters) {
```
alexander-yevsyukov, line 157: Again, please consider inlining the `childFilter` and removing `if` check to make this code less nested.

alexander-yevsyukov (changes requested):
Please see my comments and questions (213/583).

server/src/main/java/io/spine/server/storage/RecordStorageDelegate.java
```diff
@@ -0,0 +1,224 @@
...
+        return delegate.isMultitenant();
+    }
+
+    @Override
+    protected void checkNotClosed(String message) throws IllegalStateException {
+        delegate.checkNotClosed(message);
+    }
+
+    @Override
+    protected void checkNotClosed() throws IllegalStateException {
```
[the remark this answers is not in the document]
alexander-yevsyukov: Please also consider my comments made in `AbstractStorage`. I guess I was in a rush when defined that class.

server/src/main/java/io/spine/server/entity/storage/EntityRecordStorage.java
```diff
@@ -0,0 +1,339 @@
...
+        this.findActiveRecordsQuery = transform(entityQuery);
+    }
+
+    private static <I, S extends EntityState<I>> EntityRecordSpec<I, S, ?>
+    spec(Class<? extends Entity<I, S>> entityClass) {
+        return EntityRecordSpec.of(entityClass);
+    }
+
+    /**
+     * Returns the iterator over all non-archived and non-deleted entity records.
```
alexander-yevsyukov, line 111: How avoid telling that it returns `the iterator over the identifiers of all non-archived and non-deleted entity records`?

server/src/main/java/io/spine/server/entity/storage/EntityRecordStorage.java
```diff
@@ -0,0 +1,339 @@
...
+     *
+     * @throws IllegalStateException
+     *         if the storage is already closed
+     */
+    @Override
+    public Iterator<I> index() {
+        Iterator<EntityRecord> iterator = super.readAll(findActiveRecordsQuery);
```
alexander-yevsyukov, line 118: This is definitely not for this PR, but we need to make it more efficient. It looks like we're reading everything just to get all the identifiers.

alexander-yevsyukov (commented):
Please see my comments (216/583).

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+    private final I id;
+    private final R record;
+
+    /**
+     * A map of column names to the corresponding column values.
+     *
+     * @implNote It's impossible to use an {@link com.google.common.collect.ImmutableMap
+     *         ImmutableMap}, as the values may contain {@code null}s.
+     */
+    private final Map<ColumnName, @Nullable Object> storageFields;
```
alexander-yevsyukov, line 58: We don't seem to update this map. Can we make it immutable?

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+        return record;
+    }
+
+    /**
+     * Obtains the names of storage fields in the record.
+     *
+     * @return the storage field names
+     */
+    public ImmutableSet<ColumnName> columnNames() {
+        return ImmutableSet.copyOf(storageFields.keySet());
```
alexander-yevsyukov, line 130: Assuming, `storageFields` can be made immutable, its keys are going to be immutable too.

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+    public R record() {
+        return record;
+    }
+
+    /**
+     * Obtains the names of storage fields in the record.
+     *
+     * @return the storage field names
+     */
+    public ImmutableSet<ColumnName> columnNames() {
```
alexander-yevsyukov, line 129: Can we make this method `final`?

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+     * {@link #columnValue(ColumnName, ColumnMapping)} overload for convenient column value
+     * conversion.
+     *
+     * @param columnName
+     *         the column name
+     * @return the storage field value
+     * @throws IllegalStateException
+     *         if there is no column with the specified name
+     */
+    public @Nullable Object columnValue(ColumnName columnName) {
```
alexander-yevsyukov, line 150: Can we make this method `final`?

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+        return columnValue(columnName, DefaultColumnMapping.INSTANCE);
+    }
+
+    /**
+     * Obtains the value of the storage field by the specified column name.
+     *
+     * <p>The specified column mapping will be used to do the column value conversion.
+     */
+    public <V> V columnValue(ColumnName columnName, ColumnMapping<V> columnMapping) {
```
alexander-yevsyukov, line 159: Can we make this method `final`?

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+        V result = columnMapping.of(columnValue.getClass())
+                                .applyTo(columnValue);
+        return result;
+    }
+
+    /**
+     * Tells if there are any {@linkplain io.spine.query.Column columns}
+     * associated with this record.
+     */
+    public boolean hasColumns() {
```
alexander-yevsyukov, line 181: Can we make this method `final`?

server/src/main/java/io/spine/server/storage/RecordWithColumns.java
```diff
@@ -0,0 +1,211 @@
...
+        return !storageFields.isEmpty();
+    }
+
+    /**
+     * Determines if there is a column with the specified name among the storage fields.
+     */
+    public boolean hasColumn(ColumnName name) {
```
alexander-yevsyukov, line 188: Can we make this method `final`?

server/src/main/java/io/spine/server/storage/memory/RecordQueryMatcher.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Matches the records to the {@linkplain RecordQuery#subject() subject} of a {@link RecordQuery}.
+ *
+ * @param <I>
+ *         the type of the identifiers of the records
+ * @param <R>
+ *         the type of the messages stored as records
+ */
+public class RecordQueryMatcher<I, R extends Message>
```
alexander-yevsyukov, line 49: Can we make this class `final`?

server/src/main/java/io/spine/server/storage/memory/TenantRecords.java
```diff
@@ -39,119 +42,107 @@
...
 
 /**
- * The memory-based storage for {@link EntityRecord} that represents
- * all storage operations available for data of a single tenant.
+ * The memory-based storage for message records.
+ *
+ * <p>Acts like a facade API for the operations available over the data of a single tenant.
  */
-final class TenantRecords<I> implements TenantStorage<I, EntityRecordWithColumns> {
+class TenantRecords<I, R extends Message> implements TenantDataStorage<I, RecordWithColumns<I, R>> {
```
alexander-yevsyukov, line 57: It's no longer `final`, but I don't seem to find derived classes. Please check.

server/src/main/java/io/spine/server/storage/memory/TenantRecords.java
```diff
@@ -161,16 +152,29 @@ public boolean isEmpty() {
...
 
         private FieldMaskApplier(FieldMask fieldMask) {
             this.fieldMask = fieldMask;
         }
 
+        @SuppressWarnings("unchecked")
         @Override
-        public @Nullable EntityRecord apply(@Nullable EntityRecord input) {
+        public @Nullable R apply(@Nullable R input) {
+            checkNotNull(input);
```
alexander-yevsyukov, line 166: This doesn't go well together with `@Nullable` for the parameter.

server/src/main/java/io/spine/server/storage/memory/TenantRecords.java
```diff
@@ -161,16 +152,29 @@ public boolean isEmpty() {
...
+                return input;
+            }
+            if (input instanceof EntityRecord) {
+                return (R) maskEntityRecord((EntityRecord) input);
+            }
+            return applyMask(fieldMask, input);
+        }
+
+        private EntityRecord maskEntityRecord(@Nullable EntityRecord input) {
             checkNotNull(input);
```
alexander-yevsyukov, line 178: This doesn't go well together with `@Nullable` for the parameter.

alexander-yevsyukov (changes requested):
Please see my comments (237/583)

server/src/main/java/io/spine/server/tuple/EitherOf5.java
```diff
@@ -111,48 +111,109 @@ private EitherOf5(Message value, int index) {
...
     }
+
+    /**
+     * Tells whether {@code <E>} value is stored.
+     *
+     * @return {@code true} if the fifth alternative value is set, {@code false} otherwise
+     */
+    @Override
+    public boolean hasE() {
+        return index() == 4;
```
alexander-yevsyukov, line 217: Now, I think we need a class with constants shared across the package. Please consider. We have too many magic constants that are obvious, but still are “magic constants”. Can we have something like `IndexOf.E` for example?

server/src/main/proto/spine/server/aggregate/aggregate.proto
```diff
@@ -53,6 +55,14 @@ message Snapshot {
     // This field is used when storing a snapshot in `AggregateEventRecord`.
     //
     google.protobuf.Timestamp timestamp = 3;
+
+    // The lifecycle flags of the aggregate at the moment of the creation of this snapshot.
+    spine.server.entity.LifecycleFlags lifecycle = 4;
+}
+
+message AggregateEventRecordId {
```
alexander-yevsyukov, line 63: Please document.

server/src/main/proto/spine/server/aggregate/aggregate.proto
```diff
@@ -61,6 +71,8 @@ message Snapshot {
 //
 message AggregateEventRecord {
 
+    AggregateEventRecordId id = 4;
```
alexander-yevsyukov, line 74: Is it `(required)`? Assuming, this is not a state of an entity, the entity-related convention for identifiers does not apply here.

server/src/main/proto/spine/server/aggregate/aggregate.proto
```diff
@@ -71,6 +83,8 @@ message AggregateEventRecord {
         // A snapshot of the aggregate.
         server.aggregate.Snapshot snapshot = 3;
     }
+
+    google.protobuf.Any aggregate_id = 5;
```
alexander-yevsyukov, line 87: Please document. Also, in what cases we don't have a value here?

server/src/main/proto/spine/server/aggregate/aggregate.proto
```diff
@@ -95,3 +109,27 @@ message AggregateHistory {
     repeated core.Event event = 2;
 }
 
+message AggregateStateRecord {
```
alexander-yevsyukov, line 112: Please document.

server/src/main/proto/spine/server/catchup/catch_up.proto
```diff
@@ -75,12 +83,15 @@ enum CatchUpStatus {
     // Not started.
     CUS_UNDEFINED = 0;
 
+    // The process has been started.
+    STARTED = 1;
```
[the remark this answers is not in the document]
alexander-yevsyukov: I don't think we need to be _that_ strict here. Catch-up processes do not span across versions of the framework.

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -87,3 +88,16 @@ message ShardMaintenance {
...
+// to dispatch messages.
+//
+// The system-wide processes, such as `CatchUp` need to know the details on how the messages
+// were dispatched by `Delivery`. Therefore, `Delivery` records some of its parameters
+// when it is delivering the signals, and passes this data on to all interested parties.
+//
+message DeliveryRunInfo {
+
+    // The catch-up jobs observed by the Delivery.
+    repeated spine.server.catchup.CatchUp catch_up_job = 1;
```
alexander-yevsyukov, line 102: Why do we need a full package name here?

server/src/main/proto/spine/server/delivery/delivery_events.proto
```diff
@@ -12,10 +12,35 @@ option java_package = "io.spine.server.delivery.event";
...
-    ShardIndex id = 1;
+    ShardIndex index = 1 [(required) = true, (validate) = true];
+
+    // The run-time data of the delivery at the moment when this event has been dispatched.
+    DeliveryRunInfo run_info = 2;
+
+    // An identifier of a `CatchUp` process in scope of which the shard processing
+    // was originally requested.
+    //
+    spine.server.catchup.CatchUpId process = 3 [(required) = true];
```
alexander-yevsyukov, line 31: Please consider shortening the package reference here and below.

alexander-yevsyukov (changes requested):
Please see my comments and questions (280/583)

server/src/test/java/io/spine/server/route/given/user/UserRepository.java
```diff
@@ -18,23 +18,22 @@
...
-        messageId = id;
-    }
+public final class UserRepository extends AggregateRepository<UserId, UserAggregate, RUser> {
 
     @Override
-    public InboxMessageId recordId() {
-        return messageId;
+    protected void setupEventRouting(EventRouting<UserId> routing) {
+        super.setupEventRouting(routing);
+        routing.route(RUserSignedIn.class, (e, ctx) -> ImmutableSet.of(e.getUserId()));
```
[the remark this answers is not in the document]
alexander-yevsyukov: This made me also have yet another look at the event routing. Here's [the result](https://github.com/SpineEventEngine/core-java/pull/1317) :) I'm not suggesting to migrate to it in this PR. 
But we finally have the unicast routing simplified!

server/src/test/java/io/spine/server/stand/given/StandTestEnv.java
```diff
@@ -51,14 +109,213 @@ public static Stand newStand(boolean multitenant) {
...
+    }
+
+    public static int storeSampleProject(StandTestProjectionRepository repository) {
+        ProjectId id = projectIdFor(42);
+        int projectVersion = 42;
+        storeSampleProject(repository, id, "Some sample project", projectVersion);
+        return projectVersion;
+    }
+
+    public static void storeSampleProject(StandTestProjectionRepository repository,
```
alexander-yevsyukov, line 149: Please align the parameters.

server/src/test/java/io/spine/server/storage/EntityRecordStorageTest.java
```diff
@@ -0,0 +1,603 @@
...
+    protected EntityRecordStorage<StgProjectId, StgProject> newStorage() {
+        StorageFactory factory = ServerEnvironment.instance()
+                                                  .storageFactory();
+        ContextSpec ctxSpec = ContextSpec.singleTenant(EntityRecordStorageTest.class.getName());
+        return new EntityRecordStorage<>(ctxSpec, factory, TestCounterEntity.class);
+    }
+
+    @Override
+    protected EntityRecord newStorageRecord(StgProjectId id) {
+        return EntityRecordStorageTestEnv.newStorageRecord(id, newState(id));
```
alexander-yevsyukov, line 107: Can we static-import this call, please?

server/src/test/java/io/spine/server/storage/EntityRecordStorageTest.java
```diff
@@ -0,0 +1,603 @@
...
+            EntityRecord entityRecord = optional.get();
+
+            Message unpacked = unpack(entityRecord.getState());
+            assertFalse(isDefault(unpacked));
+        }
+
+        @SuppressWarnings("MethodWithMultipleLoops")
+        @Test
+        @DisplayName("multiple records according to the given field mask")
+        void multipleRecords() {
```
alexander-yevsyukov, line 195: Please consider refactoring this method into a nested class. It's way too complicated.

server/src/test/java/io/spine/server/storage/EntityRecordStorageTest.java
```diff
@@ -0,0 +1,603 @@
...
+
+            archive(archivedEntity);
+
+            EntityRecordStorage<StgProjectId, StgProject> storage = storage();
+
+            storage.write(create(activeEntity, activeRecord));
+            storage.write(create(archivedEntity, archivedRecord));
+
+            StgProject.Query query =
+                    StgProject.query()
```
alexander-yevsyukov, line 246: Please fix the alignment.

server/src/test/java/io/spine/server/storage/EntityRecordStorageTest.java
```diff
@@ -0,0 +1,603 @@
...
+                                               .where(archived.lifecycle(), true)
+                                               .build();
+            assertQueryHasSingleResult(query, archivedRecord, storage);
+        }
+    }
+
+    @Nested
+    @DisplayName("filter records")
+    class Filtering {
```
alexander-yevsyukov, line 255: Please consider splitting this long test suite into several ones that would focus on reading, filtering, etc.

server/src/test/java/io/spine/server/storage/RecordStorageDelegateTest.java
```diff
@@ -0,0 +1,477 @@
...
+/**
+ * Tests of the API provided by {@link RecordStorageDelegate} to the descendant classes.
+ *
+ * <p>Sample storage implementation used in the test is {@link StgProjectStorage}.
+ *
+ * <p>The aim of this test is to ensure that any storage implementations built on top of
+ * the {@code RecordStorageDelegate} is able to utilize the API with the expected results.
+ */
+@DisplayName("A `RecordStorageDelegate` descendant should")
+public class RecordStorageDelegateTest
```
alexander-yevsyukov, line 76: I don't seem to find descendants of this class. Do we need it `public`?

server/src/test/java/io/spine/server/storage/RecordStorageDelegateTest.java
```diff
@@ -0,0 +1,477 @@
...
+        void delete() {
+            assertThrows(IllegalStateException.class,
+                         () -> storage().delete(newId())
+            );
+        }
+
+        @Test
+        @DisplayName("`deleteAll(IDs)` method")
+        void deleteAll() {
+            assertThrows(IllegalStateException.class,
```
alexander-yevsyukov, line 464: Please extract asserting of throwing ISE into a method.

alexander-yevsyukov (changes requested):
Please see my comments and questions (486/583)

testutil-server/src/main/java/io/spine/testing/server/blackbox/SingleTenantContext.java
```diff
@@ -33,7 +33,7 @@
 /**
  * Test fixture for single-tenant Bounded Contexts.
  */
-final class SingleTenantContext extends BlackBoxContext {
+final class SingleTenantContext extends BlackBox {
```
alexander-yevsyukov, line 36: Please consider renaming this class to `SingleTenantBlackBox` or `StBlackBox`. It could be also `BlackBoxSt` as well. This way the classes would go together in alpha sorting.

testutil-server/src/main/java/io/spine/testing/server/blackbox/SingleTenantContext.java
```diff
@@ -50,7 +50,7 @@
     }
 
     @Override
-    public BlackBoxContext withTenant(TenantId tenant) {
+    public BlackBox withTenant(TenantId tenant) {
```
alexander-yevsyukov, line 53: Can this method be package-private?

alexander-yevsyukov (changes requested):
Please see remaining comments and suggestions. I'm done (583/583). What a journey!

Kudos to @armiol for this enormous amount of work done alone. Let's polish it and get into life soon. This PR improves the codebase a lot.

@armiol, if you need some assistance in addressing the comments, I'm willing to help.

server/src/main/java/io/spine/server/aggregate/UncommittedHistory.java
```diff
@@ -0,0 +1,206 @@
...
+ *         instances from the storage
+ */
+final class UncommittedHistory {
+
+    private final Supplier<Snapshot> makeSnapshot;
+    private final List<AggregateHistory> historySegments = new ArrayList<>();
+    private final List<Event> currentSegment = new ArrayList<>();
+
+    private int eventCountAfterLastSnapshot;
+    private @Nullable Integer snapshotTrigger = null;
```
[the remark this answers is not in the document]
armiol: The reason is that the `UncommittedHistory` is a part of an `Aggregate` instance. But the value of a snapshot trigger is a knowledge outer to the `Aggregate`. It is brought by a corresponding repository at a time of applying the events inside the `AggregateEndpoint`. In turn, the repository also has this value mutable, so that end-users are able to re-define it.

Therefore it's not possible to instantiate an Aggregate instance—and thus its `UncommittedHistory`—with the snapshot trigger set right away.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -48,28 +48,44 @@
...
+ *
+ * <p>If {@code Staging} is {@link Environment#type() enabled}, the specified value is going to be
  * returned on {@link #storageFactory()}.
+ *
+ * <h3>Relation to Bounded Contexts</h3>
+ *
+ * <p>A {@code ServerEnvironment} instance is a singleton. It means that all Bounded Contexts
+ * which are deployed within the same JVM share the configuration of the server environment.
+ * Namely, under the same environment type, all Bounded Contexts will share the same storage
+ * factory, transport factory, delivery and tracing tools.
```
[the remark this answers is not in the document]
armiol: > Now it seems that one has to have separate deployable components for two Bounded Contexts if they use different databases.

This _is_ the intention. We strongly recommend having different deployable artifacts in this case. See the paragraph next to one you have commented on, it explicitly tells to do so.

server/src/main/java/io/spine/server/delivery/GroupByTargetAndDeliver.java
```diff
@@ -58,12 +56,11 @@
...
     @Override
     public DeliveryErrors executeFor(List<InboxMessage> messages) {
-        Map<String, List<InboxMessage>> messagesByType = groupByTargetType(messages);
-
+        List<Segment> segments = groupByTargetType(messages);
         DeliveryErrors.Builder errors = DeliveryErrors.newBuilder();
-        for (String typeUrl : messagesByType.keySet()) {
-            ShardedMessageDelivery<InboxMessage> delivery = inboxDeliveries.get(typeUrl);
-            List<InboxMessage> deliveryPackage = messagesByType.get(typeUrl);
+        for (Segment segment : segments) {
```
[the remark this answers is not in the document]
armiol: > Can we run this in parallel (i.e. via segments.parallelStream()?

Unfortunately, no. As we must deliver the messages exactly in the order, in which they are listed. 

Different segments are produced wisely, so that across the list of segments all the messages are ordered by their time-of-emission. In this way we guarantee that two signals will be dispatched — no matter to which destination — exactly in the order they were emitted. 

Also, different entities to which the messages are dispatched, may read one another. So even from this point of view, we must not update one entity earlier than it is expected to be updated, since we may break the end-user logic.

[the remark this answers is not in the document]
armiol: I have extracted and improved some bits here and there, including switching to `ImmutableList` when producing segments.

However, the `Segment`-grouping code is hardly simplify-able (I tried!) with the same amount of data structures involved. And we need to keep up the performance here, as the delivery should be as fast as possible. So I refrained from modifying it as such.

armiol:
@alexander-yevsyukov @dmdashenkov The vast majority of the comments have been addressed.  

I have created a couple of issues to address the remaining comments in the upcoming PRs. The list of these issues follows.

* #1324 [2.x] Create entity-specific descendants of `EntityLifecycle`.
* #1325 [2.x] Count all stored `Entity` instances as those having `archived`, `deleted` and `version` system columns.
* #1326 [2.x] Implement a migration assistant for legacy data that used `Mirror`.

This pull request becomes hardly maintainable due to a huge number of commits and comments. I am seeing the GitHub unicorns more and more often. Therefore, I indent to merge it down and continue working according to this simple plan:

— Back-merge from 1.x.
— Deal with the remaining comments/issues one-by-one.

Please have a look at my changes since your last review. Please feel free to comment. I will create new issues to address in 2.x branch if something new comes up.

armiol:
@alexander-yevsyukov @dmdashenkov FYI. One more issue to work in the following PRs:

* #1328 [2.x] Sharpen the contract of `QueryableRepository`.

dmdashenkov (commented):
1/3 way there. Please see my comment. Also, there are at least 2 empty files: `CommandOutputBus.java` and `OutputDispatcherRegistry`. Please delete them.

alexander-yevsyukov (approved):
LTGM, with few minor requests.

server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
```diff
@@ -0,0 +1,72 @@
...
+     * expected to be instantiated at all. See the {@link RecordColumns} docs for more details on
+     * this approach.
+     */
+    private CatchUpColumn() {
+    }
+
+    /**
+     * Returns all the column definitions.
+     */
+    public static ImmutableList<RecordColumn<CatchUp, ?>> definitions() {
```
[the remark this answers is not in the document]
alexander-yevsyukov: I understand that we have `RecordColumns` annotation already. Let's review this matter once again a bit later. Perhaps, we'd be able to find an approach that would make it look a bit more compact.

Perhaps we can have `@Columns(ofType=Event.class, record=true)` or something so that we can have `RecordColumns` type. Or, something like that.

armiol:
@dmdashenkov see #1329. 
Also, on the empty files — I killed them sometime in April, I think. They are [still in the `master` branch though](https://github.com/SpineEventEngine/core-java/tree/master/server/src/main/java/io/spine/server/outbus). Not sure why GitHub displays them this way in this changeset.

server/src/main/java/io/spine/server/delivery/CatchUpColumn.java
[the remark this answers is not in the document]
armiol: @alexander-yevsyukov see #1330. 

Let's discuss in person later. I think will need to code-generate :)

Issue #1316, opened 2020-10-23
Migrate `EmptyClassTest` to `SingletonTest`


alexander-yevsyukov:
... after it is [available](https://github.com/SpineEventEngine/base/issues/186).

PR #1317, opened 2020-10-23
Simplify unicast event routing

alexander-yevsyukov:
This PR adds `unicast()` methods for `EventRouting` to simplify the cases when an event target is only one entity.

Previously, it was required to wrap the ID of the target entity into `EventRoute.withId()` utility call.

Only some of the `withId()` calls were replaced with simplified calls to make this PR merge sooner. 
We'll migrate to simpler API where appropriate when corresponding framework areas or tests are updated.

### DIFF
--- a/server/src/main/java/io/spine/server/route/EventFnRoute.java
+++ b/server/src/main/java/io/spine/server/route/EventFnRoute.java
@@ -18,24 +18,43 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.server.aggregate.given.aggregate;
+package io.spine.server.route;
 
-import io.spine.server.route.EventRouting;
-import io.spine.test.aggregate.ProjectId;
-import io.spine.test.aggregate.event.AggProjectPaused;
-import io.spine.test.aggregate.event.AggTaskStarted;
+import io.spine.base.EventMessage;
+import io.spine.core.EventContext;
+
+import java.util.Set;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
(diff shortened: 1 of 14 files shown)


server/src/main/java/io/spine/server/route/EventFnRoute.java
```diff
@@ -0,0 +1,60 @@
...
+ *         the type of the event message
+ */
+final class EventFnRoute<I, E extends EventMessage> implements EventRoute<I, E> {
+
+    private static final long serialVersionUID = 0L;
+    private final BiFunction<E, EventContext, I> fn;
+
+    EventFnRoute(Function<E, I> fn) {
+        checkNotNull(fn);
+        this.fn = (e, ctx) -> fn.apply(e);
```
yuri-sergiichuk, line 48: Maybe use `this((e, ctx) -> fn.apply(e));` instead?

yuri-sergiichuk (approved):
@alexander-yevsyukov LGTM.

server/src/main/java/io/spine/server/route/EventFnRoute.java
```diff
@@ -0,0 +1,60 @@
...
+import io.spine.core.EventContext;
+
+import java.util.Set;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An event route with one target entity which ID is obtained as a function
```
dmitrykuzmin, line 33: I think, the grammatically correct form would be `ID of which is obtained...`.

armiol (approved):

alexander-yevsyukov:
51fecc6 also addresses #1316.

PR #1319, opened 2020-11-12
Improve `MethodSignature` and `WithTime`

alexander-yevsyukov:
This PR:

### Adds `localDate()` and `localDateTime()` methods to the `WithTime` interface
... thus saving on efforts required for obtaining popular temporals from events or commands.

Previously, only `instant()` was exposed which required calling `LocalDate.from(TemporalAccessor)` or `LocalDateTime.from(TemporalAccessor)`. This in turn, caused Error Prone warning because not all `TemporalAccessor`s provide date information (causing `DateTimeException`). 

New methods address removes the whole hassle because `Instant` does have both date and time information, so the exception would never occur.

### Bumps `base` and `config`
  * Eliminating calls to Hamcrest.
  * Addressing new ErrorProne warnings.

### Refactors `MatchCriterion`
I encountered Java compiler error telling “unable to instantiate” for declaration of `MatchCriterion.PROHIBITED_EXCEPTION` member. So, I extracted common bits and non-trivial message composition required for `PROHIBITED_EXCEPTION`, and the compiler swallowed it.

### Makes `SignatureMismatch.create()` return `Optional<>`
... as this is the only case we use.

### Makes `MethodSignature` return only one `AccessModifier`
We use this only for recommending (via warning) correct modifiers during compile time. 
It would be strange to recommend two or more things because it would cause a question, “Which one to use when?”
The previous signature was:

```java
protected ImmutableSet<AccessModifier> modifiers(); 
```
Now it's:

```java
protected AccessModifier modifier();
```

### Encapsulate allowed types of a `MethodSignature`
Before:
```java
public ImmutableSet<? extends ParameterSpec<EventEnvelope>> paramSpecs() { ... }
```
After:
```java
public AllowedParams<EventEnvelope> paramSpecs() { ... }
```

### Encapsulates return types of method signatures

Previously, `MethodSignature.returnTypes()` returned `ImmutableSet<TypeToken<?>>`. 
Now it's just `ReturnTypes`, and the class encapsulates the checking against a `Method`.

### DIFF
--- a/server/src/main/java/io/spine/server/entity/storage/InterfaceBasedColumn.java
+++ b/server/src/main/java/io/spine/server/entity/storage/InterfaceBasedColumn.java
@@ -39,14 +39,14 @@ import java.util.function.Function;
  *         int32 year_of_registration = 8 [(column) = true];
  *         </pre>
  *     <li>Implemented using custom Spine-generated interface:
- *         <pre>{@code
+ *         <pre>
  *         interface UserProfileWithColumns extends EntityWithColumns {
  *             int getYearOfRegistration();
  *         }
  *
  *         // ...
  *         class UserProfileProjection
- *             extends Projection<UserId, UserProfile, UserProfile.Builder>
+ * {@literal            extends Projection<UserId, UserProfile, UserProfile.Builder>  }
  *             implements UserProfileWithColumns {
  *
  *            {@literal @}Override
@@ -54,7 +54,7 @@ import java.util.function.Function;
  *                 return yearOfRegistration;
  *             }
  *         }
- *         }</pre>
+ *         </pre>
  *     <li>Extracted from the entity and propagated to the entity state at the moment of
  *         transaction commit. The column value is finalized by this moment.
  * </ol>
--- a/server/src/main/java/io/spine/server/entity/storage/InterfaceBasedColumn.java
+++ b/server/src/main/java/io/spine/server/entity/storage/InterfaceBasedColumn.java
@@ -39,7 +39,7 @@ import java.util.function.Function;
  *         int32 year_of_registration = 8 [(column) = true];
  *         </pre>
  *     <li>Implemented using custom Spine-generated interface:
- *         <pre>
+ *         <pre>{@code
  *         interface UserProfileWithColumns extends EntityWithColumns {
  *             int getYearOfRegistration();
  *         }
@@ -54,7 +54,7 @@ import java.util.function.Function;
  *                 return yearOfRegistration;
  *             }
  *         }
- *         </pre>
+ *         }</pre>
  *     <li>Extracted from the entity and propagated to the entity state at the moment of
  *         transaction commit. The column value is finalized by this moment.
  * </ol>
--- a/server/src/main/java/io/spine/server/model/MessageFormatter.java
+++ b/server/src/main/java/io/spine/server/model/MessageFormatter.java
@@ -20,8 +20,6 @@
 
 package io.spine.server.model;
 
-import io.spine.string.Diags;
-
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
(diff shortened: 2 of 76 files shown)


server/src/main/java/io/spine/server/model/MessageFormatter.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * An error message formatting helper.
+ */
+final class MessageFormatter {
+
+    private MessageFormatter() {
+    }
+
+    static Collector<CharSequence, ?, String> toStringEnumeration() {
```
armiol, line 34: Please document.

server/src/main/java/io/spine/server/model/ReturnTypes.java
```diff
@@ -0,0 +1,112 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Specifies types of values a handler methods may return.
+ */
+@Immutable
+public final class ReturnTypes {
+
+    /** Shared constant. */
```
armiol, line 39: Let's tell that this is
```
A return type of a  {@code void} handler method.
```
Or something like that.

server/src/test/java/io/spine/server/storage/RecordStorageTest.java
```diff
@@ -162,12 +161,9 @@ void filterByColumns() {
         EntityRecord notFineRecord1 = buildStorageRecord(idWrong1, newState(idWrong1));
         EntityRecord notFineRecord2 = buildStorageRecord(idWrong2, newState(idWrong2));
 
-        EntityRecordWithColumns recordRight =
-                create(fineRecord, matchingEntity, storage);
-        EntityRecordWithColumns recordWrong1 =
-                create(notFineRecord1, wrongEntity1, storage);
-        EntityRecordWithColumns recordWrong2 =
-                create(notFineRecord2, wrongEntity2, storage);
+        EntityRecordWithColumns recordRight = create(fineRecord, matchingEntity);
```
armiol, line 164: I will be merging with this forever. You have asked me to modify this in 2.x branch already.

Please rollback.

armiol (changes requested):
@alexander-yevsyukov LGTM in general. I really liked the shortcut types for the numerous `Immutable...` collections.

Please see my comments though.

server/src/main/java/io/spine/server/entity/storage/InterfaceBasedColumn.java
```diff
@@ -39,7 +39,7 @@
  *         int32 year_of_registration = 8 [(column) = true];
  *         </pre>
  *     <li>Implemented using custom Spine-generated interface:
- *         <pre>
+ *         <pre>{@code
```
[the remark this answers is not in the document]
alexander-yevsyukov: `<pre>` only doesn't make new Error Prone happy:

```
/Users/sanders/Projects/Spine/core-java/server/src/main/java/io/spine/server/entity/storage/InterfaceBasedColumn.java:49: warning: [UnescapedEntity] Javadoc is interpreted as HTML, so HTML entities such as &, <, > must be escaped.
```

In my earlier “exercises” with Checkstyle it suggested to use `{@code}`, and it worked. 

What's wrong with `<pre>/{@code}` combination?
Do you suggest using escapes all over, assuming our heavy usage of generics?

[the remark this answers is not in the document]
armiol: As discussed vocally, we'll be using `<pre>` for the multi-lined code snippets. And the corresponding Error Prone inspection should be disabled.

For more information on the reasoning please see [this article](https://reflectoring.io/howto-format-code-snippets-in-javadoc/).

server/src/test/java/io/spine/server/storage/RecordStorageTest.java
alexander-yevsyukov: As discussed vocally, this change was needed to address `BadImport` warning on usage of statically imported `EntityRecordWithColumns.create()`.

armiol (approved):

Issue #1320, opened 2020-11-18
Rework `MessageId`


alexander-yevsyukov:
As the analysis of this code related to this issue (#1189) reveals, the `MessageId` type and the code around it is puzzling. The name of the type conflicts with what we have already in the framework language. We have already `CommandId`, and `EventId`. Also, we promote typed entity IDs. And when the public API shows also `MessageId`, it raises a question: “What the hell is this? Why do you need it, and why do I need to know about it?”

More, the “guts” of the type are more than ID: we have type URL of the identified message, as well as the version. Sometimes, `MessageId` produces a copy of itself with a newer version. We definitely need a better name.

Also, exposure of `MessageId` to the client API (via `EventContextMixin`) is unnecessary. Presumably, the `MessageId` types serves for diagnostics, and as such should be isolated to the servers-side code.

Issue #1189, opened 2019-11-07
Remove `required` option for `type_url` in `MessageId`


alexander-yevsyukov:
In some cases we have to put `TypeUrl.of(Empty.class).value()` into this field, which is meaningless. See #1188 for more details on this issue.

It would be good to remove the field requirement.

alexander-yevsyukov:
More work than just addressing the `type_url` being non-required is needed. Closing this issue in favour of #1320 to be addressed in v2.

PR #1321, opened 2020-11-18
Encapsulate `MessageId` creation

alexander-yevsyukov:
This PR refactors creation of `MessageId` in some cases slightly consolidating work with this type.

The issue (#1189) which cased this PR was closed without addressing the need, because just making the `type_url` non-required would not improve things much. More work (#1320) would be needed (in v2.0) to make this type and the code around it much easier to understand and support.

### DIFF
--- a/core/src/main/proto/spine/core/diagnostics.proto
+++ b/core/src/main/proto/spine/core/diagnostics.proto
@@ -14,18 +14,16 @@ import "google/protobuf/any.proto";
 import "spine/core/actor_context.proto";
 import "spine/core/version.proto";
 
-// An identifier of a domain message.
-//
-// May represent a signal or an entity state.
-//
+// An identifier of a domain message, such as a signal, or an entity state.
 message MessageId {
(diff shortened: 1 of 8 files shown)


core/src/main/proto/spine/core/diagnostics.proto
```diff
@@ -14,18 +14,16 @@ import "google/protobuf/any.proto";
...
     option (is).java_type = "io.spine.core.MessageIdMixin";
 
     // The ID of the message.
     //
     // If the message is a signal (an event of a command), `id` has the value of the signal ID (an
-    // `EventId` or a `CommandId`). If the message is an entity state, `id` had the value of
-    // the entity identifier.
+    // `EventId` or a `CommandId`).
+    //
+    // If the message is an entity state, `id` had the value of the entity identifier.
```
armiol, line 26: `has the value`.

armiol (approved):
@alexander-yevsyukov LGTM with a typo to address.

PR #1322, opened 2020-11-18
Fail tests on runtime exceptions

### LINKED ISSUE #1314: Fail tests if runtime exceptions are thrown in BBC
Currently, it's possible to receive a false-positive test when one uses BBC tests and verifies the expected state of an entity with a sequence of events that first do a change and then undo the change if runtime exceptions are thrown in the middle of the events processing.

Such runtime exceptions are always wrapped into `HandlerFailedUnexpectedly` event and one may subscribe to it which is fine when we're running in Production. But while testing the app, in most cases we do want to fail a test case if smth went wrong.

The proposed solution is to allow configuring the BBC behavior with "strict" and "lenient" modes. By default, the BBC should be instantiated in "strict" mode when *any* runtime issue fails the test case. One should be able to switch to the "lenient" mode, e.g. if she wants to test that a particular `HandlerFailedUnexpectedly` event is thrown. In the "lenient" mode we only log runtime exceptions but do not fail the test cases automatically (the current behavior of the BBC).

yuri-sergiichuk:
This PR resolves #1314.

## Potential breaking change for tests

From now on the BlackBox tests are gonna be failed if some signal handler failed in the background. Such a change is desired while it eliminates potential untested erroneous flows.

## `FailedHandlerGuard`

The new `FailedHandlerGuard` is responsible for tracking `HandlerFailedUnexpectedly` events and fail the tests as soon as such an event occurs. The guard can be configured to tolerate such handler failures and simply log the failure over failing a test.

The BBC users can use `tolerateFailures` method to configure the guard accordingly and fall back to the previous behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,18 @@ hs_err_pid*
 
 .packages
 pubspec.lock
+
+
+#
+# The gradle.properties file should contain settings specific to a developer's workstation.
+#
+# See sample file for a Mac OS X workstation below.
+# -------
+# # Set Java home to point to JDK8. This is need to generate classes working with Java8 API.
+# # Otherwise the following warning appears during the build:
+# #     warning: [options] bootstrap class path not set in conjunction with -source 1.8
+# #
+# # suppress inspection "UnusedProperty"
+# org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk1.8.0_51.jdk/Contents/Home/
+# -------
+gradle.properties
--- a/.gitignore
+++ b/.gitignore
@@ -1,48 +1,95 @@
+#
+# Copyright 2020, TeamDev. All rights reserved.
+#
+# Redistribution and use in source and/or binary forms, with or without
+# modification, must retain the above copyright notice and the following
+# disclaimer.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+#
+# This file is used for two purposes:
+#  1. ignoring files in the `config` project.
+#  2. ignoring files in the projects that import `config` as a sub-module.
+#
+# Therefore, instructions below are superset of instructions required for all the projects.
+
 # IntelliJ IDEA modules and interim config files.
 *.iml
-.idea/libraries
-.idea/shelf
 .idea/*.xml
 .idea/.name
+.idea/artifacts
+.idea/libraries
+.idea/modules
+.idea/shelf
 
-# Do not ignore code style settings.
-!/.idea/codeStyleSettings.xml
+# Do not ignore the following IDEA settings
+!.idea/misc.xml
+!.idea/codeStyleSettings.xml
+!.idea/codeStyles/
+!.idea/copyright/
 
 # Gradle interim configs
-.gradle/
-
-# Maven deployment build directory
-/core-java/target
+**/.gradle/**
 
 # Generated source code
-generated/
+**/generated/**
 
 # Gradle build files
-build/
+**/build/**
 
-# IDEA build files
-out/
+# Build files produced by the IDE
+**/out/**
 
-# Spine temporary artifact storage
-.spine/
+# Ignore Gradle GUI config
+gradle-app.setting
 
-# Credentials to Maven repositories and Google Cloud Storage used for Travis build reports
+# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
+!gradle-wrapper.jar
+
+# Cache of project
+.gradletasknamecache
+
+# # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
+# gradle/wrapper/gradle-wrapper.properties
+
+# Spine internal directory for storing intermediate artifacts
+**/.spine/**
+
+# Spine model compiler auto-generated resources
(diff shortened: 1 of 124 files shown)


yuri-sergiichuk:
@alexander-yevsyukov @dmdashenkov PTAL.

.gitignore
```diff
@@ -1,48 +1,95 @@
...
-# See sample file for a Mac OS X workstation below.
-# -------
-# # Set Java home to point to JDK8. This is need to generate classes working with Java8 API.
-# # Otherwise the following warning appears during the build:
-# #     warning: [options] bootstrap class path not set in conjunction with -source 1.8
-# #
-# # suppress inspection "UnusedProperty"
-# org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk1.8.0_51.jdk/Contents/Home/
-# -------
-gradle.properties
```
alexander-yevsyukov, line 48: Why do we want `gradle.properties` to be version-controlled now?

server/src/main/java/io/spine/server/model/ExternalCommandReceiverMethodError.java
```diff
@@ -49,11 +50,15 @@
...
+    /**
+     * Creates a new exception for the command handler that violates the {@code @Command} semantic.
+     */
+    @Internal
     public ExternalCommandReceiverMethodError(
             CommandHandlingClass<?, ?> classWithViolation,
             Collection<? extends CommandAcceptingMethod<?, ?>> invalidMethods) {
-        super("The class `%s` declares `external` command receiver methods for command types: %s."
-                      + " Only event accepting methods should be marked as `external`.",
+        super("The class `%s` declares `external` command receiver methods for command types: %s. " +
```
alexander-yevsyukov, line 60: Please move `+` to next line. We do not check it automatically yet. But soon we'd do using Checkstyle.

server/src/test/java/io/spine/server/procman/ProcessManagerTest.java
```diff
@@ -124,7 +124,7 @@
     private TestProcessManager processManager;
 
     @BeforeEach
-    void setUp() {
+    void initProcessManager() {
```
alexander-yevsyukov, line 127: A better name would be `initContextAndProcessManager()`. Just to play nicely with `closeContext()`.

client/src/main/java/io/spine/client/QueryFactory.java
```diff
@@ -204,62 +201,64 @@ private static void checkSpecified(Class<? extends EntityState> entityClass) {
         return builder;
     }
 
+    @Internal
```
alexander-yevsyukov, line 204: Why do we want this annotation on the package-private code?

client/src/main/java/io/spine/client/QueryFactory.java
```diff
@@ -204,62 +201,64 @@ private static void checkSpecified(Class<? extends EntityState> entityClass) {
...
-        Query.Builder builder = Query.newBuilder()
-                                     .setTarget(target);
-        return builder;
+        return composeQuery(target, 0, null, fieldMask);
     }
 
-    Query composeQuery(Target target,
-                       OrderBy orderBy,
-                       @Nullable FieldMask fieldMask) {
+    @Internal
```
alexander-yevsyukov, line 210: Why do we want this annotation on the package-private code?

client/src/main/java/io/spine/client/QueryFactory.java
```diff
@@ -204,62 +201,64 @@ private static void checkSpecified(Class<? extends EntityState> entityClass) {
...
-        Query query = newQuery(builder);
-        return query;
+        return composeQuery(target, 0, orderBy, fieldMask);
     }
 
-    Query composeQuery(Target target,
-                       OrderBy orderBy,
-                       int limit,
-                       @Nullable FieldMask fieldMask) {
+    @Internal
```
alexander-yevsyukov, line 217: Why do we want this annotation on the package-private code?

testutil-server/src/main/java/io/spine/testing/server/blackbox/DiagnosticLogging.java
```diff
@@ -0,0 +1,77 @@
...
+        boolean loggingEnabled = severeLogger.isEnabled();
+        if (loggingEnabled) {
+            severeLogger.log(msg);
+            severeLogger.log(toJson(event));
+        } else {
+            @SuppressWarnings("UseOfSystemOutOrSystemErr")
+            // Edge case for disabled/misconfigured logging .
+            PrintStream stderr = System.err;
+            stderr.println(msg);
+            stderr.println(toJson(event));
```
alexander-yevsyukov, line 74: Please extract `toJson(Event)` into a variable. We use it in both `if` branches.

testutil-server/src/main/java/io/spine/testing/server/blackbox/FailedHandlerGuard.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>The {@code BlackBoxBoundedContext} does not consume external events.
+     */
+    @Override
+    public ImmutableSet<EventClass> externalEventClasses() {
+        return ImmutableSet.of();
```
alexander-yevsyukov, line 111: How about introducing  `EventClass.emptySet()` so that it's easier to find places where return empty sets?

testutil-server/src/main/java/io/spine/testing/server/blackbox/HandlerFailureTolerance.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Identifies that exceptions must be only logged.
+     */
+    LOG,
+
+    /**
+     * Identifies that exceptions must be raised and tests must be failed.
+     */
+    RAISE
```
alexander-yevsyukov, line 37: Consider naming it `RAISE_AND_FAIL`.

testutil-server/src/main/java/io/spine/testing/server/blackbox/HandlerFailureTolerance.java
```diff
@@ -0,0 +1,38 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.testing.server.blackbox;
+
+/**
+ * Configures the behavior of the {@link BlackBoxContext} on handling runtime exceptions
+ * thrown from the handler methods.
+ */
+public enum HandlerFailureTolerance {
```
alexander-yevsyukov, line 27: Do we really need it to be public?

alexander-yevsyukov (changes requested):
Please see my comments and questions.

yuri-sergiichuk:
@alexander-yevsyukov PTAL again.

testutil-server/src/main/java/io/spine/testing/server/blackbox/HandlerFailureTolerance.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+enum HandlerFailureTolerance {
+
+    /**
+     * Identifies that exceptions must be only logged.
+     */
+    LOG,
+
+    /**
+     * Identifies that exceptions must be raised and tests must be failed.
```
dmdashenkov, line 35: Maybe "thrown"?

yuri-sergiichuk: I'd keep "raised" as we do not really intend to rethrow the exception.

dmdashenkov:
@yuri-sergiichuk, LGTM as long as build passes.

dmdashenkov (approved):
@yuri-sergiichuk, LGTM, as long as @alexander-yevsyukov 's comments are resolved.

alexander-yevsyukov (approved):
LGTM

Issue #1315, opened 2020-10-23
API improvements for `ServerEnvironment`


dmitrykuzmin:
1. `ServerEnvironment.use(...)` should now only be usable in a context like

    ```java
    ServerEnvironment.when(SomeEnv.class)
                     .use(envType -> storageFactory())
                     .use(envType -> delivery()) // ...
    ``` 
    That is, the `use` method won't require `EnvironmentType` as a second argument anymore (superseded by `when(...)`), and will accept lazily-initialized values instead of plain objects. The `envType` is the active environment type supplied for logging or other purposes.

    The old `use(T, EnvironmentType)` API should be deprecated.

2. `ServerEnvironment` should supply a new public method `type()` which will return the currently active `EnvironmentType`. 

    This way, the user code will be able to easily access the info about the currently active environment when the `ServerEnvironment` is already configured.

3. `@Internal`-ize the Environment.register(...) and improve `EnvironmentType` documentation. 
  It is not immediately apparent what exactly to do with a custom `EnvironmentType` once you've created it.
  For example, with the current implementation, it's possible to employ Environment.register(EnvironmentType) to register a custom env type which actually accepts some dependencies on creation which help to deduce whether it's enabled or not. And then it would be logical to check for such env presence with `when(...)`, but this would throw an exception.

alexander-yevsyukov:
@armiol @dmitrykuzmin how about having it shorter for configuration phase?

```java
ServerEnvironment.when(SomeEnv.class)
                 .use(envType -> storageFactory())
                 .use(envType -> delivery()) // ...
```

dmitrykuzmin:
I personally support this change.

armiol:
@alexander-yevsyukov LGTM.

alexander-yevsyukov:
@armiol @dmitrykuzmin 

Because of limitation of Java generics, we cannot have multiple `use(Function<T, R>)` at the same level. So, the following is not possible:

```java
ServerEnvironment.when(SomeEnv.class)
                 .use(envType -> storageFactory())
                 .use(envType -> delivery())  // ...
```

We can have:

```java
ServerEnvironment.when(SomeEnv.class)
                 .useStorageFactory(envType -> storageFactory())
                 .useDelivery(envType -> delivery())  // ...

```

Or, abandon the idea of passing functions to the `use()` API at all.

Thoughts?

alexander-yevsyukov:
[...]

PR #1327, opened 2020-11-19
Improve `ServerEnviornment` configuration

alexander-yevsyukov:
This PR addresses the first part of the issue #1315. In particular:

 * `ServerEnvironment` API deprecated prior to this PR was removed. Namely `configureDelivery()`, `configureStorageForTests()`, `configureStorage()`.
 * `ServerEnviornment.when(Class<? extends EnvironmentType> type)` method was introduced, returing `TypeConfigurator` instance.
* `TypeConfigurator` now replaces the group of `ServerEnviornment.use(V, Class)`  methods that were deprecated in favour of the methods of this new class.

Other improvements described in  #1315 will be introduced in the following PRs.


server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -48,25 +49,30 @@
...
+ *
+ * <p>A custom environment type may also be used:
  * <pre>
+ *
  *     final class StagingEnvironment extends EnvironmentType {
  *         ...
  *     }
  *
- *     // Can also be
- *     // `ServerEnvironment.use(inMemoryStorageFactory, new Staging(stagingServiceDiscovery));`,
```
dmitrykuzmin, line 67: Why do we drop API which allows passing of `EnvironmentType`-s with dependencies, like the one here? Usage looks more or less realistic to me.

armiol (approved):

server/src/main/java/io/spine/server/ServerEnvironment.java
alexander-yevsyukov: It's going to be addressed in the following PR by introducing `use(Function<Class<? extends EnvironmentType>>, V>)`.

PR #1331, opened 2020-11-20
More improvements to `ServerEnvironment` configuration

alexander-yevsyukov:
This PR continues addressing #1315.

Enhancements introduced:
  * It is now possible to configure `ServerEnvironment ` features using `useSomething()` methods that accept a `Function<Class<? extends EnvironmentType>, Something>`.
 * In order to make the declaration a bit more compact `ServerEnvironment.Fn` functional interface was introduced. It's just a function, from the previous point, but with first generic argument hidden.
 * Tests of `ServerEnvironment` that related to the configuration were extracted into a separate test suite.

### DIFF
--- a/server/src/main/java/io/spine/server/ServerEnvironment.java
+++ b/server/src/main/java/io/spine/server/ServerEnvironment.java
@@ -363,11 +363,17 @@ public final class ServerEnvironment implements AutoCloseable {
     public StorageFactory storageFactory() {
         Class<? extends EnvironmentType> type = environment().type();
         StorageFactory result = storageFactory.optionalValue(type)
-                .orElseThrow(() -> newIllegalStateException(
-                        "The storage factory for environment `%s` was not configured."
-                                + " Please call `.when(environmentType).use(storageFactory)`.",
-                        type.getSimpleName()));
+                .orElseThrow(() -> suggestConfiguringStorageFactory(type));
         return result;
+    }
+
+    private static IllegalStateException
+    suggestConfiguringStorageFactory(Class<? extends EnvironmentType> type) {
+        String typeName = type.getSimpleName();
+        return newIllegalStateException(
+                "The storage factory for the environment `%s` was not configured."
+                        + " Please call `ServerEnvironment.when(%s.class).use(storageFactory);`.",
+                typeName, typeName);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


dmitrykuzmin:
I don't know if PR in `base` is actually planned in scope of these changes, but I'd also propose to `@Internal`-ize the `Environment.register(...)` and maybe clarify `EnvironmentType` doc a bit, because from my experience it's not immediately apparent what exactly to do with a custom `EnvironmentType` once you've created it.

For example, with the current implementation, it's possible to employ `Environment.register(EnvironmentType)` to register a custom env type which actually accepts some dependencies on creation which help to deduce whether it's enabled or not. And then it would be logical to check for such env presence with `when(...)`, but this would throw an exception.

alexander-yevsyukov:
@dmitrykuzmin, that would go into another PR, because `Environment` and `EnvironmentType` are in `base`. I'll add your notes to the governing issue.

server/src/main/java/io/spine/server/ServerEnvironment.java
```diff
@@ -393,14 +400,20 @@ public StorageFactory storageFactory() {
...
-                        "Transport factory is not assigned for the current environment `%s`."
-                                + " Please call `.when(environmentType).use(transportFactory)`.",
-                        type.getSimpleName()));
+                .orElseThrow(() -> suggestConfiguringTransportFactory(type));
 
         return result;
     }
 
+    private static IllegalStateException
+    suggestConfiguringTransportFactory(Class<? extends EnvironmentType> type) {
```
armiol, line 409: Should we have the same methods for the tracer factory and delivery?

If so, maybe we should arrange a private nested class `Suggest` with the names like `configuringTransportFactory`?

armiol (approved):
@alexander-yevsyukov LGTM with a suggestion to consider.

server/src/main/java/io/spine/server/ServerEnvironment.java
armiol: Or event `SuggestConfiguring.transportFactory(..)`?

alexander-yevsyukov:
@dmitrykuzmin 
> And then it would be logical to check for such env presence with when(...), but this would throw an exception.

`TypeConfigurator` now does this automatically in its constructor. I am planning to address improvements you suggest shortly in the PRs following in `base` and `core-java`.

PR #1340, opened 2020-12-03
Improve dependencies management

yuri-sergiichuk:
This PR improves our Gradle dependencies management a bit and applies the new scope to the dependencies that are used by the code analysis software only.

The change is related to the https://github.com/SpineEventEngine/config/issues/156 issue.


alexander-yevsyukov (approved):
LGTM

PR #1341, opened 2020-12-07
Remove empty `grand_origin`

dmdashenkov:
In this PR we avoid setting the `Origin.grand_origin` field to an empty value. This allows us to generate intuitive queries in languages other than Java filtering by the message origin.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #1343, opened 2020-12-10
Improve signals dispatching

### LINKED ISSUE #1318: `@Subscribe`-r of the rejection caused by a particular command causes the dispatching failure in some cases
# Landscape

It is possible to subscribe to a rejection which is caused by some command:

```java
public class CancellationWatch extends AbstractEventSubscriber {

    @Subscribe
    void on(Rejections.OrderAlreadyCompleted r, CancelOrder c) {
        ...
    }
}
```
In this example, a subscriber wants to consume only those `OrderAlreadyCompleted` rejections which are thrown while handling order cancellations (hence, `CancelOrder` command is the second argument).

At the same time, it is also possible to declare several command handlers which would throw a rejection of the same type:

```java
public class OrderAggregate extends Aggregate<OrderId, Order, Order.Builder> {

    @Assign
    OrderCompleted handle(CompleteOrder c) throws OrderAlreadyCompleted {
        ...
    }

    @Assign
    OrderCancelled handle(CancelOrder c) throws OrderAlreadyCompleted {
        ...
    }
}
```

In the example above, the `OrderAggregate` rejects all commands with `OrderAlreadyCompleted` if the order _is_ already completed.

# Action 

Now, if some _completed_ order handles `CompleteOrder` command, the aggregate throws a `OrderAlreadyCompleted` rejection. In theory, if anyone has subscribed to this rejection, they would receive it. 

But, one thing for sure, `CancellationWatch` is *not* interested in this rejection—because the origin command is not what it expects. 

So if no subscribers to `OrderAlreadyCompleted` rejection is found, it should just pass through the event bus, just as any other rejection or event. 

# Issue

In reality, the framework attempts to select the handler for this rejection anyway, and fails to do so:
```
java.lang.IllegalStateException: Unable to find handler with the key: DispatchKey{messageClass=`com.acme.order.Rejections$OrderAlreadyCompleted}.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:589)
	at io.spine.server.model.HandlerMap.handlersOf(HandlerMap.java:145)
	at io.spine.server.model.HandlerMap.handlersOf(HandlerMap.java:174)
	at io.spine.server.event.model.EventReceivingClassDelegate.handlersOf(EventReceivingClassDelegate.java:127)
	at io.spine.server.event.model.EventSubscriberClass.subscribersOf(EventSubscriberClass.java:79)
	at io.spine.server.event.model.SubscribingClass.subscriberOf(SubscribingClass.java:47)
	at io.spine.server.event.AbstractEventSubscriber.canDispatch(AbstractEventSubscriber.java:140)
	at io.spine.server.event.AbstractEventSubscriber.canDispatch(AbstractEventSubscriber.java:60)
	at io.spine.server.bus.DispatcherRegistry.lambda$dispatchersOf$0(DispatcherRegistry.java:105)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174)
	at java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1556)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
	at io.spine.server.bus.DispatcherRegistry.dispatchersOf(DispatcherRegistry.java:107)
	at io.spine.server.bus.DeadMessageFilter.filter(DeadMessageFilter.java:56)
	at io.spine.server.bus.FilterChain.filter(FilterChain.java:57)
	at io.spine.server.bus.Bus.filter(Bus.java:296)
	at io.spine.server.bus.Bus.filter(Bus.java:272)
	at io.spine.server.bus.Bus.filterAndPost(Bus.java:150)
	at io.spine.server.bus.Bus.post(Bus.java:146)
```

The reason is as follows. From the dispatcher registry, the framework knows that someone _does_ handle the rejections of this type. And from this point, it is going to get that someone. There is no option to discover that in fact that "someone" does not match the "origin message" criterion.

yuri-sergiichuk:
This PR fixes #1318.

In the PR I have relaxed the `HandlerMap` check on missing handlers while it is totally OK to **not** have a specific handler within a specific class. It's then up to the caller to verify and check how to behave when a handler is missing. In our case, all the respective callers already do check and filter out situations when no handlers are available. The only change one may notice is the error text available via the `HandlerFailedUnexpectedly` event, instead of the `IllegalStateException`, we'll see the `EventValidationError` with the `UNSUPPORTED_EVENT` code.

### DIFF
--- a/server/src/main/java/io/spine/server/event/model/EventReceivingClassDelegate.java
+++ b/server/src/main/java/io/spine/server/event/model/EventReceivingClassDelegate.java
@@ -144,14 +144,14 @@ public class EventReceivingClassDelegate<T extends EventReceiver,
             return ImmutableSet.of();
         }
         ImmutableSet<M> stateHandlers = handlers.handlersOf(updateEvent);
-        ImmutableSet<StateClass> result = stateHandlers
-                .stream()
-                .filter(StateSubscriberMethod.class::isInstance)
-                .map(StateSubscriberMethod.class::cast)
-                .filter(external ? HandlerMethod::isExternal : HandlerMethod::isDomestic)
-                .map(StateSubscriberMethod::stateType)
-                .map(StateClass::from)
-                .collect(toImmutableSet());
+        ImmutableSet<StateClass> result =
+                stateHandlers.stream()
+                             .filter(StateSubscriberMethod.class::isInstance)
+                             .map(StateSubscriberMethod.class::cast)
+                             .filter(external ? HandlerMethod::isExternal : HandlerMethod::isDomestic)
+                             .map(StateSubscriberMethod::stateType)
+                             .map(StateClass::from)
+                             .collect(toImmutableSet());
         return result;
     }
 }
--- a/server/src/main/java/io/spine/server/event/model/EventReceivingClassDelegate.java
+++ b/server/src/main/java/io/spine/server/event/model/EventReceivingClassDelegate.java
@@ -120,8 +120,6 @@ public class EventReceivingClassDelegate<T extends EventReceiver,
 
     /**
      * Obtains the method which handles the passed event class.
-     *
-     * @throws IllegalStateException if there is such method in the class
      */
     public ImmutableSet<M> handlersOf(EventClass eventClass, MessageClass originClass) {
         return handlers.handlersOf(eventClass, originClass);
@@ -130,7 +128,8 @@ public class EventReceivingClassDelegate<T extends EventReceiver,
     /**
      * Obtains the method which handles the passed event class.
      *
-     * @throws IllegalStateException if there is such method in the class
+     * @throws IllegalStateException
+     *         if there is no such method in the class
      */
     public M handlerOf(EventClass eventClass, MessageClass originClass) {
         return handlers.handlerOf(eventClass, originClass);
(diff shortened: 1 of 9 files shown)


yuri-sergiichuk:
@armiol PTAL.

server/src/main/java/io/spine/server/event/model/EventReceivingClassDelegate.java
```diff
@@ -145,14 +144,14 @@ public M handlerOf(EventClass eventClass, MessageClass originClass) {
...
         ImmutableSet<M> stateHandlers = handlers.handlersOf(updateEvent);
-        ImmutableSet<StateClass> result =
-                stateHandlers.stream()
-                        .filter(h -> h instanceof StateSubscriberMethod)
-                        .map(h -> (StateSubscriberMethod) h)
-                        .filter(external ? HandlerMethod::isExternal : HandlerMethod::isDomestic)
-                        .map(StateSubscriberMethod::stateType)
-                        .map(StateClass::from)
-                        .collect(toImmutableSet());
+        ImmutableSet<StateClass> result = stateHandlers
```
armiol, line 147: Could you please format it like this:
```java
        ImmutableSet<StateClass> result =
                stateHandlers
                        .stream()
                        .filter(StateSubscriberMethod.class::isInstance)
                        .map(StateSubscriberMethod.class::cast)
                        .filter(external ? HandlerMethod::isExternal : HandlerMethod::isDomestic)
                        .map(StateSubscriberMethod::stateType)
                        .map(StateClass::from)
                        .collect(toImmutableSet());
```

armiol (approved):
@yuri-sergiichuk LGTM with a minor comment to address.

PR #1336, opened 2020-12-01
`ServerEnvironment` improvements

alexander-yevsyukov:
This PR finishes addressing #1315. Notable changes:
  * Deprecated API in `ServerEnvironment` was removed.
  * Introduced shortcut method for obtaining current environment type from `ServerEnvironment` instance.

PR #1308, opened 2020-10-16
Improve multitenancy support

yuri-sergiichuk:
This PR improves multitenancy support within Inbox delivery. The introduced change ensures each batch of the delivery is running within a tenant-aware operation.

PR #1344, opened 2020-12-14
Release v1.7.0

yuri-sergiichuk:
This PR does the release of the library version `1.7.0`.

This release brings numerous API improvements, as well as fixes and infrastructure updates to the framework.

## Breaking Changes

1. The `BlackBoxContext`-based tests now fail if a runtime exception was thrown within the signal handlers.

    In order to address #1314, we've decided to enforce the fail-fast approach within the BBC tests done in #1322. From now on, if a test case had any runtime exceptions thrown from signal handlers the test is failed by the BBC. While it may be a breaking change for some, we believe it worth fixing such issues right away than hiding them under the carpet.

    If one requires to fall back to the previous behavior, the BBC instance can be configured using the newly introduced `tolerateFailures` method. 

2. The `grand_origin` field is no longer set to a default instance for the signal `Origin` if no grand origin is present (see #1341 for details).

3. The `ServerEnvironment` API is improved and changed as a result of the #1315 and related discussions in a series of PRs (#1327, #1331, #1336).

    The previously deprecated `configure...()` API is removed in favor of the new fluent `when().use()` API.

    So now, instead of smth like this:

    ```java
    ServerEnvironment.use(productionStorageFactory, Production.class)
                     .use(testingStorageFactory, Tests.class)
                     .use(memoizingTracerFactory, Production.class)
    ```
    
    One should use the following snippet:

    ```java
    ServerEnvironment.when(Production.class)
                     .use(productionStorageFactory)
                     .use(memoizingTracerFactory);
    ServerEnvironment.when(Tests.class)
                     .use(testingStorageFactory);
    ```

## API Changes

1. `Change` type validation requirements are relaxed for primitive type changes.

    The `new_value` field is no longer a required one for `StringChange` and `BytesChange` types. See #1307 for details.

2. Introduced simplified `unicast()` methods in the `EventRouting`.

    The new `unicast()` API allows simplifying and prettifying the even routing for the cases where a singular ID is used.

    Now, instead of smth like this:

    ```java
     @OverridingMethodsMustInvokeSuper
     @Override
     protected void setupEventRouting(EventRouting<AirportId> routing) {
         super.setupEventRouting(routing);
         routing.route(FlightScheduled.class, (e, ctx)-> EventRoute.withId(e.getAirport()));
     }
    ```
   
    One can use the following API:
   
    ```java
     @OverridingMethodsMustInvokeSuper
     @Override
     protected void setupEventRouting(EventRouting<AirportId> routing) {
         super.setupEventRouting(routing);
         routing.unicast(FlightScheduled.class, FlightScheduled::getAirport)
     }
    ```
   
    To find out more details on the new API please check #1317.

3. Added `localDate` and `localDateTime` helpers to the `WithTime` interface.

    This feature allows accessing `localDate` and `localDateTime` methods from within all the signals. From now on, if one requires a Java `LocalDate` or `LocalDateTime` instances over the Protobuf `Timestamp` or Java `Instant` that were previously available for signals, they may use the new API to simplify such a conversion and access to the signals `timestamp` field.

    See #1319 for additional details.

## Fixes

1. Improved multitenant delivery support by ensuring the tenant is properly propagated within the delivery (see #1308).

2. Fixed a typo in the `io.spine.client.Client` `shutdownTimeout` method (see #1339).

3. Fixed dispatching of rejections caused by a particular command (see #1318 and #1343).

## Infrastructure

1. The libraries now do not use `implementation` for compile-only annotations like `errorprone` annotations but use the newly [introduced](https://docs.gradle.org/6.7.1/release-notes.html#new-dependency-management-features) `compileOnlyApi` configuration for such dependencies (see #1340).


armiol:
@yuri-sergiichuk There is a typo here:
```
... if not grand origin
```
Should be `no grand origin`.

armiol:
@yuri-sergiichuk Can we please change
```
We should use the following snippet:
```
to
```
One should use the following snippet:
```

This is going to be more inline with how we tell about things throughout this text.

armiol:
@yuri-sergiichuk I also think you may easily drop all "PR"s when referencing a pull request. Just "see #4211" works fine, with less letters to read :)

armiol:
@yuri-sergiichuk There is a typo here:

```java
routing..route(FlightScheduled.class, (e, ctx)-> EventRoute.withId(e.getAirport()));
```
Let's remove the range operator `..` in favour of calling a method (`.`).

armiol (approved):
@yuri-sergiichuk LGTM except for my comments to the PR description.

Please see them posted.

PR #1345, opened 2020-12-18
Kotlin extension `TransactionalEntity.update`

alexander-yevsyukov:
This PR introduces `TransactionalEntity.update()` extension function that allows to access `builder()` properties in the code like this:

```kotlin
@Apply
fun event(e: TaskCreated) {
    update {
        title = e.title
        description = e.description
    }
}
```

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="web" url="file://$PROJECT_DIR$" />
+  </component>
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
@@ -35,5 +38,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
 </project>
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -128,9 +128,7 @@ subprojects {
         targetCompatibility = JavaVersion.VERSION_1_8
     }
 
-    val jvmVersion = "1.8"
-
-   tasks.withType<KotlinCompile>.configureEach {
+   tasks.withType<KotlinCompile>().configureEach {
       kotlinOptions {
           jvmTarget = JavaVersion.VERSION_1_8.toString()
       }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,9 +59,9 @@ plugins {
     kotlin("jvm") version "1.4.21"
     idea
     @Suppress("RemoveRedundantQualifierName") // Cannot use imports here.
-    id("com.google.protobuf").version(io.spine.gradle.internal.Deps.versions.protobufPlugin)
-    @Suppress("RemoveRedundantQualifierName") // Cannot use imports here.
-    id("net.ltgt.errorprone").version(io.spine.gradle.internal.Deps.versions.errorPronePlugin)
+    id("com.google.protobuf") version io.spine.gradle.internal.Deps.versions.protobufPlugin
+    @Suppress("RemoveRedundantQualifierName")
+    id("net.ltgt.errorprone") version io.spine.gradle.internal.Deps.versions.errorPronePlugin
 }
 
 apply(from = "version.gradle.kts")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -130,15 +130,11 @@ subprojects {
 
     val jvmVersion = "1.8"
 
-    val compileKotlin: KotlinCompile by tasks
-    compileKotlin.kotlinOptions {
-        jvmTarget = jvmVersion
-    }
-
-    val compileTestKotlin: KotlinCompile by tasks
-    compileTestKotlin.kotlinOptions {
-        jvmTarget = jvmVersion
-    }
+   tasks.withType<KotlinCompile>.configureEach {
+      kotlinOptions {
+          jvmTarget = JavaVersion.VERSION_1_8.toString()
+      }
+   }
 
     DependencyResolution.defaultRepositories(repositories)
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -18,10 +18,10 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import io.spine.gradle.internal.DependencyResolution
 import io.spine.gradle.internal.Deps
 import io.spine.gradle.internal.PublishingRepos
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 buildscript {
 
@@ -31,6 +31,7 @@ buildscript {
     @Suppress("RemoveRedundantQualifierName") // Cannot use imports here.
     val dependencyResolution = io.spine.gradle.internal.DependencyResolution
 
+    val kotlinVersion: String by extra
     val spineBaseVersion: String by extra
     val spineTimeVersion: String by extra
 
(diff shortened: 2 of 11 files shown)


server/src/test/java/io/spine/server/entity/StubTransaction.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Stub implementation of {@code Transaction} which behaves as told in the passed parameters.
+ */
+final class StubTransaction<I,
+                            E extends TransactionalEntity<I, S, B>,
+                            S extends EntityState<I>,
+                            B extends ValidatingBuilder<S>>
+        extends Transaction<I, E, S, B> {
+
+    StubTransaction(E entity, boolean active, boolean stateChanged) {
```
armiol, line 35: We also may be un-deleting or un-archiving the entity. So please consider picking up these changes as well.

armiol (approved):
@alexander-yevsyukov LGTM with a single suggestion to consider.

.idea/misc.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="web" url="file://$PROJECT_DIR$" />
+  </component>
```
yuri-sergiichuk, line 5: I wonder why this part is always being added/removed...

.idea/misc.xml
```diff
@@ -35,5 +38,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
 </project>
```
yuri-sergiichuk, line 42: Maybe we should not bother ourselves with EOLs in IDEA configs? The IDE is always removing it.

build.gradle.kts
```diff
@@ -52,6 +56,7 @@ buildscript {
 
 plugins {
     java
+    kotlin("jvm") version "1.4.21"
```
yuri-sergiichuk, line 59: Let's use the same approach we already use for the other plugins.

```suggestion
    kotlin("jvm").version("1.4.21")
```

Also, can we rely on the `Deps` object for the version?

build.gradle.kts
```diff
@@ -121,12 +128,25 @@ subprojects {
...
+
+    val compileKotlin: KotlinCompile by tasks
+    compileKotlin.kotlinOptions {
+        jvmTarget = jvmVersion
+    }
+
+    val compileTestKotlin: KotlinCompile by tasks
+    compileTestKotlin.kotlinOptions {
+        jvmTarget = jvmVersion
+    }
```
yuri-sergiichuk, line 141: These 2 declarations look identical to me. How does it determine that one is for tests?

I'd better have it like this:
```suggestion
   tasks.withType<KotlinCompile>.configureEach {
      kotlinOptions {
          jvmTarget = JavaVersion.VERSION_1_8.toString()
      }
   }
```

server/src/main/kotlin/io/spine/server/entity/extensions.kt
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2020, TeamDev. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
```
yuri-sergiichuk, line 8: It looks like an oldish disclaimer. Do you think we can start using the new one? It is available in master already and in the `config` master as well.

server/src/main/kotlin/io/spine/server/entity/extensions.kt
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * For example, a method that applies an event may look like this:
+ *
+ * ```kotlin
+ * @Apply
+ * fun event(e: TaskCreated) {
+ *     update {
+ *         title = e.title
+ *         description = e.description
+ *     }
```
yuri-sergiichuk, line 38: I was tinkering with this example multiple times and it looks like we're not talking this way.
I mean, to me it really read like: "Update TaskCreated event" or "Apply TaskCreated event update". In any interpretation, it's not clear that we're actually updating the entity state and not the event itself.

I guess, it'd be much easier to understand if we have the following example:
```kotlin
@Apply
fun on(event: TaskCreated) {
     update {
         title = event.title
         description = event.description
     }
}
```

Also, the `update` itself looks really nice. But it may be again not obvious what we're trying to update. It'd be better to have it as a `state {}` closure, but I know that we already have the `state()` method.

And just thinking out loud here, I do understand that `updateState { ... }` is a bit too much too.

yuri-sergiichuk: @armiol @alexander-yevsyukov what do you think on the matter?

.idea/misc.xml
alexander-yevsyukov: I think it's because of different IDEA versions used by comitters.

.idea/misc.xml
alexander-yevsyukov: What if we start using pre-commit hooks that do this, instead? Or, Gradle plugins similarly to how JUnit does it?

build.gradle.kts
alexander-yevsyukov: It's a bit easier to read with spaces instead of dots and braces as delimiters. So, I'd suggest doing the other way around.

build.gradle.kts
alexander-yevsyukov: This is the way IDEA adds Kotlin to projects in an automatic way. Delegates, if memory serve, accept the descriptor of the declaration, which includes the name. But your suggestion is more compact, indeed.

server/src/main/kotlin/io/spine/server/entity/extensions.kt
alexander-yevsyukov: I would use `on()` for subscribers. 

Having `e` as the parameter name is shorter. The whole method declaration almost reads as English. We don't even need KDocs for applier methods like this, unless they do something non-trivial.

We update things on the left of assignments. It's the same thing we do for all the variables.

But removing the `state` or `builder` things, we make it easier to 1) read and 2) understand for novice users. It feels as if these properties belong to the entity.

alexander-yevsyukov: @yuri-sergiichuk, I understand it feels a bit new and unusual, but Kotlin is indeed a really different language. Have a look around bigger examples of Kotlin, and you'll see that these people “speak” and think differently. And it's more efficient to work this way.

alexander-yevsyukov: Let me mark these things `@Experimental` for a while. This way we'll play with it and see how it goes.

yuri-sergiichuk: OK, after the weekend I'm trying to read it again and if I add a `.` it really looks better: `Apply event "TaskCreated". Update "title" and "description"`.

I'm totally fine with not using `state` and `builder` directly, it just was a bit odd to me.

yuri-sergiichuk: I was reading mine example as: "On event "TaskCreated", update "title" and "description".`

alexander-yevsyukov: Now I understand. To me, when we get down into a method body, we switch the “context” of the conversation, and start speaking in terms of the entity. And referencing entity properties via `e.` helps to see the input too.

alexander-yevsyukov: Also, I'm thinking of applying event properties in a code-generated way (e.g. via `inline` functions) like this:

```kotlin
@Apply fun event(e: TaskCreated) = apply(e);
```

But I'm not clear on all the details of the implementation, and do not want to resort to using reflection.

server/src/main/kotlin/io/spine/server/entity/extensions.kt
alexander-yevsyukov: This is so because the target branch `2.x-jdk8-master`, not `master` (with v1.7.0 in it).

server/src/test/java/io/spine/server/entity/StubTransaction.java
alexander-yevsyukov: Added #1346.

PR #1347, opened 2020-12-23
More Kotlin extensions

alexander-yevsyukov:
This PR improves Kotlin support adding `under()` function, which allows to write something like this:

```kotlin
under<Production> {
    use(productionStorageFactory)
    use(memoizingTracerFactory)
}

under<Tests> {
    use(testingStorageFactory)
}
```
Build scripts were also improved using `with()` function.


build.gradle.kts
```diff
@@ -52,16 +57,17 @@ buildscript {
             )
         }
     }
-}
+} // buildscript
```
yuri-sergiichuk, line 60: Hm. Do we really need this comment? It's not that big...

server/src/test/kotlin/io/spine/server/entity/ExtensionsTest.kt
```diff
@@ -33,8 +34,14 @@ internal class ExtensionsTest {
         val entity = createEntity()
         val str = randomString()
         entity.apply(str)
-        assertThat(entity.value())
-            .isEqualTo(str)
+        assertThat(entity.value()).isEqualTo(str)
+    }
+
+    @Test
+    fun `add 'under' function for 'ServerEnvironment' configuration`() {
```
yuri-sergiichuk, line 41: So we're not back-ticking anymore, right?

yuri-sergiichuk (approved):
@alexander-yevsyukov LGTM.

I do like the updated build config and the `with` usage a lot.

armiol (approved):
@alexander-yevsyukov LGTM, but I am looking forward to seeing your further conversation with @yuri-sergiichuk :)

server/src/test/kotlin/io/spine/server/entity/ExtensionsTest.kt
alexander-yevsyukov: We cannot backtick here because the whole method name is backticked. So here I use single quotes to still highlight the code-related names, which I believe is important for readability of the programming-related texts.

PR #1355, opened 2021-01-20
Unify lifecycle and version columns, eliminate `@SystemColumn`

armiol:
In the previous 2.x branch changes, a new concept was introduced in order to unify the process of storing and querying the data by the pre-defined attributes called "Columns". In particular, a new `Column` type hierarchy was created. The `Column` implementations are parameterized with the types of the column value and the queried record, allowing to use themselves in scope of the Query DSL. More information on this is available is the corresponding PR description (see #1305).

Prior to this changeset, the columns defining `Entity` lifecycle (i.e. `archived` and `deleted`) and the `version` column were declared for each `Entity` via `@SystemColumn` annotation and the complex machinery to wire the annotated `Entity` methods with the actual column types. As long as this code could exist only in `server` module (i.e. next to the `Entity` types), the definitions of lifecycle/version columns were not available for re-use on the client-side. In other words, client queries could use the copies of the lifecycle/version columns, and the whole mechanism worked only because the column names of the server column definitions (e.g. "version") were the same as the column names of their client-side copies.

Another issue was that in fact, each `Entity` stored in real life always had lifecycle/version columns declared, as `Aggregate`, `Projection` and `ProcessManager` are the only options for stored entities, and each of them is a descendant of types containing methods with the corresponding `@SystemColumn` annotations. However, the implementations of storage were written in the way, as if it was possible to declare an `Entity` without one or more lifecycle/version attributes. It brought unneccessary complexity into the code, as there were no real cases of that neither in the past nor in foreseeable future.

To eliminate the inconsistency and the code duplicates, this PR performs a few API changes:

1. `ArchivedColumn`, `DeletedColumn` and `VersionColumn` are now defined in `io.spine.client` package of `client` module. They are just a generic definitions of the queryable attributes and do not have any dependencies on the data source, such as `Entity`. They may be used in a query like this:

```java
   Query query =
       MyProjection.query()
                   // ...
                   .where(ArchivedColumn.is(), true)   // Only include archived.
                   .build(...);
   ... result = execute(query);
```

2. `Server` module provides wrappers for these column types, which hard-wire each of the generic columns with the actual data source. In particular, `Scanner` utility now returns all three lifecycle/version columns for any passed `Entity`. The changes are internal to the `server` module of framework and aren't visible through the public API.

3. `@SystemColumn` annotation was removed from the corresponding methods. The `SystemColumn` type was deleted.

The framework version is set to `2.0.0-SNAPSHOT.1`.


armiol:
@alexander-yevsyukov @dmdashenkov @yuri-sergiichuk PTAL.

client/src/main/java/io/spine/client/DeletedColumn.java
```diff
@@ -0,0 +1,83 @@
...
+ *      SomeProjection.query()
+ *                    // ...
+ *                    .where(DeletedColumn.is(), true)   // Only include deleted.
+ *                    .build(...);
+ *  ... result = execute(query);
+ * </pre>
+ *
+ * <p>This type is a singleton.
+ */
+public class DeletedColumn extends EntityLifecycleColumn<Boolean> {
```
yuri-sergiichuk, line 46: Can we make it `final`?

client/src/main/java/io/spine/client/VersionColumn.java
```diff
@@ -24,37 +24,52 @@
...
+ *                    // ...
+ *                  .where(Version.is(), previousVersion)
+ *                  .build(...);
+ *  ... = execute(query);
+ * </pre>
+ *
+ * <p>This type is a singleton.
  */
-final class VersionColumn extends CustomColumn<Entity<?, ?>, Version> {
+public class VersionColumn extends EntityLifecycleColumn<Version> {
```
yuri-sergiichuk, line 48: Can we make it `final`?

yuri-sergiichuk, line 48: Let's add the `nameAsString` static method for this column as we have in `Archived` and `Deleted`.

client/src/main/java/io/spine/client/DeletedColumn.java
```diff
@@ -0,0 +1,83 @@
...
+        return instance;
+    }
+
+    /**
+     * Returns a singleton instance of this type.
+     *
+     * <p>Serves as a more DSL-friendly alternative to {@link #instance() instance()}.
+     */
+    public static DeletedColumn is() {
+        return instance();
```
yuri-sergiichuk, line 71: There's probably no need to call the method. We can access the instance directly. I'd personally prefer such accessors. Otherwise, let's have the same styling for all columns.

server/src/main/java/io/spine/server/entity/storage/AsEntityRecordColumn.java
```diff
@@ -69,10 +71,32 @@ private AsEntityRecordColumn() {
...
+     *         the column to create a view for
+     * @param typeOfValues
+     *         the type of the column values
+     * @param getter
+     *         the getter returning the value of the resulting column view
+     * @param <V>
+     *         the type of the column values
+     * @return a view on the column
+     */
+    @SuppressWarnings("BadImport")       // `create` looks fine in this context.
```
alexander-yevsyukov, line 93: Assuming it's used only once, I think it would be less trouble not to use the bad import. Now it requires a reader to link the warning suppression, with the static import statement, and then with the usage in the `return` statement (also pipping at the method signature about the returned type). Just using `RecordColumn.create()` would be easier to understand, and one line less of code. Please consider.

server/src/main/java/io/spine/server/storage/MessageRecordSpec.java
```diff
@@ -153,5 +154,14 @@ protected I idValueIn(R source) {
     @Immutable
     @FunctionalInterface
     public interface ExtractId<R extends Message, I> extends Function<R, I> {
+
+        /**
+         * {@inheritDoc}
+         *
+         * This method differs from its parent by the fact it never returns {@code null}.
+         */
+        @CanIgnoreReturnValue
```
alexander-yevsyukov, line 163: Please have `@Override` first.

alexander-yevsyukov (approved):
LGTM, with a couple of minor requests.

server/src/main/java/io/spine/server/entity/storage/Scanner.java
```diff
@@ -119,15 +156,18 @@
...
-     * or {@code null} if the entity declares no columns
+     *         or {@code null} if the entity declares no columns
      */
     private static @Nullable Class<?> findColumnsClass(Class<? extends EntityState<?>> stateClass) {
         Class<?>[] innerClasses = stateClass.getDeclaredClasses();
         Class<?> columnClass = null;
         for (Class<?> aClass : innerClasses) {
-            if(aClass.getSimpleName().equals(COLS_NESTED_CLASSNAME)) {
+            if (aClass.getSimpleName()
+                      .equals(COLS_NESTED_CLASSNAME)) {
```
dmdashenkov, line 170: We should probably have a marker interface or an annotation for that.

dmdashenkov (approved):
LGTM with a single comment.

server/src/main/java/io/spine/server/entity/storage/EntityRecordStorage.java
```diff
@@ -294,48 +273,38 @@ private static boolean hasNoIds(Query<?, ?> query) {
...
             RecordQueryBuilder<I, EntityRecord> builder = query.toBuilder();
-            if(hasArchivedColumn) {
-                builder.where(archived.asRecordColumn(Boolean.class)).is(false);
-            }
-            if(hasDeletedColumn) {
-                builder.where(deleted.asRecordColumn(Boolean.class)).is(false);
-            }
+            builder.where(archived).is(false)
+                   .where(deleted).is(false);
             result = builder.build();
```
yuri-sergiichuk, line 282: Do we really need the builder variable here?

Let's simplify to smth like:
```java
result = 
    query.toBuilder()
        .where(archived).is(false)
        .where(deleted).is(false)
        .build();
```

server/src/main/java/io/spine/server/entity/storage/EntityRecordStorage.java
```diff
@@ -294,48 +273,38 @@ private static boolean hasNoIds(Query<?, ?> query) {
...
-        return FindActiveEntites.newBuilder(idType(), stateType(),
-                                            hasArchivedColumn, hasDeletedColumn);
+    private static <I> boolean hasNoLifecycleCols(Query<I, EntityRecord> query) {
+        Subject<I, ?> subject = query.subject();
+        ImmutableList<? extends QueryPredicate<?>> predicates = subject.predicates();
+        for (QueryPredicate<?> predicate : predicates) {
+            boolean hasColumn =
+                    predicate.parameters()
+                             .stream()
+                             .anyMatch((param) -> isLifecycleColumn(param.column()));
```
yuri-sergiichuk, line 294: What if we use `map(Param::column).anyMatch(::isLifecycleColumn)` instead?

server/src/main/java/io/spine/server/entity/storage/Scanner.java
yuri-sergiichuk, line 170: Let's invert. We'll avoid having an extra line then.

server/src/main/java/io/spine/server/storage/Columns.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Columns stored for a particular record along with their names.
+ *
+ * @param <R>
+ *         the type of the stored record
+ * @param <C>
+ *         the type of columns
+ */
+public final class Columns<R extends Message, C extends Column<R, ?>> implements Iterable<C> {
```
yuri-sergiichuk, line 49: We probably should mark this class as Internal while it has lots of public APIs.

yuri-sergiichuk (approved):
@armiol please consider some minor things below.

client/src/main/java/io/spine/client/VersionColumn.java
yuri-sergiichuk: We should also make it `final`.

client/src/main/java/io/spine/client/DeletedColumn.java
armiol: We want to preserve the behaviour of `instance()`, as stated in the docs. So, any transformations made — potentially, in future — to `instance()` would be the same for `is()`.

server/src/main/java/io/spine/server/entity/storage/EntityRecordStorage.java
armiol: @yuri-sergiichuk I tried your version of this piece and was looking at it for quite some time :) To my taste, a single `anyMatch()` is easier to understand than `map()` + `anyMatch()` with method references as their arguments. So, thanks for the suggestion, but maybe next time.

server/src/main/java/io/spine/server/storage/Columns.java
armiol: @yuri-sergiichuk Well, as it turns out, the truth is it should be killed :)

client/src/main/java/io/spine/client/VersionColumn.java
```diff
@@ -24,37 +24,52 @@
...
-    @Override
-    public Version valueIn(Entity<?, ?> entity) {
-        return entity.version();
+    /**
+     * Returns a singleton instance of this type.
+     *
+     * <p>Serves as a more DSL-friendly alternative to {@link #instance() instance()}.
+     */
+    public static VersionColumn is() {
+        return instance;
```
yuri-sergiichuk, line 73: So we should use `instance()` here over the direct access just as in the other columns.

server/src/main/java/io/spine/server/entity/storage/Scanner.java
```diff
@@ -26,24 +26,49 @@
...
-import io.spine.query.CustomColumn;
+import io.spine.client.ArchivedColumn;
+import io.spine.client.DeletedColumn;
+import io.spine.client.VersionColumn;
+import io.spine.query.Column;
 import io.spine.query.EntityColumn;
 import io.spine.server.entity.Entity;
 import io.spine.server.entity.model.EntityClass;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.jetbrains.annotations.NotNull;
```
yuri-sergiichuk, line 39: I guess you were going to use the one from the checker Framework.

server/src/test/java/io/spine/server/aggregate/given/repo/AggregateWithLifecycle.java
```diff
@@ -55,14 +55,18 @@ Evaluated handle(Evaluate commandMessage) {
...
+        if (namedAfterColumn(msg, DeletedColumn.instance())) {
             setDeleted(true);
         }
         builder().setValue(eventMessage.hashCode());
     }
+
+    private static boolean namedAfterColumn(String msg, EntityLifecycleColumn<?> column) {
+        return column
+                .toString()
+                .equalsIgnoreCase(msg);
```
yuri-sergiichuk, line 70: We can have a one-liner if we flip the args.

yuri-sergiichuk (approved):

armiol:
@dmdashenkov as discussed, I am writing an issue #1356. It takes more time than I originally anticipated, so I am merging this PR now, and will continue working towards the description as per the result of our vocal discussion.

PR #1366, opened 2021-05-19
Allow inherited methods with generics and Kotlin methods as signal handlers

dmdashenkov:
In this PR we adjust the method scanning process.

## Generics and inheritance

In Java, when overriding a method that accepts a parameter of a generic type with a method that accepts a parameter with a concrete type, the compiler, for technical reasons, generates a "bridge" method with a similar signature and the same annotations.
If such a method is made a signal handler (by adding an annotation), an error occurred, as the bridge method was also found to be a handler. For example:
```java
interface MyFancySubscriber<E extends EventMessage> {

    void on(E event);
}

class MyFancyImpl implements MyFancySubscriber<ProjectCreated> {

    @Override
    @Subscribe
    void on(ProjectCreated event) { ... }

    // Here, the compiler generates a bridge method:
    //
    // @Override
    // @Subscribe
    // void on(EventMessage event) {
    //     on((ProjectCreated) event);
    // }
}
```

Before this PR, Spine used to fail to scan such a class, as `on(EventMessage)` is not a valid subscriber. Now, we ignore all the compiler-generated methods.

## Kotlin and package-private

When writing a signal handler class (an abstract subscriber, handler, reactor, or commander, or an entity) in Kotlin, the developer always had to deal with console warnings informing them that the used access modifier is incorrect and they should use "package-private" instead. Kotlin, of course, does not have "package-private" visibility.

Starting from this PR, Kotlin users will be able to use `internal` wherever Spine requires "package-private". That is, all the signal handler methods, except event appliers, which ought to be private regardless of the source language.


server/src/main/java/io/spine/server/model/J2Kt.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Optional;
+import java.util.function.Predicate;
+
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * A utility for working with Java class reflective representations.
+ */
+final class J2Kt {
```
armiol, line 48: As discussed, let's move this utility to some place in `base`, as it hardly belongs to `server.model`.

server/src/main/java/io/spine/server/model/J2Kt.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Prevents the utility class instantiation.
+     */
+    private J2Kt() {
+    }
+
+    static Optional<KCallable<?>> findKotlinMethod(Method javaMethod) {
```
armiol, line 56: Again, as discussed, let's document this method and, also, check for `null` values.

armiol (changes requested):
@dmdashenkov please see my comments.

dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #1367, opened 2021-05-25
Extension function `alter`

dmdashenkov:
This PR introduces an extension function `alter(..)` for `TransactionalEntity`-s.

`alter(..)`, just like her sister `update(..)` allows the user to change properties of the entity's builder without the hassle of calling Java-style setter methods.

Unlike `update(..)`, alter does not return any value, allowing it to be used as an expression body in handler methods. Example:
```kotlin
internal class DailyReport : Projection<..> {

    @Subscribe
    internal fun on(e: WorkLogged) = alter {
        workMinutes += e.minutes
    }
}
```

It's recommended to keep the call to `alter(..)` at the same line with the method signature, so that the whole construction resembles a regular method.

### DIFF
--- a/buildSrc/src/main/groovy/update-gh-pages.gradle
+++ b/buildSrc/src/main/groovy/update-gh-pages.gradle
@@ -55,6 +55,21 @@ ext {
     repositoryTempDir = Files.createTempDirectory("repoTemp")
 }
 
+/**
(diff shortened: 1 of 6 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #1371, opened 2021-06-17
More `public` API

armiol:
This changeset opens some of API endpoints to `public`.

The main reason for these changes is the experience we gained through using Spine 2.x in the real-world apps.

In particular,

* `io.spine.client.MessageConsumer` was made `public`, as long as it is referenced from the public API of `io.spine.client.Client`.
* Record specifications now expose their column definitions. It is needed in order to pre-configure storage servers in our `rdbms` library. Previously, the column definitions weren't available via `public` API.

Also, some of the missed use-cases were tested.

The version of the library has been set to `2.0.0-SNAPSHOT.26`.


armiol:
@dmdashenkov PTAL.

dmdashenkov (approved):

PR #1375, opened 2021-07-26
Smarter signal handlers

### LINKED ISSUE #1342: `EventReactorSignature` allows to declare a `RejectionMessage` as a returning type
According to the rules applies to `@React`or methods, one must declare it according to certain rules. In particular, the returning type is expected to be a descendant of an `EventMessage` — it is also possible to declare an `Iterable` or an `Optional` of `EventMessages`, but the gist stays the same. An `EventReactorSignature` is the framework type responsible for this.

The problem with this approach is that the rejections generated by Spine compiler are marked as `RejectionMessage`s. In turn, this interface is derived from `EventMessage`. Therefore, all rejection messages are `EventMessage`s, and may pass through the gate verifying signatures of `@React`ing methods.

In scope of this issue we should prohibit the usage of rejection messages as returning types of `@React`or handlers.

### LINKED ISSUE #1373: Only `@Subscribe`-r methods support `@Where` filtering
In event subscribers, users can define a field filter, which only allows a certain subset of events of the given type though:

```java
@Subscribe
void on(@When(field = "direction", equals = "SW") LightlingHit event) { ... }
```

However, `@React`-ors and other signal handling methods do not support this filtering.
Worst of all, the filter is silently swallowed, and all the events are dispatched to the method.

We should support field filters in all signal handlers.

dmdashenkov:
In this PR we improve the API and behaviour of signal handler methods.

# Argument filters

Fixes #1373.

Filtering events which are dispatched to a handler is now possible for all kinds of event handler methods. Previously, the only kind of handlers which allowed that was event subscriber.

Field filtering is **not** allowed for:
 - entity state subscriptions (for a technical reason);
 - methods that accept commands (not to confuse users — a command **must** always be handled);
 - event appliers in Aggregates (events are applied in-place to the aggregate; if the user wants to handle two cases differently, there probably should be two event types, not one).

Here is an example of filtering events using the `@Where` annotation:
```java
final class ProjectAggregate /* ... */ {

    // ...

    @React
    private ProjectPromoted onInternal(@Where(field = "owner.organization.id", equals = "internal-org"))
                                      ProjectCreated event) {
        // Internal projects require special treatment.
    }
}
```

# Reactors and rejections

Fixes #1342.

Handler methods that return produce events now test that the returned event is not a rejection.

Previously, it was technically possible to emit rejection events as regular events.
This is confusing and may lead to unexpected behaviour, so now we check that the users don't do it.

# Deprecated fields in annotations

This is a **BREAKING CHANGE**.

Fields `external` and `byField` from handler method annotations have been deprecated for a long time.
Now we delete them.

The recommended API is `@External` and `@Where` method argument annotations.


server/src/test/java/io/spine/server/model/FilterTest.java
```diff
@@ -0,0 +1,176 @@
...
+            storageFactory.close();
+            ServerEnvironment.instance().reset();
+            dropAllModels();
+        }
+
+        @Test
+        void beforeDelivery() {
+            BoundedContextBuilder context = BoundedContextBuilder
+                    .assumingTests()
+                    .add(new ProjectTasksRepository());
```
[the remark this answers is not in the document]
dmdashenkov: @sonatype-lift help

dmdashenkov:
@armiol, PTAL.

server/src/test/java/io/spine/server/model/FilterTest.java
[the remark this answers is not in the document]
dmdashenkov: @sonatype-lift ignore

dmdashenkov:
@armiol, PTAL again.

server/src/main/java/io/spine/server/Ignored.java
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * <p>The ignoring reason points towards the {@linkplain Where argument filters}.
+     *
+     * @param handler
+     *         the class of the event handler
+     * @param event
+     *         the dispatched event
+     * @return {@code ignored} outcome
+     */
+    public static DispatchOutcome ignored(ModelClass<?> handler, EventEnvelope event) {
```
armiol, line 61: Let's check the args for not being `null`.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -347,6 +347,12 @@ public final void dispatch(CommandEnvelope cmd) {
                                       .toHandler(id));
     }
 
+    @Override
+    public final boolean canDispatch(CommandEnvelope envelope) {
+        // Always dispatch commands with the correct type.
```
armiol, line 352: Please move this to the Javadocs section.

server/src/main/java/io/spine/server/command/AbstractCommander.java
```diff
@@ -107,11 +108,11 @@ public void dispatch(CommandEnvelope command) {
 
     @Override
     public void dispatchEvent(EventEnvelope event) {
-        CommandReactionMethod method = thisClass.commanderOn(event.messageClass());
-        DispatchOutcomeHandler
-                .from(method.invoke(this, event))
+        Optional<CommandReactionMethod> method = thisClass.commanderOn(event);
```
armiol, line 111: The question here is why the runtime execution has come to this point while the filtering rules may have not matched. I would think it should have been figured out at the dispatching stage.

It's not too scary with the commanders, but with `Entity` instances it's much worse, as we are going to load them from their storage.

Let's of course keep returning `Optional` here, but in case it's `empty`, there should be some consequences. Not silence.

server/src/test/java/io/spine/server/model/given/filter/ModSplitCommandAggregate.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * An invalid aggregate declaration. Assign methods cannot have {@link Where} filers.
+ */
+public final class ModSplitCommandAggregate extends Aggregate<String, ModProject, ModProject.Builder> {
+
+    private static final String FANCY_ID = "1";
+    private static final String INTERNAL_FANCY_ID = "very fancy project";
+
+    @Assign
+    ModProjectCreated handleFancy(@Where(field = "id", equals = FANCY_ID)
```
armiol, line 46: I thought we agreed on removing this option in this version.

server/src/test/java/io/spine/server/model/given/filter/ModSplitEventAggregate.java
```diff
@@ -0,0 +1,63 @@
...
+                .build();
+    }
+
+    @Apply
+    private void onFancy(@Where(field = "id", equals = INTERNAL_FANCY_ID) ModProjectCreated e) {
+        builder().setProjectId(FANCY_ID);
+    }
+
+    @Apply
+    private void on(ModProjectCreated e) {
```
armiol, line 60: I am interested in knowing what happens if

a) There will be just two appliers like these:

```java
@Apply
private void hanky(@Where(field = "id", equals = "hanky") ModProjectCreated e) { ... }

@Apply
private void panky(@Where(field = "id", equals = "panky") ModProjectCreated e) { ... }

// NOTE: no "general" option.
```

Will it blend?

or b) — just two appliers like this one

```java
@Apply
private void one(@Where(field = "id", equals = "copy-paste") ModProjectCreated e) { ... }

```java
@Apply
private void another(@Where(field = "id", equals = "copy-paste") ModProjectCreated e) { ... }

// NOTE: copy-pasted methods with different names, but the same `equals` part!
```

Maybe, this one blends?

armiol (changes requested):
@dmdashenkov please see my comments.

server/src/test/java/io/spine/server/model/given/filter/ModSplitCommandAggregate.java
dmdashenkov: This is test env for a negative case.

Such a definition causes an error.

server/src/test/java/io/spine/server/model/given/filter/ModSplitEventAggregate.java
dmdashenkov: Same as above.

Al

alexander-yevsyukov:
>  `private void onIntenal`

Assuming the comment which follows the method signature, I resume, the method was meant to be named `onInternal()`

core/src/main/kotlin/io/spine/core/Where.kt
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Filters events delivered to a handler method.
+ *
+ * To apply filtering to an signal handler method, annotate the first parameter of the method.
+ *
+ * For example, the following method would be invoked only if the owner of the created
+ * project is `admin@ackme.net`:
+ * ```
+ *   @Assign
+ *   ProjectDeletedForever on(@Where(field = "owner.email", equals = "admin@ackme.net")
```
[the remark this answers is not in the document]
dmdashenkov: Updated to a `@React` method.

dmdashenkov:
@armiol, PTAL again.

armiol (approved):

PR #1376, opened 2021-07-27
Improve the performance of gRPC-related code in concurrent environments

armiol:
This changeset updates the server part of Spine. 

In particular:

* Pushing the updates, errors and completion notices through the `Subscription` observer is now `synchronized`. Previously, this piece of code was using the underlying gRPC mechanism directly, which wasn't thread-safe. To do so, a new `ThreadSafeObserver` was introduced, which acts as a `synchronized` delegate.

* It is now possible to configure the `Executor` for the gRPC server inside of `GrpcContainer`. Previously, only gRPC-default values were used — namely, `directExecutor` for in-process mode, and some hard-coded shared executor for the "remote" mode.

The library was updated with the latest `config` code.

The version is now set to `2.0.0-SNAPSHOT.42`.


server/src/main/java/io/spine/server/GrpcContainer.java
```diff
@@ -117,15 +120,41 @@ private GrpcContainer(Builder builder) {
     /**
      * Starts the service.
      *
+     * <p>A gRPC-default executor is used for the server routines. I.e., in-process gRPC server
```
dmdashenkov, line 123: Please reformat into a list:
```
A gRPC-default executor is used for the server routines:
 1. In-process gRPC server is going to ...
 2. A server exposed at a port will ...
```

It reads clearer and avoids the awkward "I.e." at the beginning of a sentence.

server/src/main/java/io/spine/server/stand/SubscriptionCallback.java
```diff
@@ -47,8 +47,10 @@
      */
     static SubscriptionCallback forwardingTo(StreamObserver<SubscriptionUpdate> observer) {
         return update -> {
-            checkNotNull(update);
-            observer.onNext(update);
+            synchronized (observer) {
+                checkNotNull(update);
+                observer.onNext(update);
```
dmdashenkov, line 52: Let's use a dedicated object to synchronize on.

Also, it's not like we promise timestamp-sorter delivery. We might as well use a non-fair reentrant lock. The promise of it is that the synchronization overhead is minimized.

From the codebase, it seems like the relationship between the `observer`s and `SubscriptionCallback`s is 1:1. So, it shouldn't matter if we choose to lock on the observer or an object (or a `Lock`) that belongs to the callback.

alexander-yevsyukov:
@armiol 

> The version is now set to 2.0.0-SNAPSHOT.27

In #1379 I'm advancing the version to ...40. So that it's in sync. numbers with `base` and `time`. Just FYI.

armiol:
@dmdashenkov PTAL, again.

client/src/test/java/io/spine/client/ThreadSafeObserverTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static io.spine.testing.TestValues.nullRef;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@DisplayName("`ThreadSafeObserver` should")
+class ThreadSafeObserverTest {
+
+    @Test
+    @DisplayName("not accept `null`s in its ctor")
+    void notAcceptNullsInCtor() {
+        assertThrows(NullPointerException.class, () -> new ThreadSafeObserver<>(nullRef()));
```
dmdashenkov, line 50: Please use `NullPointerTester`.

dmdashenkov (approved):
@armiol, LGTM with single comment.

PR #1380, opened 2021-08-04
Protected `@ContractFor(..)` handler methods

dmdashenkov:
In this PR we allow to declare `protected` handler methods, but only if the method is overridden from a superclass.

This will enable users to create hierarchies of handling types without seeing annoying log warnings.
The experience of ProtoData shows that there are advantages to such hierarchies.

In order to implement a method from a superclass, the base method must be declared as `@ContractFor(handler = ...)` with a matching handler type.

Here is a simple example with an event subscriber method.

The abstract base that defines the contract method:
```java
public abstract class ProjectCreatedSubscriber extends AbstractEventSubscriber {

    @ContractFor(handler = Subscribe.class)
    protected abstract void on(ProjectCreated event);

    // Utility methods required by descendants.

    protected final UserId ownerFrom(Project project) { /* ... */ }
    protected final boolean isAdmin(UserId user) { /* ... */ }
}
```
The implementation:
```java
public final class LoggingSubscriber extends ProjectCreatedSubscriber {

    @Subscribe
    protected void on(ProjectCreated event) {
        // ...
    }
}
```

Note the use of annotations. The contract method has the `@ContractFor` annotation, while the concrete method has the appropriate signal handler annotation. The `handler` of the `@ContractFor` matches the handler annotation type.

For commands, there cannot be more than one handler method per command type. However, a generalized abstract base can still be useful sometimes.
Consider the following abstract commander:
```java
public abstract class SplittingCommander<C extends CommandMessage> extends AbstractCommander {

    @ContractFor(handler = Command.class)
    protected abstract Pair<FreezeProject, NotifyAdmin> on(C cmd);
}
```
And here are the implementations:
```java
public final class PaymentSafety extends SplittingCommander<InvalidatePaymentMethod> {

    @Command
    protected abstract Pair<FreezeProject, NotifyAdmin> on(InvalidatePaymentMethod cmd) {
        // ...
    }
}

// ...

public final class OrgCleansing extends SplittingCommander<DivideOrganization> {

    @Command
    protected abstract Pair<FreezeProject, NotifyAdmin> on(DivideOrganization cmd) {
        // ...
    }
}
```

The axiomatic rule regarding a single handler for one command is preserved, while the abstract base is used to define the second part of the contract: which signals are emitted as the result of the command.

Same rules apply also to `@React` and `@Assign` methods. 
However, `@Apply` methods must always be private, and, therefore, there can be no applier contracts.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL again.

version.gradle.kts
```diff
@@ -40,7 +40,7 @@
 /**
  * Version of this library.
  */
-val coreJava = "2.0.0-SNAPSHOT.27"
+val coreJava = "2.0.0-SNAPSHOT.28"
```
armiol, line 43: Please make sure to advance the version before merging.

core/src/main/java/io/spine/core/ContractFor.java
```diff
@@ -0,0 +1,59 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Marks an abstract handler method.
+ *
+ * <p>When there is a need to define an abstract base for a handler method, the base method must be
```
armiol, line 39: Can we please also have an example in this portion of the docs?

armiol (approved):
@dmdashenkov LGTM with a few comments.

Issue #1328, opened 2020-11-19
[2.x] Sharpen the contract of `QueryableRepository`


armiol:
At the moment, the newly introduced `QueryableRepository` (see 2.x branch) speaks only in terms of `EntityRecord` search. However, all `RecordBasedRepository` instances are now capable of searching for `Entity` instances by the passed `EntityQuery`.

We need to address this discrepancy by defining the `QueryableRepository` API on searching for `Entity` instances as well.

armiol:
Addressed by #1382.

PR #1382, opened 2021-08-18
`QueryableRepository` to respond to `EntityQuery`

armiol:
This changeset addresses #1328.

While the original issue speaks of appending the `QueryableRepository` with a method to return `Entity` instances, this changeset appends it with an endpoint to search for `Entity` states. 

The reason behind it is that is it hardly possible to make the `Aggregate` bulk search efficient — as it would involve executing a number of very different queries all at once.

Therefore, it is just possible to search for `Entity` states via `EntityQuery` now, in addition to the combination of `TargetFilters` and `ResponseFormat`.

The library version is set to `2.0.0-SNAPSHOT.43`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.42`
+# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.43`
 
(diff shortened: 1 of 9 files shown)


armiol:
@dmdashenkov PTAL.

dmdashenkov (approved):

PR #1385, opened 2021-09-01
Allow parameter type covariance in methods implementing `@ContractFor`

dmdashenkov:
`@ContractFor`, introduced in a past PR, marks a base method for handler methods. This is useful when building little domain-specific hierarchies out of entities/abstract handlers.

In this PR we allow the methods that implement (or override) a contract method to specify a more precise parameter type. For example, the contract method may accept an `EventMessage`:
```java
@ContractFor(handler = Subscribe.class)
protected abstract void on(EventMessage e);
```
and the implementation specifies the concrete method:
```java
@Override
@Subscribe
protected void on(ProjectCreated e) {
    // ...
}
```

### DIFF
--- a/server/src/test/java/io/spine/server/event/model/EventReactorSignatures.java
+++ b/server/src/test/java/io/spine/server/event/model/EventReactorSignatures.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2021, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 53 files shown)


server/src/test/java/io/spine/server/event/model/EventReactorSignatures.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public final class EventReactorSignatures {
+
+    /**
+     * Prevents the utility class instantiation.
+     */
+    private EventReactorSignatures() {
+    }
+
+    public static MethodSignature<?, ?> createForTest() {
```
alexander-yevsyukov, line 45: Please document and annotate with `@VisibleForTesting`.

alexander-yevsyukov: Also, having the utility class just for this feels as a little overkill. Don't we have other places in the package where such a thing would look more natural?

alexander-yevsyukov (changes requested):
Please see my comments. 

Also, I'm not sure about the naming. E.g. `handler = React.class` does not feel right to me. A handler is the method. `React.class` is the class of the annotation. Let's discuss vocally.

@armiol, FYI.

alexander-yevsyukov:
```
@Subscribe
protected void on(ProjectCreated e) {
    // ...
}
```
We probably need the `@Override` annotation below `@Subscribe`. It would make the example more obvious.

alexander-yevsyukov:
@dmdashenkov, I gave this some thought, and I must admit I don't understand the idea behind the _previous_ PR, which introduced this feature. Here why.

Message-handing methods are invoked indirectly via late-binding specified by their declarations. They are not called by end user's code directly. It's how OOP of Alan Key's differs from one Bjarne Stroustrup.

Introducing the `ContractFor` feature you suggest a mix. Since the feature is based on `abstract` methods, it's assumed that the method is called from the class which declares it. Or, the method is used from the same package, exploiting the way because of how `protected` treated in Java.

But why do we need such a trick it in the first place?

alexander-yevsyukov (approved):
LGTM

PR #1386, opened 2021-09-01
Set the license type explicitly

armiol:
In this changeset, we attempt to use GitHub's mechanism to associate a license type with this repository explicitly.

To do that, for some reason, we must use the license template. This PR puts our existing license file in accordance with this template for Apache 2.0 license.


alexander-yevsyukov (approved):
LGTM

PR #1390, opened 2021-09-02
Fix readme

alexander-yevsyukov:
This PR:
  * Updates the readme file with correct version numbers.
  * Enables 0.05% threshold on code coverage to avoid false positives on rounding issues in Codecov.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ the [Spine Examples][spine-examples] GitHub organization.
 
 Building Spine modules and libraries requires JDK 8. 
 
-The support of JDK 11 is planned for the [version 2.0][v2] of the framework.
+The migration of the codebase to JDK 11 is planned for the [version 2.0][v2] of the framework.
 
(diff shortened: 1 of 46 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ the [Spine Examples][spine-examples] GitHub organization.
 
 Building Spine modules and libraries requires JDK 8. 
 
-The support of JDK 11 is planned for the [version 3.0][v3] of the framework.
+The support of JDK 11 is planned for the [version 2.0][v2] of the framework.
```
yuri-sergiichuk, line 38: We actually work with Java 11 and Spine 1.7.x. By support, do we mean that the framework is going to be compiled with Java 11 source code target?

yuri-sergiichuk (approved):
@alexander-yevsyukov LGTM. I have one comment though which requires a bit of clarification I guess.

PR #1399, opened 2021-09-26
[1.7.6] Avoid clash between Catch-up Processes at runtime

armiol:
This PR addresses two issues discovered in production:

1. Catch-up processes may be incorrectly initialized with non-matching repository instances at run-time. To prevent that, now, a `CatchUpProcess` always loads the required `ProjectionRepository` from its `BoundedContext` according to the type URL of the projection state. That value is passed at run-time with each event, so it is guaranteed that the loaded repository will match the state of the catch-up process.

2. `MirrorProjection` could participate in catch-up. In fact, catching up mirrors makes very little sense, as they are not operating on top of domain events, but rather they subscribe to system events — which are not available for recall. This changeset prevents `MirrorProjection`s from registering their catch-up processes, and from starting the catch-up.

The library version is set to `1.7.6`.

NOTE: this PR is only made to illustrate the changes made to the codebase in relation to `1.7.5`. Actual merge will NOT happen, as for `1.x` versions we do not operate with `master` branch.

### DIFF
--- a/server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
+++ b/server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
@@ -607,9 +607,9 @@ public final class CatchUpProcess<I>
     /**
      * {@inheritDoc}
      *
-     * <p>Tells if the passed envelope can be dispatched to this instance of the process
-     * by verifying that the projection type URL passed with the event matches the one of the
-     * projection repository configured for this instance.
+     * <p>Ensures that the passed signal is an instance of {@link CatchUpSignal} and hence
+     * is suitable for the upcoming {@linkplain #route(EventEnvelope) ID extraction
+     * at the routing stage}.
      */
     @Override
     public boolean canDispatch(EventEnvelope envelope) {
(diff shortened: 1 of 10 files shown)


server/src/main/java/io/spine/server/delivery/CatchUpProcess.java
```diff
@@ -570,29 +576,45 @@ private EventStreamQuery toEventQuery(CatchUp.Request request,
...
+     *
+     * <p>Such a mechanism ensures that this process always has its ID and underlying
+     * repository matching each other.
+     *
+     * <p>The result of execution is cached, so that next execution goes faster.
+     */
+    private ProjectionRepository<I, ?, ?> repository() {
+        String type = builder().getId()
+                               .getProjectionType();
+        synchronized (repos) {
```
dmdashenkov, line 592: Can we use a concurrent map of some sort with a `computeIfAbsent(..)`?

dmdashenkov (approved):
LGTM with a small suggestion.

armiol:
Closing this PR, as it is now approved.
See [1.7.6](https://github.com/SpineEventEngine/core-java/releases/tag/v1.7.6).

PR #1409, opened 2021-10-28
Make `BlackBox` provide a `Client` linked to the context under-the-test

yevhenii-nadtochii:
It's a migration of #1407  to the 2.x branch.

**Problem**

Previously, in scope of 2.x versions it was impossible to obtain a `Client` to some `BlackBox` context. That was a problem, since in a multi-Bounded-Context environment, the contexts often communicated with each other with the `Client`s. And the black-boxing of either clients made it very difficult to run the integration flow — as no proper `Client` could be created for the black-boxed Bounded Context.

**Solution**

It is now possible to create an in-process `Client` to any `BlackBox`:
```java

BlackBox box = ...

// Creates the instance with the same `TenantId` as the one currently set in the `box`.
Client client = box.clients().withMatchingTenant();

// For multi-tenant Contexts only, it is also possible to create a `Client` with a specific tenant:
Client client = box.clients().create(myTenant);
```

The created `Client` instances are running through the in-process gRPC server. The lifecycle of this server is similar to the lifecycle of the associated `BlackBox`: once the box is closed, the server is closed as well.

**Other Changes**

The versions of the dependencies and the own version of `core-java` modules were set to the latest available.


armiol:
@yevhenii-nadtochii PTAL. Should you have any comments, let's discuss them vocally.

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBox.java
```diff
@@ -456,15 +465,26 @@ private BlackBoxSetup setup() {
...
      *
      * <p>Instead of a checked {@link java.io.IOException IOException}, wraps any issues
      * that may occur while closing, into an {@link IllegalStateException}.
      */
     @Override
     public final void close() {
+        if(!isOpen()) {
+            return;
+        }
+
```
dmdashenkov, line 484:
```suggestion
        if (!isOpen()) {
            return;
        }
```

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBoxClients.java
```diff
@@ -0,0 +1,87 @@
...
+
+import io.spine.client.Client;
+import io.spine.core.TenantId;
+import io.spine.testing.client.TestActorRequestFactory;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Creates the {@link Client}s to a particular {@link BlackBox} instance.
```
dmdashenkov, line 37: Please check the grammar around articles here.
I'd expect:
```
Creates Clients for a ...
```

testutil-server/src/main/java/io/spine/testing/server/blackbox/ClientSupplier.java
```diff
@@ -0,0 +1,171 @@
...
+import static io.spine.util.Exceptions.illegalStateWithCauseOf;
+import static java.util.Objects.isNull;
+
+/**
+ * Creates {@code Client} instances for the passed Bounded Context.
+ *
+ * <p>As {@code Client} requires a gRPC server to be connected to, the supplier assembles
+ * and starts a {@link GrpcContainer}. The server and the created clients are working in-process.
+ */
+class ClientSupplier implements Closeable {
```
dmdashenkov, line 56: This should be called `Clients` so that it's coherent with the `BlackBoxClients`. Maybe, `ClientFactory`, but then the `BlackBoxClients` should also be renamed.

This type is too complicated to be a "supplier".

yevhenii-nadtochii (commented):
Please see my questions

testutil-server/src/main/java/io/spine/testing/server/blackbox/BlackBox.java
```diff
@@ -673,4 +695,13 @@ public SubscriptionFixture subscribeTo(Topic topic) {
         result.activate();
         return result;
     }
+
+    /**
+     * Returns a factory of {@link io.spine.client.Client Client}s
+     * to this instance of {@code BlackBox}.
+     */
+    public BlackBoxClients clients() {
+        BlackBoxClients result = new BlackBoxClients(this, clientSupplier);
```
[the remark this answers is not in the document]
armiol: As discussed, that is because we want to explicitly close the factory of clients in our `BlackBox.close()` method. And having both `blackBoxClients` and `clientSupplier` seems redundant. 

Besides, `BlackBoxClients` consumes little to none resources by itself. It's rather a method object, which just groups and exposes the part of the `BlackBox`'s public API.

PR #1410, opened 2021-10-29
Prototype `javac-args.gradle` in Kotlin

yevhenii-nadtochii:
## Note

This is not going to be merged as approved. A separate PR into `config` repository would be created.

## What have been done

1. `javac-args.gradle` was split into `JavaCompileConfig.kt` and `ErrorProneConfig.kt`;
2. `ErrorProne` was added on a `buildSrc` classpath to have `ErrorProneConfig.kt` typed. Thus, it'd be on a classpath of each build file and no version should be specified in `plugins.apply` block. Old scripts that specify version would fail, it's a breaking change.


buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/ErrorProneConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+ * tasks {
+ *     withType<JavaCompile> {
+ *         configureErrorProne()
+ *     }
+ * }
+ * ```
+ */
+object ErrorProneConfig {
+
+    private val CODE_ANALYZING_FLAGS = listOf(
```
alexander-yevsyukov, line 49: `CODE_ANALYZING_FLAGS` — does this name have to be that long? Can it be just `flags` or, better, `args`?

yevhenii-nadtochii: Yes, It can. 
But I would like to go with more specific name as long as we add them to the list.
```
.errorproneArgs
.addAll(CODE_ANALYZING_FLAGS)

// => Let's add flags that tune code analysis to the list of arguments.
```
```
.errorproneArgs
.set(ARGUMENTS)

// => Let's set the arguments for ErrorProne.
```

The original script in Groovy was adding them so I have chosen the first option.

buildSrc/build.gradle.kts
```diff
@@ -69,4 +77,5 @@ dependencies {
     implementation("com.google.guava:guava:$guavaVersion")
     api("com.github.jk1:gradle-license-report:$licenseReportVersion")
     implementation("org.ajoberstar.grgit:grgit-core:${grGitVersion}")
+    implementation("net.ltgt.gradle:gradle-errorprone-plugin:${errorProneVersion}")
```
yevhenii-nadtochii, line 80: One thing I would like to point out.
In order to configure `ErrorProne` in a type-safe manner we need it on a classpath of `buildSrc`.  Thus, it'd be on a classpath of each build file and no version should be specified in `plugins.apply` block.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/ErrorProneConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+import net.ltgt.gradle.errorprone.errorprone
+import org.gradle.api.Project
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.kotlin.dsl.withType
+
+/**
+ * Configures `ErrorProne` plugin.
+ *
+ * ```
```
armiol, line 39: Please don't leave this sample hanging without a foreword.

Also, it would be nice to describe what the utility does. It says in its name that it configures ErrorProne, so let's be more specific in the documentation.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/ErrorProneConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+     * it is actually configured through the [JavaCompile] task.
+     */
+    fun JavaCompile.configureErrorProne() {
+        options.errorprone
+            .errorproneArgs
+            .addAll(CODE_ANALYZING_FLAGS)
+    }
+}
+
+// Pick up the API that fits best to your code base.
```
armiol, line 78: There is no "you" in the documentation.

Let's have a documentation section at the start of this file describing the available options.

```
/**
 * This file declares ......
 *
 * ...
 */
```

buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/ErrorProneConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Applies [ErrorProneConfig] to all [JavaCompile] tasks in this container.
+ */
+fun TaskContainer.configureErrorProne() {
```
armiol, line 83: I suggest that there should be just one public entry point in this file. The rest of them should be either `private` or `internal`. Currently, you have all of them declared public. Please adjust.

armiol, line 83: On another matter. The implementation of this extension duplicates the one you have listed as an example for `ErrorProneConfig`. So it's not clear which one a developer should pick. Let's straighten the way to use this utility by narrowing down the number of options to one.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/JavaCompileConfig.kt
```diff
@@ -0,0 +1,120 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.internal.gradle.javacompile
```
armiol, line 27: All comments I left for `ErrorProneConfig` are applicable here.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/ErrorProneConfig.kt
```diff
@@ -0,0 +1,114 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.internal.gradle.javacompile
```
armiol, line 27: To me, that's `io.spine.internal.gradle.javac`, as it's not about the compilation of Java code in general, but rather about configuring the `javac` tool.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/ErrorProneConfig.kt
armiol: As long as this is a private constant, and it virtually contains _everything_ that we ever add, we should aim for brevity. Please adjust your naming.

armiol (changes requested):

buildSrc/build.gradle.kts
```diff
@@ -53,6 +53,14 @@ val googleAuthToolVersion = "2.1.2"
 val licenseReportVersion = "1.16"
 val grGitVersion = "3.1.1"
 
+/**
+ * The version of ErrorProne Gradle plugin.
+ *
+ * Always use the same version as the one specified in `io.spine.internal.dependency.ErrorProne`.
+ * Otherwise, when testing Gradle plugins, clashes may occur.
```
alexander-yevsyukov, line 60: We're digging deeper into the needs of manual verification of versions. Someday it would break. It's not the matter of “if”, it's a matter of “when”.

How can we avoid this?

@armiol, FYI.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javacompile/ErrorProneConfig.kt
alexander-yevsyukov: I concur. In most cases people need one simple way of doing things. 

Only in very care cases people want and _really_ need “flexibility“. This is definitely not the case.

alexander-yevsyukov (changes requested):
Almost LGTM. Please address my comments.

buildSrc/build.gradle.kts
yevhenii-nadtochii: Well, I have little experience with Gradle but still can say what I think :)

As far as I understand `buildSrc/src/.../dependency` supplies typed versions to build files in concrete projects. `buildSrc/build.gradle.kts` also needs that versions but can't access types for assembling of which it's responsible. 

1. Straight patch solution. Versions that are also needed in `buildSrc/build.gradle.kts` are to be put in something like `buildSrc/versions.properties`. Affected types from `buildSrc/src/.../dependency` and `buildSrc/build.gradle.kts` now use that file as a source of versions;
2. Big change. Gradle has an incubating feature for [version catalogs](https://melix.github.io/blog/2021/03/version-catalogs.html).

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,80 @@
...
+
+import org.gradle.api.GradleException
+import org.gradle.api.JavaVersion
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.process.CommandLineArgumentProvider
+
+/**
+ * The knowledge that is required to set up `javac`.
+ */
+internal object JavacConfig {
```
armiol, line 37: I would assume this is `private`.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * ```
+ * tasks {
+ *     withType<JavaCompile> {
+ *         configureJavac()
+ *     }
+ * }
+ * ```
+ */
+fun JavaCompile.configureJavac() {
```
armiol, line 66: And this goes first in this file. 

Beside the documentation it already has, it should document a few more points.

* Explain the idea of running with JDK 8.
* Tell what it does to `javac`.

The idea is that when someone navigates to this file, the first thing she sees is the entry point. Which is documented in a way that the gist of what's going to happen is already there. 

Now, it's not quite so: the private type goes first, the entry point goes last, and — for some reason — ping-pongs the reader back the the first type.

This comment applies to `ErrorProne` implementation as well, as it's pretty much the same as this one.

armiol (changes requested):
@yevhenii-nadtochii I like the idea of extending `JavaCompile`. This really minimizes the extra efforts.

However, we are not there with the code yet. Please see my comments.

Let's discuss the point you have raised regarding the `Toolchain` and all, once we are back to the office.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,91 @@
...
+ *  1. Ensuring JDK 8 would be used for compilation;
+ *  2. Specifying arguments for `javac` invocations;
+ *  3. Specifying the `UTF-8` encoding to be used when reading source files.
+ *
+ * ### Supporting other JDKs.
+ *
+ * `Spine Event Engine` can be built with JDK 8 only. Supporting JDK 11 and above
+ * at build-time is planned in 2.0 release.
+ *
+ * &nbsp;
```
alexander-yevsyukov, line 48: Let's not use in-place formatting like this.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * ### Here's an example of how to use it:
+ *
+ *     tasks {
```
alexander-yevsyukov, line 52: How about fencing this code with triple backtics?

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,91 @@
...
+ *  3. Specifying the `UTF-8` encoding to be used when reading source files.
+ *
+ * ### Supporting other JDKs.
+ *
+ * `Spine Event Engine` can be built with JDK 8 only. Supporting JDK 11 and above
+ * at build-time is planned in 2.0 release.
+ *
+ * &nbsp;
+ *
+ * ### Here's an example of how to use it:
```
alexander-yevsyukov, line 50: This doesn't sound like a good header text.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Sets up `javac` by applying [JavacConfig].
+ *
+ * Which means:
+ *
+ *  1. Ensuring JDK 8 would be used for compilation;
+ *  2. Specifying arguments for `javac` invocations;
+ *  3. Specifying the `UTF-8` encoding to be used when reading source files.
+ *
+ * ### Supporting other JDKs.
```
alexander-yevsyukov, line 43: Normally, English people do not put periods at the end of headers. The only exception to this rule I know is legal texts.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,91 @@
...
+        compilerArgumentProviders.add(JavacConfig.ARGUMENTS)
+    }
+}
+
+/**
+ * The knowledge that is required to set up `javac`.
+ */
+private object JavacConfig {
+    const val SOURCE_FILES_ENCODING = "UTF-8"
+    val EXPECTED_JAVA_VERSION = JavaVersion.VERSION_1_8
```
alexander-yevsyukov, line 79: We need to expose the version of Java we use here and in the builds. Otherwise, we're keeping a timebomp.

alexander-yevsyukov (changes requested):
Please see my comments.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
yevhenii-nadtochii: I answered above, employing of Gradle `Toolchain` should be considered. It would eliminate the necessity of specifying Java version in builds. But not in this PR.

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
yevhenii-nadtochii: It is possible to make a code snippet by triple backticks and four spaces indent. For me, it looks clearer without backticks. Is three backticks a preferable way?

alexander-yevsyukov: Since we're inside KDoc, we need a way to separate the code block from the rest of the text. Backticks is the standard way to do it.

alexander-yevsyukov (approved):
LGTM

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/ErrorProne.kt
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+package io.spine.internal.gradle.javac
+
+import net.ltgt.gradle.errorprone.errorprone
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.process.CommandLineArgumentProvider
+
+/**
+ * Sets up `Error Prone` by applying [ErrorProneConfig].
```
armiol, line 34: Let's make it a bit more readable.
```
Configures Error Prone for this `JavaCompile` task.

Specifies the arguments for the compiler invocations. In particular, this configuration overrides a number of Error Prone defaults. See [ErrorProneConfig] for the details.

Please note that while `ErrorProne` is a standalone Gradle plugin, it still has to be configured through `JavaCompile` task options.

Here's an example of how to use it: ...
```

buildSrc/src/main/kotlin/io/spine/internal/gradle/javac/Javac.kt
```diff
@@ -0,0 +1,93 @@
...
+
+package io.spine.internal.gradle.javac
+
+import org.gradle.api.GradleException
+import org.gradle.api.JavaVersion
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.process.CommandLineArgumentProvider
+
+/**
+ * Sets up `javac` by applying [JavacConfig].
```
armiol, line 35: Same here.

```
Configures the `javac` tool through this `JavaCompile` task.

There are several steps performed:

* Ensures JDK 8 is used for compilation.
* Passes a couple of arguments to the compiler. See [JavacConfig] for more details.
* Sets the UTF-8 encoding to be used when reading Java source files.

Please note that Spine Event Engine can be built with JDK 8 only. Supporting JDK 11 and above at build-time is planned in 2.0 release.

Here's an example of how to use it:
...
```

armiol (changes requested):

yevhenii-nadtochii:
Closing this draft as it's approved and an associated PR into `config` repository is created.
See [Migrate javac-args.gradle to Kotlin](https://github.com/SpineEventEngine/config/pull/289)

PR #1412, opened 2021-11-05
Prototype `slow-tests.gradle` and `test-output.gradle` in Kotlin

yevhenii-nadtochii:
This is not going to be merged as approved. A separate PR into `config` repository would be created.


buildSrc/src/main/kotlin/io/spine/internal/gradle/test/Test.kt
```diff
@@ -0,0 +1,111 @@
...
+ *
+ *```
+ * tasks {
+ *     withType<Test> {
+ *         configureLogging()
+ *     }
+ * }
+ *```
+ */
+fun Test.configureLogging() {
```
armiol, line 78: Let's reformat the whole PR into two files:

* `Logging.kt`, which would define the `configureLogging()` extension.
* `Tasks.kt`, which would contain all three routines: both of the implementations, and the `TaskContainer` extension. I assume that would allow us to reduce the visibility for `FastTest` and `SlowTest` types — for some reason, they are both `public` now, while we don't presume they should be used from outside.

buildSrc/src/main/kotlin/io/spine/internal/gradle/test/task/FastTest.kt
```diff
@@ -0,0 +1,45 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.internal.gradle.test.task
+
+import org.gradle.api.tasks.testing.Test
+
+/**
+ * Executes all JUnit tests but the ones tagged as `slow`.
```
armiol, line 32: Let's say the same, but in a bit more explicit way:

```
Excludes the tests tagged as `slow` from the execution by JUnit.
```

buildSrc/src/main/kotlin/io/spine/internal/gradle/test/task/SlowTest.kt
```diff
@@ -0,0 +1,45 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.internal.gradle.test.task
+
+import org.gradle.api.tasks.testing.Test
+
+/**
+ * Executes JUnit tests tagged as `slow`.
```
armiol, line 32: This isn't completely true. It _includes_ slow tests into the execution. So, all other tests are executed as well.

Let's rewrite this piece:

```
Includes the tests tagged as `slow` into execution by JUnit.
```

buildSrc/src/main/kotlin/io/spine/internal/gradle/test/task/Environment.kt
```diff
@@ -0,0 +1,36 @@
...
+package io.spine.internal.gradle.test.task
+
+/**
+ * Name of a tag for annotating a test class or method that is known to be slow and
+ * should not normally be run together with the main test suite.
+ *
+ * @see io.spine.testing.SlowTest
+ * @see org.junit.jupiter.api.Tag
+ */
+internal const val SLOW_TAG = "slow"
```
armiol, line 36: This should go to `Tasks.kt` (as per my previous comment).

armiol (changes requested):

buildSrc/src/main/kotlin/io/spine/internal/gradle/test/task/FastTest.kt
yevhenii-nadtochii: `gradle fastTest` - will run JUnit tests filtering out the ones marked `slow`. It's a full-fledged but a little tuned `Test` task. It still **executes/runs** tests. Suggested doc implies the next usage `gradle slowTest test` to run tests filtering out `slow`. No?

buildSrc/src/main/kotlin/io/spine/internal/gradle/test/task/SlowTest.kt
yevhenii-nadtochii: Not the best naming but `includeTags(...)` seems to execute **only** the tests marked with the passed tags. I've tested it out.
[+ article with an example](https://www.codingeek.com/tutorials/junit/junit-tag-annotation-filter-example/)

armiol (approved):

yevhenii-nadtochii:
Closing this draft as it's approved and an associated PR into `config` repository is created.
See [Migrate slow-tests.gradle and test-output.gradle to Kotlin](https://github.com/SpineEventEngine/config/pull/290)

PR #1418, opened 2021-11-10
Update option defenitions

dmdashenkov:
We update the option definitions by bumping the `base` and `time` dependencies.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

Issue #1403, opened 2021-10-21
`Migration` should overwrite entity state instead of merging new state into old one


dmitrykuzmin:
`Migration` allows to update one or more entities in the repository by applying the user-supplied state mapping.

At the moment, the `newState` returned as the result of the mapping is propagated to the entity using `mergeFrom(...)` call on the old state. This causes `repeated` and message-type fields to be overwritten only partially.

The `Migration` logic should be changed so that the `newState` completely overwrites the old one within the migration transaction.

See [this line](https://github.com/SpineEventEngine/core-java/blob/ef79164ba77298581990392dd30b1e679c62a547/server/src/main/java/io/spine/server/entity/Migration.java#L295).

armiol:
Addressed in #1405 and #1419.

PR #1419, opened 2021-11-16
[2.x] Perform clear-and-replace when applying entity `Migration`

armiol:
This changeset addresses #1403, now for 2.x `master` branch. In short, previously during the entity migration the old state was being merged with the new state returned by the `Migration` instance. That led to a number of unwanted side effects and bugs.

Now, the entity `Migration` routines fully replace the old state with the new state, as per its original API design.

([a similar PR](https://github.com/SpineEventEngine/core-java/pull/1405) has been merged into 1.x branch)

The library version has been set to `2.0.0-SNAPSHOT.78`.

### DIFF
--- a/server/src/test/java/io/spine/server/projection/given/RandomFillProjection.java
+++ b/server/src/test/java/io/spine/server/projection/given/RandomFillProjection.java
@@ -24,28 +24,47 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.server.procman.given.repo;
+package io.spine.server.projection.given;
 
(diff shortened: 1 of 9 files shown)


server/src/test/java/io/spine/server/projection/given/RandomFillProjection.java
```diff
@@ -0,0 +1,70 @@
...
+                .setDescription(newUuid())
+                .vBuild();
+    }
+
+    private static TaskId taskId() {
+        return TaskId.newBuilder()
+                .setId(newUuid().hashCode())
+                .build();
+    }
+
```
dmdashenkov, line 69: Please kill this line.

dmdashenkov (approved):
LGTM with a single comment.

PR #1420, opened 2021-11-16
[2.x] Improve Catch-up by caching `ProjectionRepository` instances

armiol:
This changeset is a port of #1406 to 2.x branch.

Previously, `CatchUpProcess` used a direct reference to the `ProjectionRepository` to dispatch the domain events to the `Projection` instances. However, as the catch-up itself could be distributed over different server nodes, the reference to the repository sometimes led to the repository mismatch.

This changeset addresses introduces caching of `ProjectionRepository` instances by the type URL of the `Projection` instances.

The library version is bumped to `2.0.0-SNAPSHOT.79`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.78`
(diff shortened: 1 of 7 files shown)


alexander-yevsyukov (approved):
LGTM

PR #1421, opened 2021-11-18
[1.x] Lazy `ServerEnvironment.when(..).use(..)`

### LINKED ISSUE #1414: The server environment configurator has a lazy API but is not actually lazy
The server environment provides API to configure storage, delivery, etc using functions instead of passing variables.
```
ServerEnvironment.when(Production.class)
                 .useDelivery(env -> delivery())
		 .useStorage(env -> storage())
		...
```
This API looks lazy and you expect this function to be called only if the environment meets the `when(...)` condition. But actually, it is not lazy and invokes immediately inside `use...` method. Such kind of API is useless and probably should be removed or replaced with lazy invocation.

armiol:
Previously, `ServerEnvironment` provided two kinds of API calls for the conditional settings:

```java
    ServerEnvironment.when(Staging.class)
                     // This call is handy when `myStorageFactory` is already available.
                     .use(myStorageFactory)

    ServerEnvironment.when(PreProduction.class)
                     // And this one looks lazy, so that it is only executed when and _if_ requested.
                     .use((env) -> createMySqlStorage())
```

The "lazy" option looks as one allowing to postpone the initialization until it is requested. I.e., the `createMySqlStorage()` would only make sense if run in the pre-production environment, so the call may fail if executed when running, say, local tests.

However, in fact, there was no "lazy" behavior, which caused numerous workarounds to actually _postpone_ the initialization of environment-specific settings until they start to make sense.

This changeset addresses the issue by making the behavior truly "lazy".

Closes #1414.

The library version is set to `1.7.7-SNAPSHOT.6`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.7.7-SNAPSHOT.5`
+# Dependencies of `io.spine:spine-client:1.7.7-SNAPSHOT.6`
 
 ## Runtime
 1. **Group:** com.google.android **Name:** annotations **Version:** 4.1.1.4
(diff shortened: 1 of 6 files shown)


alexander-yevsyukov (approved):
LGTM

PR #1422, opened 2021-11-18
[2.x] Lazy `ServerEnvironment.when(..).use(..)`

armiol:
This is a migration of #1421 to the `master` branch.

### DIFF
--- a/server/src/main/java/io/spine/server/EnvSetting.java
+++ b/server/src/main/java/io/spine/server/EnvSetting.java
@@ -28,6 +28,7 @@ package io.spine.server;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.spine.environment.EnvironmentType;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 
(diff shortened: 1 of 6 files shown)


server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -193,13 +197,31 @@ void reset() {
...
+     *
+     * <p>The supplier will not be invoked unless someone requests the value under
+     * the matching environment.
+     *
+     * @param value
+     *         supplier of the value to assign to one of environments
+     * @param type
+     *         the type of the environment
+     */
+    void lazyUse(Supplier<V> value, Class<? extends EnvironmentType> type) {
```
dmdashenkov, line 215: Let's just call it `use`. The compiler will allow it because the args are different and, I think, it reads better as well.

Also, let's put the functional interface param last. This way, in Kotlin, we can use the DSL-like syntax.

dmdashenkov (approved):
LGTM with a comment.

Issue #1370, opened 2021-06-17
Events produced by `AbstractCommandHandler` are not registered in the context `EventRegistry`


yuri-sergiichuk:
If one subscribes to an event produced by `AbstractCommandHandler`, the following warning is emitted by the `SubscriptionService`:

```
WARNING: Unable to find a Bounded Context for type `<some-time>`. Creating a subscription in contexts: <some-context>.
```

The reason behind such a behavior resides in the `InMemoryEventRegistry` implementation. The registry picks up outgoing events only from repositories that implement `EventProducingRepository`.

This issue also leads to being not able to cancel a subscription returned by the Spine Client.

Consider the following code as an example that leads to the noticed behavior:

```java
final class SomeHandler extends AbstractCommandHandler implements Logging {

    @Assign
    SmthHappened handle(DoSmth c, CommandContext context) {
        return SmthHappened.newBuilder()
                .setMessage("Command handled by abstract handler.")
                .vBuild();
    }
}

@Test
@DIsplayName("observing an event produced by stateless handler leads to warnings and issues with subscriptions")
void observeHandlerEvent(){
    var client = Client.inProcess("localServer").build()
    var eventObserved = new CountDownLatch(1);
    client.asGuest()
          .command(DoSmth.newBuild().setMessage("Handle command using stateless handler.").vBuild())
          .observe(SmthHappened.class, (e) -> {
              eventObserved.countDown();
          })
          .post();
    eventObserved.await();
    client.shutdown();
}
```

**A temporary workaround**: use singleton Process Managers instead of abstract command handlers.

armiol:
Closed via #1423 and #1424.

Issue #1349, opened 2021-01-05
`ConcurrentModificationException: null` while closing a context 


yuri-sergiichuk:
In one of our products we've seen a `ConcurrentModificationException` being thrown from inside of the `ThirdPartyContext#close` routine with the following stacktrace:

```
java.util.ConcurrentModificationException: null
at io.spine.server.integration.ConfigurationChangeObserver.findStale (ConfigurationChangeObserver.java:149)
at io.spine.server.integration.ConfigurationChangeObserver.clearStaleSubscriptions (ConfigurationChangeObserver.java:125)
at io.spine.server.integration.ConfigurationChangeObserver.handle (ConfigurationChangeObserver.java:95)
at io.spine.server.integration.AbstractChannelObserver.onNext (AbstractChannelObserver.java:98)
at io.spine.server.integration.AbstractChannelObserver.onNext (AbstractChannelObserver.java:46)
at io.spine.server.transport.Subscriber.callObservers (Subscriber.java:103)
at io.spine.server.transport.Subscriber.onMessage (Subscriber.java:98)
at io.spine.server.transport.memory.InMemoryPublisher.publish (InMemoryPublisher.java:61)
at io.spine.server.integration.ConfigurationBroadcast.send (ConfigurationBroadcast.java:79)
at io.spine.server.integration.IntegrationBroker.notifyOthers (IntegrationBroker.java:265)
at io.spine.server.integration.ConfigurationChangeObserver.handle (ConfigurationChangeObserver.java:98)
at io.spine.server.integration.AbstractChannelObserver.onNext (AbstractChannelObserver.java:98)
at io.spine.server.integration.AbstractChannelObserver.onNext (AbstractChannelObserver.java:46)
at io.spine.server.transport.Subscriber.callObservers (Subscriber.java:103)
at io.spine.server.transport.Subscriber.onMessage (Subscriber.java:98)
at io.spine.server.transport.memory.InMemoryPublisher.publish (InMemoryPublisher.java:61)
at io.spine.server.integration.ConfigurationBroadcast.send (ConfigurationBroadcast.java:79)
at io.spine.server.integration.IntegrationBroker.notifyOthers (IntegrationBroker.java:265)
at io.spine.server.integration.ConfigurationChangeObserver.handle (ConfigurationChangeObserver.java:98)
at io.spine.server.integration.AbstractChannelObserver.onNext (AbstractChannelObserver.java:98)
[...]

PR #1423, opened 2021-11-22
[1.x] Make `IntegrationBroker` safer. Adjust `SubscriptionService` for standalone event producers.

armiol:
This changeset addresses #1349 and #1370.

In particular, the internals of `IntegrationBroker` were made more thread-safe.

Also, `SubscriptionService` now properly locates the Bounded Context when subscribing to events produced by standalone producers, such as `AbstractCommandHandler` or `AbstractEventReactor` descendants. Previously, these dispatchers did not register themselves in `Stand`. Therefore, `SubscriptionService` was failing to find the target Bounded Context and always ended up creating a number of redundant event subscriptions in each known Bounded Context.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.7.7-SNAPSHOT.6`
+# Dependencies of `io.spine:spine-client:1.7.7-SNAPSHOT.7`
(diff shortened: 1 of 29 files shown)


dmdashenkov (approved):
LGTM once the tests pass.

PR #1424, opened 2021-11-23
[2.x] Make `IntegrationBroker` safer. Adjust `SubscriptionService` for standalone event producers.

armiol:
This is a port of #1423 to the `master` branch.

This changeset addresses #1349 and #1370.

In particular, the internals of `IntegrationBroker` were made more thread-safe.

Also, `SubscriptionService` now properly locates the Bounded Context when subscribing to events produced by standalone producers, such as `AbstractCommandHandler` or `AbstractEventReactor` descendants. Previously, these dispatchers did not register themselves in `Stand`. Therefore, `SubscriptionService` was failing to find the target Bounded Context and always ended up creating a number of redundant event subscriptions in each known Bounded Context.

As a part of the updates, the event producers related to catch-up were hidden from `Stand` — as we don't want the events
emitted by them become available for subscribing.

Lastly, a minor change was made to the pieces which attempted to use a raw version of `QueryableRepository`. In order to avoid the erasure, they were supplied with the corresponding generic param values.

The version of the library was set to `2.0.0-SNAPSHOT.81`.

### DIFF
--- a/server/src/main/java/io/spine/server/EventProducer.java
+++ b/server/src/main/java/io/spine/server/EventProducer.java
@@ -51,6 +51,11 @@ public interface EventProducer {
     Version version();
 
     /**
+     * Obtains classes of the events produced by this object.
+     */
+    ImmutableSet<EventClass> producedEvents();
+
+    /**
      * Obtains the {@link io.spine.server.model.Nothing} event message.
      *
      * <p>This event should be returned if there is no value for the domain to produce an actual
@@ -62,9 +67,4 @@ public interface EventProducer {
     default Nothing nothing() {
         return Nothing.getDefaultInstance();
     }
-
-    /**
-     * Obtains classes of the events produced by this object.
-     */
-    ImmutableSet<EventClass> producedEvents();
 }
--- a/server/src/main/java/io/spine/server/EventProducer.java
+++ b/server/src/main/java/io/spine/server/EventProducer.java
@@ -26,9 +26,11 @@
 
 package io.spine.server;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.protobuf.Any;
 import io.spine.core.Version;
 import io.spine.server.model.Nothing;
+import io.spine.server.type.EventClass;
 
 /**
  * An object with identity which produces events.
@@ -60,4 +62,9 @@ public interface EventProducer {
     default Nothing nothing() {
         return Nothing.getDefaultInstance();
     }
+
+    /**
+     * Obtains classes of the events produced by this object.
+     */
+    ImmutableSet<EventClass> producedEvents();
 }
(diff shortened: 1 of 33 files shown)


server/src/main/java/io/spine/server/EventProducer.java
```diff
@@ -60,4 +62,9 @@
     default Nothing nothing() {
         return Nothing.getDefaultInstance();
     }
+
+    /**
+     * Obtains classes of the events produced by this object.
+     */
+    ImmutableSet<EventClass> producedEvents();
```
dmdashenkov, line 69: Please move this declaration above `nothing()` to keep the important methods together and separate them from the convenience shortcut that is `nothing()`.

server/src/test/java/io/spine/server/SubscriptionServiceTest.java
```diff
@@ -287,6 +322,42 @@ private Subscription checkSubscribesTo(Class<? extends Message> aClass) {
...
+        @DisplayName("state of entities that belong to Context")
+        void entityType() {
+            assertTargetFound(AggProject.class);
+        }
+
+        private void assertTargetFound(Class<? extends Message> targetType) {
+            Target target = Targets.allOf(targetType);
+            Optional<BoundedContext> result = subscriptionService.findContextOf(target);
+            assertThat(result).isPresent();
+            assertThat(result.get().name()).isEqualTo(context.name());
```
dmdashenkov, line 357:
```suggestion
            assertThat(result)
                    .isPresent();
            assertThat(result.get().name())
                    .isEqualTo(context.name());
```

server/src/test/java/io/spine/server/stand/given/StandTestEnv.java
```diff
@@ -123,6 +124,14 @@ public static Stand newStand(boolean multitenant, Repository<?, ?>... repositori
         return context.stand();
     }
 
+    public static Stand newStand(CommandDispatcher... dispatchers) {
+        BoundedContextBuilder builder = BoundedContextBuilder.assumingTests();
+        Arrays.stream(dispatchers)
+              .forEach(builder::addCommandDispatcher);
```
dmdashenkov, line 130: Let's use a loop instead.

I'd accept this if there was at least one intermediate operation (e.g. `map`, `flatMap`, `filter`).

dmdashenkov: Also, this should've been a `reduce()` or maybe a `collect()`, not a `forEach()`

server/src/test/proto/spine/test/subscriptionservice/report.proto
```diff
@@ -0,0 +1,40 @@
...
+
+package spine.test.subscriptionservice;
+
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.test.subscriptionservice";
+option java_outer_classname = "ReportDefsProto";
+option java_multiple_files = true;
+
```
dmdashenkov, line 36: Please kill the empty line.

dmdashenkov (approved):
LGTM with a few comments to address.

PR #1428, opened 2021-11-29
Migrate to JDK 11 and Java 11 syntax [part 4]

armiol:
This PR is the follow-up to #1427.

It completes the `var` replacement for the half of `server` test sources

Some of the existing warnings were addressed as well. 

The library version is set to `2.0.0-SNAPSHOT.85`.

PR #1429, opened 2021-11-30
Migrate to JDK 11 and Java 11 syntax [part 5]

armiol:
This PR is the follow-up to #1428.

It completes the `var` replacement for the rest of `server` test sources

Some of the existing warnings were addressed as well. 

The library version is set to `2.0.0-SNAPSHOT.86`.


server/src/test/java/io/spine/server/trace/TracingTest.java
```diff
@@ -127,7 +123,7 @@ private void assertReceiver(Class<? extends Message> messageType,
     }
 
     private void post(CommandMessage command) {
-        Command cmd = requests.command()
+        var cmd = requests.command()
                               .create(command);
```
alexander-yevsyukov, line 127: Please fix wrapping here.

alexander-yevsyukov (approved):
LGTM, with a couple of comments.

PR #1432, opened 2021-12-08
Expose gRPC dependencies at `api` level

alexander-yevsyukov:
This updates the project model so that gRPC dependencies used by the `client` and `server` artifacts are exposed using the `api` (rather than `implementation`) level. 

Such an exposure allows avoiding inevitable dependency declarations in the projects that has services declarations in the `.proto` files. One of such cases is the dependencies handled by the `bootstrap` plugin. We need to isolate the knowledge of what a client- or a server-side project needs to `core-java`, rather than duplicate this knowledge at the `bootstrap` side.

Latest `config` and `tool-base` were also applied.


buildSrc/src/main/kotlin/io/spine/internal/markup/MarkdownDocument.kt
```diff
@@ -113,11 +113,11 @@ class MarkdownDocument {
     fun ol(): MarkdownDocument = nl().text("1. ")
 
     /**
-     * Appends the passed [file] with the contents of this document.
+     * Writes the content of this document to the passed file.
      *
-     * The previous contents of the file are NOT overwritten.
+     * If the file exists, it becomes overwritten.
```
dmdashenkov, line 118:
```suggestion
     * If the file exists, it is overwritten.
```

dmdashenkov (approved):
LGTM with a suggestion.

armiol (approved):
@alexander-yevsyukov LGTM with a comment.

PR #1433, opened 2021-12-13
[1.x] Make `AbstractWorkRegistry` respect the current worker of a node

yevhenii-nadtochii:
Previously, `ShardProcessingSession` used `NodeId` to indicate who is currently processing which shard. 

On an application node, several workers (threads) can be used to deliver messages. Thus, their identifiers should be used in registry to clearly say who is processing the shard.

This PR updates `AbstractWorkRegistry ` to request the current  node's `WorkerId` to use it in `ShardProcessingSession`. `WorkerId` stores information about the node it belongs and an a worker identifier itself (typically, a thread ID).

The library version is bumped to `1.7.7-SNAPSHOT.8`.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
```
alexander-yevsyukov, line 3: The most recent is 7.3.1 already.

armiol: @alexander-yevsyukov This is 1.x branch. We cannot use Gradle 7 here.

server/src/test/java/io/spine/server/delivery/given/DeliveryTestEnv.java
```diff
@@ -78,6 +79,14 @@ public static NodeId generateNodeId() {
                 .vBuild();
     }
 
+    public static WorkerId generateWorkerId() {
+        return WorkerId
+                .newBuilder()
+                .setNodeId(generateNodeId())
+                .setValue(String.valueOf(Thread.currentThread().getId()))
```
armiol, line 86: Please extract the thread ID into the variable.

server/src/main/java/io/spine/server/delivery/AbstractWorkRegistry.java
```diff
@@ -45,50 +45,66 @@
...
+        ShardSessionRecord updatedRecord = updateNode(record, worker);
+        return Optional.of(asSession(updatedRecord));
     }
 
-    private static boolean hasPickedBy(ShardSessionRecord record) {
-        return !NodeId.getDefaultInstance().equals(record.getPickedBy());
+    /**
+     * Returns an identifier of the current worker that is now going to process the shard.
+     *
+     * Typically, it would a thread ID.
```
armiol, line 82: Same as discussed vocally.

Also, in Java we use `<p>`.

armiol (approved):
@yevhenii-nadtochii LGTM with some things to fix according to the written comments and vocal discussion.

PR #1434, opened 2021-12-15
[2.x] Make `AbstractWorkRegistry` respect the current worker of a node 

yevhenii-nadtochii:
This is a port of #1433 to the `master` branch.

Previously, `ShardSessionRecord` used `NodeId` to indicate who is currently processing which shard. 

On an application node, several workers (threads) can be used to deliver messages. Thus, their identifiers should be used in registry to clearly say who is processing the shard.

This PR updates `AbstractWorkRegistry` to request the current  node's `WorkerId` to use it in `ShardSessionRecord`. `WorkerId` stores information about the node it belongs and an a worker identifier itself (typically, a thread ID).

The library version is bumped to `2.0.0-SNAPSHOT.89`.

### DIFF
--- a/server/src/main/java/io/spine/server/delivery/memory/InMemoryShardedWorkRegistry.java
+++ b/server/src/main/java/io/spine/server/delivery/memory/InMemoryShardedWorkRegistry.java
@@ -59,7 +59,7 @@ public final class InMemoryShardedWorkRegistry extends AbstractWorkRegistry {
 
     @Override
     protected WorkerId currentWorkerFor(NodeId node) {
-        WorkerId worker = WorkerId
+        var worker = WorkerId
                 .newBuilder()
                 .setNodeId(node)
                 .setValue(String.valueOf(Thread.currentThread().getId()))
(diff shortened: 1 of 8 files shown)


server/src/main/java/io/spine/server/delivery/memory/InMemoryShardedWorkRegistry.java
```diff
@@ -52,8 +53,18 @@
...
+    public synchronized Optional<ShardProcessingSession> pickUp(ShardIndex index, NodeId node) {
+        return super.pickUp(index, node);
+    }
+
+    @Override
+    protected WorkerId currentWorkerFor(NodeId node) {
+        WorkerId worker = WorkerId
+                .newBuilder()
+                .setNodeId(node)
+                .setValue(String.valueOf(Thread.currentThread().getId()))
```
armiol, line 65: Please extract the thread ID into a variable. And document the details in the `currentWorkerFor` Javadoc.

armiol (changes requested):
@yevhenii-nadtochii LGTM in general. Please see my comment.

armiol (approved):

PR #1439, opened 2022-01-04
Fix uncommitted list

yevhenii-nadtochii:


server/src/test/java/io/spine/server/aggregate/given/AggregateCachingTestEnv.java
```diff
@@ -0,0 +1,60 @@
...
+import io.spine.core.Event;
+import io.spine.server.BoundedContext;
+import io.spine.server.BoundedContextBuilder;
+import io.spine.server.entity.Repository;
+
+import java.util.List;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class AggregateCachingTestEnv {
```
armiol, line 39: This type also seems `final` to me.

Issue #1326, opened 2020-11-19
[2.x] Implement a migration assistant for legacy data that used `Mirror`


armiol:
In 2.x. branch, the Aggregate mirroring feature has been replaced by the new functionality of `AggregateStorage`. 

We need to ease the transition for the existing users of Spine 1.x and implement a configurable assistant that could help migrate the Aggregate data stored as `MirrorProjection`s.

PR #1442, opened 2022-01-20
Implement a migration assistant for legacy data that used `Mirror`

yevhenii-nadtochii:
This PR addresses #1326.

`Mirror` projection was deprecated in Spine 2.0. Previously, it was used to store aggregates' states to allow their querying. A single projection stored states of **all** aggregates in a Bounded Context.

 Now, `EntityRecordStorage` is used to store the aggregate's states when it is open for querying. This storage is created on a **per-aggregate** basis. This enables storing of queryable state-based columns along the state itself. 

In order to ease the transition for the existing users of Spine 1.x, `MirrorMigration` assistant was implement. It performs the migration on a per-aggregate basis. For this reason, an aggregate class is required to create an instance of this class. State-based column values are fetched automatically.

`MirrorMigrationMonitor` is required to actually run the process. It dictates the used batch size, receives notifications about the course of the migration,  and can interrupt it when needed. 

Below is an example of how to use it:

```java
// Create an instance of `MirrorMigration`.

var context = ContextSpec.singleTenant("...");
var factory = ServerEnvironment.instance().storageFactory();
var aggToMigrate = ParcelAgg.class;
var migration = new MirrorMigration(context, factory, aggToMigrate);

// Create an instance of `MirrorMigrationMonitor`. We are going to use
// a default implementation, which always continues the migration.
// Also, the mirrors will be migrated in batches of 500 records.

var batchSize = 500;
var monitor = new MirrorMigrationMonitor(batchSize);

// Run the migration.

migration.run(monitor);
```

The library version is bumped to `2.0.0-SNAPSHOT.92`.


server/src/main/java/io/spine/server/migration/mirror/MirrorMapping.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param <I>
+ *         the type of aggregate's identifiers
+ * @param <S>
+ *         the type of aggregate's state
+ * @param <A>
+ *         the type of aggregate
+ */
+@Immutable
+final class MirrorMapping<I, S extends EntityState<I>, A extends Aggregate<I, S, ?>> {
```
armiol, line 57: There isn't much of mapping going on in this type. We have just a single transformation.

I would say, that's a `MirrorToRecord` which extends a `Function<Mirror, EntityRecordWithColumns<I>>`. And, correspondingly, instead of `toEntityRecord` is has its `apply(...)` method.

server/src/main/java/io/spine/server/migration/mirror/MirrorMigration.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.Iterator;
+
+/**
+ * Migrates {@linkplain Mirror} projections into {@linkplain EntityRecordWithColumns}.
+ *
+ * <p>{@code Mirror} projection was deprecated in Spine 2.x. Previously, it was used to store
+ * aggregates' states to allow their querying. A single projection stored states of <b>all</b>
+ * aggregates in a Bounded Context.
+ *
+ * <p>Now, {@linkplain EntityRecordStorage} is used to store the aggregate's states when it is
```
armiol, line 51: Let's also say something about the order in which the migration is performed.

server/src/main/java/io/spine/server/migration/mirror/MirrorMigration.java
```diff
@@ -0,0 +1,186 @@
...
+ * open for querying. This storage is created on a <b>per-aggregate</b> basis. This enables
+ * storing of queryable state-based columns along the state itself. For this reason, the migration
+ * is also done on a per-aggregate basis.
+ *
+ * <p><b>An example usage</b>
+ *
+ * <p>1. Create an instance of `MirrorMigration`:
+ * <pre>
+ * var contextSpec = ContextSpec.singleTenant("...");
```
armiol, line 60: Please format the code pieces, so that they are more readable when viewed as plain comments.

server/src/main/java/io/spine/server/migration/mirror/MirrorMigration.java
```diff
@@ -0,0 +1,186 @@
...
+        this.aggregateType = AggregateClass.asAggregateClass(aggClass)
+                                          .stateTypeUrl()
+                                          .value();
+    }
+
+    /**
+     * Migrates {@linkplain Mirror} projections
+     * to the aggregate's {@linkplain EntityRecordStorage}.
+     */
+    public void run(MigrationSupervisor supervisor) {
```
armiol, line 122: In addition to this method, I would expect an endpoint which takes the `MigrationStep`, e.g., `continueFrom(MigrationStep step, MigrationSupervisor supervisor)`.

server/src/main/java/io/spine/server/migration/mirror/MirrorMigration.java
```diff
@@ -0,0 +1,186 @@
...
+        supervisor.onStepStarted();
+
+        var batchSize = supervisor.stepSize();
+        Collection<EntityRecordWithColumns<I>> entityRecords = new ArrayList<>(batchSize);
+        Collection<Mirror> migratedMirrors = new ArrayList<>(batchSize);
+
+        fetchNextBatch(batchSize)
+                .forEachRemaining(mirror -> {
+
+                    var entityRecord = mapping.toEntityRecord(mirror);
```
armiol, line 146: Please extract this lambda into a nested (or just inner?) type.

server/src/main/java/io/spine/server/migration/mirror/MirrorMigration.java
```diff
@@ -0,0 +1,186 @@
...
+                                 .where(Mirror.Column.wasMigrated()).is(false)
+                                 .sortAscendingBy(Mirror.Column.wasMigrated())
+                                 .limit(batchSize)
+                                 .build();
+        var iterator = mirrorStorage.readAll(query);
+        return iterator;
+    }
+
+    @VisibleForTesting
+    EntityRecordStorage<I, S> entityRecordStorage() {
```
armiol, line 178: As long as it's no `private`, please document it. 

Same goes for the method below.

server/src/main/java/io/spine/server/migration/mirror/MirrorStorage.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * A contract for storages of {@code Mirror} projections.
+ *
+ * <p>This storage exposes methods for performing read and write operations
+ * in batches.
+ */
+public class MirrorStorage extends MessageStorage<MirrorId, Mirror> {
```
armiol, line 46: I would assume it's `final`.

Also, can we please make the `StorageFactory` creating it?

server/src/main/proto/spine/system/server/mirror.proto
```diff
@@ -64,6 +64,16 @@ message Mirror {
...
+
+    // A flag which tells whether this message was migrated to an `EntityRecord`.
+    //
+    // `Mirror` was deprecated in Spine 2.x. Now, `EntityRecord`s are used to persist
+    // the aggregate's state
+    //
+    // The presence of this flag pops out of the need to perform a step-by-step migration.
+    //
+    // See `io.spine.server.migration.mirror.MirrorMigration`.
+    bool wasMigrated = 7 [(column) = true];
```
armiol, line 76: We don't use camelCased variables in Protobuf. Please rename to `was_migrated`.

armiol (changes requested):

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -102,7 +102,7 @@ public void onDeliveryStarted(ShardIndex index) {
     }
 
     /**
-     * Returns an instance of {@code DeliveryMonitor} which always tells to continue.
+     * Returns an instance of {@code DeliveryMonitor} which always says to continue.
```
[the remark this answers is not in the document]
yevhenii-nadtochii: This one was reported by a spell checker. I try to eliminate those warnings.

server/src/main/java/io/spine/server/migration/mirror/MirrorMigration.java
```diff
@@ -0,0 +1,228 @@
...
+ * // Create an instance of `MirrorMigration`.
+ *
+ * var context = ContextSpec.singleTenant("...");
+ * var factory = ServerEnvironment.instance().storageFactory();
+ * var aggToMigrate = ParcelAgg.class;
+ * var migration = new MirrorMigration(context, factory, aggToMigrate);
+ *
+ * // Create a migration monitor. Here we are going to use a default
+ * // implementation, which always continues the migration.
+ * // Wel will work with storages via batches of 500 records.
```
armiol, line 77: Please fix the typos.

armiol (changes requested):
@yevhenii-nadtochii we are getting closer. Please see my comments.

server/src/main/java/io/spine/server/entity/storage/EntityRecordWithColumns.java
```diff
@@ -104,6 +107,45 @@ private EntityRecordWithColumns(I id, EntityRecord record, Map<ColumnName, Objec
...
+     *         the type of the entity's identifier
+     * @param <S>
+     *         the type of the entity's state
+     * @param <E>
+     *         the type of the entity
+     */
+    @SuppressWarnings("unchecked") // see the docs.
+    public static <I, S extends EntityState<I>, E extends Entity<I, S>> EntityRecordWithColumns<I>
+    create(EntityRecord record, Class<E> entityClass) {
+
```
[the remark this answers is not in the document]
yevhenii-nadtochii: Do we still need runtime checks when the package is annotated with `@ParametersAreNonnullByDefault`?

server/src/main/java/io/spine/server/migration/mirror/MirrorMigration.java
```diff
@@ -72,9 +75,9 @@
  * var aggToMigrate = ParcelAgg.class;
  * var migration = new MirrorMigration(context, factory, aggToMigrate);
  *
- * // Create a migration monitor. Here we are going to use a default
- * // implementation, which always continues the migration.
- * // Wel will work with storages via batches of 500 records.
+ * // Create an instance of `MirrorMigrationMonitor`. We are going to use
+ * // a default implementation, which always continues the migration.
+ * // We will work with storage implementations in batches of 500 records.
```
armiol, line 80: Let's make it even more clear:
```
Also, the records will be migrated in batches of 500 records.
```

armiol (approved):
@yevhenii-nadtochii LGTM with a couple of comments.

Issue #1415, opened 2021-11-08
Expand Spine server builder to support preconfigured gRPC services


dmytro-kashcheiev:
Expand `spine.Server` builder to make it possible to build it using preconfigured `Query`, `Subscription` and `CommandService` and custom services.

**Use case 1:** 
If you want your server to run in both `HTTP`(external connection with Spine web client) and `gRPC`(internal connections). You should run both `HTTP` servlets in the servlet container and gRPC server on a different port. 

Currently, it's impossible to achieve this using `spine.Server` as far as you need `command, query, and subscription services` for HTTP servlet instantiation, and the same services will be instantiated inside `Server.build()` method, as a result, it's impossible to reuse same services for both HTTP servlets and gRPC server.

**Use case 2:** 
For `GCE` instances it's great to have `healthcheks` and `GCP` now supports `gRPC` health-checking. To implement this feature you should run an additional `gRPC` server next to `Spine` and add `HealthcheckService` directly there. It would be great to have a possibility to add custom `gRPC` service to the `io.spine.server.Server`.

PR #1443, opened 2022-03-22
Flexible server

dmdashenkov:
This is a PR to the v1.8 branch.

In this PR we expand the `io.spine.server.Server` API to allow to:
 - obtain the Query, Subscription, and Command services after the `Server` is built;
 - add extra gRPC services to the same server.

This PR is related to #1415. That issue will be closed after the same change is done to the v2.0 branch.


armiol (approved):
LGTM.

PR #1444, opened 2022-03-29
Flexible Server

### LINKED ISSUE #1415: Expand Spine server builder to support preconfigured gRPC services

dmdashenkov:
In this PR we expand the `io.spine.server.Server` API to allow to:
 - obtain the Query, Subscription, and Command services after the `Server` is built;
 - add extra gRPC services to the same server.

This PR closes #1415.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -138,11 +138,7 @@ allprojects {
 
 subprojects {
 
-    with(repositories) {
-        applyGitHubPackages("base", rootProject)
-        applyGitHubPackages("time", rootProject)
-        applyStandard()
-    }
+    repositories.applyStandard()
 
     apply {
         plugin("java-library")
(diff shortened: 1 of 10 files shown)


dmdashenkov:
@armiol, PTAL.
Even after an approval, I'll wait till the checks pass. Right now, there is a GH Actions [outage](https://www.githubstatus.com/incidents/3nrmdrmgqxp3) so the CI doesn't run.

armiol (approved):
LGTM.

PR #345, opened 2017-02-16
Exclude `@Internal` from Javadoc

dmytro-grankin:
We need a tool to exclude `@Internal` program elements (which are not part of public API) from javadocs.

Summary:

- Fix javadocs, which produce errors.
- Extend standard doclet (program that specify the content and format of the output of the Javadoc tool).

PR #344, opened 2017-02-16
Fix Javadocs to have text with class name after FQN.

alexander-aleksandrov:
This is the fix for Javadocs from {@link FQN} to {@link FQN text} format. Also some wrapping for 100 column right margin.

PR #1445, opened 2022-03-30
Update to the latest `config`

yevhenii-nadtochii:
This PR updates the repository to use the latest `config`.

Mostly, this is done in order to pull in the following changes:

1. [Make published artifacts full-fledged #335](https://github.com/SpineEventEngine/config/pull/335).
2. [Bump dependencies #344](https://github.com/SpineEventEngine/config/pull/344).
3. [Rename `exposeTestArtifacts` to `exposeTestConfiguration` #345](https://github.com/SpineEventEngine/config/pull/345).

Applying of those changes led to the following:

1. `testJar` has disappeared  in most modules. The artifact is [optional](https://github.com/SpineEventEngine/config/pull/340) now and not published by default. Publishing of this artifact was preserved only for `server` module.
2. `protoJar` artifact has also disappeared in modules that don't have any proto definitions.
4. Deleted unused imports.

Please note, the repository doesn't use the last versions of ErrorProne, `spine-mc-java` and `spine-tool-base`. See #1446 for details.

The library version is bumped to `2.0.0-SNAPSHOT.94`.


build.gradle.kts
```diff
@@ -24,11 +24,12 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+@file:Suppress("RemoveRedundantQualifierName")
```
armiol, line 27: Please don't suppress warnings for such a big scope. We don't want to skip the cases in which the qualifier name is really redundant.

armiol (approved):
@yevhenii-nadtochii please see my comment. Other than that, LGTM.

PR #1447, opened 2022-04-15
Allow supplying `ExecutorService` for `SystemWriteSide`

yevhenii-nadtochii:


yevhenii-nadtochii:
@armiol Please take a look.

server/src/main/java/io/spine/system/server/CurrentThreadExecutor.java
```diff
@@ -0,0 +1,42 @@
...
+package io.spine.system.server;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+import java.util.concurrent.Executor;
+
+/**
+ * An {@code Executor} which immediately runs all given commands in the current thread.
+ */
+class CurrentThreadExecutor implements Executor {
```
armiol, line 36: Seems `final` to me.

server/src/main/java/io/spine/system/server/SystemFeatures.java
```diff
@@ -44,10 +46,9 @@ interface SystemFeatures {
      * @return {@code true} if system events should be stored, {@code false} otherwise
      */
     boolean includePersistentEvents();
+
     /**
-     * Checks if the system events are allowed to be posted in parallel.
```
armiol, line 48: We still need a way to know if the parallel posting is enabled.

server/src/main/java/io/spine/system/server/SystemFeatures.java
```diff
@@ -44,10 +46,9 @@ interface SystemFeatures {
...
-     *
-     * @return {@code true} if it's OK to post system event is parallel, {@code false} otherwise
+     * Obtains an {@code Executor} with which system events will be posted.
      */
-    boolean postEventsInParallel();
+    Executor eventPostingExecutor();
```
armiol, line 53: Exposing an `Executor` here means to show the implementation details. It's nobody's business on the level this high which executor is used. Besides, it is also dangerous, as anyone with an access to the system features may post anything at any time.

Please rethink this change so that the executor itself remains hidden.

server/src/main/java/io/spine/system/server/SystemSettings.java
```diff
@@ -190,12 +207,12 @@ public boolean equals(Object o) {
...
                 storeEvents == settings.storeEvents &&
-                parallelPosting == settings.parallelPosting;
+                postEvents.equals(settings.postEvents);
     }
 
     @SuppressWarnings("NonFinalFieldReferencedInHashCode")
     @Override
     public int hashCode() {
-        return Objects.hashCode(commandLog, storeEvents, parallelPosting);
+        return hash(commandLog, storeEvents, postEvents);
```
armiol, line 216: Obtaining the hash code of the executor is not a good idea. Generally speaking, it's better not to hash any objects with a volatile state (executors, network connections, etc). There is no guarantee that their hash code value is consistent through their lifespan, and does not depend on their changing state.

testutil-server/src/test/java/io/spine/testing/server/blackbox/DiagnosticLoggingTest.java
```diff
@@ -69,11 +70,11 @@ void resetStderr() {
     }
 
     /**
-     * Ensures the {@code messagePart} has beed logged.
+     * Ensures the {@code messagePart} has been logged.
      */
     protected void assertLogged(String messagePart) {
         substituteErrorStream.flush();
-        assertThat(output.toString())
+        assertThat(output.toString(StandardCharsets.UTF_8))
```
armiol, line 77: I would import this value statically.

armiol (changes requested):
@yevhenii-nadtochii please see my comments. Let's discuss them vocally if needed.

yevhenii-nadtochii:
@armiol PTAL

server/src/main/java/io/spine/system/server/SystemSettings.java
```diff
@@ -124,32 +138,75 @@ public SystemSettings forgetEvents() {
...
+    @CanIgnoreReturnValue
+    public SystemSettings useCustomPostingExecutor(Executor executor) {
+        checkNotNull(executor);
+        checkState(parallelPosting);
+        this.postingExecutor = executor;
+        return this;
+    }
+
+    /**
+     * Configures the system context to post system events using {@link ForkJoinPool#commonPool()}.
```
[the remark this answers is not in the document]
armiol: Just so you know.

The thing is that the system context does not post any events on its own. The domain context acts as a client and tells the system context to do things.

This description is fine, as a lot of detail may be worse than less. But it is important that you'd understand how it really works.

PR #1449, opened 2022-04-22
Better document implemented gRPC services

yevhenii-nadtochii:


yevhenii-nadtochii:
@armiol Please take a look.

When we are done with `CommandService`, I will update `QueryService` and `SubscriptionService` accordingly.

yevhenii-nadtochii:
@armiol Please take a look on a second iteration.

server/src/main/java/io/spine/server/CommandService.java
```diff
@@ -45,8 +45,13 @@
...
 
 /**
- * The {@code CommandService} allows client applications to post commands and
- * receive updates from the application backend.
+ * The {@code CommandService} provides a synchronous way to post commands
+ * to the application backend.
+ *
+ * <p>This class is an implementation of a corresponding gRPC service. Hence, public API of this
+ * class is dictated by the {@linkplain CommandServiceGrpc generated code}. Despite the fact of its
+ * "publicity", it's not meant to be used directly. Use {@link io.spine.client.Client Client}
```
armiol, line 53: This is not true. It _is_ meant to be used directly. Otherwise, users won't be able to assembly the definitions of services for their gRPC servers at server-side. 

Please rephrase.

client/src/main/proto/spine/client/command_service.proto
```diff
@@ -30,18 +30,31 @@ package spine.client;
...
 service CommandService {
 
-    // Request to handle a command.
+    // Posts the given command to the application backend.
+    //
+    // When the command is successfully received by the application, it responds with
+    // an acknowledgement. The received `Ack` may have one of the following statuses:
+    //
+    //    1. `Ok` meaning the command is accepted for further processing.
+    //    2. `Error` when a technical error occurs on the side of the application.
```
armiol, line 50: Please be more specific. What is a technical error? What is the "side of the application"? Both client and server make the application as a whole.

server/src/main/java/io/spine/server/CommandService.java
```diff
@@ -70,14 +75,41 @@ public static Builder newBuilder() {
...
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>In the original proto definition of this service, this method is unary.
+     * Meaning, for every posted command only a single {@link Ack} is received in response:
+     *
+     * <pre>Ack post(Command);</pre>
+     *
+     * <p>But for every service, gRPC generates several clients: blocking, asynchronous,
```
armiol, line 96: There are two issues with this text.

1. It is pretty much the same for every method which is generated after gRPC. So it would be nice to write it once and "run everywhere" (meaning, reference, not copy).

2. The wording here is a bit harsh. We aren't the authors of gRPC. And for sure we have chosen to use them voluntarily. Therefore, they don't owe us anything. But the impression of this text is that they do.

client/src/main/proto/spine/client/command_service.proto
```diff
@@ -30,18 +30,31 @@ package spine.client;
...
+    //    3. `Rejection` if no technical error occurred but due to the business rules the command
+    //        should be immediately disqualified from being executed. A typical scenario would be
+    //        when the permissions of a user who made a request aren't broad enough.
+    //
+    // In case of internal gRPC errors or issues on a transport layer, an `Ack` would not be
+    // received. All errors which occur on the side of gRCP should be handled by the means
```
armiol, line 56: "gRPC"

client/src/main/proto/spine/client/command_service.proto
```diff
@@ -30,18 +30,31 @@ package spine.client;
...
+    // of the used client stub.
```
armiol, line 57: It is not "used". Maybe, "chosen", or no word at all.

client/src/main/proto/spine/client/command_service.proto
```diff
@@ -30,18 +30,31 @@ package spine.client;
...
+//
 service CommandService {
 
-    // Request to handle a command.
+    // Posts the given command to the application backend.
+    //
+    // When the command is successfully received by the application, it responds with
+    // an acknowledgement. The received `Ack` may have one of the following statuses:
+    //
+    //    1. `Ok` meaning the command is accepted for further processing.
```
armiol, line 49:
```
`OK`, meaning the ...
```

client/src/main/proto/spine/client/command_service.proto
armiol, line 50:
```
`Error`, when a ...
```

client/src/main/proto/spine/client/command_service.proto
```diff
@@ -30,18 +30,31 @@ package spine.client;
...
+    //    2. `Error` when a technical error occurs on the side of the application.
+    //    3. `Rejection` if no technical error occurred but due to the business rules the command
```
armiol, line 51:
```
`Rejection`, if no ...
```

armiol (changes requested):

Let's discuss my comments vocally once you return back to this task.

PR #1452, opened 2022-04-27
Increase the amount of available JVM memory for Gradle builds

armiol:
Recently, Dokka plugin has been enabled for the Gradle builds. However, it is known to consume lots of memory.

This PR introduces the global JVM property. Also, it attempts to address the publishing failure which occurred while merging the previous PR.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -25,4 +25,4 @@
 #
 
 # Dokka plugin eats more memory than usual. Therefore all builds should have enough.
-org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
+org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m -Xmx1024m
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,28 @@
+#
+# Copyright 2022, TeamDev. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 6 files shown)


gradle.properties
```diff
@@ -0,0 +1,28 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+# Dokka plugin eats more memory than usual. Therefore all builds should have enough.
+org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
```
dmdashenkov, line 28: I'd suggest we also add the `Xmx` value. Seems like, typically, it's set to be double the size of `MaxMetaspeceSize`.

As per the [doc](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties), it seems like this line may override all the default values. My concern is that this would make the `Xmx` value inadequately low.

dmdashenkov (approved):
@armiol, LGTM with a suggestion.

PR #1457, opened 2022-07-21
Improve the performance of `ColumnMapping` lookup

armiol:
In scope of this changeset, the peformance of `ColumnMapping` lookup is improved by caching the results per column type.

Recently, a new version of Spine `modelCompiler` has been developed. Its main feature is that the validation codegen is performed via ProtoData library with Spine's Validation extensions to it. However, at this point, the performance of the corresponding build tasks is low, dragging the build duration into dozens of minutes.

The primary effort on optimization is concentrated in Validation library (see [this PR](https://github.com/SpineEventEngine/validation/pull/29)). However, some bits may be improved in `spine-server` artifact as well — since Validation library itself is a Spine-based application.

Therefore, this changeset introduces the `ColumnMapping` performance improvements, so that they could be later used in Validation library. 

At the moment, Validation library uses an "old" (pre-ProtoData) version of `spine-server`. So, in this changeset, a minimal set of changes was made to keep the resulting `spine-server` artifact as close to the one used currently as possible — in terms of versions of transitive dependencies.

### DIFF
--- a/.github/workflows/build-on-ubuntu.yml
+++ b/.github/workflows/build-on-ubuntu.yml
@@ -7,16 +7,15 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'true'
 
-      - uses: actions/setup-java@v2
+      - uses: actions/setup-java@v3
         with:
           java-version: 11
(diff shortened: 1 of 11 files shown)


armiol:
@dmdashenkov PTAL.

dmdashenkov (approved):
LGTM.

PR #1458, opened 2022-07-25
Direct `Delivery`

armiol:
This changeset introduces a "direct" mode of `Delivery`.

Prior to these changes, in any mode of its operation, `Delivery` operated by processing the `InboxMessage`s which were stored and read from `InboxStorage`. It guaranteed that no concurrent changes could be made to the same entity state.

However, in some cases a Spine-based application may look for more performance in dispatching the signals. In this changeset, a direct `Delivery` mode is now available:

```java
        ServerEnvironment.when(...)
                         .use(Delivery.direct());
```

Using this mode has the following impact onto the application:

* `Inbox`es and `InboxStorage` are skipped altogether. They no longer participate in dispatching of any signals. No sharding is performed as well. It improves the application performance by eliminating the `Inbox`-related I/O operations.

* Any concurrent modifications made to the entities are now out of control. I.e. simultaneous handling of signals by the same entity instance (loaded either in different threads or in different JVMs) is now possible.

A typical scenario is a single-threaded processing of a huge number of signals (such as importing an event log via third-party events) in a single-threaded application. In Spine-based tools, this feature is designed for ProtoData CLI tool.

Please note, this is an **experimental** API.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.99`
+# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.100`
 
 ## Runtime
 1.  **Group** : com.google.android. **Name** : annotations. **Version** : 4.1.1.4.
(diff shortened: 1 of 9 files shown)


armiol:
@dmdashenkov PTAL.

dmdashenkov (approved):
LGTM.

PR #1459, opened 2022-08-21
New model compiler

armiol:
This changeset updates `core-java` with the latest `modelCompiler`. 

Proto messages are now validated according to the new codegen approach. Previously, the validation implementation was running purely operating the runtime context. New validation code is generated for each Proto message, taking into account the corresponding validation options. It speeds up the process of validating messages, since no runtime discovery is performed. Instead, the process is hard-coded for each message.

Each Proto message is now validated upon invoking `MyMessage.Builder.build()`. In case one needs to compose an "invalid" (or partial) `Message` instance, she should still invoke `... .buildPartial()` as previously.

Here is the list of changes made in scope of this PR:

* New `mc-java` artifact is now used for both Gradle plugin and `protoc` runtime (`io.spine.tools:spine-mc-java-plugins:${mcJavaVersion}:all`). It reduces the size of artifacts (in MB) downloaded upon Gradle build.

* New ProtoData and Validation artifacts are now used (as configured via `modelCompiler`). Each of these is a fat JAR which has no dependencies. It allows to run ProtoData in isolation without bumping into conflicts with Spine libraries-under-development.

* Conflicts previously occurring with Error Prone and its transitive dependencies are now resolved.

* The code used in tests has been adjusted, since previously `Message` instances built were not valid. Since now the validation is always performed upon invoking `build()` method, approximately 720-730 tests were failing.

* `FieldMasks` utility now returns `@NonValidated` message instances. They are always built with `buildPartial()`, since required fields are very likely to be stripped.

* `ProducedEvents` and `ProducedCommands` Spine-internal types now do not require their wrapped `Event`/`Command` collections to be filled in, since Spine handler methods may return `Optional<..>` values.

* `ExternalMessage` no longer has `actor_context` field. Since the message introduction, this field has never been filled properly. And as this message is used in various scenarios, the `actor_context` field value cannot be meaningfully set in general.

* Fields of type `base.Error` now do not have `(validate) = true` option set, since the definition of `Error` has no validation constraints.

The library version has been set to `2.0.0-SNAPSHOT.108`.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
```
alexander-yevsyukov, line 3: @armiol, how about 7.5.1. They recommend it instead.

pom.xml
```diff
@@ -227,17 +247,17 @@ all modules and does not describe the project structure per-subproject.
...
-    <version>1.6.20</version>
+    <version>1.6.21</version>
   </dependency>
   <dependency>
     <groupId>org.jetbrains.dokka</groupId>
     <artifactId>dokka-core</artifactId>
-    <version>1.6.20</version>
+    <version>1.6.21</version>
   </dependency>
   <dependency>
```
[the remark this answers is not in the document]
armiol: @sonatype-lift ignore

armiol:
@alexander-yevsyukov @dmdashenkov PTAL.

In the following PRs I will be moving `base/validate` to the Validation library, and afterwards, `core-java` will be migrated to this new Validation artifact one more time.

build.gradle.kts
```diff
@@ -154,9 +158,14 @@ subprojects {
...
     java {
-        tasks.withType<JavaCompile>().configureEach {
-            configureJavac()
-            configureErrorProne()
+        tasks {
+            withType<JavaCompile>().configureEach {
+                configureJavac()
+                configureErrorProne()
+            }
+            withType<org.gradle.jvm.tasks.Jar>().configureEach {
```
dmdashenkov, line 166: Please import this type if possible.

client/src/main/java/io/spine/client/Targets.java
```diff
@@ -148,6 +148,9 @@ private static <I> IdFilter toIdFilter(I... id) {
     }
 
     private static IdFilter idFilter(List<Any> ids) {
+        if(ids.isEmpty()) {
```
dmdashenkov, line 151:
```suggestion
        if (ids.isEmpty()) {
```

server/src/main/proto/spine/server/dispatch/dispatching.proto
```diff
@@ -118,17 +118,25 @@ message Success {
...
 
 // Commands produced during a signal handling.
+//
+// Spine supports the handler methods returning `Optional` values,
+// so there may be no produced commands.
+//
 message ProducedCommands {
     option (is).java_type = "ProducedCommandsMixin";
 
-    repeated core.Command command = 1 [(required) = true, (validate) = true];
```
dmdashenkov, line 131: Why are validation rules in this file changed?

dmdashenkov (commented):
Please see my comments.

server/src/main/proto/spine/server/dispatch/dispatching.proto
armiol: @dmdashenkov that's because there may be no produced commands. In our code, we allow to return `Optional<Cmd>` in the commanding-splitting methods. Previously, the validation of instances of this type was never performed. Now, as the validation code is executed upon `build()`, it is always validated, and lots of tests fail.

build.gradle.kts
armiol: Done.

client/src/main/java/io/spine/client/Targets.java
armiol: Addressed.

server/src/test/java/io/spine/server/projection/StateRoutingTest.java
```diff
@@ -68,21 +68,24 @@ void setupContext() {
     @Test
     @DisplayName("support explicit routing")
     void explicit() {
+        var manifesto = Manifesto
+                .newBuilder()
```
alexander-yevsyukov, line 72: Please move `newBuilder()` to the previous line.

alexander-yevsyukov (approved):
LGTM, with one minor request.

alexander-yevsyukov:
@armiol, I had to bump Gradle to 7.6 nightly in order to avoid `StackOverflow` in Gradle (https://github.com/gradle/gradle/issues/21997).

armiol:
I had that one, too. But just for a couple of times. Then it disappeared. 

I guess, we're expecting the patch release of Gradle shortly.

PR #1462, opened 2022-08-24
Fix gRPC dependency version and CPU-dependent test

alexander-yevsyukov:
This PR addresses two issues discovered when attempting to build under macOS running under M1 Max.

### Forced gRPC version
The first issue is that `mc-java` applied in the `:core` module of the project sets the artifact `protoc-gen-grpc-java-1.38.0-osx-aarch_64.exe` (because that version of `mc-java` depends on gRPC v 1.38). Due to some reason there is no such exe file available on Maven Plugin portal, which resulted in the error:
```
Execution failed for task ':core:generateProto'.
> Could not resolve all files for configuration ':core:protobufToolsLocator_grpc'.
   > Could not find protoc-gen-grpc-java-1.38.0-osx-aarch_64.exe (io.grpc:protoc-gen-grpc-java:1.38.0).
     Searched in the following locations:
         https://plugins.gradle.org/m2/io/grpc/protoc-gen-grpc-java/1.38.0/protoc-gen-grpc-java-1.38.0-osx-aarch_64.exe
``` 
Another version of gRPC is used by the `:client` module. Forcing this version in the main code _and_ in the `:model-verifier` tests fixes the issue.

### `MirrorMigrationTest` was altered to not depend on CPU performance
Before this PR the test used a stub `MirrorMigrationMonitor` which failed when 5000 records were not converted during 3 seconds. To make this stub instance fail a small batch size of 10 records and corresponding timing were used.

Such an approach is suboptimal because of two reasons:
  1. It slows tests down. 3 seconds is not much if it comes to one test, but with thousands of tests, seconds here and there add up.
  2. The assumption that migration wouldn't every go through with such a small batch is wrong. CPUs become faster eventually. Tests can be time-based if there's time involved in the business logic.

The test was updated to just count number of migration steps.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,6 +25,7 @@
  */
 
 import io.spine.internal.dependency.ErrorProne
+import io.spine.internal.dependency.Grpc
 import io.spine.internal.dependency.JUnit
 import io.spine.internal.gradle.publish.IncrementGuard
 import io.spine.internal.gradle.VersionWriter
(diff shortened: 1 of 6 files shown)


alexander-yevsyukov:
@yevhenii-nadtochii, invited you to review so that you can see the comments on the 2nd issue addressed in this PR.

armiol (approved):
@alexander-yevsyukov LGTM.

PR #1465, opened 2022-08-24
Do not allow internal types sent from client-side

alexander-yevsyukov:
This PR extends the client-side implementation with prohibiting sending types marked with `internal_type` option.

Such types are meant to be used only within a bounded context which declares it. Thus, sending them to the context from a client should not be allowed.

This PR also fixes the hierarchy of `ClientRequest` types by introducing `ClientRequestBase`. Previous implementation leaked unwanted API into the classes that were derived from `ClientRequest` (and now are derived from `ClientRequestBase`).

### DIFF
--- a/client/src/main/java/io/spine/client/ClientRequest.java
+++ b/client/src/main/java/io/spine/client/ClientRequest.java
@@ -55,6 +55,10 @@ import static io.spine.type.MessageExtensions.requirePublished;
 @SuppressWarnings("ClassReferencesSubclass")
 // we want to have DSL for calls encapsulated in this class.
 public class ClientRequest extends ClientRequestBase {
+
+    /**
+     * Creates a new instance with the given user ID and the instance of the {@code client}.
+     */
     ClientRequest(UserId user, Client client) {
         super(user, client);
     }
--- a/client/src/main/java/io/spine/client/ClientRequest.java
+++ b/client/src/main/java/io/spine/client/ClientRequest.java
@@ -27,18 +27,14 @@
 package io.spine.client;
 
 import com.google.common.collect.ImmutableList;
-import com.google.errorprone.annotations.CanIgnoreReturnValue;
-import com.google.errorprone.annotations.OverridingMethodsMustInvokeSuper;
 import io.spine.base.CommandMessage;
 import io.spine.base.EntityState;
 import io.spine.base.EventMessage;
 import io.spine.core.UserId;
 import io.spine.query.EntityQuery;
-import org.checkerframework.checker.nullness.qual.Nullable;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static io.spine.type.MessageExtensions.requirePublished;
-import static io.spine.util.Preconditions2.checkNotDefaultArg;
 
 /**
  * Entry point for creating client requests.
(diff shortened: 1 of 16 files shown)


client/src/main/java/io/spine/client/ClientRequest.java
```diff
@@ -56,31 +54,17 @@
...
-public class ClientRequest {
-
-    private final UserId user;
-    private final Client client;
-
-    private @Nullable ErrorHandler streamingErrorHandler;
-    private @Nullable ServerErrorHandler serverErrorHandler;
-
+public class ClientRequest extends ClientRequestBase {
     ClientRequest(UserId user, Client client) {
```
armiol, line 58: Please allow an empty line above the ctor and please document it.

client/src/main/java/io/spine/client/ClientRequestBase.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Assigns a handler for an error occurred on the server-side (such as validation error)
+     * in response to posting a request.
+     */
+    @CanIgnoreReturnValue
+    @OverridingMethodsMustInvokeSuper
+    public ClientRequestBase onServerError(ServerErrorHandler handler) {
+        checkNotNull(handler);
+        this.serverErrorHandler = handler;
```
armiol, line 102: Let's make this more compact:
```java
this.serverErrorHandler = checkNotNull(handler);
```

client/src/main/java/io/spine/client/ClientRequestBase.java
```diff
@@ -0,0 +1,113 @@
...
+    private ErrorHandler streamingErrorHandler = null;
+    @Nullable
+    private ServerErrorHandler serverErrorHandler = null;
+
+    /**
+     * Creates a new instance with the given user ID and the {@code client} instance.
+     */
+    ClientRequestBase(UserId user, Client client) {
+        this.user = user;
+        checkNotDefaultArg(user);
```
armiol, line 55: I would check that prior to assignment.

client/src/main/java/io/spine/client/CommandFactory.java
```diff
@@ -106,11 +111,14 @@ public Command create(CommandMessage message, int targetVersion) throws Validati
...
-     * @param context the command context to use as a base for the new command
+     * @param message
+     *         the command message
+     * @param context
+     *         the command context to use as a base for the new command
      * @return new command instance
      * @throws ValidationException
-     * if the passed message does not satisfy the constraints set for it in its Protobuf definition
+     *         if the passed message does not satisfy the constraints set for it in its Protobuf
+     *         definition
```
armiol, line 121: Let's move "in its Protobuf" to the next line.

armiol (approved):
@alexander-yevsyukov LGTM with some minor comments to be addressed before merging.

server/src/test/kotlin/io/spine/client/ClientRequestPublishedLanguageTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import io.spine.test.client.ClientTestContext
+import io.spine.test.unpublished.Locomotive
+import io.spine.test.unpublished.command.Halt
+import io.spine.test.unpublished.event.WheelsKnocked
+import io.spine.type.UnpublishedLanguageException
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+internal class `'ClientRequest' should prohibit using 'internal_type' messages when` :
```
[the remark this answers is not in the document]
alexander-yevsyukov: It does not apply to test suite classes because such names are much more convenient than `@DisplayName` annotations.

PR #1468, opened 2022-08-26
Filter internal commands posted to `CommandService`

alexander-yevsyukov:
This PR extends the `CommandService` with an ability to reject commands with messages marked as `internal_type`.

Also creating `CommandService` without any bounded context associated cases now warning in logging. The ability to create such ”empty” command services may be useful for stubbing (which was employed in the test added by this PR), and for special cases when, for example, a server-side listens only to events.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -619,7 +619,7 @@
 
 The dependencies distributed under several licenses, are used according their commercial-use-friendly license.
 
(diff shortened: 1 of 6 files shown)


alexander-yevsyukov:
> @armiol, strangely, I'm getting catch-up tests failing under Windows. I've updated the test grouping them by time precision. But I'm not sure it would fix that :)

It did! :) But let's keep an eye on it.

armiol (approved):
@alexander-yevsyukov LGTM.

As to failed catch-up tests on Windows, I'd rather ignore and let it go. They are indeed somewhat dependent on the inter-thread communication, which together with JUnit sometimes results in a weird behaviour. Perhaps, in the future I will be able to improve the stability of these tests even more than I did for a few times already. But for now it seems 95/5 effort, really.

PR #1469, opened 2022-08-28
Extract common parts in service creation

alexander-yevsyukov:
This PR extracts the common parts of the code that was used to create `QueryService` and `SubscriptionService`.

An abstract base for service builders was introduced. Also `TypeDictionary` utility class was created to simplify mapping a type to a bounded context to which this type belongs.

### DIFF
--- a/server/src/main/java/io/spine/server/AbstractServiceBuilder.java
+++ b/server/src/main/java/io/spine/server/AbstractServiceBuilder.java
@@ -42,6 +42,9 @@ public abstract class AbstractServiceBuilder<T, B extends AbstractServiceBuilder
 
     private final Set<BoundedContext> contexts = Sets.newHashSet();
 
+    /**
+     * Returns this instance with the overloaded type.
+     */
     abstract B self();
 
     /**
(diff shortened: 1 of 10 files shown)


server/src/main/java/io/spine/server/AbstractServiceBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+ * Abstract for builder of service classes.
+ *
+ * @param <T>
+ *         the type of service to be produced
+ * @param <B>
+ *         the self-type of the builder for return type covariance
+ */
+public abstract class AbstractServiceBuilder<T, B extends AbstractServiceBuilder<T, B>> {
+
+    private final Set<BoundedContext> contexts = new HashSet<>();
```
armiol, line 48: Let's make sure we have both `equals()` and `hashCode()` defined for `BoundedContext`.

From what I can see now, we don't.

armiol (changes requested):
@alexander-yevsyukov LGTM except for a single comment. Please have a look.

server/src/main/java/io/spine/server/AbstractServiceBuilder.java
alexander-yevsyukov: I made them comparable by the names too. We use this already at least once, for diagnostics.

alexander-yevsyukov:
@armiol, I'm going to bump `base` to the latest one. I'm on it.

armiol (approved):
@alexander-yevsyukov LGTM.

PR #1470, opened 2022-09-01
Allow services serving no types

alexander-yevsyukov:
This PR unifies the convention on creating "empty" services. Previously, we allowed `CommandService` serving no types, but disallowed it for `QueryService` and `SubscriptionService`.

Now it is possible to create a service with no served types (because no bounded contexts were added to its builder).
It would allow to handle unusual server-side implementations and creating service stubs for testing.

If an empty service is created, a warning would be logged.

Also `CommandService` implementation was migrated to `TypeDictionary`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.105`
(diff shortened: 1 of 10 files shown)


armiol (approved):
@alexander-yevsyukov LGTM.

PR #1471, opened 2022-09-16
More internal type filtering

alexander-yevsyukov:
This PR teaches `QueryService` and `SubscriptionService` to reject calls referring types marked as `internal_type` (either directly in a type option or via a file option `internal_all`).

To extract common logic of processing, an abstract `ServiceDelegate` class was introduced.
`CommandService`, `QueryService`, and `SubscriptionService` were modified to processing calls to classes derived from `ServiceDelegate`.

Also `UnknownEntityTypeException` was renamed to `UnknownEntityStateTypeException` to better reflect the cause of the error.


yevhenii-nadtochii (commented):
@alexander-yevsyukov thank you. I've checked this out and left several suggestions. LGTM, but I would preserve the power of a formal approve to other reviewers.

armiol (changes requested):
@alexander-yevsyukov please see the comments left by @yevhenii-nadtochii   

Other than those, LGTM.

armiol (approved):
@alexander-yevsyukov LGTM.

PR #1474, opened 2022-09-28
Checking inclusion of standard types

alexander-yevsyukov:


alexander-yevsyukov:
@armiol, this is the branch based on version from January 6th, with updated dependencies to avoid the missing gRPC compiler for M1 process.

As you can see, the Ubuntu build passes. I've also checked recent `base` artifacts for the presence of standard Google Protobuf types. All artifacts do contain these types, so `TypeAttractor` is not necessary. 

I suspect there's something happening with the classpath in recent versions of `core-java`.

armiol:
@alexander-yevsyukov so basically, that something happened with the most recent `base` changes? I am asking since this "New model-compiler" PR was passing the build stages before the migration.  Can [this](https://github.com/SpineEventEngine/base/blob/master/validate/build.gradle.kts#L53) be it?

alexander-yevsyukov:
@armiol, even latest `base` artifacts contain `EnumValue` in the descriptors. 

What we do in `excludeGoogleProtoFromArtifacts()` is filtering out Google's source code when creating source code archive.

alexander-yevsyukov:
This experimental PR is no longer needed.

PR #1477, opened 2022-10-04
Move model tools

alexander-yevsyukov:
This PR removes the `model-assember` and `model-verifier` modules as they moved to the [`model-tools` repository](https://github.com/SpineEventEngine/model-tools/).


armiol (changes requested):
@alexander-yevsyukov as discussed, please remove the files of the modules as well.

PR #1478, opened 2022-10-06
Cleaner console

alexander-yevsyukov:
This PR updates dependencies on `base` and `validation` thus avoiding console output that was a result of applying runtime validation to the generated code.


dmdashenkov (approved):
LGTM, given that we've decided to reinstate the `validation-java-runtime` artifact.

alexander-yevsyukov:
> LGTM, given that we've decided to reinstate the `validation-java-runtime` artifact.

Yes. It has a name which people can understand.

PR #1480, opened 2022-10-06
Fix publishing

armiol:
This PR temporarily removes explicit dependencies of `dokkaHtml` and `sourcesJar` onto `launchProtoDataMain`, since they break publishing. 

It happens because `launchProtoDataMain` (and a bunch of other tasks) is re-launched once again right before sources and-or documentation are being assembled. In turn, it leads to non-compilable sources processed, and ultimately to the compilation failure. Yes, compilation failure, on a stage of assembling the documentation and sources.


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #1487, opened 2022-10-23
More tests to Kotlin

alexander-yevsyukov:
It's a pull request employing generated Kotlin DSL for Protobuf types.

Notable changes:
  * `ActorRequestFactoryTestEnv` class which was shared among several test suites was eliminated in favor of encapsulated test environements.
  * Several test suites were converted to Kotlin.


client/src/test/java/io/spine/client/QueryBuilderSpec.kt
```diff
@@ -0,0 +1,598 @@
...
+import org.junit.jupiter.api.Assertions.fail
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import com.google.protobuf.Any as AnyProto
+
+@DisplayName("`QueryBuilder` should")
+internal class QueryBuilderSpec {
```
dmdashenkov, line 81: Is there a reason for us to use `@DisplayName` instead of the class name is backticks?

I'm not against using `..Spec` in Java tests, but Kotlin gives us, IMHO, a better option.

client/src/test/java/io/spine/client/QueryBuilderSpec.kt
```diff
@@ -0,0 +1,598 @@
...
+
+    companion object TestEnv {
```
dmdashenkov, line 83: I don't see this companion object used inside this test suite by name. We could either turn it into a normal object or ditch the name altogether.

dmdashenkov (approved):
LGTM with a couple of suggestions regarding tests.

client/src/test/java/io/spine/client/QueryBuilderSpec.kt
alexander-yevsyukov: @armiol, @dmdashenkov, your thoughts?

armiol: Yes, that makes sense. We'll need to include this explanation into our (future?) Kotlin codestyle guidelines.

alexander-yevsyukov: https://github.com/SpineEventEngine/documentation/issues/28

Issue #1466, opened 2022-08-25
Current tenant when routing a signal


dmdashenkov:
When routing an event via an `EventRoute` configured through a repository, the current tenant is not set.

Since the repository is responsible for routing signals, it's natural that the users will want to run queries upon that repository in order to find the target of a signal.

We should either:
 - set the tenant automatically from the signal context;
 - on the API level, ensure the user provides a tenant instead of relying on the `CurrentTenant`;
 - discourage using `protected` API of a repository and instead provide an easy way for the user to query a repository from within.

Which option to choose is a matter of discussion.

Issue #905, opened 2018-12-06
Fallback routing by type of fields


alexander-yevsyukov:
Suppose, we have an aggregate with ID of type `EntityOneId`. It produces events that have the first field of this type (and, of course, `producerId` in the `EventContext`). Some of these events are related to another entity with IDs of `TypeTwoId`. And these relations are reflected in the fields that have indexes starting from 2.

We want to react to such events in the entity with IDs of `EntityTwoId`, and in order to achieve this, we need to implement custom routing in the corresponding repository. It would look like this:

```java
    public EntityTwoRepository() {
        super();
        getEventRouting().route(
                SomethingHappenedInEntityOne.class, 
                (event, context) -> ImmutableSet.of(event.getEntityTwoId())
        );
    }
```

What if we implement automatic fall-back routing which would use Proto reflection to find a field which matches the type of the IDs of entities managed by the repository? If there is such a field, and its value is not empty, it would do the `ImmutableSet.of()` thing so that the user does not have to tune the routing which seems trivial and obvious.

PR #1491, opened 2022-10-29
Add `@Route` annotation

alexander-yevsyukov:
This PR introduces `@Route` annotation for static methods of an entity class that would be a preferred way for arranging routing.  Currently we have to create a repository class _just_ for arranging routing. It it's too much code.

This PR only renames the `Route` functional interface to `RouteFn`, freeing up the name for the new annotation interface.
It does that very minimum to introduce this breaking change sooner.

Following PRs would:

  * Verify the correctness of applying the `@Route` annotation (presumably via using Java Annotation Processor).

  * Extend `BoundedContextBuilder` for creating a routing function over a default repository for a class, which has static methods annotated with `@Route`.


alexander-yevsyukov:
@armiol, @dmdashenkov, please have a look at how I suggest to improve routing.

I have not implemented the functional part of it. I'd like to finalize the concept first.

alexander-yevsyukov:
We also probably need to discuss #1466 and #905 in the context of this PR.

armiol:
@alexander-yevsyukov I like the idea and the implementation.

It would be nice to have a Java-based test migrated to this approach as well.
Another thing, it would be nice to have a more detailed description of the feature in the PR description.

alexander-yevsyukov:
> It would be nice to have a Java-based test migrated to this approach as well.
@armiol, I'm glad and will proceed with the implementation. But I'm not sure which ”test” did you mean? Or did you mean tests that currently employ repos just for routing?

alexander-yevsyukov:
> Some of them redefined their routing as well. We could migrate one of them — whichever seemed convenient — to this approach.

@armiol, got it. Will do.

armiol (approved):
Apparently, I have already reviewed most of the files in this very PR. 

LGTM.

Issue #1472, opened 2022-09-27
Handle failed aggregate loading during message delivery


dmytro-kashcheiev:
Delivery behavior, when it's impossible to load Aggregate for some reason (a coding mistake, storage issues, etc.), is unclear. It looks like all commands and events that point to corrupted `Aggregate` stays in `InboxStorage` forever. And re-delivered on each shard delivery iteration.

**Steps to reproduce:** 
1. Throw an exception in the `@Apply` method of the aggregate. 
2. Send `FirstCommand` to the aggregate and emmit the event that will lead to the exception. 
3. Send `SecondCommand` to the aggregate and emmit the event that will lead to the exception. 
4. Send one more command with the same `ShardIndex` to another entity.

**What happens:**
- Aggregate failed to apply the event and threw an exception;
- Diagnostic event `CannotDispatchDuplicateCommand` thrown for  `SecondCommand`.

**What I expect:** 
- Aggregate failed to apply the event and threw an exception; 
- ~~Diagnostic event `CannotDispatchDuplicateCommand` thrown for  `SecondCommand`~~;
- Another entity successfully handles the command.

alexander-yevsyukov:
From v2.0.0 we won't support event sourcing for aggregates. Aggregates will be loaded by their persisted state types.

See #1647 for details.

PR #1496, opened 2022-11-24
[1.x] Allow to handle any `RuntimeException`s occurring when delivering an `InboxMessage`

armiol:
This changeset addresses #1472 and brings other improvements into the codebase.

## Failure of signal reception, and `DeliveryMonitor` to respond

`InboxMessage`s which are processed in scope of each `Delivery` run, are eventually unpacked, and the respective signals are dispatched to their targets. In case a signal target is an `Entity`, it is possible that the signal reception does not go well for various reasons. In particular, an aggregate may have "broken" `@Apply`-er, making it impossible to load the aggregate state. Or, some other unexpected error may occur upon loading or storing an entity.

Previously, it was impossible to trace such failures of signal reception. And the `InboxMessage`s causing the errors were left in their inboxes in `TO_DELIVER` status forever, being delivered over and over again — most likely with the same reception failure repeating.

Now, the `Delivery` API allows to subscribe for any failures which occur during the reception of each signal. Additionally, end-users may now choose the way to handle the reception failures in terms of action in respect to the `InboxMessage` of interest. 

Out-of-the-box, there are two actions provided:
* mark the `InboxMessage` as delivered — so that it does not block further delivery of messages;
* repeat the dispatching of the `InboxMessage`. Please note, this is a _synchronous_ and immediate action in terms of a delivery run.

Alternatively, end-users may implement their own way of handling the reception failure.

The corresponding functionality is provided via the API of `DeliveryMonitor`.

Here is a code sample:

```java

public final class MarkFailureDelivered extends DeliveryMonitor {

    /**
     * In case the reception of the {@code InboxMessage} failed,
     * mark it as {@code DELIVERED} anyway.
     */
    @Override
    public FailedReception.Action onReceptionFailure(FailedReception reception) {

    	//// Error details are available as well:
        // InboxMessage msg = reception.message();
        // Error error = reception.error();
        // notifyOf(msg, error);

        return reception.markDelivered();
    }
}

// ...

// Plugging the monitor into the Delivery:

DeliveryMonitor monitor = new MarkFailureDelivered();

Delivery delivery = Delivery.newBuilder()
        .setMonitor(monitor)
        // ...
        .build();

ServerEnvironment
        .when(MyEnvironment.class)
        .use(delivery);
```

By default, `InboxMessage`s are marked as `DELIVERED` in case of failure of their reception. This is so, because otherwise the issue described in #1472 arises.

Internally, such a functionality became possible by making `MessageDispatcher.dispatch(envelope)` to return `DispatchOutcome`. Previously, it was `void`, which was not suitable for handling any of the issues synchronously — as it was required for `InboxMessage` to allow users choose the way to handle the reception failures.

## Other changes

It was reported, that at run-time the routing of events by their interfaces may lead to `ConcurrentModificationException`. 
This PR addresses the issue for `v1` by making the `Map` of signal routes concurrency-friendly. See [`io.spine.server.route.MessageRouting`](https://github.com/SpineEventEngine/core-java/pull/1496/files#diff-e25c04493dd2dccb2090c2ae6038eff8b0ecfadd2bc7f96ea42196a1b61550f6R69) for more details.

Even more minor changes are made to fix typos and remove the unused code.

The library version is set to `1.9.0-SNAPSHOT.6`.


.github/workflows/build-on-ubuntu.yml
```diff
@@ -17,7 +17,6 @@ jobs:
         with:
           java-version: 8
           distribution: zulu
-          cache: gradle
```
alexander-yevsyukov, line 20: Why do we turn the cache off?

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -623,10 +631,12 @@ protected A play(I id, AggregateHistory history) {
         tx.commitIfActive();
         if (!success) {
             lifecycleOf(id).onCorruptedState(outcome);
-            throw newIllegalStateException("Aggregate `%s` (ID: %s) cannot be loaded.%n",
+            String outcomeDetails = Json.toJson(outcome);
+            throw newIllegalStateException("Aggregate `%s` (ID: %s) cannot be loaded.%n%s",
```
alexander-yevsyukov, line 635: Can we have something explaining the trailing JSON, please?

server/src/main/java/io/spine/server/delivery/AbstractMessageEndpoint.java
```diff
@@ -0,0 +1,119 @@
...
+     * @param targetId
+     *         the ID of the target to which the signal should be dispatched
+     * @return an outcome of the dispatch operation
+     */
+    protected abstract DispatchOutcome performDispatch(I targetId);
+
+    private DispatchOutcome onFailedDispatch(Exception failure, I targetId) {
+        Error error = failureToError(failure, targetId);
+        DispatchOutcome outcome = DispatchOutcome
+                .newBuilder()
```
alexander-yevsyukov, line 91: Can we have `newBuilder()` on the previous line, please?

server/src/main/java/io/spine/server/delivery/FailedReception.java
```diff
@@ -0,0 +1,140 @@
...
+ *     <li>{@linkplain #markDelivered() mark} the message as delivered;
+ *     <li>{@linkplain #repeatDispatching() repeat dispatching} of the message immediately.
+ * </ul>
+ *
+ * <p>Alternatively, end-users may choose to define their own way of reacting
+ * to reception failures by implementing a custom {@code FailedReception.Action},
+ * and returning it via the corresponding
+ * {@linkplain DeliveryMonitor#onReceptionFailure(FailedReception) API} of {@code DeliveryMonitor}.
+ */
+public final class FailedReception {
```
alexander-yevsyukov, line 49: Don't we need `hashCode()`, `equals()`, and `toString()` in this class?

server/src/main/java/io/spine/server/delivery/FailedReception.java
```diff
@@ -0,0 +1,140 @@
...
+        void execute();
+    }
+
+    /**
+     * An internal interface used to specify the action,
+     * which immediately repeats the {@code InboxMessage} dispatching.
+     */
+    @Internal
+    @FunctionalInterface
+    public interface RepeatDispatching {
```
alexander-yevsyukov, line 133: It says that it is an action. Why doesn't it extend the `Action` interface above?

server/src/test/java/io/spine/server/delivery/given/NoOpEndpoint.java
```diff
@@ -26,17 +26,25 @@
...
 import static io.spine.testing.Tests.nullRef;
 
 public class NoOpEndpoint implements MessageEndpoint<String, CommandEnvelope> {
 
     @Override
-    public void dispatchTo(String targetId) {
-        // do nothing.
+    public DispatchOutcome dispatchTo(String targetId) {
+        MessageId id = MessageId
+                .newBuilder()
```
alexander-yevsyukov, line 44: Can we have `newBuilder()` on the prev. line, please?

server/src/test/java/io/spine/server/delivery/given/ReceptionistAggregate.java
```diff
@@ -0,0 +1,94 @@
...
+        extends Aggregate<String, Receptionist, Receptionist.Builder> {
+
+    public static final String FAILURE_MESSAGE = "Receptionist failed to apply an event.";
+
+    private static boolean failInAppliers = false;
+
+    @Assign
+    ConditionerTurnedOn handler(TurnConditionerOn cmd) {
+        return ConditionerTurnedOn
+                .newBuilder()
```
alexander-yevsyukov, line 50: Can we have `newBuilder()` on the prev. line here, and in other proto types below, please?

server/src/test/proto/spine/test/delivery/receptionist_events.proto
```diff
@@ -0,0 +1,44 @@
...
+package spine.test.delivery;
+
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.test.delivery.event";
+option java_outer_classname = "ReceptionistEventsProto";
+option java_multiple_files = true;
+
+
```
alexander-yevsyukov, line 37: Please remove this extra line.

alexander-yevsyukov (commented):
Please see my comments and questions.

.github/workflows/build-on-ubuntu.yml
armiol: @alexander-yevsyukov 
That's because it causes the build to fail sometimes, as it uses the cached `.class` files from the previous workflow runs. I don't want to take any chance with that, especially when publishing the library.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
armiol: @alexander-yevsyukov 
That's because this JSON will start with `DispatchOutcome`. I thought it was going to be enough.
I will add some "label" before it.

server/src/main/java/io/spine/server/delivery/FailedReception.java
armiol: @alexander-yevsyukov not sure why we would need those. Let's discuss vocally.

armiol:
@alexander-yevsyukov PTAL again.

alexander-yevsyukov (approved):
LGTM

PR #1500, opened 2023-01-17
Improve filtering of internal messages

alexander-yevsyukov:
This PR applies latest `base` for improving the reliability of filtering internal messages that may be between computation nodes. Previously, the filtering relied only on the presence of `@Internal` annotation of the generated Java class. 

Now, in the absence of the annotation, descriptors of the corresponding message types are also checked for:
  * `option (internal_type)=true` set in the message, and if the option is not declared,
  * `option (internal_all)=true` in the declaring file.

Other changes:
  * Kotlin bumped to `1.8.0`, detekt -> `1.22.0`.
  * Raw generic type usage warnings were addressed.
  * Test fixtures for the routing features of the `server` module were moved under corresponding source set.


armiol (approved):
LGTM.

Issue #1499, opened 2023-01-17
[1.x] Empty aggregate `state()` just after snapshot 


dmytro-kashcheiev:
In the `@Apply` method handler just after the snapshot `state()` object is a `DefaultInstance`.  

Steps to reproduce: 
1. Aggregate, one `@Assign` and one `@Apply` method;
2. Update state via `builder()` in the `@Apply` method;
3. Emit 100 commands to this aggregate;
4. Emit one more command to the aggregate;
5. The `@Apply` method is called to build the state;
6. Check `state()`, it will be `DefaultInstance`.

armiol:
@dmytro-kashcheiev I have analysed this use-case, and it turns out the `state()` behaves as designed in this case. It is impossible to expect something other than a default state until the transaction is committed. While technically it is possible for the framework-internal code to populate both `state()` and `builder()` when restoring from a snapshot, such an approach breaks a nature of the ongoing transaction (e.g., making it impossible to perform a rollback).

Therefore, we just prohibit calling `state()` from within an applier method. See #1501.

armiol:
Closing as resolved via #1501.

PR #1501, opened 2023-02-02
[1.x] Prohibit calling Aggregate's `state()` from `@Apply`-er methods

armiol:
This changeset addresses #1499 by disallowing to call `state()` method from aggregate appliers.

Previously, end-users could rely on `Aggregate.state()` from `@Apply`-ers. However, in some cases (such as the one described in #1499) `state()` does not reflect the last state of an aggregate. This is so because appliers are invoked in scope of a single transaction — be it when loading an aggregate instance from storage, or when applying the just-emitted events during the command dispatching. Until such a transaction is completed, `state()` does not reflect the model updates corresponding to the respective domain events. Using the returned state value most likely leads to bugs.

The designed way has always been to use `builder()` instead of `state()`, because it is kept up-to-date throughout the active aggregate transaction.

Therefore, once this PR is merged, any calls to `state()` from `@Apply`-annotated method will lead to an `IllegalStateException`.


alexander-yevsyukov (approved):
LGTM

PR #1502, opened 2023-02-03
[1.x] Fix delivering the signals to aggregate `Mirror`s in a multi-Bounded Context environment

armiol:
This changeset allows to properly access the read-side of Aggregates which are registered in different Bounded Contexts, and are exposed to read-side via the corresponding `entity` visibility.

In a Spine-based application, it is possible to expose an Aggregate state to the read-side. Under-the-hood it is done via system-internal `Mirror` projection, which accumulates all changes for all such Aggregates in scope of a single Bounded Context. At the same time, `Delivery` (which performs signal dispatching) relies onto the type URL of entities to which the dispatched signals are being delivered.

When several Bounded Contexts have their Aggregates "visible" (i.e. exposed via `Mirror`), `Delivery` is confused with multiple `Mirror` entity types which technically are distinct, but at the same time have exactly the same type URL. Such a use-cases leads to failures when Aggregate state on read-side is updated by the framework code.

This changeset alters the type URL, under which `Mirror` projections register themselves in `Delivery`. The new type URL value includes the name of the Bounded Context — which makes this type URL invalid in terms of type discovery, but addresses the issue.

An important thing to know is that such a workaround makes sense for 1.x codebase only, because Aggregate mirroring was already completely redesigned in scope of Spine 2.x.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.9.0-SNAPSHOT.6`
+# Dependencies of `io.spine:spine-client:1.9.0-SNAPSHOT.8`
 
 ## Runtime
 1. **Group:** com.google.android **Name:** annotations **Version:** 4.1.1.4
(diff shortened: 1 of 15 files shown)


armiol:
@dmytro-kashcheiev FYI.

alexander-yevsyukov (approved):
LGTM

PR #1503, opened 2023-02-07
[1.x] Import events from 3rd-party with respect for tenancy

armiol:
When it comes about importing the third-party events, the framework provides a `ThirdPartyContext` tool. 

However, in a multi-tenant application, the imported events were dispatched in a straightforward manner, without specifying the `TenantId` via `TenantAwareRunner`.

The corresponding test was also flawed: in fact, it was dispatching the events to a single-tenant Bounded Context.

This PR addresses the issue by fixing the test and making `IntegrationBroker` (an intermediary between `ThirdPartyContext` and "domestic" Bounded Contexts) to explicitly use `TenantAwareOperation` during the event dispatch.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
     
-# Dependencies of `io.spine:spine-client:1.9.0-SNAPSHOT.8`
+# Dependencies of `io.spine:spine-client:1.9.0-SNAPSHOT.9`
 
 ## Runtime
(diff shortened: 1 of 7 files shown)


armiol:
@yevhenii-nadtochii PTAL.

yevhenii-nadtochii (approved):
LGTM.

PR #1505, opened 2023-03-29
[1.x] Distinguish shard pick-up results

ndolhii:
In this PR we start to distinguish the shard pick-up results. In particular, it is now possible to find out the reason of an unsuccessful shard pick-up. In particular, there may be some runtime issues, or the shard may already be picked-up by another worker.

Two new API endpoints were added to the `DeliveryMonitor` to provide end-users with some control over such cases:

1. `FailedPickUp.Action onShardAlreadyPicked(AlreadyPickedUp failure)`

This method will be invoked if the shard could not be picked as it's already picked by another worker. This method receives the `ShardIndex` of the shard that could not be picked, the `WorkerId` of the worker who owns the delivery session, and the `Timestamp` when the shard was picked. It is required to return an action to take in relation to this case. By default, it returns an action which just accepts this case (and ends the delivery session without any processing), but end-users may return a predefined action retrying the shard pick-up:
```java
final class MyDeliveryMonitor extends DeliveryMonitor {

    ...

    @Override
    public FailedPickUp.Action onShardAlreadyPicked(AlreadyPickedUp failure) {
        return failure.retry();
    }

    ...

}
```

2. `FailedPickUp.Action onShardPickUpFailure(RuntimeFailure failure)`

This method is invoked if the shard could not be picked for some runtime technical reason, such as a runtime exception. This method receives the `ShardIndex` of the shard that could not be picked, and the instance of the occurred `Exception`. It also requires to return an action to handle this case. By default, such failures are just rethrown as `RuntimeException`, but end-users may choose to retry the pick-up:
```java
final class MyDeliveryMonitor extends DeliveryMonitor {

    ...

    @Override
    public FailedPickUp.Action onShardPickUpFailure(RuntimeFailure failure) {
        return failure.retry();
    }

    ...

}
```

# Breaking changes

The API of the `ShardedWorkRegistry` has been changed.

In particular, a new `PickUpOutcome pickUp(ShardIndex index, NodeId node)` method is introduced. Note, it returns an explicit result instead of `Optional`, as previously. This outcome contains **either of two**:
* a `ShardSessionRecord` — meaning that the shard is picked successfully,  
* a `ShardAlreadyPickedUp` — a message that contains a `WorkerID` of the worker who owns the session at the moment, and the `Timestamp` when the shard was picked. This outcome means the session cannot be obtained as it's already picked.

Also, there is a new `void release(ShardSessionRecord session)` method that releases the passed session. 

Here is a summary of code changes for those using `ShardedWorkRegistry`:

## Before:
``` java
Optional<ShardProcessingSession> session = workRegistry.pickUp(index, currentNode);
if (session.isPresent()) { // Check if shard is picked.
   // ...
   session.get().complete(); // Release shard.
}
```

## After:
``` java
PickUpOutcome outcome = workRegistry.pickUp(index, currentNode);
if (outcome.hasSession()) { // Check if shard is picked
    // ...
    workRegistry.release(outcome.getSession()); // Release shard.
}
```
Also, the new API allows getting the `WorkerId` of the worker who owns the session in case if the shard is already picked by someone else and the `Timestamp` when the shard was picked:
``` java
PickUpOutcome outcome = workRegistry.pickUp(index, currentNode);
if (outcome.hasAlreadyPickedBy()) {
    WorkerId worker = outcome.getAlreadyPicked().getWorker();
    Timestamp whenPicked = outcome.getAlreadyPicked().getWhenPicked();
    // ...
}
```

### DIFF
--- a/server/src/main/java/io/spine/server/delivery/AlreadyPickedUp.java
+++ b/server/src/main/java/io/spine/server/delivery/AlreadyPickedUp.java
@@ -68,9 +68,9 @@ public final class AlreadyPickedUp extends FailedPickUp {
     }
 
     /**
-     * Returns the {@code WorkerId} that owns the delivery session.
+     * Returns the {@code ShardAlreadyPickedUp} outcome.
      */
-    public ShardAlreadyPickedUp sessionOwner() {
+    public ShardAlreadyPickedUp outcome() {
         return alreadyPickedUp;
     }
 }


server/src/main/java/io/spine/server/delivery/PickUpOutcome.java
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * <p>The outcome contains one of 3 following results:
+ * <p>1. Shard picked up successfully. In this case the outcome will contain
+ * a new {@link ShardProcessingSession}.
+ * <p>2. Shard could not be picked up as it's already picked by another worker. In this case
+ * the outcome will contain the {@code WorkerId} of the worker owning the session.
+ * <p>3. Shard could not be picked up for some technical problem (Exception). In this case
+ * the outcome will contain the occurred {@code Exception}.
+ */
+public final class PickUpOutcome {
```
armiol, line 50: I think this should be a Proto message, possibly with an interface providing all the convenience methods. 

Otherwise it would be difficult to pass this data over the wire from Liquor.

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -95,6 +95,34 @@ public void onDeliveryStarted(ShardIndex index) {
...
+     * <p>Pay attention, this method is not called if the shard could not be picked up because
+     * it's already picked by another worker. This situation doesn't lead to an exception in
+     * {@code ShardedWorkRegistry} so it is not considered as a failure.
+     *
+     * @param index
+     *         an index of the shard that could not be picked up due to an exception
+     */
+    @SuppressWarnings({"WeakerAccess", "unused"})  // This SPI method is designed for descendants.
+    public void onShardPickUpFailure(ShardIndex index) {
+        // do nothing.
```
armiol, line 111: Could you please change this (and every other case in this file to a capitalized version:
```java
// Do nothing.
```

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -95,6 +95,34 @@ public void onDeliveryStarted(ShardIndex index) {
         // do nothing.
     }
 
+    /**
+     * Called if an {@code Exception} occurred when the {@code Delivery} attempted
+     * to pick up a shard.
+     *
+     * <p>Pay attention, this method is not called if the shard could not be picked up because
```
armiol, line 102: Let's improve:
```
<p>Please note, this callback is <em>not</em> invoked in case the shard cannot be picked
if it's already picked by another worker. It is so, because such a use case is a part of normal flow,
and thus does not provoke a {@code RuntimeException}.
```

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -95,6 +95,34 @@ public void onDeliveryStarted(ShardIndex index) {
...
+     * it's already picked by another worker. This situation doesn't lead to an exception in
+     * {@code ShardedWorkRegistry} so it is not considered as a failure.
+     *
+     * @param index
+     *         an index of the shard that could not be picked up due to an exception
+     */
+    @SuppressWarnings({"WeakerAccess", "unused"})  // This SPI method is designed for descendants.
```
armiol, line 109: Let's unify the suppression comment with others:
```
/* Part of public API. */
```

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -95,6 +95,34 @@ public void onDeliveryStarted(ShardIndex index) {
...
+
+    /**
+     * Called if {@code Delivery} could not pick up a shard because it was already picked
+     * by another worker.
+     *
+     * @param index
+     *         a shard index that could not be picked
+     */
+    @SuppressWarnings({"WeakerAccess", "unused"})  // This SPI method is designed for descendants.
+    public void onShardAlreadyPicked(ShardIndex index) {
```
armiol, line 122: Do we know the worker? If yes, let's include it into the API.

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -95,6 +95,34 @@ public void onDeliveryStarted(ShardIndex index) {
...
+     *
+     * <p>Pay attention, this method is not called if the shard could not be picked up because
+     * it's already picked by another worker. This situation doesn't lead to an exception in
+     * {@code ShardedWorkRegistry} so it is not considered as a failure.
+     *
+     * @param index
+     *         an index of the shard that could not be picked up due to an exception
+     */
+    @SuppressWarnings({"WeakerAccess", "unused"})  // This SPI method is designed for descendants.
+    public void onShardPickUpFailure(ShardIndex index) {
```
armiol, line 110: I thought we discussed returning an "action" (e.g. "ignore", or "retry") here. 

Please review.

server/src/main/java/io/spine/server/delivery/PickUpAckMixin.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A mixin of the {@link PickUpAck} that provides convenient methods for accessing the result.
+ */
+@GeneratedMixin
+public interface PickUpAckMixin {
```
armiol, line 41: When posting a command, it was an acknowledgement that we return, because we wanted to say: "we acknowledge the command you posted, and it will be dispatched; but at this point there cannot be neither a result nor an outcome here yet".

Here, it's not the acknowledgement. It's rather an event, or just an "outcome".

Besides, here in this class you mix `PickUpOutcome` and `PickUpAck`, while only one of them is really the thing.

Please review.

server/src/main/java/io/spine/server/delivery/PickUpAckMixin.java
```diff
@@ -0,0 +1,114 @@
...
+    static PickUpAck alreadyPickedBy(WorkerId worker) {
+        checkNotNull(worker);
+        return PickUpAck
+                .newBuilder()
+                .setAlreadyPickedBy(worker)
+                .vBuild();
+    }
+
+    /**
+     * Calls the given {@code consumer} with the {@code ShardProcessingSession} if the shard is
```
armiol, line 70: In this case (and in cases below), please specify what happens if "not".

server/src/main/java/io/spine/server/delivery/PickUpAckMixin.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Returns {@code WorkerId} if this outcome indicates that shard could not be picked
+     * as it's already picked by another worker, or empty {@code Optional} otherwise.
+     */
+    default Optional<WorkerId> alreadyPickedBy() {
+        return Optional.ofNullable(getAlreadyPickedBy());
+    }
+
+    ShardSessionRecord getSession();
```
armiol, line 111: Please document. And also, I think there may be some warnings by IDEA. See which annotations we use to mark these methods, in similar mixins.

Same below.

server/src/main/java/io/spine/server/delivery/memory/InMemoryShardedWorkRegistry.java
```diff
@@ -53,10 +53,17 @@ public final class InMemoryShardedWorkRegistry extends AbstractWorkRegistry {
...
     @Override
-    public synchronized Optional<ShardProcessingSession> pickUp(ShardIndex index, NodeId nodeId) {
+    public synchronized PickUpAck pickUp(ShardIndex index, NodeId nodeId) {
         return super.pickUp(index, nodeId);
     }
 
+    @Override
+    public void release(ShardSessionRecord session) {
+        ShardSessionRecord record = workByNode.get(session.getIndex());
+        // Clear the node ID value and release the session.
```
armiol, line 63: Why do we need this comment inlined? Can we use `@implNote` on a method level instead?

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,24 @@ message WorkerId {
     // An example of such an identifier could be ID of the thread which processes a shard.
     string value = 2 [(required) = true];
 }
+
+// An outcome of the `pickUp()` method of method of `ShardedWorkRegistry` indicating
+// the result of the pickup attempt.
+//
+// The outcome contains one of 2 following results:
+// 1. Shard picked up successfully. In this case the outcome will contain
+// a new `ShardSessionRecord`.
```
armiol, line 117: Please allow an empty line below.

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,24 @@ message WorkerId {
     // An example of such an identifier could be ID of the thread which processes a shard.
     string value = 2 [(required) = true];
 }
+
+// An outcome of the `pickUp()` method of method of `ShardedWorkRegistry` indicating
```
armiol, line 112: Please describe the message in terms of domain language, not in terms of some Java API.

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,24 @@ message WorkerId {
...
+// 2. Shard could not be picked up as it's already picked by another worker. In this case
+// the outcome will contain the `WorkerId` of the worker owning the session.
+message PickUpAck {
+
+    option (is).java_type = "io.spine.server.delivery.PickUpAckMixin";
+
+    oneof ack {
+      option (is_required) = true;
+
+      ShardSessionRecord session = 1;
```
armiol, line 127: Please document the fields.

Also, there is NO(!) camel-case in Proto definitions.

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,24 @@ message WorkerId {
...
+// The outcome contains one of 2 following results:
+// 1. Shard picked up successfully. In this case the outcome will contain
+// a new `ShardSessionRecord`.
+// 2. Shard could not be picked up as it's already picked by another worker. In this case
+// the outcome will contain the `WorkerId` of the worker owning the session.
+message PickUpAck {
+
+    option (is).java_type = "io.spine.server.delivery.PickUpAckMixin";
+
+    oneof ack {
```
armiol, line 124: I think this is just `value`.

server/src/test/java/io/spine/server/delivery/DeliveryTest.java
```diff
@@ -215,12 +218,39 @@ public void returnOptionalEmptyIfPicked() {
...
+        ShardedWorkRegistry registry = new ThrowingWorkRegistry();
+        FixedShardStrategy strategy = new FixedShardStrategy(shardCount);
+        MonitorUnderTest monitor = new MonitorUnderTest();
+        Delivery delivery = Delivery.newBuilder()
+                                    .setStrategy(strategy)
+                                    .setWorkRegistry(registry)
+                                    .setMonitor(monitor)
+                                    .build();
+        ServerEnvironment.when(Tests.class)
+                         .use(delivery);
```
armiol, line 245: Please review the test code to make sure the delivery is reset after this test (no matter it failed or passed).

armiol (changes requested):
@nick-dolgiy please see my comments.

server/src/test/java/io/spine/server/delivery/DeliveryTest.java
nick-dolgiy: The base class has the 
``` java
@AfterEach
public void tearDown() {
    ServerEnvironment.when(Tests.class)
                     .use(originalDelivery);
}
```
method that resets the delivery.

nick-dolgiy:
@armiol PTAL.

armiol:
@nick-dolgiy could you please update the PR description, so that it includes methods with their full signatures (including return values)? It would help to understand the explanations better.

server/src/main/java/io/spine/server/delivery/AlreadyPickedUp.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Represents a scenario when shard could not be picked as it's already picked by another worker.
+ */
+public final class AlreadyPickedUp extends FailedPickUp {
+
+    private final WorkerId sessionOwner;
+
+    /**
+     * Creates  anew {@code AlreadyPickedUp} failure.
```
armiol, line 41: Please fix the typo.

server/src/main/java/io/spine/server/delivery/AlreadyPickedUp.java
```diff
@@ -0,0 +1,72 @@
...
+    AlreadyPickedUp(ShardIndex shard, WorkerId owner, RetryDelivery retry) {
+        super(shard, retry);
+        checkNotNull(owner);
+        sessionOwner = owner;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>In case when shard is already picked up “do nothing” means stop the delivery.
```
armiol, line 59:
```
...picked up, returning this action 
will make the delivery process to finish without any processing done.
```

server/src/main/java/io/spine/server/delivery/AlreadyPickedUp.java
```diff
@@ -0,0 +1,72 @@
...
+     * Creates  anew {@code AlreadyPickedUp} failure.
+     *
+     * @param shard
+     *         a shard that could not be picked
+     * @param owner
+     *         a worker who owns the delivery session
+     * @param retry
+     *         a way to retry delivery
+     */
+    AlreadyPickedUp(ShardIndex shard, WorkerId owner, RetryDelivery retry) {
```
armiol, line 50: If we count on end-users to actually be able to instantiate this type, it must not be package-private.

If your idea was different, and package-private visibility is set intentionally, please explain this choice in `@apiNote` section.

server/src/main/java/io/spine/server/delivery/Delivery.java
```diff
@@ -436,6 +441,27 @@ public Optional<DeliveryStats> deliverMessagesFrom(ShardIndex index) {
         return Optional.of(stats);
     }
 
+    /**
+     * Notifies the {@code DeliveryMonitor} about the technical failure and executes
+     * the failure handler {@code Action}.
+     */
+    private Optional<DeliveryStats> handleTechPickUpFailure(ShardIndex shard, RuntimeException e) {
```
armiol, line 448: Let's shorten to `onRuntimeFailure`.

I don't think that saying "technical <something>" really explains something. It's rather an "unexpected runtime <something>". This goes for both the documentation of this method, and for the `TechFailure` name.

server/src/main/java/io/spine/server/delivery/Delivery.java
```diff
@@ -436,6 +441,27 @@ public Optional<DeliveryStats> deliverMessagesFrom(ShardIndex index) {
...
+    }
+
+    /**
+     * Notifies the {@code DeliveryMonitor} that shard is already pocked up and executes
+     * the failure handler {@code Action}.
+     */
+    private Optional<DeliveryStats> handleAlreadyPickedUp(ShardIndex shard, WorkerId owner) {
+        AlreadyPickedUp failure =
+                new AlreadyPickedUp(shard, owner, () -> deliverMessagesFrom(shard));
+        return monitor.onShardAlreadyPicked(failure)
```
armiol, line 461: Please extract the result into variable prior to returning.

It's a common practice to do so, especially with the chained calls like this one. Otherwise, it's really hard to inspect any issues in scope of debugging.

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -87,12 +87,52 @@ public boolean shouldContinueAfter(DeliveryStage stage) {
...
+     * to pick up a shard.
+     *
+     * <p>Please note, this callback is <em>not</em> invoked in case the shard cannot be picked
+     * if it's already picked by another worker. It is so, because such a use case is a part
+     * of normal flow, and thus does not provoke a {@code RuntimeException}.
+     *
+     * <p>Returns an action to take in relation to the failure.
+     *
+     * <p>By default this callback returns an Action that does nothing. This means the occurred
+     * exception will be throw out of the {@code deliverMessagesFrom()} method.
```
armiol, line 109: Let's have a `{@link ...}` instead of `{@code ..}` here.

server/src/main/java/io/spine/server/delivery/DeliveryMonitor.java
```diff
@@ -87,12 +87,52 @@ public boolean shouldContinueAfter(DeliveryStage stage) {
...
+     *
+     * @param failure
+     *         contains an information about the occurred failure and gives access to
+     *         predefined {@code Action}s to handle the error.
+     */
+    @SuppressWarnings({"WeakerAccess", "unused"})  /* Part of public API. */
+    public FailedPickUp.Action onShardPickUpFailure(TechFailure failure) {
+        return failure.doNothing();
```
armiol, line 117: I don't really like the idea to have `.doNothing()` meaning two different things. In this scenario, returning this value means "rethrow (or propagate) the exception", and in the scenario below it really means "do nothing".

Maybe we could have a more specific action-method for this exception-related scenario?

server/src/main/java/io/spine/server/delivery/FailedPickUp.java
```diff
@@ -0,0 +1,92 @@
...
+        this.shard = shard;
+    }
+
+    /**
+     * Returns an {@code Action} that will not perform any attempts to recovery after the failure.
+     */
+    public abstract Action doNothing();
+
+    /**
+     * Returns an {@code Actioon} that will retry the delivery from the shard.
```
armiol, line 58: Let's fix a typo.

server/src/main/java/io/spine/server/delivery/FailedPickUp.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public Action retry() {
```
armiol, line 60: I think this (and maybe some others) method is `final`.

server/src/main/java/io/spine/server/delivery/FailedPickUp.java
```diff
@@ -0,0 +1,92 @@
...
+        /**
+         * Executes the {@code Action}.
+         */
+        Optional<DeliveryStats> execute();
+    }
+
+    /**
+     * Specifies a way to retry the delivery.
+     */
+    interface RetryDelivery {
```
armiol, line 85: I am not sure why this isn't a sub-type of `Action`? Or why do we need this interface at all.

server/src/main/java/io/spine/server/delivery/FailedPickUp.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Returns a {@code ShardIndex} of the not picked shard.
+     */
+    public ShardIndex shard() {
+        return shard;
+    }
+
+    /**
+     * Action to take in relation n to failed pick-up.
```
armiol, line 72: Please do the proof-read.

server/src/main/java/io/spine/server/delivery/PickUpOutcomeMixin.java
```diff
@@ -0,0 +1,112 @@
...
+                .vBuild();
+    }
+
+    /**
+     * Calls the given {@code consumer} with the {@code ShardProcessingSession} if the shard is
+     * successfully picked, or does nothing otherwise.
+     */
+    @CanIgnoreReturnValue
+    default PickUpOutcomeMixin ifPicked(Consumer<ShardSessionRecord> consumer) {
+        if (getSession() != null) {
```
armiol, line 75: If I understand correctly, `getSession()` is a Proto-generated method. How does it ever return `null`?

It's the same question raising throughout this file for various fields.

server/src/main/java/io/spine/server/delivery/TechFailure.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+package io.spine.server.delivery;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Represents a scenario when shard could not be picked because of some occurred exception.
+ */
+public class TechFailure extends FailedPickUp {
```
armiol, line 34: Please see my comments above regarding the name.

And also, this type seems `final`.

server/src/main/java/io/spine/server/delivery/TechFailure.java
```diff
@@ -0,0 +1,72 @@
...
+     * Creates a new {@code TechFailure}.
+     *
+     * @param shard
+     *         a shard that could not be picked
+     * @param exception
+     *         an occurred exception
+     * @param retry
+     *         a way to retry delivery
+     */
+    TechFailure(ShardIndex shard, RuntimeException exception, RetryDelivery retry) {
```
armiol, line 48: See my comment on `public` vs. package-private above. It's the same here.

server/src/main/java/io/spine/server/delivery/TechFailure.java
```diff
@@ -0,0 +1,72 @@
...
+        super(shard, retry);
+        checkNotNull(exception);
+        this.exception = exception;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>In cease of an exception “do nothing” means propagate the thrown exception.
```
armiol, line 57: "cease"?

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,27 @@ message WorkerId {
     // An example of such an identifier could be ID of the thread which processes a shard.
     string value = 2 [(required) = true];
 }
+
+// An outcome of the shard pick-up attempt indicating the result.
+//
+// The outcome contains one of 2 following results:
+// 1. Shard picked up successfully. In this case the outcome will contain
```
armiol, line 115: Please allow an empty line above.

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,27 @@ message WorkerId {
...
+// a new `ShardSessionRecord`.
+//
+// 2. Shard could not be picked up as it's already picked by another worker. In this case
+// the outcome will contain the `WorkerId` of the worker owning the session.
```
armiol, line 119: Please allow an empty `//` line below.

server/src/test/java/io/spine/server/delivery/ThrowingWorkRegistry.java
```diff
@@ -26,33 +26,31 @@
...
+import com.google.protobuf.Duration;
 import io.spine.server.NodeId;
 
 /**
- * The session of processing the messages, which reside in a shard.
+ * Throws the {@code IllegalStateException} on each operation.
  *
- * <p>Starts by {@linkplain ShardedWorkRegistry#pickUp(ShardIndex, NodeId) picking up}
- * the shard to process.
+ * <p>Such an implementation is usefully in tests for testing error handling.
```
armiol, line 35: "usefully" is not quite English in this context.

server/src/test/java/io/spine/server/delivery/ThrowingWorkRegistry.java
```diff
@@ -26,33 +26,31 @@
...
  */
-@SPI
-public abstract class ShardProcessingSession {
+public class ThrowingWorkRegistry implements ShardedWorkRegistry {
```
armiol, line 37: Seems `final` to me.

armiol (changes requested):
@nick-dolgiy please see my comments.

nick-dolgiy:
@armiol PTAL.

armiol (changes requested):
@nick-dolgiy almost LGTM.

One more thing with this changeset. Could you please update the PR description with the notion of breaking API changes? The thing is that we now assume a different way of releasing the session. So please have a section in PR telling about that, and with "before" and "after" code snippets.

It will be extremely helpful for creating the release notes for 1.9.0 release.

armiol:
@nick-dolgiy I have significantly updated the PR description, and left two `TODO:` statements for you to fill in.

Also note there is no "I" in the PR description now, and most of the sentences were updated.

server/src/main/java/io/spine/server/delivery/FailedPickUp.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An abstract base for shard pick-up-related failures.
+ */
+@Internal
```
armiol, line 39: It it is really `@Internal`, please add a paragraph stating that this type is internal because it is not designed to be extended. But it exposes public `Action`, which is in fact `@SPI` — and end-users should use this action to provide their custom implementations if needed.

Let's discuss this point vocally.

armiol (changes requested):
@nick-dolgiy please see my comments to both PR description and the changeset.

armiol:
@YegorUdovchenko @dmitrykuzmin PTAL at the PR description. I wonder if this API addresses any of your current issues. Please let me know.

YegorUdovchenko:
@armiol It looks sufficient to handle the failures we had.

dmitrykuzmin:
LGTM. A single question/suggestion: could the `AlreadyPickedUp` message include `when_picked`? It seems like a good spot to release a stale shard and retry the delivery.

nick-dolgiy:
@armiol PTAL.

server/src/main/java/io/spine/server/delivery/Delivery.java
```diff
@@ -436,6 +441,29 @@ public Optional<DeliveryStats> deliverMessagesFrom(ShardIndex index) {
...
+     */
+    private Optional<DeliveryStats> onRuntimeFailure(ShardIndex shard, RuntimeException e) {
+        RuntimeFailure failure = new RuntimeFailure(shard, e, () -> deliverMessagesFrom(shard));
+        Optional<DeliveryStats> result = monitor.onShardPickUpFailure(failure)
+                                                .execute();
+        return result;
+    }
+
+    /**
+     * Notifies the {@code DeliveryMonitor} that shard is already pocked up and executes
```
armiol, line 456: "picked".

:(

server/src/main/java/io/spine/server/delivery/FailedPickUp.java
```diff
@@ -0,0 +1,90 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An abstract base for shard pick-up-related failures.
+ *
+ * <p>This class is internal and is not designed to be extended by the framework users. The actual
+ * inheritor of this class will be created by framework and passed to the corresponding failure
+ * handler.
+ *
+ * <p>The internal {@linkplain Action} may be implemented by the framework users to provide
```
armiol, line 43:
```
On the other hand, nested {@link Action} is designed to be implemented 
by the framework users ...
```

server/src/main/java/io/spine/server/delivery/FailedPickUp.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An abstract base for shard pick-up-related failures.
+ *
+ * <p>This class is internal and is not designed to be extended by the framework users. The actual
```
armiol, line 39:
```
Although not marked {@code Internal} explicitly, this type <em>is<em> internal to framework
in terms of extensibility. End-users should not extend it directly. 
Its descendants are provided by the framework itself.
```

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,45 @@ message WorkerId {
...
 }
+
+// One of the possible outcomes of shard pick-up request stating that the shard is already picked
+// by another worker.
+message ShardAlreadyPickedUp {
+
+    // A worker that owns the session.
+    WorkerId worker = 1 [(required) = true];
+
+    // Time when the `worker` has picked the shard.
```
armiol, line 119: "picked up".

It's everywhere, I think.

"To pick" is "to choose". "To pick up" is closer to "grab and possess".

server/src/main/proto/spine/server/delivery/delivery.proto
```diff
@@ -107,3 +108,45 @@ message WorkerId {
...
+    //
+    // Pay attention this time is not synchronized between servers, the time on each instance is
```
armiol, line 121:
```
Pay attention, this value is not synchronized between application servers by default. 
The wall clock on each instance is generally slightly different. Please be careful
when comparing this value (potentially instantiated on another application node),
with the wall clock on your current application node.
```

armiol (changes requested):
@nick-dolgiy please see my comments.

server/src/main/java/io/spine/server/delivery/AlreadyPickedUp.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Returns an action that will make the delivery process to finish without any processing done.
+     */
+    public Action doNothing() {
+        return Optional::empty;
+    }
+
+    /**
+     * Returns the {@code WorkerId} that owns the delivery session.
```
armiol, line 71: I think, it's an outdated description. Please review.

armiol (approved):
LGTM!

PR #1510, opened 2023-05-31
Allow fine-tuning of gRPC server underneath of `GrpcContainer`

armiol:
This changeset is a port of #1454 to `master`.

### Disclaimer

Also, this is just a first of such "porting" PRs. To keep things going, we are going to use the _current_ state of `core-java:master`, rather than wait until the whole machinery (ProtoData et al) unwinds in its full blossom. Therefore, `config`, dependencies and stuff will NOT be updated in scope of "porting".

### What's Changed

Prior to this changeset, there was no convenient way to set some "native" properties of the gRPC server, which is running underneath Spine's `GrpcContainer`. 

In particular, it is often needed [to set](https://grpc.github.io/grpc-java/javadoc/io/grpc/ServerBuilder.html#maxInboundMessageSize-int-) the maximum size for the inbound messages. And it was nearly not possible to achieve, other than using a test-only `GrpcContainer.injectServer()` method. There is also a number of other configuration endpoints provided by the gRPC's `ServerBuilder`, which could not be accessed.

In this PR, the gRPC's `ServerBuilder` API becomes exposed like this:

```java

GrpcContainer.atPort(1654)
             // `server` is an instance of `io.grpc.ServerBuilder`.
             .withServer((server) -> server.maxInboundMessageSize(16_000_000))
             // ...
             .build();
```

So, rather than copying the `ServerBuilder`'s API in `GrpcContainer`, the direct access to the builder is provided. That will allow to be always up-to-date with the configuration capabilities of the gRPC `Server`, rather than copying and chasing their API in the gRPC releases to come.

However, as of this PR, the `GrpcContainer.Builder.withServer(...)` method is marked as `@Experimental`. The reason is that the "native" `ServerBuilder` API allows to perform some destructive actions. Therefore, the decision to expose it as-is is not yet final.

### DIFF
--- a/server/src/test/java/io/spine/server/GrpcContainerTest.java
+++ b/server/src/test/java/io/spine/server/GrpcContainerTest.java
@@ -38,6 +38,7 @@ import java.util.Set;
 
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.Truth8.assertThat;
+import static io.spine.server.given.service.GivenCommandService.noOpCommandService;
 import static io.spine.testing.TestValues.randomString;
 import static java.util.concurrent.Executors.newSingleThreadExecutor;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -69,7 +70,7 @@ class GrpcContainerTest {
 
         var count = 3;
         for (var i = 0; i < count; i++) {
-            var service = CommandService.newBuilder().build();
+            var service = noOpCommandService();
             builder.addService(service);
         }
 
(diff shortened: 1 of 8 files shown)


server/src/test/java/io/spine/server/GrpcContainerTest.java
```diff
@@ -104,6 +103,40 @@ void startServerWithExecutor() throws IOException {
                 .isNotNull();
     }
 
+    @Test
+    @DisplayName("configure underlying gRPC server")
+    void configureUnderlyingGrpcServer() {
+        var port = 1654;
+        var service = CommandService
```
alexander-yevsyukov, line 110: Can we have `.newBuilder()` on the previous line, please? 

Probably, we need to have a test fixture which would read something like `CommandSerice.noOp()` or `CommandService.servingNoCommands()`.

alexander-yevsyukov (approved):
LGTM, which one minor suggestion.

server/src/test/java/io/spine/server/GrpcContainerTest.java
armiol: I have refrained from creating a fully-fledged fixture set, as we only have such cases thrice in two different tests. But I have created a new `Given...` on this matter.

PR #1512, opened 2023-06-02
Allow to handle `RuntimeException`s when delivering `InboxMessage`s

armiol:
This PR is a port of #1496 into `master` branch. Here is a full description, for reviewer's convenience.

This changeset addresses #1472 and brings other improvements into the codebase.

## Failure of signal reception, and `DeliveryMonitor` to respond

`InboxMessage`s which are processed in scope of each `Delivery` run, are eventually unpacked, and the respective signals are dispatched to their targets. In case a signal target is an `Entity`, it is possible that the signal reception does not go well for various reasons. In particular, an aggregate may have "broken" `@Apply`-er, making it impossible to load the aggregate state. Or, some other unexpected error may occur upon loading or storing an entity.

Previously, it was impossible to trace such failures of signal reception. And the `InboxMessage`s causing the errors were left in their inboxes in `TO_DELIVER` status forever, being delivered over and over again — most likely with the same reception failure repeating.

Now, the `Delivery` API allows to subscribe for any failures which occur during the reception of each signal. Additionally, end-users may now choose the way to handle the reception failures in terms of action in respect to the `InboxMessage` of interest. 

Out-of-the-box, there are two actions provided:
* mark the `InboxMessage` as delivered — so that it does not block further delivery of messages;
* repeat the dispatching of the `InboxMessage`. Please note, this is a _synchronous_ and immediate action in terms of a delivery run.

Alternatively, end-users may implement their own way of handling the reception failure.

The corresponding functionality is provided via the API of `DeliveryMonitor`.

Here is a code sample:

```java

public final class MarkFailureDelivered extends DeliveryMonitor {

    /**
     * In case the reception of the {@code InboxMessage} failed,
     * mark it as {@code DELIVERED} anyway.
     */
    @Override
    public FailedReception.Action onReceptionFailure(FailedReception reception) {

    	//// Error details are available as well:
        // InboxMessage msg = reception.message();
        // Error error = reception.error();
        // notifyOf(msg, error);

        return reception.markDelivered();
    }
}

// ...

// Plugging the monitor into the Delivery:

var monitor = new MarkFailureDelivered();

var delivery = Delivery.newBuilder()
        .setMonitor(monitor)
        // ...
        .build();

ServerEnvironment
        .when(MyEnvironment.class)
        .use(delivery);
```

By default, `InboxMessage`s are marked as `DELIVERED` in case of failure of their reception. This is so, because otherwise the issue described in #1472 arises.

Internally, such a functionality became possible by making `MessageDispatcher.dispatch(envelope)` to return `DispatchOutcome`. Previously, it was `void`, which was not suitable for handling any of the issues synchronously — as it was required for `InboxMessage` to allow users choose the way to handle the reception failures.

## Other changes

It was reported, that at run-time the routing of events by their interfaces may lead to `ConcurrentModificationException`. 
This PR addresses the issue by making the `Map` of signal routes concurrency-friendly. See [`io.spine.server.route.MessageRouting`](https://github.com/SpineEventEngine/core-java/pull/1512/files#diff-e25c04493dd2dccb2090c2ae6038eff8b0ecfadd2bc7f96ea42196a1b61550f6) for more details.

Even more minor changes are made to fix typos and remove the chunks of unused code.

The library version is set to `2.0.0-SNAPSHOT.144`.

### DIFF
--- a/server/src/test/java/io/spine/server/delivery/given/ReceptionistAggregate.java
+++ b/server/src/test/java/io/spine/server/delivery/given/ReceptionistAggregate.java
@@ -52,7 +52,7 @@ public final class ReceptionistAggregate
     }
 
     @Apply
-    @SuppressWarnings("ResultOfMethodCallIgnored")  /* Using Proto builder. */
+    @SuppressWarnings("ResultOfMethodCallIgnored" /* Using Proto builder. */)
     private void apply(ConditionerTurnedOn event) {
         maybeFail();
         int newValue = builder().getHowManyCmdsHandled() + 1;
@@ -69,7 +69,7 @@ public final class ReceptionistAggregate
     }
 
     @Apply
-    @SuppressWarnings("ResultOfMethodCallIgnored")  /* Using Proto builder. */
+    @SuppressWarnings("ResultOfMethodCallIgnored" /* Using Proto builder. */)
     private void apply(ConditionerTurnedOff event) {
         maybeFail();
         var newValue = builder().getHowManyCmdsHandled() + 1;
--- a/server/src/test/java/io/spine/server/delivery/given/ReceptionistAggregate.java
+++ b/server/src/test/java/io/spine/server/delivery/given/ReceptionistAggregate.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2023, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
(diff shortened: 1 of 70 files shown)


armiol:
@alexander-yevsyukov PTAL.

server/src/test/java/io/spine/server/delivery/given/ReceptionistAggregate.java
```diff
@@ -0,0 +1,93 @@
...
+        maybeFail();
+        var newValue = builder().getHowManyCmdsHandled() + 1;
+        builder().setId(event.getReceptionist())
+                 .setHowManyCmdsHandled(newValue);
+    }
+
+    @Assign
+    ConditionerTurnedOff handler(TurnConditionerOff cmd) {
+        return ConditionerTurnedOff
+                .newBuilder()
```
alexander-yevsyukov, line 66: Please move the `.newBuilder()` call to the previous line.

alexander-yevsyukov (approved):
LGTM, with one minor request.

PR #1516, opened 2023-06-04
Import events from 3rd-party with respect for tenancy

armiol:
This changeset is a port of #1503 to `master`. Here is a full PR description for reviewer's convenience.

When it comes about importing the third-party events, the framework provides a `ThirdPartyContext` tool. 

However, in a multi-tenant application, the imported events were dispatched in a straightforward manner, without specifying the `TenantId` via `TenantAwareRunner`.

The corresponding test was also flawed: in fact, it was dispatching the events to a single-tenant Bounded Context.

This PR addresses the issue by fixing the test and making `IncomingEventObserver` (part of `IntegrationBroker`, an intermediary between `ThirdPartyContext` and "domestic" Bounded Contexts) to explicitly use `TenantAwareOperation` during the event dispatch.

The version is set to `2.0.0-SNAPSHOT.146`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.145`
+# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.146`
 
 ## Runtime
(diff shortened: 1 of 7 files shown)


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #1519, opened 2023-06-09
Improve test `Sample` generation

armiol:
This changeset is a port of #1506 into `master`. Here is the full description of changes made.

In server-side test code, there is a utility called `Sample`. It allows to generate sample values for any given Proto message. However, it does a lousy job with Proto `string`s. In particular, it just creates them as eight random bytes treated as UTF-8-encoded Java `String`.

Such an approach is problematic if such values are stored either as IDs, or as entity column values into some real DB. For instance, MySQL 5.7 is not able to store such strings as-is — which was discovered when running Spine tests against MySQL in scope of `spine-rdbms` development.

This changeset updates this `Sample` tool, so that `string`-typed fields in Proto messages get UUID-generated values instead of random bytes.

The library version is set to `2.0.0-SNAPSHOT.149`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.148`
+# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.149`
 
 ## Runtime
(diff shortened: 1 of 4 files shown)


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #1523, opened 2023-07-11
Cancel `Subscription`s properly

armiol:
This is a port of #1521 to `master` branch. Here is the full description of the original PR (modified to be in accordance with versions), for reviewer's convenience.

This changeset is meant to address a production-discovered issue related to `Subscription` cancellation.

It addresses [web#202](https://github.com/SpineEventEngine/web/issues/202), and also it is a counterpart of [web#203](https://github.com/SpineEventEngine/web/pull/203).

General idea is to make the `Subscription` cancellation operate the ID of the subscription, not the `Subscription` instance as a whole. Previous approach was based on attempting to find the exact `Subscription` instance before it can be cancelled. However, as `Subscription` instances are received as arguments from outside `core-java` libraries, they may be modified, and thus may differ from those stored in `SubscriptionRegistry`. One of the discovered use-cases was a subscription keep-up process, which led to the changes in the timestamp of corresponding `Subscription` instances.

Public API of `SubscriptionRegistry` was not changed.

The library version is set to `2.0.0-SNAPSHOT.150`.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.149`
+# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.150`
 
 ## Runtime
 1.  **Group** : com.google.android. **Name** : annotations. **Version** : 4.1.1.4.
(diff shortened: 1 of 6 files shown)


armiol:
@alexander-yevsyukov PTAL.

PR #1525, opened 2023-07-28
Add Quodana checks

alexander-yevsyukov:


alexander-yevsyukov:
Too much hassle for now. Let's return to this matter sometime later this year.

PR #1528, opened 2023-07-31
`select()` method for `ProcessManager`

alexander-yevsyukov:
This PR extends `ProcessManagers` with an ability to run `ClientSideQuery` on projection via new `select()`  method. 

### In details:
 * `ProessManagerRepository` was updated to use its `configure(ProcessManager)` method for injecting `BoundedContext` into newly created instance of `ProcessManager`. Previously, the method was no-op. Now, as it does essential job, it got the annotation `@OverridingMethodsMustInvokeSuper`.
 * Correspondingly, the `ProcessManager` class got `context` property, which can be set only once via `injectContext()`.

### Other notable changes:
  * Added `actorRequestFactory(UserId)` function for brevity of creating `ActorRequestFactory`, primarily in tests.
  * Tests for `ProcessManager` were migrated to Kotlin.
  * Migrated some assertions to Kotest.
  * Improved diagnostic message in `EventContextMixin`.
  * Deprecated `find()` and `withId()` methods, which returned `Optional` (in Kotlin `QueringClient` class) in favor of `findWithId()` method which returns `T?`.
  * Fixed naming of proto properties in test data for `ProcessManager` tests (removing `_id` suffix).
  * Improved names of test classes in `ProcessManager` tests.
  * Fixed the package of Kotlin functions associated with `BlackBox`.


alexander-yevsyukov:
@armiol @dmdashenkov, I'm not 100% sure the deprecation of `find()` and `withId()` methods is not suggested too early. 

The main reason is that having `Optional` in Kotlin class is weird, and I've heard a lot, “We hate `Optional`s”, from our colleagues in the integration team recently.

What's your take?

armiol:
I say we deprecate and not remove for a while. Let's try to live with the new API until we are 100% sure.

server/src/main/kotlin/io/spine/server/query/QueryingClient.kt
```diff
@@ -43,40 +43,49 @@ import java.util.*
...
+    /**
+     * Obtains a state of an entity by its ID.
+     *
+     * The value of the ID must be one of the [supported types][io.spine.base.Identifier].
+     *
+     * @return the state of the entity or empty `Optional` if the entity with the given ID
+     *         was not found.
+     * @throws IllegalArgumentException
+     *          if the given ID is not of one of the supported types.
+     * @see [id]
```
armiol, line 84: Not sure I understand this reference. 

Maybe, that's a typo, and you wanted to reference an alternative method? If not, and you really wanted to reference the `id` parameter, please have some explanatory text next to the reference itself. 

It's the same issue below, in the next newly deprecated method.

server/src/test/kotlin/io/spine/server/procman/given/pm/QuizProcess.kt
```diff
@@ -0,0 +1,176 @@
...
+                    stats = it
+                }
+            })
+        } else {
+            EitherOf2.withA(nothing())
+        }
+    }
+}
+
+internal class QuizRepository : ProcessManagerRepository<PmQuizId, QuizProcess, PmQuiz>()
```
armiol, line 154: Not sure why we need it. A default repo would do, with nothing customized in the behavior of this type.

server/src/test/kotlin/io/spine/server/procman/given/pm/QuizProcess.kt
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Subscribe
+    @Suppress("UNUSED_PARAMETER")
+    fun on(event: PmQuestionFailed) = alter {
+        failedQuestions += 1
+    }
+}
+
+internal class QuizStatsRepository : ProjectionRepository<PmQuizId, QuizStatsView, PmQuizStats>()
```
armiol, line 176: Same as above.

server/src/test/proto/spine/test/procman/project.proto
```diff
@@ -75,6 +75,10 @@ message Task {
 }
 
 // Remembers the commands and events.
+//
+// They say that elephants never forget.
+// Our process remembers only the last handled signal.
```
armiol, line 80: "Fitzelephant" :)

armiol (changes requested):
Please see my comments.

server/src/main/kotlin/io/spine/server/query/QueryingClient.kt
alexander-yevsyukov: The method I introduced initially was called `id()`. But it didn't look well in the tests. For example, in a `ProcessManager` it would be like:
```kotlin
val loaded = select(PmQuizStats::class.java).id(id())
```
So I turned it to be a longer name with the verb.

alexander-yevsyukov: I'm fixing the refs.

server/src/test/kotlin/io/spine/server/procman/given/pm/QuizProcess.kt
alexander-yevsyukov: Good point. I was keeping them for the case of custom routing. But it's not needed for now.

dmdashenkov (approved):
LGTM.

armiol (approved):

PR #1530, opened 2023-08-22
Bump dependencies

alexander-yevsyukov:
This PR updates the dependency on latest `base` and on external dependencies.


alexander-yevsyukov:
The build depends on merging (this PR in base)[https://github.com/SpineEventEngine/base/pull/802].

armiol (approved):
LGTM, however, `Spine.kt` still lists `153` as its own version. I guess, that does not matter in this repo, though.

PR #1531, opened 2023-08-26
Make some of `CatchUpTest` tests overridable again

armiol:
Previously, some of `CatchUpTest` methods were grouped into `@Nested` tests. It made impossible to override/disable them in our "descendant" libraries, such as `gcloud-java`.

This PR restores the original behavior.


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #1537, opened 2023-09-20
Extension properties for `TransactionalEntity`

alexander-yevsyukov:
This PR introduces the following extension properties of `TransactionalEntity`:
  * `id`
  * `version`
  * `state`.

The properties are shortcuts allowing to avoid braces when calling corresponding Java methods in Kotlin code.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
@@ -34,7 +34,7 @@ package io.spine.internal.dependency
 @Suppress("unused", "ConstPropertyName")
 object ProtoData {
     const val version = "0.11.6"
-    const val dogfoodingVersion = "0.9.11"
+    const val dogfoodingVersion = "0.11.6"
     const val group = "io.spine.protodata"
     const val compiler = "$group:protodata-compiler:$version"
 
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
@@ -33,7 +33,8 @@ package io.spine.internal.dependency
  */
 @Suppress("unused", "ConstPropertyName")
 object ProtoData {
-    const val version = "0.9.11"
+    const val version = "0.11.6"
+    const val dogfoodingVersion = "0.9.11"
     const val group = "io.spine.protodata"
     const val compiler = "$group:protodata-compiler:$version"
 
--- a/server/src/main/java/io/spine/server/entity/AbstractEntity.java
+++ b/server/src/main/java/io/spine/server/entity/AbstractEntity.java
@@ -104,8 +104,12 @@ public abstract class AbstractEntity<I, S extends EntityState<I>>
      *
      * <p>Assigned either through the {@linkplain #AbstractEntity(Object) constructor which
      * accepts the ID}, or via {@link #setId(Object)}. Is never {@code null}.
+     *
+     * @apiNote The field is named with the underscore prefix to avoid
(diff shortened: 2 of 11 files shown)


buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
```diff
@@ -33,7 +33,8 @@ package io.spine.internal.dependency
  */
 @Suppress("unused", "ConstPropertyName")
 object ProtoData {
-    const val version = "0.9.11"
+    const val version = "0.11.6"
+    const val dogfoodingVersion = "0.11.6"
```
dmdashenkov, line 37: Not sure if we need two versions of ProtoData in `core-java`.

I introduced `dogfoodingVersion` in `mc-java` to separate between the version we apply to build `mc-java` — the dogfooding version — and the version we apply to the `mc-java`'s client projects — the regular version.

server/src/main/kotlin/io/spine/server/entity/TransactionalEntityExts.kt
```diff
@@ -95,3 +96,45 @@ public fun <I, E : TransactionalEntity<I, S, B>, S : EntityState<I>, B : Validat
...
+ * This is a shortcut for `state()`.
+ *
+ * @param I the type of the entity identifiers.
+ * @param E the type of the transactional entity.
+ * @param S the type of the entity state.
+ * @param B the type of the entity state builder.
+ */
+public val <I, E : TransactionalEntity<I, S, B>, S : EntityState<I>, B : ValidatingBuilder<S>>
+        E.state: S
+    get() = state()
```
dmdashenkov, line 140: This does not look like a property to me. The access to the value is synchronized in `AbstractEntity`.

dmdashenkov (changes requested):
Please see my comments.

On a broader note, I'm not sure why would we need the extension properties at all. If it's only to avoid braces, I'm not sure if this is a worthwhile sacrifice. 

To me, the fields starting with `_`, specifically because there are some external accessors that clash with their names, indicate a blurring of the agency of the Java class.

buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
armiol: I second this comment.

server/src/main/java/io/spine/server/entity/AbstractEntity.java
```diff
@@ -135,9 +151,16 @@ public abstract class AbstractEntity<I, S extends EntityState<I>>
      */
     private volatile boolean lifecycleFlagsChanged;
 
+    /**
+     * A context for the logging operations performed by the entity in a receptor.
+     *
+     * <p>Is {@code null} if the entity is not within one of its receptors.
```
armiol, line 157:
```
<p>This field is {@code null}, if the entity is not being accessed 
through a receptor.
```

armiol (changes requested):
@alexander-yevsyukov please see some minor comments.

buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
dmdashenkov: Turns out this is just what `config` brings now. I guess, better have it and not use it that not have it and need it.

server/src/main/kotlin/io/spine/server/entity/TransactionalEntityExts.kt
alexander-yevsyukov: Yes, but we want it for Kotlin code, in tests in particular.

armiol (approved):

armiol (approved):

PR #1539, opened 2023-10-12
Inline functions for routing

alexander-yevsyukov:
This PR converts `CommandRouting` and `EventRouting` to Kotlin, extending these classes with inline functions for idiomatic routing configuration in Kotlin.

Other notable changes:
  * Updated internal dependencies, and Logging in particular.
  * Addressed console warnings for over exposed receptors.


server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -426,7 +426,9 @@ public DispatchOutcome dispatchEvent(EventEnvelope event) {
 
     private Set<I> route(EventEnvelope event) {
         var route = route(eventRouting(), event);
-        return route.orElse(ImmutableSet.of());
+        @SuppressWarnings("unchecked")
+        var result = (Set<I>) route.orElse(ImmutableSet.of());
```
armiol, line 430: You can just create an immutable set with a generic parameter. No need to cast here:

```java
return route.orElse(ImmutableSet.<I>of());
```

server/src/main/java/io/spine/server/entity/EventDispatchingRepository.java
```diff
@@ -136,6 +136,8 @@ public final DispatchOutcome dispatch(EventEnvelope event) {
      */
     protected Set<I> route(EventEnvelope event) {
         var targets = route(eventRouting(), event);
-        return targets.orElse(ImmutableSet.of());
+        @SuppressWarnings("unchecked")
+        var result = (Set<I>) targets.orElse(ImmutableSet.of());
```
armiol, line 140: Same as above.

armiol (changes requested):
@alexander-yevsyukov Almost LGTM. Please see my comments.

armiol (approved):
@alexander-yevsyukov LGTM.

As discussed, Kotlin `Set<I>` means `Set<? extends I>` in Java, since it has a built-in covariance for the generic type. In routing, we expect exactly `I`-typed IDs. So the explicit cast is better in terms of semantics.

server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
alexander-yevsyukov: As we investigated, this is an issue with covariance between Kotlin and Java. We're safe to suppress here. It is better to have `Set<I>` here rather than `Set<? extends I>` because we have a fixed identifier type (`I`) per entity class.

PR #1541, opened 2023-10-31
Introduce `BoundedContext.Probe`

alexander-yevsyukov:
This PR introduces a diagnostic `Probe` interface which can be inserted into `BoundedContext`.

The goal is to encourage `Blackbox` users pass `BoundedContext` instance instead of `BoundedContextBuilder`. The method `BlackBox.from(BoundedContextBuilder)` was not deprecated because of the large number of current usages. We will gradually phase out the method via providing handy utilities for creating `BlackBox` which was already started in this PR.

The code of `BlackBox` was adjusted accordingly, introducing `io.spine.testing.blackbox.probe` package with the `Probe` class which implements `BoundedContext.Probe`.

Other notable changes:
  * **Methods `close()` no longer throw `Exception`.** 1) The declared exception made `try` with resources over `BlackBox` cumbersome because of the need to handle an exception never thrown. 2) Since we're going towards Kotlin, checked exceptions should go anyway.
  * `BoundedContext` got `onBeforeClose` callback for performing necessary clean-up of custom resources used by the context.
  * `Closeable` got `closeIfOpen()` convenience method.
  * `DelegatingDispatcher` and `DispatcherDelegate` interfaces were introduced.
  * `BlackBox` got static factory methods for creating instances without nesting `BoundedContext.assumingTests()` calls.
  * Two Kotlin extension functions were added for `BoundedContext` to simplify check of entities in tests.
  * `DeliveredMessages` class was renamed to `DeliveredMessagesCache` to highlight the fact that it contains not all the messages.
  * Tests now close `BoundedContext` and `BlackBox` more thoroughly. 
  * Guava, Jackson, and Protobuf were bumped to latest versions.


server/src/main/java/io/spine/server/bus/DispatcherDelegate.java
```diff
@@ -0,0 +1,39 @@
...
+import io.spine.server.type.MessageEnvelope;
+import io.spine.type.MessageClass;
+
+/**
+ * A delegate of a {@link DelegatingDispatcher}.
+ *
+ * @param <C> the type of the dispatched message class
+ * @param <E> the type of the message envelope
+ */
+public interface DispatcherDelegate<C extends MessageClass<?>, E extends MessageEnvelope<?, ?, ?>> {
```
yevhenii-nadtochii, line 38: These two `DelegatingDispatcher` and `DispatcherDelegate` are a bit difficult to understand. Maybe we could say more about them in docs?

For example, why `DispatcherDelegate` is not `MessageDispatcher` itself? We need to separate them for some reason.

armiol (changes requested):
@alexander-yevsyukov please see my comments.

PR #1542, opened 2023-11-03
Make `whenever()` return `Iterable<Message>` again

alexander-yevsyukov:
This PR reverts the previously made change to the `Policy.whenever()` method. Now it returns `Iterable<Message>` (instead of `Iterable<EventMessage>`) again.

The initial intent was to make `Policy.whenever()` have a more strict contract on the returned `Iterable`.  But it turned out that this would not allow to use `Either` classes as the declared results of the overridden `whenever()` methods.

The problem is that `Either` implements `Iterable<Message>`. Descendants of `Either` have two or more generic parameters bound by `Message` (not `EventMessage`).  So, `Iterable<EventMessage>` will not be a super-type for anything `EitherOfX`.

There is no common interface for either components other than `Message`. Adding such would require having one more generic parameter for the common interface, which is a lot of pain for users. So, the most obvious way is to revert the change, which I did, documenting this in API notes for `Policy.whenever()` method.

Other changes:
 * Improved documentation of the `Policy` class.
 * `Policy` now checks that it handles only one event in the constructor. This is to fail faster.
 * Added type aliases `NothingHappened ` and `NoReaction` for `Nothing` event in order to help avoiding name clash with `kotlin.Nothing`.
 * Added test to a `Policy` which attempts to react to more than one event.
 * the static method `isOptionalPresent()` was moved into `Tuple` thus eliminating the `Values` utility class.


alexander-yevsyukov:
I had to turn it back to draft because there's the same issue with `Either`, but it's more complicated than with `Just`.

armiol (approved):

PR #1544, opened 2023-11-13
Avoid `MIN_VALUE` and `MAX_VALUE` from `Timestamps`

armiol:
This changeset addresses test code, which used easily-reached constants `Timestamps.MIN_VALUE` and `Timestamps.MAX_VALUE`.

The problem with using these constants is that on a Storage level, we convert those to nanoseconds — as it is a part of our mechanism on providing signal ordering within a single JVM (see emulated nanos in `Time`). But the mentioned constants defined by `Timestamps` cannot be converted to nanos, as the proposed Java type is `long`, and it is not long enough to store such values.

Ultimately, in Storage implementations, using these constants led to `long` overflow.

Besides, `MIN_VALUE` corresponds to 1 CE, and `MAX_VALUE` corresponds to year 9999. Both of which aren't widely used in event-sourced applications at the moment. We'll see how it goes, but for now it's definitely an overkill.

This PR migrates the test code to using more meaningful `Timestamp` values.

### DIFF
--- a/license-report.md
+++ b/license-report.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.174`
+# Dependencies of `io.spine:spine-client:2.0.0-SNAPSHOT.175`
 
 ## Runtime
 1.  **Group** : com.google.android. **Name** : annotations. **Version** : 4.1.1.4.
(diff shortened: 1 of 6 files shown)


alexander-yevsyukov (approved):
LGTM

PR #1550, opened 2024-05-27
[1.x] Expose the `EnvSetting` class

Artem-Semenov-dev:
The `EnvSetting` class is a tool that allows configuring the value that may differ across environment types. Previously it was an internal tool, but now it is available for all Spine projects.
Usage example:
```java 
    EnvSetting<UserReader> setting = new EnvSetting<>();
    setting.use(projectsUserReader(), Production.class)
           .use(localUserReader(), Local.class)
           .use(testsUserReader(), Tests.class);
    
    // ...

    // The instance is returned according to the current `Environment` type.
    final UserReader reader = setting.value(); 
```

This changeset includes the following changes in the `EnvSetting`:
 - introduced method chaining for a more fluent and convenient API (as shown in the example);
 - added a new method that allows retrieval of the stored value for the current environment type of the application;
 - made the API thread-safe, allowing concurrent read operations while ensuring exclusive access for write operations to maintain data integrity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ This repository contains the code of:
 The project is under active ongoing development. 
 You are welcome to experiment and [provide your feedback][email-developers].
 
-The latest stable version is [1.9.0][latest-release].  
+The latest stable version is [1.9.1][latest-release].  
 
 ## Quick start and examples
 
@@ -57,7 +57,7 @@ latest version of the configuration files.
 If you need to use API with one of these annotations, please [contact us][email-developers].
 
 [email-developers]: mailto:developers@spine.io
-[latest-release]: https://github.com/SpineEventEngine/core-java/releases/tag/v1.9.0
+[latest-release]: https://github.com/SpineEventEngine/core-java/releases/tag/v1.9.1
 [spine-site]: https://spine.io/
 [quick-start]: https://spine.io/docs/quick-start
 [spine-examples]: https://github.com/spine-examples


server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -194,10 +227,28 @@ void reset() {
...
         checkNotNull(value);
         checkNotNull(type);
-        this.environmentValues.put(type, new Value<>(value));
+        lockWriteOperation(() -> {
+            this.environmentValues.put(type, new Value<>(value));
+        });
+    }
+
+    /**
+     * This is a test-only method that allows setting the value
```
armiol, line 239: In which case it should probably be `@VisibleForTesting`?

server/src/test/java/io/spine/server/EnvSettingTest.java
```diff
@@ -222,6 +246,135 @@ private void testLazyForEnv(Class<? extends EnvironmentType<?>> type) {
...
+            assertThat(setting.value(Local.class)).isEqualTo(rewrittenValue);
+        }
+
+        @AfterEach
+        void tearDown() {
+            try {
+                readWriteExecutors.shutdownNow();
+                readWriteExecutors.awaitTermination(500, MILLISECONDS);
+            } catch (InterruptedException e) {
+                throw illegalStateWithCauseOf(e);
```
armiol, line 307: I am not sure we want that, I would just ignore. 

If you still think we need to re-throw, please have a method-level comment explaining the behaviour.

server/src/main/java/io/spine/server/EnvSetting.java
```diff
@@ -194,10 +227,28 @@ void reset() {
      * @param type
      *         the type of the environment
      */
-    void use(V value, Class<? extends EnvironmentType> type) {
+    public void use(V value, Class<? extends EnvironmentType> type) {
```
armiol, line 230: Maybe we could return `this`? For call chaining, that is.

I am asking since the example from the PR description is a bit mouthful. I think I would look better like this:
```java

    EnvSetting<UserReader> setting = new EnvSetting<>();
    setting.use(projectsUserReader(), Production.class)
           .use(localUserReader(), Local.class)
           .use(testsUserReader(), Tests.class);
```

armiol (changes requested):
@Artem-Semenov-dev Mostly LGTM. Please see my comments.

armiol:
@Artem-Semenov-dev 
```java
   UserReader currentUserReader = setting.use(projectsUserReader(), Production.class)
                                          .use(localUserReader(), Local.class)
                                          .use(testsUserReader(), Tests.class)
                                          .value();
```
This should be an illustration of some real-world usage. In which (on Earth) scenario, one would call `.value()` right after `.use()`?

armiol (approved):

PR #1563, opened 2025-01-08
Remove usages of `(if_invalid)` option

yevhenii-nadtochii:
This PR removes usages of `(if_invalid)` option.

PR #1564, opened 2025-01-09
Disappearing `(validate)` constraint

yevhenii-nadtochii:
This PR showcases unexpected behavior of codegen validation in `core-java`. Please note, it is based on `class-based-routing` branch.

For some reason, when `(if_invalid)` is applied for `Subscription.topic` field, its generated code doesn't contain the `(validate)` check in `validate()` method. 

In `StandTest`, the modified line should have contained `buildPartial()` from the very beginning.

Take a look at comparison of the generated `validate()` method with and without `(if_invalid)` option:

image

Discovered in no longer needed #1563.

### DIFF
--- a/client/src/main/proto/spine/client/subscription.proto
+++ b/client/src/main/proto/spine/client/subscription.proto
@@ -190,7 +190,7 @@ message Subscription {
     SubscriptionId id = 1 [(required) = true];
 
     // Represents the original topic for this subscription.
-    Topic topic = 2 [(required) = true, (validate) = true, (if_invalid).msg_format = "Invalid topic"];
+    Topic topic = 2 [(required) = true, (validate) = true];
 }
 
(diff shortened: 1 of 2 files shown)


yevhenii-nadtochii:
@alexander-yevsyukov I've reproduced it in `validation`.

Caused by use of the deprecated `(if_invalid).msg_format` instead of `(if_invalid).error_msg`. Hard to notice. As a result, the error message remained empty, whereas `(if_invalid)` was discovered by the corresponding projection and expected a non-empty message in `error_msg`. The validation library throws an exception in this case.

There are the following problems:

1) It looks like we are going to keep `msg_format` for first releases of V2, so it either should be a somehow workable option with a warning, or an explicit compilation error.
2) Runtime exceptions from `validation` are hidden, but they are actually compilation errors.
3) Deprecation warnings from `protoc` are hidden, if they exist at all.

PR #1571, opened 2025-02-14
Create an extension upon `EventMessage` to create `Just` instance

yevhenii-nadtochii:
This PR adds `EventMessage.just()` shortcut to be aligned with `asA()` and `asB()` shortcuts, allowing to avoid unnecessary code nesting when returning an event.


yevhenii-nadtochii:
@armiol @alexander-yevsyukov PTAL

For some reason, Windows tests don't want to pass today.

PR #1584, opened 2025-05-19
Avoid casts to `GeneratedMessageV3`

alexander-yevsyukov:
This PR eliminates the casts and usage of `GeneratedMessageV3` class so that we can migrate to Protobuf v4.31.0, in which this type is deprecated, and the code generated by `protoc` no longer extends this class.

### Other notable changes
 * Pattern variables were used in the code when suggested by IDEA.

### DIFF
--- a/server/src/main/java/io/spine/server/tuple/Either.java
+++ b/server/src/main/java/io/spine/server/tuple/Either.java
@@ -26,14 +26,13 @@
 
 package io.spine.server.tuple;
 
-import com.google.protobuf.GeneratedMessageV3;
 import com.google.protobuf.Message;
 
+import java.io.Serial;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Objects;
-import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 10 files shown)


server/src/main/java/io/spine/server/tuple/Either.java
```diff
@@ -102,17 +99,16 @@ public final boolean equals(Object obj) {
         if (this == obj) {
             return true;
         }
-        if (!(obj instanceof Either)) {
+        if (!(obj instanceof Either other)) {
```
armiol, line 102: This one makes no sense to me. I would expect IDEA to file a warning here.

server/src/main/java/io/spine/server/tuple/Tuple.java
```diff
@@ -166,10 +160,9 @@ public final boolean equals(Object obj) {
      */
     static boolean isOptionalPresent(Object value) {
         checkNotNull(value);
-        if(!(value instanceof Optional)) {
+        if(!(value instanceof Optional<?> asOptional)) {
```
armiol, line 163: Same as above.

armiol (approved):
@alexander-yevsyukov LGTM with a couple of comments to consider.

server/src/main/java/io/spine/server/tuple/Either.java
alexander-yevsyukov: This is new Java feature called "pattern variable".

yevhenii-nadtochii (approved):

server/src/main/java/io/spine/server/tuple/Either.java
alexander-yevsyukov: IDEA suggested changing the code this way. So, I did to avoid the warnings :)

PR #1615, opened 2025-11-14
Introduce `EnvSetting` utility

armiol:
This changeset migrates the changes [made to Spine 1.x](https://github.com/SpineEventEngine/core-jvm/pull/1550) on making `EnvSetting` utility available for public use.

Now, `EnvSetting` allows configuring the value that may differ per environment type:
```java
    var setting = new EnvSetting<UserReader>();
    setting.use(projectsUserReader(), Production.class)
           .use(localUserReader(), Local.class)
           .use(testsUserReader(), Tests.class);
    
    // ...

    // The instance is returned according to the current `Environment` type.
    var reader = setting.value(); 
```

In the example above, there are different implementations of `UserReader` required in different usage scenarios. Previously, such a behaviour could only be achieved via numerous `if ... else ...` statements.

Another improvement made by this PR is the chaining of `use(..)` calls. Previously, when `EnvSetting` was an internal utility, `use()` methods were `void`.

Please note, `EnvSetting` is left non-thread-safe, since it is able to utulise `Supplier`s as the value providers. It makes it impossible to provide a reliable deadlock-free environment without overcomplicating things.

### DIFF
--- a/server/src/main/java/io/spine/server/EnvSetting.java
+++ b/server/src/main/java/io/spine/server/EnvSetting.java
@@ -294,6 +294,8 @@ public final class EnvSetting<V> {
          *
          * @param value
          *         the value to use in this operation
+         * @throws Exception
+         *         in case of a failure
          */
         void accept(V value) throws Exception;
     }


armiol:
@alexander-yevsyukov PTAL. That's NOT urgent by any means, given how much time we've spent having fun with copilot.

Issue #838, opened 2018-09-04
A snapshot of an aggregate may be corrupted due to eventually consistent history


dmytro-grankin:
Currently, it is possible, that a corrupted aggregate snapshot will be written to a storage.

The problem happens when an eventually consistent storage is used (e.g. Datastore).

Let's look at the problem using a task creation example. A normal scenario is the following: `TaskCreated` and `TaskAssigned` events are already stored in the aggregate history. Then, the aggregate is loaded and the events are played, `StartTask` command is dispatched and `TaskStarted` event is emitted and applied to the aggregate. When the aggregate is stored, the snapshot trigger is reached and a snapshot is stored (see the picture below).

![image](https://user-images.githubusercontent.com/25301769/45034309-cba88200-b05f-11e8-866a-47154e5cb7eb.png)

But, after an event was stored in an aggregate history, it may be unavailable during the next history read operations. A problem scenario: `TaskCreated` and `TaskAssigned` events are already stored in the aggregate history. Then, the aggregate is loaded and only `TaskCreated` event is played; `TaskAssigned` event is not returned from the history backward due to eventual consistency. `StartTask` command is dispatched and `TaskStarted` event is emitted and applied to the aggregate. When the aggregate is stored, the snapshot trigger is reached and a snapshot is stored. But, because `TaskAssigned` event was not available and hence not played before applying of `TaskStarted`, the task snapshot has a missing assignee (see the picture below).

![image](https://user-images.githubusercontent.com/25301769/45035265-4377ac00-b062-11e8-9bc2-88de29ae6fab.png)

In other words, if the problem happens, the following is true: the number of played events (excluding a snapshot) don't equal to the event count after the last snapshot (`AggregateStorage.readEventCountAfterLastSnapshot(...)`).

Also, when using Datastore, it is possible, that an event from the middle of an aggregate history won't be available. So, the fix should take this into account.

The framework should not store a wrong snapshot as in the example above or somehow deal with the eventual consistency of an aggregate history.

alexander-yevsyukov:
Hopefully this is going to be addressed by #1259. @armiol is adjusting the way snapshots are made.

alexander-yevsyukov:
After some discussion with @armiol we decided to postpone the fix. We cannot fix it under 1.x without significant performance penalty. The problem does not manifest often. The delay with the fix is not going to impact the current users of the framework.

alexander-yevsyukov:
[...]

PR #1639, opened 2026-07-01
Prevent corrupted aggregate snapshots under eventually consistent history reads

alexander-yevsyukov:
## What & why

Addresses the core of #838: under an eventually consistent storage (e.g. Datastore), an event already written to an aggregate's history can be temporarily unavailable during a later backward read. The aggregate was then reconstructed from a **partial** history and, once the snapshot trigger fired, a **corrupted snapshot was persisted permanently**, poisoning all future loads.

### Root cause

The load path trusted the eventually-consistent *query* (`ReadOperation` / `historyBackward`) as ground truth. `ReadOperation.checkRecord` only asserted "snapshot or non-empty events", and `UncommittedHistory.onAggregateRestored` seeded the post-snapshot event count from `history.getEventCount()` — the size of whatever was read. The existing `AggregateHistoryCorrupted` diagnostic only fires when an applier *throws*, so a silently-short history slipped through.

### The fix

`AggregateStorage.read` now reconciles the number of events read against the aggregate's **stored state version**. That version is written on every store (`writeAll` → `writeState`, which always sets the version) and looked up **by key**, so it is strongly consistent on Datastore-class stores — unlike the query-based history read. Because aggregate event versions are strictly contiguous (`VersionSequence`), a shortfall proves the read is incomplete (missing **tail, middle, or head** events — including the issue's own tail-loss example, where the newest stored event is the one invisible). The load is refused with a new `IncompleteHistoryException` instead of proceeding on partial state — so no corrupted snapshot is written and no new events are emitted from wrong state.

The check is deliberately conservative:

- **skipped** when no authoritative version is available (legacy data written before the state was persisted), and
- **accepts** histories with *more* events than the stored version implies (a state write lagging its events).

New/changed production files (`server/src/main/java/io/spine/server/aggregate/`):

- `HistoryCompleteness.java` — pure reconciliation (`expected = authoritativeVersion − baseVersion`).
- `IncompleteHistoryException.java` — dedicated, catchable failure signal.
- `AggregateStorage.java` — wiring in `read(id, batchSize)`.

### Tests

- `HistoryCompletenessSpec.kt` — unit coverage of the reconciliation (accept / skip / reject; tail, middle, and head loss; snapshot base; whole-history-invisible; legacy skip; over-complete acceptance).
- `AggregateRepositoryTest.throwWhenHistoryIncomplete` (+ a `NewestEventDroppingStorage` stub) — end-to-end wiring test proving `read` actually fetches the authoritative version and throws when the backward read drops the newest event.

## Follow-up (not in this PR)

Fail-fast **prevents corruption** (the issue's subject) but does not yet **retry**. The delivery layer catches the throw and, by default, `DeliveryMonitor.onReceptionFailure` marks the message delivered (drops it) rather than retrying. `IncompleteHistoryException` is a dedicated, catchable type so a custom `DeliveryMonitor` can `repeatDispatching()`; a first-class delayed-redelivery / retrying monitor is a recommended follow-up.

## Also included on this branch

Routine maintenance that accompanied the work (per the repo's per-branch flow):

- Version bump → `2.0.0-SNAPSHOT.384` (with the `by extra(...)` → `extra.set(...)` migration).
- `Update config` — floated shared submodules and the distributed `buildSrc`.
- `Update dependency reports`.

## Verification

`./gradlew clean build` and `./gradlew build dokkaGenerate` green (all modules, detekt, dokka, static analysis). Reviewed by `spine-code-review`, `kotlin-engineer`, and `review-docs` — all APPROVE (must-fix items applied).


server/src/main/java/io/spine/server/aggregate/HistoryCompleteness.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        var authoritative = authoritativeVersion.getNumber();
+        var baseVersion = history != null && history.hasSnapshot()
+                          ? history.getSnapshot()
+                                   .getVersion()
+                                   .getNumber()
+                          : 0;
+        var eventsRead = history == null ? 0 : history.getEventCount();
+        var expected = authoritative - baseVersion;
+        if (eventsRead < expected) {
```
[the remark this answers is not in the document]
alexander-yevsyukov: Clamped `expected` to `Math.max(0, authoritative - baseVersion)` so it can no longer go negative when a lagging stored state version trails the snapshot the read starts from, and added a regression test for that case. Behavior is unchanged (that case was already accepted) — the intent is just explicit now. Fixed in d7f0839.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/test/kotlin/io/spine/server/aggregate/HistoryCompletenessSpec.kt
```diff
@@ -0,0 +1,174 @@
...
+import io.spine.core.Event
+import io.spine.core.Version
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@DisplayName("`HistoryCompleteness` should")
+internal class HistoryCompletenessSpec {
+
+    @Nested
```
alexander-yevsyukov, line 40: Use the following format for nested classes:

```kotlin
@Nested inner class
`accept a history that` {
```

Please correct here and below.

alexander-yevsyukov: Done — converted all three nested classes to backticked class names (`accept a history that`, `skip the check when`, `reject a history that`) and dropped the `@DisplayName` annotations, matching the repo convention (e.g. `IncrementGuardTest`). Pushed in cddc7f1.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

armiol (changes requested):
@alexander-yevsyukov let's postpone these changes until we decide on Aggregate's API changes.

PR #1641, opened 2026-07-01
Retry delivery on `IncompleteHistoryException` after a back-off (#838 follow-up)

alexander-yevsyukov:
## Summary

Follow-up to #838. This PR is **stacked on #1639** (its base is `claude/trusting-bardeen-56899a`). The #838 fix makes an incomplete, eventually-consistent aggregate-history read throw `IncompleteHistoryException` instead of persisting a corrupted snapshot — but the delivery layer then **drops the signal**: the default `DeliveryMonitor.onReceptionFailure` marks it delivered and it is never retried. The only existing retry hook, `FailedReception.repeatDispatching()`, retries **immediately and synchronously**, which cannot let eventual consistency settle and can spin.

This PR adds an **opt-in** way to retry such a signal after a non-blocking, backing-off delay, once storage has likely reached consistency. The default behavior is unchanged (drain).

### Delivery core — make `onReceptionFailure` authoritative

`LiveDeliveryStation` used to mark **every** dispatched message `DELIVERED` *after* the monitor had already acted, so a monitor could never leave a message for a later retry (only `markDelivered()` or the synchronous `repeatDispatching()` worked). Now:

- **`Conveyor`** — `keepForRedelivery` / `isKeptForRetry`; a deferred message keeps its `TO_DELIVER` status (no storage write).
- **`LiveDeliveryStation`** — marks delivered only the messages **not** kept for retry. The non-retry path is unchanged (`delivered == toDispatch`).
- **`FailedReception.keepForRedelivery()`** — new SPI action leaving the message `TO_DELIVER`.
- **`DeliveryMonitor.shouldDeliverNow(InboxMessage)`** — new default-`true` SPI predicate.
- **`Delivery`** — gates the framework's immediate end-of-run self re-trigger through `shouldDeliverNow`, so a deferred message doesn't spin while it waits out its back-off.

`MonitoringDispatcher` and `CatchUpStation` are intentionally untouched — catch-up is projection-only, so the aggregate exception cannot arise there; this is documented on `keepForRedelivery()`.

### The opt-in monitor

**`IncompleteHistoryRetryMonitor`** (aggregate package) recognizes `IncompleteHistoryException` via `Error.type`, keeps the signal and reschedules the shard on a **daemon `ScheduledExecutorService`** (with idle time-out) after an **exponential, capped back-off**, and drains after a configurable maximum number of attempts. Per-message state is a bounded Guava cache; the monitor is `AutoCloseable`. Install it via `DeliveryBuilder.setMonitor(...)`.

## Tests

- **`IncompleteHistoryRetryMonitorSpec`** (Kotlin / Kotest) — recognizes the real exception type (rejecting others, including its superclass) and the exponential-back-off math; deterministic (injected clock and scheduler).
- **`IncompleteHistoryRetryTest`** (Java, `@SlowTest`) — end-to-end through the real delivery pipeline: retries until the storage settles then handles the signal; gives up and drains after the maximum attempts; does not block another entity in the same shard; ignores an unrelated (non-retryable) failure. `ReceptionistAggregate` gained additive per-id failure control.

## Notes for reviewers

- **Stacked on #1639.** Base is `claude/trusting-bardeen-56899a`. Only this PR's own commits appear in the diff; GitHub will retarget it to `master` when #1639 merges.
- Back-off / attempt state is **in-memory, per node** — sufficient for eventual-consistency recovery (any node re-reading after the data settles succeeds; the back-off only paces one node). A persistent, restart-safe variant (an `inbox.proto` `deliver_after` + attempt count and a time-gated redelivery station) is a possible future enhancement.
- One new `@SPI` method (`DeliveryMonitor.shouldDeliverNow`, default `true`) — backward compatible.
- Version bumped `2.0.0-SNAPSHOT.384` → `.385`.

## Verification

`./gradlew build dokkaGenerate` — green across all modules (full test suite, PMD / checkstyle / detekt, Kover coverage verification, Dokka). Reviewed by the repository's `spine-code-review`, `kotlin-engineer`, and `review-docs` reviewers — all **APPROVE**.

### DIFF
--- a/.agents/tasks/838-incomplete-history-retry.md
+++ b/.agents/tasks/838-incomplete-history-retry.md
@@ -1,0 +1,83 @@
+# Task: #838 follow-up — retry delivery on `IncompleteHistoryException`
+
+Status: **implemented & tested** on branch `claude/determined-cerf-ecbbd6`.
+Delete this file on merge to master.
+
+Issue: https://github.com/SpineEventEngine/core-jvm/issues/838
+
+## Context
+
+The #838 fix (merged here from `claude/trusting-bardeen-56899a`) makes an incomplete,
+eventually-consistent aggregate-history read throw `IncompleteHistoryException` from
+`AggregateStorage.read` instead of persisting a corrupted snapshot. But the delivery layer
+then **drops the signal**: `DeliveryMonitor.onReceptionFailure` defaults to `markDelivered()`,
+so the message is never retried. The only existing retry hook,
+`FailedReception.repeatDispatching()`, retries **immediately and synchronously**, which cannot
+let eventual consistency settle and can spin.
+
+This task adds an **opt-in** way to retry such a signal after a **backing-off delay**, without
+blocking the single-threaded-per-shard delivery. Default behavior stays "drain" (decided with
+the maintainer; see also the #1472 task notes).
+
+## What changed
+
+### Delivery core — make `onReceptionFailure` authoritative (the #1472 "Option B" clobber)
+
+`LiveDeliveryStation.process()` used to mark **every** dispatched message `DELIVERED`
+*after* the monitor already acted, so a monitor could never leave a message for a later retry.
+
+- `Conveyor` — new `keepForRedelivery(InboxMessage)` / `isKeptForRetry(InboxMessageId)`; a
+  deferred message keeps its `TO_DELIVER` status (no storage write needed).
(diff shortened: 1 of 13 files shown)


alexander-yevsyukov:
@codex review.

armiol (changes requested):
@alexander-yevsyukov let's postpone these changes until the decision on Aggregate shape is made.

PR #1650, opened 2026-07-10
Introduce entity state history for Aggregates (Phase E)

alexander-yevsyukov:
Implements **Phase E** of the de-event-sourcing plan (`.agents/tasks/de-event-sourcing-plan.md`): an opt-in history of recent `EntityRecord` versions per entity, **readable by the aggregates themselves** ("state at time T"). Detailed task file: `.agents/tasks/entity-state-history.md`. Version: `2.0.0-SNAPSHOT.430 → .440` (**breaking** — see the compatibility notes below).

## What changed

- **`EntityStateKey`** (new proto, `spine/server/entity/state_key.proto`) — the composite record key: entity + version number. Re-writing the same version of the same entity overwrites one record instead of duplicating. *(Named `Key`, not `Id`, to avoid reading like the `I` type parameter of `EntityState`.)*
- **`HistoryStorage<I, M>`** — the abstract base of both per-entity histories: it owns the shared `entity_id` / `created` / `version` column contract (properties named after the stored columns, per the `RecordColumns` convention), the `historyBackward` window reads (with `startingFrom` pagination), the per-entity `trim`, and the count/date `truncate` maintenance.
- **`EntityStateHistoryStorage`** (Kotlin, `io.spine.server.entity.storage`, beside its sibling `EntityEventStorage`; both extend `HistoryStorage`) — stores `EntityRecord`s **as-is** with the same queryable columns as the event journal; `created` is the `Version` timestamp, i.e. the time the state became current. Its `trim` is identifier-only: the record keys carry the version, so ranking needs no record payloads.
- **Retrieval by Aggregates** — the point of the feature: `Aggregate.stateAt(Timestamp): Optional<S>` and `Aggregate.stateHistoryBackward(int): Iterator<S>`, typed to the aggregate state. The event-side reads are now qualified to match: `historyBackward`/`historyContains` → **`eventHistoryBackward`/`eventHistoryContains`** (the deprecated parameterless forms keep their names and point at the replacements).
- **Unified recent histories** — a generic `RecentHistory<T, L>` base owns the loader handling; `RecentEventHistory` and `RecentStateHistory<S>` are its two symmetric views on `TransactionalEntity`, fed by `EventHistoryLoader`/`StateHistoryLoader` (shared `HistoryLoader<R>` base). **Both loaders are lazy** — the event path dropped its double `ImmutableList` materialization. Phase F wires only the `ProcessManager` repository side. Aggregate **parts** get the same wiring: both `create(id)` paths install the loaders.
- **`AggregateRepository.recordStateHistory()`** — opt-in, **off by default**, with `stopRecordingStateHistory()` for the runtime flip (volatile flag; single per-dispatch decision). Each successful dispatch appends the state record — in the per-dispatch `store(A)`, not the batch-deferred `doStore()`, so a batched delivery records its intermediate versions too. **The history is not trimmed automatically**: the framework spends no maintenance query on the dispatch path; retention is the application's duty via `truncate(keepMostRecent[, olderThan])` / `trim(entityId, keepMostRecent)`, as documented on the enabling method. The `stateHistory()` accessor and the entity-side reads **fail fast** while recording is disabled; a bare instance outside a repository reads empty.
- **Storage SPI**:
  - `StorageFactory.createEntityEventStorage(context, entityClass)` and `createEntityStateHistoryStorage(context, entityClass)` — both accept the **entity class** ("create a storage for the repository serving this entity class"), symmetric with `createEntityRecordStorage`.
  - **`StorageGroup`** (new, beside `StorageFactory`) — a named group differentiating the record storages that hold records of the same type. Both histories pass `StorageGroup.of(entityClass)`: the qualified Protobuf name of the entity state.
  - **`createRecordStorage(context, recordSpec, @Nullable StorageGroup)`** is now the single vendor seam and the abstract method of the factory; the two-argument overload became a `default` forwarding "no group". The physical identity a vendor allocates by — a table, a kind — is the **`(group, recordType)` pair**; naming is at the vendor's discretion. Latest-state storages arrive with no group, so they never collide with the state history despite sharing the `EntityRecord` record type. Per-instance isolation makes the group ignorable in-memory; **JDBC and Datastore must honor it** (tracked as a Phase G obligation).
  - **`RecordSpec` slimmed back down**: the short-lived entity-class parameter and the `HistorySpec` intermediary are gone — the histories compose a plain `RecordSpec`, and `sourceType` is a constructor-supplied value again: the entity state class for the state history, and `Event` itself for the journal, which stores its records as-is.

## Semantics worth a look

- **"State at time T"**: the retained record with the highest version whose `created` time is `<= T` (inclusive); same-instant ties resolve to the higher version. Honest about retention: empty when `T` precedes the oldest retained record — or the entity itself — never a guess. The history is read in newest-first batches, so the cost is bounded by the position of the answer, not by the length of the retained history.
- **Trimming is skip-based** over a newest-first read, not version arithmetic: retention gaps (late enabling, purge-and-resume) would break `version <= newest − keep`.
- The journal keeps enrichments out on **both** write paths; the keyed `write(id, …)` overloads of both histories require the identifier to match the record.
- A state record must carry its entity id, **state**, version, and version timestamp — validated at the write boundary.
- Nothing in the new types is aggregate-specific — Phase F wires the same contracts to Process Managers.

## Compatibility notes (`.430 → .440`)

- `Aggregate.historyBackward(int)`/`historyContains(int, Predicate)` → `eventHistoryBackward`/`eventHistoryContains`; `AggregateRepository.historyDepth()`/`setHistoryDepth` → `eventHistoryDepth()`/`setEventHistoryDepth`. The migration guide is updated.
- `RecentHistory` is reshaped into the generic base; `TransactionalEntity.setRecentHistoryLoader` → `setEventHistoryLoader`.
- `StorageFactory.createRecordStorage` gained a `@Nullable StorageGroup` parameter — the three-argument form is now the abstract vendor seam; `createEntityEventStorage` changed its signature (entity class); `createEntityStateHistoryStorage` is new. Vendor factories must implement the new seam and take the group into account when allocating physical storage.
- **Vendor data caveat (pre-GA)**: the journal is now identified per entity class — journal rows written by the `.430` shared-identity layout become invisible to per-type reads, the same class of caveat as the #1649 record-kind change.

## Testing

- `EntityStateHistoryStorageSpec` — window reads, idempotent overwrite, keyed write (matching and mismatching), incomplete-record rejection (including a missing state), `stateAt` (exact hit / floor / same-instant tie / honest-empty), version-over-timestamp ordering, trim (including the identifier-only gap case), truncate at scale, maintenance deletes.
- `EntityEventStorageSpec` — journal semantics incl. enrichment clearing on both write paths and the keyed-write guard.
- `HistoryStorageIdentitySpec` — the `(StorageGroup, recordType)` identity: two entity classes ⇒ separate journals and separate state histories; one entity class ⇒ the two histories distinct from each other and from the latest-state storage, which arrives at the seam with no group; the group survives the system-aware wrapper.
- `AggregateRepositoryStateHistorySpec` — disabled-by-default, fail-fast, one record per successful dispatch (incl. deferred write-through), failed dispatch appends nothing, per-tenant isolation (content-level), receptor reads mid-dispatch, stop/resume, purge recipe, unbounded retention, close-with-repository; `AggregatePartTest` pins the part-side loader wiring.
- `AggregateRepositoryTest` — the configuration accessors (`idempotencyGuardEnabled`, event-history depth group).
- Full `./gradlew clean build dokkaGenerate` green.

## Process

Reviewed twice by multi-agent sessions (48 and 60 agents): the first pass produced 10 findings, the re-review 9 confirmed + 7 split — all confirmed items are addressed on the branch (storage identity, part-repository wiring, keyed-write coverage, retention docs, the version reclassification to `.440`); the split-verdict test/doc improvements are applied, and the append-before-post ordering remains documented fail-loud behavior. After the second review, the storage seam was redesigned in-branch: the dedicated `createHistoryStorage`/`HistorySpec` pair gave way to the `StorageGroup` parameter of `createRecordStorage`. The plan document carries dated amendments for every settled decision.


server/src/main/java/io/spine/server/aggregate/AggregateRepository.java
```diff
@@ -319,6 +335,22 @@ protected void doStore(A aggregate) {
         var history = aggregate.uncommittedHistory();
         aggregateStorage().writeAll(aggregate, history.get());
         aggregate.commitEvents();
+        if (stateHistoryEnabled) {
+            appendStateHistory(aggregate);
```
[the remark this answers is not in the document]
alexander-yevsyukov: Moved the history append from the flush-time `doStore()` to the per-dispatch `store(A)` in 8d45708c43 — a batched delivery now records every intermediate version. A new spec test pins the behavior by deferring the write-through in the fixture.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/main/kotlin/io/spine/server/entity/storage/EntityStateHistoryStorage.kt
```diff
@@ -0,0 +1,177 @@
...
+     * @throws IllegalArgumentException If the record is incomplete (see the
+     *   one-argument [write]), or if the identifier does not match the record.
+     */
+    public override fun write(id: EntityStateId, message: EntityRecord) {
+        validate(message)
+        require(id == message.stateId()) {
+            "The passed identifier does not match the entity and the version of the record."
+        }
+        super.write(id, message)
+    }
```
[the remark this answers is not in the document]
alexander-yevsyukov: Added `@Synchronized` to the `write(id, message)` override in 5b3f8c60c5 — the method-level locking contract of `MessageStorage.write(I, M)` is restored; the monitor is reentrant, so the `super.write` call keeps working unchanged.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/test/kotlin/io/spine/server/entity/storage/HistoryStorageIdentitySpec.kt
```diff
@@ -0,0 +1,112 @@
...
+ * they receive, so the histories of two entity types must reach the
+ * vendor with two different source types. In-memory storages are isolated
+ * per instance and cannot observe a shared table, hence the assertions
+ * capture the specifications at the vendor seam,
+ * [StorageFactory.createRecordStorage].
+ */
+@DisplayName("Per-entity history storages should")
+internal class HistoryStorageIdentitySpec {
+
+    private val context = ContextSpec.singleTenant("`HistoryStorageIdentity` tests")
```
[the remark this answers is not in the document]
alexander-yevsyukov: No change: this file is test code (`server/src/test/kotlin/…/HistoryStorageIdentitySpec.kt`) — the intended audience of `@VisibleForTesting` `ContextSpec.singleTenant` — and uses the same construction as the sibling storage specs; the rule misclassified the test source set.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/main/kotlin/io/spine/server/entity/storage/EntityStateHistoryStorage.kt
```diff
@@ -0,0 +1,209 @@
...
+     *   one-argument [write]), or if the identifier does not match the record.
+     */
+    @Synchronized
+    public override fun write(id: EntityStateKey, message: EntityRecord) {
+        validate(message)
+        require(id == message.stateKey()) {
+            "The passed identifier does not match the entity and the version of the record."
+        }
+        super.write(id, message)
+    }
```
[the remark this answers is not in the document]
alexander-yevsyukov: No change: the override _is_ synchronized — `@Synchronized` (`kotlin.jvm`) on the line above the flagged range marks the generated JVM method `ACC_SYNCHRONIZED`, exactly equivalent to the Java `synchronized` modifier; the analyzer did not credit the Kotlin annotation.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/main/java/io/spine/server/aggregate/AggregateStorage.java
```diff
@@ -218,7 +210,7 @@ public Iterator<I> index() {
     public Optional<EntityEventHistory> read(I id, int batchSize) {
         checkNotClosedAndArguments(id, batchSize);
         checkPositive(batchSize);
-        var events = readHistoryBackward(id, batchSize);
+        var events = ImmutableList.copyOf(historyBackward(id, batchSize));
```
armiol, line 213: I think this is redundant because `Lists.reverse` does not seem to mutate. Please re-check.

server/src/main/java/io/spine/server/storage/RecordSpec.java
```diff
@@ -102,31 +125,41 @@ public final class RecordSpec<I, R extends Message> {
...
+     * @apiNote This ctor is internal to framework, and used to create the record
+     *         specifications serving an entity class: those of the latest entity states
+     *         stored as {@code EntityRecord}s (see
+     *         {@link io.spine.server.entity.storage.SpecScanner SpecScanner}), and those
+     *         of the per-entity histories (see
+     *         {@link io.spine.server.entity.storage.HistorySpec HistorySpec}).
      */
     @Internal
-    public RecordSpec(Class<I> idType,
+    public RecordSpec(@Nullable Class<? extends Entity<?, ?>> entityClass,
```
armiol: Let's not dismantle this type's API. Instead of having this parameter as `null`-able, let's create some other means to achieve the goal.

server/src/main/kotlin/io/spine/server/entity/storage/EntityStateHistoryStorage.kt
```diff
@@ -0,0 +1,241 @@
...
+     * @param at The point in time to look at.
+     * @return The record effective at the given time, or `null` if the history
+     *   does not retain it.
+     * @throws IllegalArgumentException If the type of [entityId] is not supported
+     *   by the framework.
+     */
+    public fun stateAt(entityId: Any, at: Timestamp): EntityRecord? {
+        var startingFrom: Version? = null
+        while (true) {
+            val batch = historyBackward(entityId, STATE_AT_BATCH, startingFrom)
```
armiol, line 122: This will not work in real life on a real storage. We must make `stateAt` mandatory for override in underlying storage implementation. Let's discuss.

armiol (changes requested):

server/src/main/java/io/spine/server/aggregate/AggregateStorage.java
alexander-yevsyukov: The copy is not defensive — since the loaders went lazy, `historyBackward` returns an `Iterator`, and the copy materializes it into a list. The code now says so: a purpose comment, and the reversal via `ImmutableList.reverse()` on the copied list (`e5c33a9066`).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/main/java/io/spine/server/storage/RecordSpec.java
alexander-yevsyukov: Restored: the `@Internal` constructor requires a non-null entity class again; the nullability moved into a private canonical constructor both public constructors delegate to, and the `null` semantics stay documented on the field (`8090fefb5e`).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/main/kotlin/io/spine/server/entity/storage/EntityStateHistoryStorage.kt
alexander-yevsyukov: No code change yet — opening the discussion as asked. Proposal: make the point read a first-class part of the history seam — `createHistoryStorage` returns a dedicated `HistoryRecordStorage` (extending `RecordStorage`) with an **abstract** `latestBefore(timestamp)`-style read; `EntityStateHistoryStorage.stateAt` delegates to it, the in-memory implementation keeps the backward scan, and every real vendor is forced to implement it natively (`WHERE created <= ? ORDER BY version DESC LIMIT 1`). This reopens the earlier "no dedicated storage type" call, which I believe your point about real storages justifies. Softer alternative — an optional capability interface with the scan as fallback — would not give the mandatory guarantee you are asking for.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

server/src/main/java/io/spine/server/storage/RecordSpec.java
alexander-yevsyukov: Follow-up: taken further per the product owner — the entity-class parameter is now removed from `RecordSpec` entirely (`dc26197c53`), rather than hidden behind a private constructor. `RecordSpec` keeps its `(sourceType, idType, recordType, columns, extractId)` API unchanged; `HistorySpec` derives the state class into the source type, and `SpecScanner` passes the scanned state class. The type is no longer touched by the entity-class concept.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

armiol (approved):

PR #1659, opened 2026-07-17
State history for all entity types (Phase F)

alexander-yevsyukov:
## Summary

Generalizes the opt-in entity **state history** so every entity type — not only aggregates — records and reads it, per the reshaped **Phase F** of the de-event-sourcing plan. The feature moves up the class hierarchy; each repository/entity type inherits it with no per-type wiring. Aggregates, Process Managers, and Projections all gain it.

## Logical units (commits)

1. **Planning** — archive the completed de-event-sourcing sub-tasks, reshape Phase F (state history for *all* entity types), add Phase G (event journal for PMs + Aggregates), lock the decisions.
2. **`AbstractEntity` → Kotlin** — behavior- and API-preserving prerequisite so the entity-side member move is Kotlin-to-Kotlin (the `TransactionalEntity` #1644 / `Transaction` #1645 pattern).
3. **Pull state history up** to `AbstractEntity` / `AbstractEntityRepository`.
4. **Expose `id`/`version`/`state`** as member properties of `AbstractEntity` (from the former `TransactionalEntity` extensions).

## What changed

- **Entity side (`AbstractEntity`)** — gains `stateAt(Timestamp)` / `stateHistoryBackward(int)`, the `RecentStateHistory` field, and the `setStateHistoryLoader` seam (from `Aggregate` / `TransactionalEntity`).
- **Repository side (`AbstractEntityRepository`)** — gains the recording config (`recordStateHistory` / `stopRecordingStateHistory` / `stateHistoryEnabled`), the fail-fast `stateHistory()` accessor, the `createStateHistoryStorage` seam, the lazy `stateHistoryStorage()`, a **`final afterStore(E)`** that appends when enabled (final so an override can't silently disable recording), and loader installation on `create(I)` / `toEntity(EntityRecord)`.
- **Bulk-store fix** — `RecordBasedRepository.store(Collection)` now fires `afterStore` per entity; the bulk path previously bypassed it, so a batched `applyMigration` on a recording repository dropped intermediate history entries.
- `AggregateRepository` slimmed to event-journal-only; `Aggregate.toRecord()` deleted (`appendStateHistory` uses the generic `StorageConverter.toEntityRecord`, which builds the identical record).
- `id`/`version`/`state` extension properties become co-located `public val` members; `TransactionalEntityExts.kt` removed.

## Verification

- `./gradlew build dokkaGenerate` green — **1968 server tests**, detekt / ErrorProne / checkstyle / Dokka clean.
- Reviewed by `spine-code-review` (approve), `kotlin-engineer` and `review-docs` (approve-with-changes; all applied).
- New `ProjectionStateHistorySpec` proves the feature on a non-aggregate repository incl. the bulk path; the existing aggregate suite now runs on the inherited machinery.

## Reviewer notes

- **Version**: the branch bumps to `2.0.0-SNAPSHOT.491`, which is shared with the `update-config` branch. If that publishes `.491` first, this branch must re-bump to `.492` before merge.
- **AGENTS.md code-review filter**: the `.agents/**` planning docs are included for context — skip them per the filter.
- **Deferred tests**: dispatch-based acceptance tests (PM "state at time T", projection catch-up idempotency) were deferred to avoid event-routing flakiness; the generalization and bulk-store fix are proven via a routing-free `store()`-based projection spec.
- Two long-standing team-memory cautions (the Fir2Ir ICE on `internal` inherited members, and the `getVersion()`/`HasVersionColumn` "collision") were empirically shown to be over-cautious and corrected.
- The `config` submodule pointer is intentionally left untouched.


server/src/main/kotlin/io/spine/server/entity/Transaction.kt
```diff
@@ -610,7 +610,7 @@ public abstract class Transaction<I : Any,
             val currentState = entity.state()
             @Suppress("UNCHECKED_CAST") // Ensured by the argument of `<E>`.
             val result = currentState.toBuilder() as B
-            if (currentState == entity.defaultState()) {
+            if (currentState == entity.modelClass().defaultState()) {
```
armiol: `entity.defaultState()` is a shortcut for `entity.modelClass().defaultState()`. Consider removing `entity.defaultState()` because it only has just one use-case. in `AbstractEntity.state`.

server/src/test/kotlin/io/spine/server/entity/TransactionTest.kt
```diff
@@ -267,7 +267,7 @@ abstract class TransactionTest<I : Any,
     @Test
     fun `not allow injecting state if entity has non-zero version already`() {
         val entity = createEntity()
-        entity.incrementVersion()
+        entity.updateVersion(Versions.increment(entity.version()))
```
armiol: Not sure why `entity.incrementVersion()` wasn't good enough. I feel the changed version is required for something, but it's not obvious for what. Please document.

armiol (changes requested):

server/src/main/kotlin/io/spine/server/entity/Transaction.kt
alexander-yevsyukov: The change was made because of the `protected` visibility, which is not accessible in the now-Kotlin code. Made `AbstractEntity.defaultState()` `public`.

server/src/test/kotlin/io/spine/server/entity/TransactionTest.kt
alexander-yevsyukov: `incrementVersion()` is now `internal` in Kotlin.

armiol (approved):

